PR #25, opened 2020-06-26
Synсronization TimeControlModeLock and stop playing while connecting

### LINKED ISSUE #24: Server fire ServerPlaying too early after ServerSendingWorldData
`EConnectionState.ServerPlaying` triggered when player watch background image after he loaded world data.
At this moment server think that player already play.
This is fraught with errors when the server tries to send some rpc method to the client.

https://github.com/Bannerlord-Coop-Team/BannerlordCoop/blob/f3b388c1669f386b3f4d2e74ae5fc3e6f5fbf110/source/Network/Infrastructure/ConnectionServer.cs#L47-L54

evorios:
related to issue #24


garrettluskey (changes requested):
Try to not commit .sln or .csproj files

source/Coop/Mod/CoopServer.cs
```diff
@@ -88,6 +95,8 @@ private void OnClientConnected(ConnectionServer connection)
             SyncedObjectStore.AddConnection(connection);
             connection.OnClientJoined += Persistence.ClientJoined;
             connection.OnDisconnected += Persistence.Disconnected;
+            connection.OnServerSendingWorldData += m_GameEnvironmentServer.LockTimeControlStopped;
+            connection.OnServerSendedWorldData += m_GameEnvironmentServer.UnlockTimeControl;
```
araex, line 99: I think that is too early to unlock. We should keep it paused until the client has loaded the state, since the synchronization will only start once the client has fully loaded into the game. I'd unlock on `connection.OnClientJoined` instead.

araex (approved):

garrettluskey (approved):

PR #26, opened 2020-06-27
Disconnect command cli and button in DebugUI

evorios:


source/Coop/Mod/CoopClient.cs
```diff
@@ -113,15 +118,13 @@ private void ConnectionCreated(ConnectionClient con)
 
             m_ReconnectAttempts = MaxReconnectAttempts;
             TryInitPersistence(con);
-            SyncedObjectStore = new RemoteStore(m_SyncedObjects, con);
```
araex: Is there an issue with the store? As it's currently not in use, i don't mind removing it. Just wondering.

araex (approved):

source/Coop/Mod/CoopClient.cs
evorios: No. It's just my fail.

PR #33, opened 2020-07-07
Replay service module for testing synchronization quality

evorios:
Blocked by pull request https://github.com/araex/RailgunNet/pull/1

How it works:
- load savefile in game
- start server
- open rgl-debug-console
- type `coop.record 1`
- do some movements on campaign map
- type `coop.stop`
- close game (there is no "stop server" button yet)
- load the same savefile in game
- start server
- open rgl-debug-console
- type `coop.play 1`
- press `normal speed` (key `2`)
- player's unit will be moving exactly same as past
- wait while in log shown `[MOVETO] Verifying has finished and exported into logs/1-verified.csv`
- export file `BannerlordCoop\mb2\bin\Win64_Shipping_Client\logs\1-verified.csv` in new excel-book
- **time_diff(ms)**: it is the in-game time interval in `ms` between related saved and replayed movement events
- **skipped**: it is the number of saved movement which did not find a related movement while playback
- **new_events**: it is the number of new movement happend while playback which did not find a related in saved movements
- you can create histogram same as follow:


.gitmodules
```diff
@@ -1,16 +1,16 @@
 [submodule "source/RailgunNet"]
 	path = source/RailgunNet
-	url = git@github.com:araex/RailgunNet.git
-	branch = netstandard
+	url = git@github.com:evorios/RailgunNet.git
```
garrettluskey: What is the reason for changing this?

source/Coop/Coop.csproj
```diff
@@ -241,6 +241,8 @@
     <Compile Include="Mod\DebugUtil\DebugManager.cs" />
     <Compile Include="Mod\Coop.cs" />
     <Compile Include="Mod\DebugUtil\DebugUI.cs" />
+    <Compile Include="Mod\DebugUtil\MoveTo.cs" />
```
garrettluskey: Why is this named MoveTo? I get the event being named that but it seems that this module is more generic than specifically movement.

source/Coop/Mod/DebugUtil/MoveTo.cs
```diff
@@ -0,0 +1,283 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.Library;
+using Path = System.IO.Path;
+
+namespace Coop.Mod.DebugUtil
+{
+    public static class MoveTo
```
garrettluskey: Add documentation comments for class and all methods

source/Coop/Mod/DebugUtil/MoveToEvent.cs
```diff
@@ -0,0 +1,55 @@
+﻿using Coop.Mod.Persistence;
+using RailgunNet.System.Encoding;
+using RailgunNet.System.Types;
+using System.Text;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.ObjectSystem;
+
+namespace Coop.Mod.DebugUtil
+{
+    public class MoveToEvent
```
garrettluskey: Add documentation comments for class and all methods

source/Coop/Mod/Persistence/MovementData.cs
```diff
@@ -84,4 +100,31 @@ private enum Field
             NumberOfFleeingsAtLastTravel = 4
         }
     }
+
+    public static class MovementDataSerializer
```
garrettluskey, line 104: Add documentation comments for class and all methods

source/Coop/Mod/Persistence/RailBitBufferExtensions.cs
```diff
@@ -31,5 +31,17 @@ public static MBGUID ReadMBGUID(this RailBitBuffer buffer)
         {
             return new MBGUID(buffer.ReadUInt());
         }
+
+        [Encoder]
+        public static void WriteCampaignTime(this RailBitBuffer buffer, CampaignTime time)
```
garrettluskey, line 36: Could you also add documentation comments for class and all methods

garrettluskey (changes requested):
Looks good except for missing comments

.gitmodules
araex: Because  https://github.com/araex/RailgunNet/pull/1 needed to be approved first. It is merged now. 

@evorios I also merged `netstandard` back to master. Can you please change the url back to what it was and the branch to `master`?

source/Coop/Coop.csproj
araex: Agreed. Maybe `Replay`? 

I would even argue that's not purely a debug utility, we could use it to catch up desynced clients.

source/Coop/Mod/DebugUtil/MoveToEvent.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        public CampaignTime time;
+        public EntityId entityId;
+        public MobileParty party;
+        public MovementData movement;
+        public bool passed = false;
```
araex: I would rename `passed` to `applied`. It more accurately describes the member.

source/Coop/Mod/DebugUtil/MoveTo.cs
```diff
@@ -0,0 +1,283 @@
...
+                        DefaultBehavior = moveTo.movement.DefaultBehaviour,
+                        Position = moveTo.movement.TargetPosition,
+                        SettlementIndex = moveTo.movement.TargetSettlement != null ?
+                            moveTo.movement.TargetSettlement.Id : MovementState.InvalidIndex,
+                        TargetPartyIndex = moveTo.movement.TargetParty != null ?
+                            moveTo.movement.TargetParty.Id : MovementState.InvalidIndex
+                    };
+
+                moveTo.passed = true;
+                Logger.Warn("[MOVETO] Moving to new position.");
```
araex: That's no a warning, `Info` is more appropriate.

source/Coop/Mod/DebugUtil/MoveTo.cs
```diff
@@ -0,0 +1,283 @@
...
+            }
+
+            if (now >= lastTick)
+            {
+                Stop(); // TODO: send message in debug console instead to send on game screen
+                Logger.Warn("[MOVETO] Playback has finished.");
```
araex: Info

source/Coop/Mod/DebugUtil/MoveTo.cs
```diff
@@ -0,0 +1,283 @@
...
+        private static void VerifyEvents()
+        {
+            var difftime = new List<long?>();
+
+            foreach (var play in PlaybackEventList)
+            {
+                var rec = RecordingEventList.FirstOrDefault(q => !q.passed && q.Equals(play));
+                if (rec != null)
+                {
+                    difftime.Add((long)(rec.time - play.time).ToMilliseconds);
```
araex: I think there's an issue here with what is being measured. `play.time` is the campaign time when the client _received_ the event during the first recording. During playback, the event was was processed server side once the campaign time was higher than that time. `rec.time` is again the time the client received the event during playback. Meaning what you're measuring here is the ingame time difference for:

- detecting the state change on the server side (1 railgun update call on server)
- sending the new server to the clients (no additional update call)
- receiving the new state on the client (1 network receive call)
- processing the new state on the client (1 railgun update call on client)

So thats 1 frame on the server and 1 frame on the client, that is the absolute minimum you can measure using this approach. Is that intended?

source/Coop/Mod/DebugUtil/MoveTo.cs
```diff
@@ -0,0 +1,283 @@
...
+            csv[0][5] = l_maxValue;
+
+            foreach (var line in csv)
+            {
+                long diff = ((long?)line[0]).Value;
+                int index = (Math.Sign(diff) * RailUtil.Log2((ulong)Math.Abs(diff)) - minValue) / colWidth;
+                csv[index][1] = (int)csv[index][1] + 1;
+            }
+
+            var report = $"time_diff(ms),cols,skipped,new_events,minValue,maxValue\r\n";
```
araex: The `time_diff(ms)` is in campaign time. Can we maybe get a real world time measurements for this? i think it would be much easier to undestand the results since 1ms in game is basically nothing in real time.

source/Coop/Mod/Persistence/EntityManager.cs
```diff
@@ -28,6 +28,8 @@ public class EntityManager
         private readonly RailServerRoom m_Room;
         private readonly RailServer m_Server;
         private RailServerPeer m_Arbiter;
+        public WorldEntityServer WorldEntityServer { get; private set; }
+        public bool DisableWarn { get; set; } = false;
```
araex: Why did you introduce that? I don't think it's a good idea to suppress warning in code. You could instead use a log receive that can filter incoming messages by logger name. Log4View for example can do that.

araex (changes requested):
Very useful! 

I think it might need some tuning on what's being measured, details in the comments.

source/Coop/Mod/DebugUtil/MoveTo.cs
evorios: @araex It's not problem for desynchronization analyzing. We just search abnormal times. **But** it will be problem if that delay in 2 ticks affects on how local game instance reacts in contrast server's instance.

source/Coop/Mod/Persistence/EntityManager.cs
evorios: @araex It is very inconvenient to catch your messages on the screen in dozens of messages about an inconsistent state.
Is there any other way not to display this on the screen?

araex: Why are there so many inconsistent state messages? This shouldn't happen that often, i would've guessed it happens only when new parties are spawned since that is not yet synchronized.

evorios: I don't know, but it's always shown in 1~2 warnings per second.

araex: During playback, recording or at all times? This shouldn't happen. We should open an issue in that case.

source/Coop/Mod/DebugUtil/MoveTo.cs
evorios: @araex Campaign time is more matter for right simulating events. The main thing is that the variance of the time intervals was to be minimal.

source/Coop/Mod/Persistence/EntityManager.cs
evorios: At all times.

araex: Ok, i opened issue #38 .

araex (approved):

source/Coop/Mod/DebugUtil/Replay.cs
```diff
@@ -0,0 +1,413 @@
...
+    /// <item><term>newEvents</term><description> it is the number of new movement happend while playback
+    /// which did not find a related in saved movements</description></item>
+    /// <item><term>cols</term><description> it is the statistical info about distribution of time intervals
+    /// calculated as <c>count(log2(time_diff), column)</c></description></item>
+    /// </list>
+    /// Constant shift in result data no matter.
+    /// It is important that the spread of values is minimal.
+    /// See <seealso>https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/33</seealso>
+    /// </summary>
+    public static class Replay
```
garrettluskey, line 29: A short summary is sufficient. This should be added to the wiki. You can reference it in remarks

garrettluskey (commented):
Looks good other than too much in one of the document comments.

garrettluskey (approved):
Looks good

Issue #45, opened 2020-07-12
Serializer is missing transfer items


garrettluskey:
Objects after transfer are missing data items. Add all valid serialized to SerializerFactory

PR #54, opened 2020-07-13
Remote procedure calls with large objects

### LINKED ISSUE #19: Provide a way to trigger a synchronized RPC with large binary payloads
RPC through persistence are very limited in the amount of data that can be packed into the event arguments itself. We need a way to send arbitrary data amounts through a second channel and delay the RPC until all client received it.

araex:
Closes #19 . This is a big one, wall of text incoming...

# Content
## Automatically generated remote procedure calls (RPC) with arbitrary arguments.
Best illustrated using an example.

### Patching a class
```C#
class Foo
{
    public void Bar(TaleWorlds.CampaignSystem.MobileParty arg0, int arg1);
}

private static readonly MethodPatch Patch = new MethodPatch(typeof(Foo)).InterceptAll();
```
The above code will do the following:
1. Create a `MethodAccess` that internally stores a snapshot of the `Foo.Bar` method as it is right now. This includes any previously patched functions.
2. Dynamically generate IL code & patch it as prefix for `Foo.Bar`. The prefix will intercept the original call, convert all arguments to an `object[]` and forward the call to `MethodAccess.InvokeOnBeforeCallHandler`. 

That alone will not do much as no OnBeforeCallHandler is registred. The generated prefix will thus just abort and execute the original method. 

### Intercepting the call
To register a handler using the RPC infrastructure:
```C#
[PatchInitializer]
public static void Init()
{
    CoopClient.Instance.OnPersistenceInitialized += persistence => 
        persistence.RpcSyncHandlers.Register(Patch.Methods, CoopClient.Instance);
}
```
This will create a create a `MethodCallSyncHandler` that registers with the generated patch. From this point on, all calls to `Foo.Bar` will be rerouted to the callback defined in `MethodCallSyncHandler.Register`. 

### Converting the intercepted call to a remote procedure call
When a call is intercepted, a `EventMethodCall` is generated which contains a `RPC.MethodCall`. The call consist of the following fields:
- `MethodId`: Globally unique ID for every method patched with `MethodPatch`. Will be indentical on all clients and server as long as they run the same mod version.
- `Instance`: The instance the method was called on. This object has to be previously known at the call site. 
- `Arguments`: List of all arguments

#### Arguments
Every argument in a `RPC.MethodCall` (including the instance) is represented as a `Argument`. `Argument` represents a type safe union that currently supports:
```
    public enum EventArgType
    {
        Null,
        MBObjectManager,
        MBObject,
        Int,
        Float,
        StoreObjectId
    }
```
The `Argument` contains the minimum amount of data to resolve the argument. Meaning the largest binary size an `Argument` will have is 4 Byte + a type identifier. Everything else should be put into the shared store and then be referenced using `EventArgType.StoreObjectId`. See also `ArgumentSerializer`.

### Distributing the RPC event
The client will send the generated RPC event to the server. There it will be passed to a `EventBroadcastingQueue`. If the RPC contains arguments shared using the remote sotre, the call will stay in the queue until all clients have all arguments in their local store. The queue will then broadcast the event to all clients.

### Executing the RPC on a client
When a client receives an RPC, it will:
1. Deserialize the `MethodCall` (including all arguments).
2. Resolve all reference type arguments (`StoreObjectId`, `MBObject`).
3. Resolve the `MethodId` to obtain the local `MethodAccess`.
4. Call `MethodAccess.CallOriginal` on the resolved instance with the resolved arguments.

## Remote store serialization
The remote store now supports custom serializers for predefined types. See `SerializableFactory.Wrap` and `SerializableFactory.Unwrap`.

## Bugfixes with the event handling in Railgun
There were some use after free issues with RailEvents causing events to be executed more than once. It also leaked memory.

## Tons of tests
Lots of features require lots of automated testing :shipit:


source/Coop/Mod/CoopClient.cs
```diff
@@ -163,15 +175,18 @@ private void CreateCharacter(ConnectionClient con)
             {
                 gameManager = new ClientCharacterCreatorManager();
                 MBGameManager.StartNewGame(gameManager);
-                ClientCharacterCreatorManager.OnLoadFinishedEvent += (object source, EventArgs e) =>
+                ClientCharacterCreatorManager.OnLoadFinishedEvent += (source, e) =>
```
garrettluskey: We are going to have issues merging #56 with this as a note.

source/Coop/Mod/CoopServer.cs
```diff
@@ -60,9 +60,12 @@ public string StartServer()
                 Server.EType eServerType = Server.EType.Threaded;
                 Current = new Server(eServerType);
 
-                SyncedObjectStore = new SharedRemoteStore();
+                SyncedObjectStore = new SharedRemoteStore(new SerializableFactory());
```
garrettluskey, line 65: Oh you implemented this? Nice!

source/Coop/Mod/Persistence/RPC/ArgumentFactory.cs
```diff
@@ -1,60 +1,121 @@
...
-                case MBObjectBase mbobj:
-                    return new Argument(mbobj.Id);
                 case int i:
                     return new Argument(i);
+                case float f:
+                    return new Argument(f);
+                case MBObjectBase mbobj:
+                    if (bTransferByValue)
+                    {
+                        goto default;
```
garrettluskey: I hesitate seeing goto statements. I recommend having the return in the if statement and having the code execute past the case on false instead.

source/Coop/Mod/Persistence/RPC/EventBroadcastingQueue.cs
```diff
@@ -0,0 +1,150 @@
...
+using Sync.Store;
+
+namespace Coop.Mod.Persistence.RPC
+{
+    /// <summary>
+    ///     Queue to broadcast events. Sends all events in order and makes sure, that all event
+    ///     arguments can be resolved on the clients. This includes waiting for any pending
+    ///     large object transfers.
+    /// </summary>
+    public class EventBroadcastingQueue : IUpdateable
```
garrettluskey, line 17: Can you add a max queue size and have the Timeout point to the server configuration

source/Coop/Mod/Serializers/SerializableFactory.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Sync.Store;
+using TaleWorlds.CampaignSystem;
+
+namespace Coop.Mod.Serializers
+{
+    /// <summary>
+    ///     Factory to create the serialization wrappers for non-serializable game objects.
+    /// </summary>
+    public class SerializableFactory : ISerializableFactory
```
garrettluskey, line 9: Doesn't contain all serializer members. Should I handle that in #45?

source/Sync/EMethodPatchFlag.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Sync
+{
+    [Flags]
+    public enum EMethodPatchFlag : byte
+    {
+        None = 0,
+        TransferArgumentsByValue = 1 << 1
```
garrettluskey, line 19: It was a little convoluted on what that bit shifting meant. Figured it out thought. Can you add "Reserved" for the other 6 bits?

garrettluskey (changes requested):
A few small things. The new feature here look great though!

source/Coop/Mod/Persistence/RPC/EventBroadcastingQueue.cs
araex: Done. I added errors and an exception (DEBUG only) when the maximum queue size is reached. 

I was thinking it might be i good idea to automatically pause the game when the maximum queue size is reached. That would stop most other events from occuring and give the cause of the delay time to "catch up". That would be a somewhat elegant way to deal with low bandwith connections. What do you think?

source/Coop/Mod/Serializers/SerializableFactory.cs
araex: I'd appreciate if you could do that.

source/Coop/Mod/Persistence/RPC/EventBroadcastingQueue.cs
garrettluskey: I'm fine with that.

garrettluskey (approved):
Looks good. Great work!

PR #55, opened 2020-07-16
Fix for deploy.ps1 and Coop.csproj

garrettluskey:
Verify that this works on your end if the old one also worked.

### DIFF
--- a/source/Coop/Coop.csproj
+++ b/source/Coop/Coop.csproj
@@ -263,8 +263,6 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="Mod\Behaviour\InitServerBehaviour.cs" />
-    <Compile Include="Mod\CoopClientSM.cs" />
-    <Compile Include="Mod\CoopServerSM.cs" />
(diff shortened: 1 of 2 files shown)


source/Coop/Coop.csproj
```diff
@@ -260,6 +263,8 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="Mod\Behaviour\InitServerBehaviour.cs" />
+    <Compile Include="Mod\CoopClientSM.cs" />
+    <Compile Include="Mod\CoopServerSM.cs" />
```
araex: These 2 file adds slipped in i guess?

araex (changes requested):
The files added to Coop.csproj are missing. I assume the adds should not be included in this pull request. Other than that it seems to work just fine on my end.

source/Coop/Coop.csproj
garrettluskey: That would be the case

araex (approved):

PR #56, opened 2020-07-18
State machine redesign

garrettluskey:


garrettluskey:
**TODO**
Add wiki pages for design. 
Impacted tests need to be updated at some point.

This merge is still ready for review in the meantime

source/Common/CoopStateMachine.cs
```diff
@@ -0,0 +1,29 @@
...
+        public CoopStateMachine(T StartingState)
+        {
+            StateMachine = new StateMachine<T, U>(StartingState);
+            State = StartingState;
+
+            StateMachine.OnTransitioned((stateMachine) =>
+            {
+                State = stateMachine.Destination;
+                Logger.Debug($"{GetType().Name} switched from {stateMachine.Source} " +
+                    $"to {stateMachine.Source} " +
```
araex: Wrong state, should be `stateMachine.Destination`.

Additionally, NLog recommends using structured logging, see https://github.com/NLog/NLog/wiki/How-to-use-structured-logging . I don't mind the `$""`, just wanted to make you aware of an alternative.

source/Coop/Mod/CoopClientSM.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        /// <summary>
+        /// Client is required to create character.
+        /// </summary>
+        RequiresCharacterCreation,
+
+        /// <summary>
+        /// Client has existing character on host.
+        /// </summary>
+        CharacterExists,
```
araex, line 49: It seems kinda odd to me that the client decides that. Shouldn't the server offer the available characters to play and the client can then either pick one of those or create a new one?

source/Coop/Mod/CoopServer.cs
```diff
@@ -154,18 +179,57 @@ public override string ToString()
...
             SyncedObjectStore?.RemoveConnection(connection);
         }
+
+        [GameServerPacketHandler(ECoopServerState.Preparing, EPacket.Client_RequestWorldData)]
+        private void ReceiveClientRequestWorldData(ConnectionBase connection, Packet packet)
+        {
+            Client_RequestWorldData info =
+                Client_RequestWorldData.Deserialize(new ByteReader(packet.Payload));
+            Logger.Info("Client requested world data.");
+            ((CoopServerSM)m_CoopServerSMs[(ConnectionServer)connection]).StateMachine.Fire(
```
araex: Couldn't you make `m_CoopServerSMs` a `Dictionary<ConnectionServer, CoopServerSM>`? That would eliminate all those casts. There's only those specific types in there anyways.

source/Network/Infrastructure/ConnectionServerSM.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        /// <summary>
+        /// Client is joining.
+        /// </summary>
+        ClientJoining,
+
+        /// <summary>
+        /// Client is ready.
+        /// </summary>
+        Ready,
```
araex, line 25: Ready for what?

source/Network/Infrastructure/ServerConfiguration.cs
```diff
@@ -9,9 +9,9 @@ public class ServerConfiguration
         public int LanPort { get; set; } = 4201;
         public IPAddress WanAddress { get; set; } = null;
         public int WanPort { get; set; } = 4200;
-        public TimeSpan KeepAliveInterval { get; set; } = TimeSpan.FromSeconds(5);
+        public TimeSpan KeepAliveInterval { get; set; } = TimeSpan.FromSeconds(500);
         public TimeSpan LanDiscoveryInterval { get; } = TimeSpan.FromSeconds(2);
-        public TimeSpan DisconnectTimeout { get; set; } = TimeSpan.FromSeconds(15);
+        public TimeSpan DisconnectTimeout { get; set; } = TimeSpan.FromSeconds(500);
```
araex, line 21: I assume thats to prevent it from disconnecting while debugging? We can leave it at that for now i guess, there's not much point in kicking clients based on that.

source/Network/Protocol/PacketDispatcher.cs
```diff
@@ -1,154 +1,192 @@
...
+                   EqualityComparer<object>.Default.Equals(Owner, pair.Owner) &&
+                   EqualityComparer<Action<ConnectionBase, Packet>>.Default.Equals(Handler, pair.Handler);
         }
 
-        public class PacketEventArgs : EventArgs
+        public override int GetHashCode()
         {
-            public readonly Packet Packet;
-            public readonly EConnectionState State;
+            int hashCode = 1730637479;
```
araex: What are those magic numbers?

araex (changes requested):
Commented on some smaller things, looks good overall. Thanks for cleaning this up! As you've already mentioned in a comment, adding the state diagram that you've shown me previously to the wiki would be helpful.

source/Common/CoopStateMachine.cs
garrettluskey: I find `$""` more readable.

source/Network/Infrastructure/ServerConfiguration.cs
garrettluskey: Client is still timing out even when this is set to 500 seconds. May create an issue for that, let me know what you think.

source/Network/Protocol/PacketDispatcher.cs
garrettluskey: Those came with visual studio quick actions. I removed those numbers.

source/Network/Infrastructure/ConnectionServerSM.cs
garrettluskey: "Ready" is kind of a placeholder for a valid state for the CoopServerSM to take control. If you have an idea of a more explicit name let me know.

source/Network/Infrastructure/ServerConfiguration.cs
garrettluskey: Recorded in #58

source/Coop/Mod/CoopClientSM.cs
garrettluskey: The client will have to control the character creation. These states are to record or skip that process based on what is returned from server.

araex (approved):

PR #68, opened 2020-08-15
Serializable factory update

garrettluskey:


araex (changes requested):
The changes to source/RailgunNet are not intentional, right? It appears you have reverted two commits.

araex (approved):

PR #70, opened 2020-08-22
Synchronize campaign time from host to clients

araex:
This PR synchronizes the hosts CampaignTime.Now to all connected clients. On the clients, the tick times will be dynamically adjusted to match the hosts CampaignTime.


garrettluskey (approved):
Looks good

PR #72, opened 2020-08-26
Player transfer bug fixes

### LINKED ISSUE #45: Serializer is missing transfer items

### LINKED ISSUE #76: Client needs to be granted control to new party
RailNetPeer needs to be assigned to newly created party.

garrettluskey:
Many issues with hero transfer fixed in this. Object Transaction system was scrapped for this update and will remain as an on-hold until we see more value in it. SyncStore works fine for now. Mainly serialization updates and manager updates to pass around newly created heros.


source/Coop/Mod/Managers/ClientManager.cs
```diff
@@ -1,51 +1,41 @@
...
+        public ClientManager(LoadResult saveGameData, string partyName) : base(saveGameData) { m_PartyName = partyName; }
 
         public delegate void OnOnLoadFinishedEventHandler(object source, EventArgs e);
         public static event OnOnLoadFinishedEventHandler OnLoadFinishedEvent;
-
         public override void OnLoadFinished()
         {
             base.OnLoadFinished();
-
+            Parallel.ForEach(MobileParty.All, (party) =>
```
araex: Feels kinda weird to express it that way. The code is not actually thread save, it only works under the assumption that only 1 party.Name == m_PartyName.

How about this instead:
```C#
MobileParty playerParty = MobileParty.All.AsParallel().SingleOrDefault(party => party.Name.ToString() == m_PartyName);
if(playerParty != null)
{
   // ...
}
```

source/Coop/Mod/Serializers/CustomSerializers/PlayerHeroSerializer.cs
```diff
@@ -159,7 +159,25 @@ public override object Deserialize()
...
 
-            return base.Deserialize(hero);
+            
+
+            ConstructorInfo ctor = typeof(HeroDeveloper).GetConstructor(BindingFlags.NonPublic | BindingFlags.Instance,
+                                    null, new Type[] { typeof(Hero) }, null);
+
+            HeroDeveloper newDeveloper = (HeroDeveloper)ctor.Invoke(new object[] { hero });
+            hero.GetType()
+                .GetField("_heroDeveloper", BindingFlags.NonPublic | BindingFlags.Instance)
```
araex, line 169: Do you happen to know of a way to get this to yield a compile error if the field does not exist? Maybe trough a public property that uses the field internally?

Since they are refactoring game DLLs, these hard coded variable names may compile, but no longer work in a new game version. Some way to at least notice that at compile time would be great.

araex (approved):
Looks good

araex:
I pushed 2 compile fixes in 32ffda5 . Most likely these are due to the latest update.

source/Coop/Mod/Serializers/CustomSerializers/PlayerHeroSerializer.cs
garrettluskey: .SetValue will throw a compiler exception due to FieldInfo being null

source/Coop/Mod/Managers/ClientManager.cs
garrettluskey: I agree

Issue #77, opened 2020-08-31
Transferred party visuals not displaying on host


garrettluskey:
When the party is transferred from character creation, the party's visuals are not displaying on the host.

PR #82, opened 2020-09-15
Client can control party

### LINKED ISSUE #81: Railgun movement on events not working as intended
When clicking to move, the event is not processed until another move to click happens.

garrettluskey:
Note there is a bug on the server where the client party is not visible after transfer. Captured in issue #77


araex (approved):
Looks good

PR #88, opened 2020-09-20
fixes 75: Campaign.IsMainPartyWaiting for multiple players

### LINKED ISSUE #75: Time not updating with player movement from separate party
Time controls are working as intended. Currently when parties move the time does not update on all clients. Event for move must not be hooked properly.

araex:
Since we're dealing with multiple player controlled parties now, the Campaign.IsMainPartyWaiting need to be patched to accurately control time flow when any party is moving.


garrettluskey (approved):
Thanks a ton for doing this :)

PR #102, opened 2020-10-20
Issue 41: Steam Interface

### LINKED ISSUE #41: Steam interface
We need a way of getting steam names and a GUID for each player.

Create a file called SteamAPI and have functions for getting Steam name and account number.

Steamworks.NET has SteamUser which probably has all the functionality needed

**Acceptance Criteria**
- Both functions implemented
- Functions are proven working with tests

tristanka:
Created a PlatformAPI class to allow easier access to TaleWorlds' Platform services classes. Intention is to provide an easy way to get player information without having to use individual platform APIs or TaleWorlds' platform services directly.

Test results:
image


araex (approved):
Looks good to me. Thanks a lot!

source/Coop/NetImpl/PlatformAPI.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            this.friendListService = friendListService;
+            this.platformServices = platformServices;
+        }
+
+        public PlatformAPI()
+        {
+            platformServices = PlatformServices.Instance;
+            platformServices.Initialize(new IFriendListService[0]);
+            friendListService = platformServices.GetFriendListServices()[0];
```
araex: Thinking about it, you might want to catch the case where no friend list service is available. We cannot really control that on our end. Right now that would lead to a nullptr access

tristanka: How do you want to handle it? Log it as an error and maybe throw an exception since the PlatformAPI is essentially broken without the friend list service? Trying to think of how this would affect our use cases for it. Maybe attempt to construct the PlatformAPI when connecting to a session and abort the connection attempt if it throws the exception?

tristanka: On second thought, the friends list service isn't required to get the UserId, only the PlatformService instance is... but then again, a given platform's UserId may not technically be unique between platforms, so maybe we avoid it anyway? (PlayerId, not UserId is TaleWorlds' ID that prefixes some stuff to differentiate between platforms and maybe some other stuff)

PR #110, opened 2020-11-21
Game pauses when EventBroadcastingQueue is full

### LINKED ISSUE #57: Automatically pause the game when pending event queue grows too large
I was thinking it might be i good idea to automatically pause the game when the maximum queue size is reached. That would stop most other events from occuring and give the cause of the delay time to "catch up". That would be a somewhat elegant way to deal with low bandwith connections. What do you think?

_Originally posted by @araex in https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/54/review_comment/create_

mihien:
Here's what I did:
-Added a function TimeControlStop() that will pause the game in GameEnvironmentServer, exposed in IEnvironmentServer.
-Changed Add() to return a boolean; true if the event was added to the queue without problem else false (the queue is full).


source/Coop/Mod/Persistence/RPC/EventBroadcastingQueue.cs
```diff
@@ -98,15 +98,17 @@ public void Update(TimeSpan frameTime)
...
-        public void Add(RailServerRoom room, EventMethodCall rpc)
+        public bool Add(RailServerRoom room, EventMethodCall rpc)
         {
             lock (m_Queue)
             {
                 Call call = new Call(room, rpc);
                 call.ObjectsToBeDistributed.RemoveAll(id => m_DistributedObjects.Contains(id));
 
+                bool isFull = false;
                 if (m_Queue.Count >= MaximumQueueSize)
```
araex: The pause should happen before an event gets lost. The condition should probably be `if(m_Queue.Count >= (MaximumQueueSize - 1)`.

source/Coop/Mod/Persistence/RPC/EventBroadcastingQueue.cs
```diff
@@ -98,15 +98,17 @@ public void Update(TimeSpan frameTime)
...
                 {
+                    isFull = true;
                     Logger.Error("Event queue is full!");
```
araex: The error should only appear when the queue is full and an event is thus lost. In the best case scenario, this will be prevented by the pause. But i would assume that the pause will most likely trigger too late.

araex (changes requested):
Looks good, thanks a bunch! But the logic when to pause is somewhat off i think. It should pause before anything gets lost, not when it happens.

source/Coop/Mod/Persistence/RPC/EventMethodCall.cs
```diff
@@ -51,7 +51,8 @@ protected override void Execute(RailRoom room, RailController sender)
                 if (room is RailServerRoom serverRoom)
                 {
                     Logger.Trace("[{eventId}] Broadcast SyncCall: {call}", EventId, Call);
-                    m_EnvironmentServer.EventQueue.Add(serverRoom, this);
+                    if(!m_EnvironmentServer.EventQueue.Add(serverRoom, this))
```
araex: Logic error. Shouldn't be inverted, should it?

mihien: Yes my bad, I've fixed it.

source/Coop/Mod/Persistence/RPC/EventBroadcastingQueue.cs
mihien: Is it okay for some event to be lost or should I found a way to have the pause trigger before it's too late?

source/Coop/Mod/Persistence/RPC/EventMethodCall.cs
araex: I can't see the commit, did you push your changes?

mihien: No I'm going a complete different route to be sure the game will pause before the overflow, but it's not yet functionnal so i haven't pushed it. In fact I have a hard time testing if my solution actually works, I've tried to use Logger.Debug() to write some info on screen (for example the size of the queue), but I doesn't seem to print the result anywhere I've looked.

DrMorphDev: Hi @mihien 

You can view the queue size in the debug info (Ctrl + `):
image

Thanks
Jack

mihien: Ah so that's where it was hidden, it's going to be insanely easier to work with this haha 
Thank you!

MaxBosse:
I merged the current development branch back so this PR is mergable.

For me it looks like this should work, but I was not able to get the queue up (I even limited it to 5, still only reached a max of 1...) so I could not really test it.

PR #141, opened 2021-01-02
Dedicated server

### LINKED ISSUE #127: Implement dedicated server
This issue is needed due to the fact that the map state is not processed while the host is in a battle/settlement. We need to create a dedicated server.

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Currently, the mod treats the server as a host/client. This presents problems when host goes out of the map instance as well as a laggy game. This mitigates both these issues.

## What is the new behaviour?

Resolves #127 

Creates a dedicated game instance without main party to act as a dedicated server.

## Other information


source/SystemTests/UnitTest1.cs
```diff
@@ -0,0 +1,70 @@
...
+using BannerlordSystemTestingLibrary.Utils;
+using System.Reflection;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Diagnostics;
+
+namespace SystemTests
+{
+    [TestClass]
+    public class UnitTest1
```
araex: Ah, here's the missing file :) 

I'm having trouble running this though. I get tons of `UnauthorizedAccessException` in `GameProcess:FindGamePath`. It's appears to be looking for a bannerlord.exe, but UnitTest1.Setup hasn't been called yet. Im not sure how this is supposed to work. I'll have a closer look at whats going on tomorrow.

araex: I tried it again using a fresh checkout of the repository on the same drive as the game installation. That worked, as in it started a server with no party & launched a client instance. The test routine stopped at that point, but i can't find anything in the code base that would continue. So the test doesn't finish. Am i correct to assume that this need be implemented yet?

araex (approved):

PR #146, opened 2021-01-04
Settlement menu open twice fix

### LINKED ISSUE #134: Menu on settlement opens twice
Steps to reproduce
  - Get into a settlement
  - When menu pops up press "Leave"

What should happen
  Menu would open again

Expected behaviour
  Menu gets closed

Possible Problem/Solution
  Unknown yet

DiMiGi:
## PR Checklist

- [X] All new classes have class-level documentation comments, if there are any at all

## PR Type

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

When leaving a settlement the menu opens again, after pressing a second time the "Leave" button the party would leave the settlement

## What is the new behaviour?
Resolves #134 

When leaving a settlement the menu opens once

## Other information

### DIFF
--- a/source/Coop/Mod/Persistence/Party/MovementData.cs
+++ b/source/Coop/Mod/Persistence/Party/MovementData.cs
@@ -85,14 +85,6 @@ namespace Coop.Mod.Persistence.Party
             return Values.GetEnumerator();
         }
 
-        public bool NearlyEquals(MovementData other)
-        {
-            return DefaultBehaviour == other.DefaultBehaviour &&
-                   TargetPosition.NearlyEquals(other.TargetPosition) &&
-                   TargetParty?.Id == other.TargetParty?.Id &&
-                   TargetSettlement?.Id == other.TargetSettlement?.Id;
-        }
-
         public override bool Equals(Object obj)
         {
             if (!(obj is MovementData))
@@ -101,8 +93,8 @@ namespace Coop.Mod.Persistence.Party
             }
 
             MovementData other = (MovementData)obj;
-            return DefaultBehaviour.Equals(other.DefaultBehaviour) &&
-                   TargetPosition.Equals(other.TargetPosition) &&
+            return DefaultBehaviour == other.DefaultBehaviour &&
+                   TargetPosition.NearlyEquals(other.TargetPosition, Compare.COORDINATE_PRECISION) &&
                    TargetParty?.Id == other.TargetParty?.Id &&
                    TargetSettlement?.Id == other.TargetSettlement?.Id;
         }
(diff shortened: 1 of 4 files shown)


source/Coop/Mod/Persistence/Party/MovementData.cs
```diff
@@ -93,6 +93,20 @@ public bool NearlyEquals(MovementData other)
...
+        public override bool Equals(Object obj)
+        {
+            if (!(obj is MovementData))
+            {
+                return false;
+            }
+
+            MovementData other = (MovementData)obj;
+            return DefaultBehaviour.Equals(other.DefaultBehaviour) &&
+                   TargetPosition.Equals(other.TargetPosition) &&
```
araex: Do you really want to exactly compare the float coordinates? I think the position should be compared using `Compare.CoordinatesEqual(Vec2 a, Vec2 b)`. As map coordinates are compressed for synchronization (`Compression.Coordinate2d`), floating point errors are to be expected. I realize you're using the new Equals method to compare local values only, but for the sake of usability i think it would be best to implement it using the coordinate compare anyways. This might prevent comparision errors in the future. Or what do you think?

Looking at the class, there's already a `MovementData.NearlyEquals(MovementData other)` which should use `Compare.CoordinatesEqual(Vec2 a, Vec2 b)` as well. I think it would be good if you combined `Equals` and `NearlyEquals` into a single method, probably just `Equals`.

araex (commented):
Looks good, but i'd like to discuss the implementation of `MovementData.Equals`. I think it should compensate for the floating point errors expected due to the synchronization compression. Even if it currently is only used to compare non-synchronized values. But i'm not sure what the best practices are in C# regarding `Equals`. 

Does anyone else have an opinion on this? Otherwise i'll leave it up to @DiMiGi to decide.

source/Coop/Mod/Persistence/Party/MovementData.cs
DiMiGi: You're right, I've changed the implementation but instead of using CoordinatesEqual I've used the NearlyEquals implemented by TaleWorlds in the Vec2 Class, using the precision in the Compare Class.

image

DiMiGi:
As far as I know `Equals` is used to compare objects and not to know if these objects are the same instance, that means `Equals` should be used in this case. Also, it brings abstraction when using `System.Object` instances, because casting won't be needed, as you could see in source/Sync/FieldChangeBuffer.cs file, where `Equals` is being used in a `Object` type instance, but instead of using the `Object` type implementation it's actually using the `Equals` of the `MovementData` type.

araex (approved):

PR #157, opened 2021-02-17
Refactoring persistence

### LINKED ISSUE #150: Implement Railgun Scoping for player parties
Implement a RailScope class to capture client parties.

araex:
## PR Checklist

## PR Type
- [x] Build related changes
- [ ] CI related changes
- [x] Documentation update
- [x] Other... Please describe: Refactoring with breaking API changes

Refactoring of the persistence integration. Main goal was to make it easier to follow the synchronization logic when navigating the code base. For example, all synchronization in regards to the campaign map party movement is handled in `MobilePartySync` which in turn is instantiated right inside the patch definition of `CampaignMapMovement`. 

### `ISynchronization`
Provides an interface for all broadcasting implementations. A default implemenation of the interface is provided in `CoopClient.Synchronization`. It can be used to issue an RPC:
```C#
CoopClient.Instance.Synchronization?.Broadcast(methodId, instance, args);
```

### `CoopManaged`
Offers an easy way to hook into `TaleWorld` objects and extend them with arbitrary data and patches. The usage of `CoopManaged` is optional, for most use cases a regular harmony patch is easier. A CoopManaged patch is configured statically using the constructor. A usage example can be found in `CampaignMapMovement` that handles all party movement on the world map.

CoopManaged patches that enable `AutoWrapAllInstances` will get automatically attached whenever an object of the managed type is either constructed or initialized through the object manager. See `ObjectManagerAdapter` for the lifetime hooking. Lifetime implementation exist for all `MBObjectBase` and `Campaign`.

### Improved debug ui
The debug UI now has a section "Harmony patches" that displays information about every patch that was registered through any harmony instance. For `ISynchronization` implementations based on `SyncBuffered`, the outgoing command history is automatically recorded. The command history can be viewed an replayed using the UI.

## Other information


araex:
I'm aware it's way too much code to actually review... But a good starting point to get a feel for it is `Coop.Mod.Patch.MobilePartyPatches.CampaignMapMovement`. Additionaly there's `Coop.Tests.CoopFramework` that uses most of the features it offers.

source/Coop/Mod/Persistence/Party/MovementData.cs
```diff
@@ -139,15 +166,15 @@ public static MovementData ReadMovementData(this RailBitBuffer buffer)
...
             {
                 DefaultBehaviour = (AiBehavior) buffer.ReadByte(),
-                TargetSettlement = (id = buffer.ReadMBGUID()) != MovementState.InvalidIndex ?
-                    (Settlement) MBObjectManager.Instance.GetObject(id) :
-                    null,
-                TargetParty = (id = buffer.ReadMBGUID()) != MovementState.InvalidIndex ?
-                    (MobileParty) MBObjectManager.Instance.GetObject(id) :
-                    null,
+                TargetSettlement = (id = buffer.ReadMBGUID()) != Coop.InvalidId
+                    ? (Settlement) MBObjectManager.Instance.GetObject(id)
```
DiMiGi, line 170: I think it would be better creating our own ObjectManager class maybe `CoopObjectManager `so that we can call it like this

`TargetSettlement = (Settlement) CoopObjectManager.GetObject( buffer.ReadMBGUID() )`

It would be much clearer and we will be avoiding validating each time for the `Coop.InvalidId`, it's just an idea what do you think?

araex: Would that `CoopObjectManager` do anything else than checking against `InvalidId`?

I was thinking about whether it would be a good idea to create a shared game object manager that autonomously synchronizes game objects across all clients. Because that is currently unsolved and i'm not sure manual synchronization with patches is the way to go, especially when considering mod support in the future.

araex: On that note, have a look at `Coop.Mod.ObjectManagerAdapter` in this PR. It's an adapter to track the lifetime of `MBObject` instances. Currently used to attach our own synchronization state objects onto the game objects. It could be extended into a bit more than just an adapter to include the automatic synchronization of all instances. What do you think?

DiMiGi: There is a lot of advanced topics in what you've done, at least in what I've seen, and I'm still trying to figure it all out, mainly because I'm working in the `ItemRoster` sync. The idea of `CoopObjectManager` would be to act as a wrapper for the `MBObjectManager`, I'm thinking that it could be used like

`TargetSettlement = CoopObjectManager.GetSettlementObject( buffer.ReadMBGUID() )`

So that if for whatever reason something changes in the `MBObjectManager` the only place to edit would be the wrapper. And the casting wouldn't be necessary.

`Coop.Mod.ObjectManagerAdapter` could be that class if the Adapter refers to the pattern, it should be the same concept. 

About to the automatic sync, for me sounds great, but I'm still struggling to understand the currently synchronization that you've implemented, so there is not much more I can contribute, at least for now.

source/Coop/Mod/Patch/MobilePartyPatches/CampaignMapMovement.cs
```diff
@@ -0,0 +1,260 @@
...
+    ///         whatever the client is doing locally and acts as a ground truth for all parties, regardless of movement
+    ///         orders.
+    /// </summary>
+    public class CampaignMapMovement : CoopManaged<CampaignMapMovement, MobileParty>, IDisposable
+    {
+        #region static patch data
+        /// <summary>
+        ///     Definition of the patch.
+        /// </summary>
+        static CampaignMapMovement()
```
WesKrug, line 41: ⛏
```suggestion
        public static CampaignMapMovement()
```

source/Common/FrameLimiter.cs
```diff
@@ -4,15 +4,21 @@
...
-        private readonly MovingAverage m_Avg;
-        private readonly long m_TargetTicksPerFrame;
-        private readonly Stopwatch m_Timer;
-        private readonly Action<long> m_WaitUntilTick;
-        public TimeSpan LastFrameTime;
-        private double m_AverageTicksPerFrame;
+        /// <summary>
+        ///     The elapsed time between the last call to <see cref="Throttle"/> and the call before that.
+        /// </summary>
+        public TimeSpan LastThrottledFrameTime;
```
WesKrug: Might be worth giving something like this a private setter?
```suggestion
        public TimeSpan LastThrottledFrameTime { get; private set; }
```

source/Common/FrameLimiter.cs
```diff
@@ -40,23 +46,41 @@ public FrameLimiter(TimeSpan targetFrameTime)
                 };
             }
 
-            m_Avg = new MovingAverage(32);
+            m_AverageActualFrameTime = new MovingAverage(32);
```
WesKrug: Just being nitpicky but magic numbers aren't generally great

source/Coop/Mod/GameEnvironmentServer.cs
```diff
@@ -1,38 +1,45 @@
...
-                {
-                    ret = MobileParty.All.SingleOrDefault(p => p.Party.Index == iPartyIndex);
-                });
+            if (!m_PartyCache.TryGetValue(guid, out MobileParty ret))
+            {
+                GameLoopRunner.RunOnMainThread(
+                    () =>
+                    {
+                        // Update the whole cache since we're already in the game loop thread. Doesn't happen that often.
+                        m_PartyCache = MobileParty.All.ToDictionary(party => party.Id);
```
garrettluskey: LINQ as parallel single would work faster afaik. Might not be due to how the threading is handled

source/Coop/Mod/MBLogTarget.cs
```diff
@@ -13,11 +13,11 @@ protected override void Write(LogEventInfo logEvent)
             Color textColor = Color.White;
             if (logEvent.Level == LogLevel.Warn)
             {
-                textColor = Color.FromUint(0xFF0000);
+                textColor = Color.FromUint(0xFFFF00);
```
garrettluskey: Put 0xFFFF00 into a readonly static field

garrettluskey (changes requested):

source/Coop/Mod/Patch/MobilePartyPatches/CampaignMapMovement.cs
araex: static constructors do not take access modifiers. See [Microsoft Docs](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-constructors#:~:text=A%20static%20constructor%20doesn't,common%20language%20runtime%20(CLR).).

source/Common/FrameLimiter.cs
araex: Absolutely, especially if they are used multiple times like in the example you pointed out. Fixed in 
7128bad.

source/Coop/Mod/MBLogTarget.cs
araex: 7128bad

source/Coop/Mod/GameEnvironmentServer.cs
araex: 0c1de41

source/Common/FrameLimiter.cs
araex: 7128bad

garrettluskey (approved):

PR #159, opened 2021-03-31
Misc 1.5.8 fixes

garrettluskey:
## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes/features)

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Serialization changed as party classes changed with update. More than one connected player was not working correctly. There is an unnecessary reference to "Epic.OnlineServices".

## What is the new behavior?
Fixed issue with serialization, multiple players, and removed unnecessary reference.

## Other information


source/Coop/Mod/Managers/ClientCharacterCreatorManager.cs
```diff
@@ -78,16 +82,22 @@ public override void OnLoadFinished()
         private void SkipCharacterCreation()
         {
             CharacterCreationState characterCreationState = GameStateManager.Current.ActiveState as CharacterCreationState;
-            bool flag = CharacterObject.PlayerCharacter.Culture == null;
+            bool flag = characterCreationState.CurrentStage is CharacterCreationCultureStage;
             if (flag)
```
WesKrug:
```suggestion
            if (characterCreationState.CurrentStage is CharacterCreationCultureStage)
```
and then remove the previous bool. No need to store this for later since it isn't used.

WesKrug: And for the same checks below.

source/Coop/Mod/Managers/ClientCharacterCreatorManager.cs
```diff
@@ -78,16 +82,22 @@ public override void OnLoadFinished()
...
-                CultureObject culture = CharacterCreationContentBase.Instance.GetCultures().FirstOrDefault<CultureObject>();
+                CultureObject culture = CharacterCreationContentBase.Instance.GetCultures().GetRandomElementInefficiently();
                 CharacterCreationContentBase.Instance.SetSelectedCulture(culture, characterCreationState.CharacterCreation);
                 characterCreationState.NextStage();
             }
             bool flag2 = characterCreationState.CurrentStage is CharacterCreationFaceGeneratorStage;
             if (flag2)
             {
+                CharacterCreationFaceGeneratorView stageView = characterCreationState.CurrentStage.Listener as CharacterCreationFaceGeneratorView;
+                BodyGeneratorView bgv = (BodyGeneratorView)stageView.GetType().GetField("_faceGeneratorView", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(stageView);
```
WesKrug: Why cast stageView as a CharacterCreationFaceGeneratorView if we're just going to coerce it as a BodyGeneratorView to reflect some things?

source/Coop/Mod/Managers/ClientCharacterCreatorManager.cs
```diff
@@ -78,16 +82,22 @@ public override void OnLoadFinished()
...
+
+                FaceGenVM facegen = bgv.DataSource;
+
+                facegen.FaceProperties.Randomize();
```
WesKrug, line 98: Is this intended to always randomize the face during character generation? Should there be a todo to handle the clients actually creating a character?

source/Coop/Mod/Persistence/EntityManager.cs
```diff
@@ -222,6 +222,18 @@ private void OnClientRemoved(RailServerPeer peer)
             }
         }
 
+        private void AddControlledParty(RailServerPeer peer, MobileParty party)
+        {
+            if (!PlayerControllerParties.ContainsKey(peer))
+            {
+
```
WesKrug: Todo here? Otherwise we should just have lines 231-234 without the else.

source/Coop/Mod/Managers/ClientCharacterCreatorManager.cs
garrettluskey: This is to skip character creation is only for debugging purposes and will be omitted in the released version. Character creation already works.

source/Coop/Mod/Managers/ClientCharacterCreatorManager.cs
garrettluskey: Good point

source/Coop/NetImpl/PlatformAPI.cs
```diff
@@ -1,5 +1,5 @@
 ﻿using Coop.NetImpl.Steam;
-using Epic.OnlineServices;
+//using Epic.OnlineServices;
```
p-JordanStiebritz: Is there a reason why you commented this out vs just removing it?

p-JordanStiebritz:
One general comment that is just a nit/stylistic, so feel free to resolve/ignore it. When instantiating variables, you can just use 'var': ex: var characters = MethodToCreateCharacterArray(). The compiler is smart enough to infer the type and act accordingly =).

WesKrug (approved):

PR #179, opened 2021-04-30
Add fixes to support game version 1.5.9

### LINKED ISSUE #180: Update Mod for 1.5.9 Compatability

MaxBosse:
## PR Checklist

## PR Type

- [x] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Currently you are unable to build the project with the game version 1.5.9

## What is the new behaviour?
This PR fixes all errors during the build of the project and allows you to start the game with the mod and host a coop campaign.

## Other information
The OnLordPartySpawned event was removed in the last update. To compensate I wrote a patch that executes a custom event after a call to InitializeLordPartyProperties.
In 1.5.8 the event was called at the end of InitializeLordPartyProperties, which is why I decided to use a suffix instead of the usualy prefix.


source/Coop/Mod/Persistence/MobilePartyEntityManager.cs
```diff
@@ -116,7 +116,7 @@ private void InitRoom(RailServerRoom room)
             CampaignEvents.OnPartyRemovedEvent.AddNonSerializedListener(this, OnPartyRemoved);
             
             CampaignEvents.MobilePartyCreated.AddNonSerializedListener(this, OnPartyAdded);
-            CampaignEvents.OnLordPartySpawnedEvent.AddNonSerializedListener(this, OnPartyAdded);
+            LordPartyComponentPatch.OnLordPartySpawned += (sender, e) => OnPartyAdded(e);
```
p-JordanStiebritz: Might be able to just get away with removing this line wholly rather than adding `LordPartyComponentPatch`. Since 1.5.9 changed it so that lord parties are just mobile parties, we would probably be handling the party being added twice for players. (Might be a reason why the main party spawned and AI took control of it?)

MaxBosse: Removing the patch indeed fixed the duplicate party. Thanks!

p-JordanStiebritz:
This PR looks fine to me. I think there is a previous existing issue where when you exit, it doesn't actually despawn you from the game. This could be fine, but requires us to not register the client's party a second time or to correctly unregister it when player exits or the game crashes. Any opinions on the persistent client party @garrettluskey ?

garrettluskey (approved):

PR #184, opened 2021-06-12
Remote store sync fix

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Resolving breaks when removing from remote store.
Bool is not encodable by ArgumentFactory.

## What is the new behavior?
Added locking to instance handlers
Added Bool, Campaign Behaviors to arg encoding
Removing from store breaks things so commented it out
Added some misc syncing
Resolves #
N/A

## Other information


source/Coop/Mod/Managers/ServerManager.cs
```diff
@@ -47,6 +48,10 @@ public override void OnLoadFinished()
                     {
                         // Hero received from client after character creation
                         Hero hero = (Hero)serializedPlayerHero.Deserialize();
+
+                        // TODO only do for new players
+                        Settlement settlement = Settlement.Find("tutorial_training_field");
```
WesKrug, line 53: I'd throw this string in a static, or maybe we should make some holder class for this. This could change in the future and magic strings are 0/10

WesKrug (approved):

PR #200, opened 2021-11-15
full launcher and installer

masesk:
## PR Checklist

## PR Type

- [x] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Adding an installer and a launcher to the mod. Installer contains all the necessary dlls and auto-detects where Mount and Blade is installed. Creates a shortcut if needed.

Launcher runs mount and blade with /server or /client (with the necessary submodules) depending on the selection from UI.
There is potential to add more launch options to this.

## What is the new behaviour?
* Install mod to auto-detected MB path
* Create launcher shortcut on desktop
* Launcher can launch the game in server or client mode.
Resolves #
https://i.imgur.com/loc9jqz.png
https://i.imgur.com/OuZTijg.png

## Other information


source/BannerlordCoopInstaller/Installer.cs
```diff
@@ -0,0 +1,264 @@
...
+{
+    public partial class BannerlordCoopInstallerForm : Form
+    {
+
+        private Dictionary<string, ResourceFile> resourceDict;
+
+        /// <summary>
+        /// Check the BannerlordPath and ensure and executable exists. If not, disable the install button
+        /// </summary>
+        private void checkBannerLordPath()
```
WesKrug:
```suggestion
        private void CheckBannerLordPath()
```

source/BannerlordCoopInstaller/Installer.cs
```diff
@@ -0,0 +1,264 @@
...
+        /// Check if MB is installed (through the registry) - Only works in x64.
+        /// </summary>
+        /// <returns>Full path where MB is installed</returns>
+        private string GetBannerLordInstalledPath()
+        {
+
+            string registry_key = @"SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall";
+            using (Microsoft.Win32.RegistryKey key = Registry.LocalMachine.OpenSubKey(registry_key))
+            {
+                foreach (string subkey_name in key.GetSubKeyNames())
```
WesKrug, line 32: This feels a bit unnecessary. Can we use LINQ to find the subkey we're looking for? Do we only expect a few subkeys in each key?  Probably fine since it's just during install...

source/BannerlordCoopInstaller/Installer.cs
```diff
@@ -0,0 +1,264 @@
...
+                {
+                    using (RegistryKey subkey = key.OpenSubKey(subkey_name))
+                    {
+                        string name = (string)subkey.GetValue("DisplayName");
+                        if (name != null && name.Equals("Mount & Blade II: Bannerlord"))
```
WesKrug:
```suggestion
                        if (!string.IsNullOrEmpty(name) && name.Equals("Mount & Blade II: Bannerlord"))
```

source/BannerlordCoopInstaller/Installer.cs
```diff
@@ -0,0 +1,264 @@
...
+                        {
+                            return subkey.GetValue("installlocation").ToString();
+                        }
+                    }
+                }
+            }
+            return null;
```
WesKrug: Since we're using this to set the initial text, should probably use an empty string instead of null. Same difference but nicer imo
```suggestion
            return string.Empty;
```

source/BannerlordCoopInstaller/Installer.cs
```diff
@@ -0,0 +1,264 @@
...
+
+        /// <summary>
+        /// Allow user to click the browse button, open a prompt to select a different path.
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        private void PathBrowse_Click(object sender, EventArgs e)
+        {
+            FolderBrowserDialog folderBrowserDialog1 = new FolderBrowserDialog();
+            if (folderBrowserDialog1.ShowDialog() == DialogResult.OK)
```
WesKrug, line 94: Does the cancel button work? Is that inherit functionality from forms?

source/BannerlordCoopInstaller/Installer.cs
```diff
@@ -0,0 +1,264 @@
...
+                        this.Status.Text = "Ensure all files and processes are closed before reinstalling.";
+                        this.BannerlordPath.Enabled = true;
+                        this.ShortcutCheckbox.Enabled = true;
+                        this.PathBrowse.Enabled = true;
+                        this.Install.Enabled = true;
+                        return;
+                    }
+                    
+                }
+                else if (dialogResult == DialogResult.No)
```
WesKrug: Else is unnecessary.
```suggestion
                else
```

source/BannerlordCoopInstaller/Installer.cs
```diff
@@ -0,0 +1,264 @@
...
+            if (Directory.Exists($@"{this.BannerlordPath.Text}\Modules\Coop"))
+            {
+                DialogResult dialogResult = MessageBox.Show("An older version exists and must be removed. Are you sure?", "Older Version Detected", MessageBoxButtons.YesNo);
+                if (dialogResult == DialogResult.Yes)
+                {
+                    try
+                    {
+                        //recursively delete the mod
+                        Directory.Delete($@"{this.BannerlordPath.Text}\Modules\Coop", true);
+                    }catch(Exception)
```
WesKrug:
```suggestion
                    }
                    catch(Exception)
```

WesKrug: Also should we log/display the error? I'd be curious to see the error message as a user

source/BannerlordCoopInstaller/Installer.cs
```diff
@@ -0,0 +1,264 @@
...
+            PerformInstall();
+
+
+
+
+
+            // creating a shortcut
+            if (this.ShortcutCheckbox.Checked)
+            {
+                IShellLink link = (IShellLink)new ShellLink();
```
WesKrug, line 202: do we need to coerce this here? Is a ShellLink not an IShellLink? Seems silly

source/BannerlordCoopInstaller/Installer.cs
```diff
@@ -0,0 +1,264 @@
...
+using System.Text;
+using System.Windows.Forms;
+using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.ComTypes;
+
+namespace BannerlordCoopInstaller
+{
+    public partial class BannerlordCoopInstallerForm : Form
+    {
+
```
WesKrug, line 18: Tons of whitespace in this class. ctrl+K+F should clean it up, but lets trim this down

source/Coop.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
```
WesKrug: Lots of scary changes in here - Are you running VS 2022? shouldn't be any reason for these project ids and such to change

source/Launcher/Launcher.cs
```diff
@@ -0,0 +1,103 @@
...
+        public BannerlordCoopLauncher()
+        {
+            InitializeComponent();
+        }
+
+        /// <summary>
+        /// Check if MB is installed (through the registry) - Only works in x64.
+        /// </summary>
+        /// <returns>Full path where MB is installed</returns>
+        private string GetBannerLordInstalledPath()
```
WesKrug: Since this is copy pasta, lets pull this out into some status util for both classes to use. If for some reason this changes in future releases, we don't want to have to change it twice

source/Launcher/Launcher.cs
```diff
@@ -0,0 +1,103 @@
...
+                        if (name != null && name.Equals("Mount & Blade II: Bannerlord"))
+                        {
+                            return subkey.GetValue("installlocation").ToString();
+                        }
+                    }
+                }
+            }
+            return null;
+        }
+
```
WesKrug, line 58: Also some white space to clean

source/Launcher/Launcher.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            Process process = new Process();
+            string mbPath = GetBannerLordInstalledPath();
+            process.StartInfo.WorkingDirectory = mbPath + @"\bin\Win64_Shipping_Client\";
+            process.StartInfo.UseShellExecute = true;
+            process.StartInfo.CreateNoWindow = false;
+            process.StartInfo.FileName = "\"" + mbPath + @"\bin\Win64_Shipping_Client\Bannerlord.exe" + "\"";
+            process.StartInfo.Arguments = "/singleplayer /server _MODULES_*Native*SandBoxCore*CustomBattle*SandBox*StoryMode*Coop*_MODULES_";
+            process.Start();
+            process.Close();
```
WesKrug: Why do we close it? I assume this works I just don't know what it does haha

source/Launcher/Launcher.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+        /// <summary>
+        /// Start a new process to run the application, then close it. Launch the game in sever mode
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        private void JoinGame_Click(object sender, EventArgs e)
+        {
+            Process process = new Process();
+            string mbPath = GetBannerLordInstalledPath();
```
WesKrug: Just pull this out into a private variable and set it in the constructor

source/Launcher/Launcher.cs
```diff
@@ -0,0 +1,103 @@
...
+        private void CloseButton_Click(object sender, EventArgs e)
+        {
+            Close();
+        }
+        /// <summary>
+        /// Start a new process to run the application, then close it. Launch the game in client mode
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        private void HostGame_Click(object sender, EventArgs e)
```
WesKrug, line 136: this is pretty identical to the join game click method, lets extract them out to a new method and set the arguments as appropriate in here

WesKrug (changes requested):

source/BannerlordCoopInstaller/Installer.cs
masesk: The functionality of the dialog box is inherited from the second parameter: MessageBoxButtons.YesNo in the declaration.

source/BannerlordCoopInstaller/Installer.cs
masesk: The status of the installer changes in the bottom left. The user is notified to close files and processes since the only exception that should occur when deleting files as privileged is IO exception. I'll change it to that.

source/BannerlordCoopInstaller/Installer.cs
masesk: Might be a cleaner way. Will investigate.

source/Coop.sln
masesk: Yup, VS 2022. I will revert the changes.

source/Launcher/Launcher.cs
masesk: Spawns a process to run MB, once it's done, the process is dead, it just needs to be closed so the resources can be freed up.

MetehanAydemir (commented):
wow

garrettluskey:
We have decided to not use a full installer, rather steam workshop or module export

PR #221, opened 2022-02-26
Improve UI debugging

### LINKED ISSUE #218: Create debug UI for shared object manager
We have the debug menu already, we should try to add in a window for our shared object manager with filters/duplication finders so we can easily debug what all is being tracked. 

We could then add features like clicking on a party in that debug list would center the camera on it. Would be helpful for verification purposes that all parties are truly synced between clients/server.

thomasDelaporte:
## PR Checklist

## PR Type

- [X] New feature (non-breaking change which adds functionality)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is new 
- Open/close (toggle) of the debugging UI through keys
- Visibility of all objects saved in the CoopObjectManager class

Resolves #218 
New Debug UI

### DIFF
--- a/source/Coop/Mod/DebugUtil/DebugUI.cs
+++ b/source/Coop/Mod/DebugUtil/DebugUI.cs
@@ -29,6 +29,7 @@ using TaleWorlds.Core;
 using TaleWorlds.Engine;
 using Registry = Sync.Registry;
 using Logger = NLog.Logger;
+using SandBox.View.Map;
 
 namespace Coop.Mod.DebugUtil
 {
@@ -36,6 +37,9 @@ namespace Coop.Mod.DebugUtil
     {
         private static readonly Logger Logger = LogManager.GetCurrentClassLogger();
         private readonly string m_WindowTitle = "Debug UI";
+
+        private HashSet<object> detailsObjects = new HashSet<object>();
+        private HashSet<object> toDeleteObjects = new HashSet<object>();
 
         public bool Visible { get; set; }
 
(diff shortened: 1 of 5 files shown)


garrettluskey:
Freezes on client for some reason. Please fix

source/Coop/Mod/DebugUtil/DebugUI.cs
```diff
@@ -49,11 +53,99 @@ public void Update(TimeSpan frameTime)
                 DisplayConnectionInfo();
                 DisplayHarmonyPatches();
                 DisplayPersistenceMenu();
+                DisplayCoopObjectManagers();
                 End();
+
+                foreach(object detailObject in this.detailsObjects)
```
garrettluskey: Move block to independant section

source/Coop/Mod/DebugUtil/DebugUI.cs
```diff
@@ -49,11 +53,99 @@ public void Update(TimeSpan frameTime)
...
+                    this.detailsObjects.Remove(toDeleteObject);
+                }
+
+                this.toDeleteObjects.Clear();
+                Imgui.EndMainThreadScope();
             }
         }
         public int Priority { get; } = UpdatePriority.MainLoop.DebugUI;
 
+        private void DisplayCoopObjectManagers()
```
garrettluskey, line 85: Add comments

garrettluskey (changes requested):
Fix on client and update code, then will be good.

thomasDelaporte:
@garrettluskey As we quickly discussed, the freeze is probably due to the gamedata and not the code quoted here.

I cleaned up the class as much as possible and added a lot of documentation for the whole class. 

The problem now is that the client has no data in CoopObjectManager so no way to check if it matches on both sides. The persistence => parties tab is also empty in the client.

garrettluskey (approved):
Nice

PR #231, opened 2022-03-06
Old xfer fix

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Initial transfer is working once again, there are still bugs with movement but those can be investigated after PR.

## What is the new behaviour?
Transfer works without crash.
Resolves (Dont have issue :))

## Other information


thomasDelaporte:
It seems to work as expected, I have the party "Joke" on the client and the server (visible in both cases). 

I think it's voluntary but the server doesn't have a MobileParty and it would be a good idea to look at it before switching to the movement system? (_Maybe it's because the server is configured as a dedicated server_).

It would also be nice to have a little explanation of what was done and why and the problem from the dev point of view.

thomasDelaporte (approved):
**It looks clean**, I think we should add documentation on methods after and I think there is way to reduce the code a bit with more regular returns and for example change the `out guidWrapper` to `out var guidWrapper` in CoopObjectManager.

The unit tests don't seem to have liked the change of project of some classes. To be seen when we will work on the tests.

PR #240, opened 2022-04-14
1.7.1 update

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Game is in 1.7.0

## What is the new behaviour?
Game is in 1.7.1

### DIFF
--- a/source/Coop/Mod/CoopServer.cs
+++ b/source/Coop/Mod/CoopServer.cs
@@ -168,20 +168,6 @@ namespace Coop.Mod
 #endif
         }
 
-        //public ServerGameManager CreateGameManager(LoadGameResult saveGameData = null)
-        //{
-        //    if (saveGameData != null)
-        //    {
-        //        gameManager = CreateGameManager(saveGameData.LoadResult);
-        //    }
-        //    else
-        //    {
-        //        gameManager = new ServerGameManager();
-        //    }
-
-        //    return gameManager;
-        //}
-
         public ServerGameManager CreateGameManager(LoadResult loadResult = null)
         {
             if (loadResult != null)
--- a/source/Coop/Mod/CoopServer.cs
+++ b/source/Coop/Mod/CoopServer.cs
@@ -157,9 +157,8 @@ namespace Coop.Mod
 #if DEBUG
             try
             {
-                LoadGameResult saveGameData = MBSaveLoad.LoadSaveGameData(
-                    saveName,
-                    Utilities.GetModulesNames());
+                LoadResult saveGameData = MBSaveLoad.LoadSaveGameData(
+                    saveName);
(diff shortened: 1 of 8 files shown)


source/Coop/Mod/CoopServer.cs
```diff
@@ -169,19 +168,19 @@ public void StartGame(string saveName)
...
-                gameManager = CreateGameManager(saveGameData.LoadResult);
-            }
-            else
-            {
-                gameManager = new ServerGameManager();
-            }
-
-            return gameManager;
-        }
+        //public ServerGameManager CreateGameManager(LoadGameResult saveGameData = null)
```
araex: I would delete it. Or do you expect this `LoadGameResult` to come back?

source/Coop/Mod/Extensions.cs
```diff
@@ -94,49 +94,48 @@ public static bool IsRemotePlayerMainParty(this MobileParty party)
...
-            string sRet = "Loading " + (loadResult.Successful ? "success. " : "failed. ");
-            if (loadResult.MetaData == null)
-            {
-                sRet += "No meta data";
-            }
-            else
-            {
-                sRet += loadResult.MetaData;
-            }
+        //public static string ToFriendlyString(this LoadResult loadResult)
```
araex: Delete

source/Coop/Mod/ObjectManagerAdapter.cs
```diff
@@ -226,14 +226,14 @@ static void Postfix(Clan __instance)
...
         }
-        [HarmonyPatch(typeof(Hero), "OnBeforeLoad")]
-        private static class Hero_OnBeforeLoad
-        {
-            static void Postfix(Hero __instance)
-            {
-                Instance.OnAfterRegisterObjectWithoutInitialization(__instance);
-            }
-        }
+        //[HarmonyPatch(typeof(Hero), "OnBeforeLoad")] Not used, replacement?
```
araex: From a quick look they progressed a bit further with their init logic refactoring. The introduced a new attribute `[LoadInitializationCallback]`. All methods with that attribute are called called in `LoadResult.InitializeObjects`.  `MBObjectBase.OnBeforeLoad` is part of that initialization. So i guess the whole workaround for all these `OnBeforeLoad` is now redundant because they fixed it themselves.

With current development, this is unused anyways. So i think it's fine to just delete and have a look again if we need it.

araex (approved):
Please delete commented out code before merging. I had to create a new "Sandbox" save game for it to work. With the old save game it crashed on load due to some game internal issues with a quest. Maybe they broke compatibility with old saves? Regardless, doesn't matter for us at this point.

Otherwise looks good and works on my machine. Thanks!

PR #255, opened 2022-06-07
Patching development branch to 1.7.2 game version

WakooMan:
## PR Checklist

## PR Type

- [x] Now the mod works with 1.7.2 game version
- [x] Most likely imports/usings changed in the mod (about 75%).
- [x] Some serializers needed to be changed to work correctly.
- [x] Added some suggestions in comment for removing empty cases in serializers in one of the serializer file.
- [x] Town Sync is almost done, but not completely, so it's commented.

The mod works with 1.7.1 game version.

## What is the new behaviour?

The mod works with 1.7.2 game version.
Resolves #

## Other information


garrettluskey:
Town sync should be separate from the 1.7.2 update. You can base a "TownSync" branch off of this one to work with 1.7.2, but since these are different features they should be in separate branches.

source/Coop/Mod/GameSync/Initializer.cs
```diff
@@ -29,6 +31,11 @@ public static void SetupSyncAfterLoad()
                 FiefSync.MakeManaged(town);
             }
 
+            //foreach (Town town in Town.AllTowns)
```
garrettluskey: Move this commented code to a different branch

source/Coop/Mod/GameSync/Town/TownCommands.cs
```diff
@@ -0,0 +1,296 @@
+﻿//using System;
```
garrettluskey: Move this code to a different branch

source/Coop/Mod/GameSync/Town/TownSync.cs
```diff
@@ -0,0 +1,103 @@
+﻿//using CoopFramework;
```
garrettluskey: Move this code to a separate branch

garrettluskey (changes requested):

source/Coop/Mod/GameSync/Initializer.cs
WakooMan: Done

source/Coop/Mod/GameSync/Town/TownCommands.cs
WakooMan: Done

source/Coop/Mod/GameSync/Town/TownSync.cs
WakooMan: Done

garrettluskey (approved):

PR #258, opened 2022-06-15
Automatic Sync initializing bugfix (Based on 1.7.2 patch PR)

### LINKED ISSUE #254: Fix Time Sync
With the changes to the coop framework, time control is no longer syncing. Investigate and fix this issue.

WakooMan:
## PR Checklist

## PR Type

- TimeControl sync does not work.
- CoopManaged child classes were not setup automatically, because constructors weren't called at deserialization.

## What is the new behaviour?

- Deserialized objects are now registered by CoopFramework, when they are loaded.
- TimeControl sync now works and TimeCommands written for testing.
## Resolves

- TimeControl bug
- Sync Automatic Setup bugfix

## Other information


source/CoopFramework/ObjectLifetimeObserver.cs
```diff
@@ -61,14 +64,46 @@ public bool PatchConstruction()
...
+        ///     in the hierarchy, that has this attribute./>.
+        /// </summary>
+        private void PatchLoadInitializationCallbacks()
+        {
+            Type current = typeof(T);
+            m_LoadInitializationPatch = new List<MethodPatch<ObjectLifetimeObserver<T>>>();
+            while (current != typeof(object))
+            {
+                    var patch = new MethodPatch<ObjectLifetimeObserver<T>>(current);
+                    var methods = current.GetMethods(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public).Where(m => m.GetCustomAttributes(typeof(LoadInitializationCallback), false).Any()).ToList();
```
garrettluskey: This should also get parent class methods, you should be able to remove the while loop and `current = current.BaseType`

garrettluskey (changes requested):

source/CoopFramework/ObjectLifetimeObserver.cs
WakooMan: Removed the while loop, but changed the whole thing. Please take a look at it. Also Movement sync does not work somehow, but I don't know if it was working before.

source/CoopFramework/CoopFramework.cs
```diff
@@ -38,6 +42,16 @@ public static void InitPatches([CanBeNull] IObjectManager objectManager, Func<bo
             }
 
             var loadedAssemblies = AppDomain.CurrentDomain.GetAssemblies();
+            loadedAssemblies.ToList().ForEach(a =>
```
garrettluskey: Split these into helper functions. Too many anonymous methods

source/CoopFramework/ObjectLifetimeObserver.cs
```diff
@@ -58,17 +62,43 @@ public void AfterUnregisterObject(object removedObject)
...
         }
+        /// <summary>
+        ///     Patches the first method with <see cref="LoadInitializationCallback"/> LoadInitializationCallback attribute, but if none is found, then we check 
+        ///     if T is subclass of MBObjectBase and patch it's first method with the given attribute.. 
+        ///     This method is needed because constructors are not called at some classes, when loading a saved game, but for these classes there is a method
+        ///     in the hierarchy, that has this attribute./>.
+        /// </summary>
+        private void PatchLoadInitializationCallbacks()
+        {
+            Type type = CoopFramework.LoadInitializationCallbacks.Keys.ToList().Find(t => typeof(T) == t);
```
garrettluskey: Try this, you can remove one of the null checks then as well assuming it works as intended

```suggestion
            Type type = CoopFramework.LoadInitializationCallbacks.Keys.ToList().Find(t => t.IsAssignableFrom(typeof(T));
```

garrettluskey (changes requested):

source/CoopFramework/ObjectLifetimeObserver.cs
WakooMan: Done

source/CoopFramework/CoopFramework.cs
WakooMan: foreach cycles are no longer anonymous methods and added one local variable for BindingFlags, so it is easier to read in my opinion.

garrettluskey (approved):

PR #259, opened 2022-06-20
CoopFramework Synchronization Trigger Change bugfix (Based on Automatic Sync Initialization PR)

### LINKED ISSUE #256: CoopFramework Synchronization Trigger Change
Currently syncronization broadcast is called on skipped functions. Change this so that sync is only broadcasted when the method is called.

See 
CallBehaviourBuilder Broadcast method
CoopManaged Dispatch method

WakooMan:
## PR Checklist

## PR Type

- When the function is skipped the call is still broadcasted.

## What is the new behaviour?

- When the function is skipped, the call is not broadcasted further.

## Question For Future Problems Probably
- In Dispatch method if one behaviour skips the method, but the list has a behaviour that doesn't skip the method, than should the method be broadcasted?
## Resolves

- CoopFramework Synchronization Trigger Change issue.

## Other information


source/CoopFramework/CoopManaged.cs
```diff
@@ -260,20 +260,25 @@ private static void OnAutoConstructed(TSelf newInstance)
             foreach (var behaviour in behaviours)
             {
                 if (!behaviour.AppliesTo(eOriginator, instanceResolved)) continue;
-                if (behaviour.DoBroadcast)
+                ECallPropagation prop = ECallPropagation.CallOriginal;
+                if (behaviour.MethodCallHandler != null)
+                { 
+                    prop = behaviour.MethodCallHandler.Invoke(new PendingMethodCall(invokableId,
```
garrettluskey:
```suggestion
                    prop = behaviour.MethodCallHandler?.Invoke(new PendingMethodCall(invokableId,
```

source/CoopFramework/CoopManaged.cs
```diff
@@ -260,20 +260,25 @@ private static void OnAutoConstructed(TSelf newInstance)
...
+                        behaviour.SynchronizationFactory,
+                        instanceResolved,
+                        args));
```
garrettluskey:
```suggestion
                        args)) ?? prop; 
```

garrettluskey (changes requested):

source/CoopFramework/CoopManaged.cs
WakooMan: Done

source/CoopFramework/CoopManaged.cs
WakooMan: Done

garrettluskey:
Refactoring entire sync framework. Changes no longer required

PR #277, opened 2022-08-11
The Refactoring (Small)

garrettluskey:
No PR Checklist required here, simply updating architecture and no lingering files remain.


source/Common/Components/ComponentContainerBase.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public bool AddComponent<T>(T component) where T : IComponent
+        {
+            if (_components.ContainsKey(component.GetType())) return false;
+
+            _components.Add(component.GetType(), component);
+            return true;
+        }
+
+        private Type GetInterface(Type t)
```
Allen-Glass, line 22: Not seeing any other uses of GetInterface in this class

source/Common/Components/ComponentContainerBase.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Common.Components
+{
+    public abstract class ComponentContainerBase : IComponentContainer
+    {
+        public IEnumerable<IComponent> Components => _components.Values;
+        private readonly Dictionary<Type, IComponent> _components = new Dictionary<Type, IComponent>();
+
+        public bool AddComponent<T>(T component) where T : IComponent
+        {
+            if (_components.ContainsKey(component.GetType())) return false;
```
Allen-Glass, line 16: You can make this simply `return _components.TryAdd()`

Allen-Glass:
I had to delete Coop.Tests obj folder but other than that it builds correctly.

Allen-Glass (approved):
Didn't dig through all the files but noticed two things on the first file that did pop up. I may dig through the rest later

PR #460, opened 2023-01-03
Added missing opponent detection

### LINKED ISSUE #433: Fix Board Games
# For New Contributors

## Description

Add robustness for forfeiting when other player crashes.

Create some logging for packet sending and receiving (maybe in a file).

If needed @garrettluskey can add server logging.

It seems that message sending or receiving is now broken.

## Intended Design

N/A

## Location

N/A

## Related Issues

N/A

## Requirements

N/A

## Additional information

N/A

## Definition of Done

### LINKED ISSUE #453: Update Boardgames to use new MessageBroker and Event arch
# For New Contributors

## Description

Network events are now sent through LiteNetP2PClient and still received via MessageBroker.

All network messages should now extend from INetworkMessage and sent via P2PClient.

Update all events in BoardGamePatches to use the MessageBroker instead.

## Intended Design

N/A

## Location

N/A

## Related Issues

Blocked by
- #451

## Requirements

N/A

## Additional information

N/A

## Definition of Done

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Players is in limbo until they surrender when opponent has left/crashed during board game.

## What is the new behaviour?

Resolves #433 

Player wins by forfeit when no opponent is detected, runs on a three second timer.

## Other information

### DIFF
--- a/source/Missions/BoardGameLogic.cs
+++ b/source/Missions/BoardGameLogic.cs
@@ -46,13 +46,13 @@ namespace Coop.Mod.Missions
             GameId = gameId;
 
             //Internal Messages
-            _messageBroker.Subscribe<StopConvoAfterGameMessage>(Handle_OnGameOver);
+            _messageBroker.Subscribe<StopConvoAfterGameMessage>(OnGameOver);
             _messageBroker.Subscribe<BoardGameMoveMessage>(OnPlayerInput);
-            _messageBroker.Subscribe<AgentDeleted>(Handle_AgentDeleted);
+            _messageBroker.Subscribe<AgentDeleted>(OnAgentDeleted);
             _messageBroker.Subscribe<OnForfeitMessage>(OnForfeitGame);
-            _messageBroker.Subscribe<OnHandlePreMovementStageMessage>(PreMovementStage);
+            _messageBroker.Subscribe<OnHandlePreMovementStageMessage>(OnPreMovementStage);
             _messageBroker.Subscribe<OnSetPawnCapturedMessage>(OnPawnCapture);
-            _messageBroker.Subscribe<PreplaceUnitsSeegaMessage>(PreplaceUnits);
+            _messageBroker.Subscribe<PreplaceUnitsSeegaMessage>(OnPreplaceUnits);
 
             //External Messages
             _messageBroker.Subscribe<ForfeitGameMessage>(Handle_ForfeitGameMessage);
@@ -61,7 +61,7 @@ namespace Coop.Mod.Missions
 
         }
 
-        private void Handle_AgentDeleted(MessagePayload<AgentDeleted> payload)
+        private void OnAgentDeleted(MessagePayload<AgentDeleted> payload)
         {
             if(payload.What.Agent.Equals(_boardGameLogic.OpposingAgent))
             {
@@ -84,13 +84,13 @@ namespace Coop.Mod.Missions
             _messageBroker.Unsubscribe<PawnCapturedMessage>(Handle_PawnCapture);
             _messageBroker.Unsubscribe<BoardGameMoveRequest>(Handle_MoveRequest);
 
-            _messageBroker.Unsubscribe<StopConvoAfterGameMessage>(Handle_OnGameOver);
+            _messageBroker.Unsubscribe<StopConvoAfterGameMessage>(OnGameOver);
             _messageBroker.Unsubscribe<BoardGameMoveMessage>(OnPlayerInput);
-            _messageBroker.Unsubscribe<AgentDeleted>(Handle_AgentDeleted);
+            _messageBroker.Unsubscribe<AgentDeleted>(OnAgentDeleted);
             _messageBroker.Unsubscribe<OnForfeitMessage>(OnForfeitGame);
-            _messageBroker.Unsubscribe<OnHandlePreMovementStageMessage>(PreMovementStage);
+            _messageBroker.Unsubscribe<OnHandlePreMovementStageMessage>(OnPreMovementStage);
             _messageBroker.Unsubscribe<OnSetPawnCapturedMessage>(OnPawnCapture);
-            _messageBroker.Unsubscribe<PreplaceUnitsSeegaMessage>(PreplaceUnits);
+            _messageBroker.Unsubscribe<PreplaceUnitsSeegaMessage>(OnPreplaceUnits);
 
         }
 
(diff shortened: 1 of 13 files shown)


source/Missions/BoardGameLogic.cs
```diff
@@ -83,8 +85,26 @@ public void StartGame(bool startFirst, Agent opposingAgent)
...
+        Task CheckOpponentConnected()
+        {
+            if (!m_BoardGameLogic.OpposingAgent.IsNetworkAgent())
+            {
+                m_BoardGameLogic.AIForfeitGame();
+                MBInformationManager.AddQuickInformation(new TextObject("You won! Your opponent has left the game"));
+                Dispose();
+            }
+
+            return Task.CompletedTask;
```
smudge202: If this doesn't need to be async, just make it void?

source/Missions/BoardGameLogic.cs
```diff
@@ -83,8 +85,26 @@ public void StartGame(bool startFirst, Agent opposingAgent)
             m_BoardGameLogic.SetBoardGame(m_BoardGameType);
             m_BoardGameLogic.SetStartingPlayer(startFirst);
             m_BoardGameLogic.StartBoardGame();
+
+            //Opponent crash detection
+            while(IsPlayingOtherPlayer)
```
smudge202: This will sit in a loop forever, even after the player has won? (`IsPlayingOtherPlayer` isn't being set to false?)

source/Missions/BoardGameLogic.cs
```diff
@@ -73,7 +75,7 @@ public void Dispose()
 
         private static readonly PropertyInfo OpposingAgentPropertyInfo = typeof(MissionBoardGameLogic).GetProperty("OpposingAgent");
 
-        public void StartGame(bool startFirst, Agent opposingAgent)
+        public async void StartGame(bool startFirst, Agent opposingAgent)
```
smudge202: async void is really bad and should be avoided at all costs.
Looks like the reason it's here is to call a `CheckOpponentConnected` which per comment below, doesn't need to be async, and to call `Task.Delay` which can probably be replaced by `Thread.Sleep`?

source/Missions/BoardGameLogic.cs
```diff
@@ -83,8 +85,26 @@ public void StartGame(bool startFirst, Agent opposingAgent)
...
+
+            //Opponent crash detection
+            while(IsPlayingOtherPlayer)
+            {
+                await Task.Delay(3000);
+                await CheckOpponentConnected();
+            }
         }
-        
+        Task CheckOpponentConnected()
```
smudge202: nit: spacing between methods and explicit modifiers :)

smudge202 (changes requested):

source/Missions/BoardGameLogic.cs
EgardA: It is being set to false, do I know how? no clue.

EgardA: Or actually, maybe not. I don't remember how the BoardGameLogic works entirely. I believe the BoardGameLogic gets removed when the board game is over, wouldn't this break the loop?

smudge202: If there's a thread still churning round in that loop, no, it'll effectively memory leak (admittedly a small leak), though worse I guess, it'd hold a thread and the .net runtime only allocates a limited number of those to the app domain

smudge202: Actually, looking at it again... I think it results in `Dispose` being called multiple times (or infinite times, not certain). Because you're inside an async void the method is not awaited anywhere, so the method probably blows up, but again, because it's an async void, the exception has nowhere to propagate to. I could pull your branch down to debug it but I suspect _something_ will be doing ugly things right now.

smudge202: Oh, the `Dispose` method is "safe" to repeat I think; nothing in there will blow up if you call it twice as best I can tell? So yeh, you probably just leave a thread sat looping around inside that `while(IsPlayingOtherPlayer)` loop.

garrettluskey: This can be turned into an event after #465 is merged by subscribing to the PeerDisconnected event

EgardA: Yep, will update after #465

garrettluskey (dismissed):

source/Missions/BoardGameLogic.cs
```diff
@@ -44,17 +45,30 @@ public class BoardGameLogic : IDisposable
             _boardGameType = gameType;
             GameId = gameId;
 
+            //Internal Messages
+            _messageBroker.Subscribe<StopConvoAfterGameMessage>(Handle_OnGameOver);
```
smudge202: nit: should we consolidate these delegate names too? So either all `OnFoo` or `Handle_Foo`? (I prefer the former for what it's worth)

source/Missions/Messages/BoardGames/BoardGameMoveMessage.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Common.Messaging;
+using SandBox.BoardGames;
+using TaleWorlds.MountAndBlade;
+
+namespace Missions.Messages.Agents
+{
+    public readonly struct BoardGameMoveMessage : IEvent
+    {
+        public Move move { get; }
```
smudge202: nit: casing (Should be `Move` for a property)
This in turn lets you change `inMove` in ctor to `move`

source/Missions/Messages/BoardGames/OnSetPawnCapturedMessage.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Common.Messaging;
+using SandBox.BoardGames.MissionLogics;
+using SandBox.BoardGames.Pawns;
+using TaleWorlds.MountAndBlade;
+
+namespace Missions.Messages.Agents
+{
+    public readonly struct OnSetPawnCapturedMessage : IEvent
+    {
+        public PawnBase pawn { get; }
```
smudge202: nit: As per previous comment, correct casing here, kill the `_` on ctor param)

source/Missions/BoardGameLogic.cs
smudge202: Resolving as outdated

source/Missions/BoardGameLogic.cs
smudge202: Resolving as outdated

smudge202 (dismissed):

smudge202 (approved):

PR #508, opened 2023-02-16
Arena character selection

masesk:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Loading into arena picks the first save and assumes the correct modules are there.

## What is the new behaviour?
A save selection appears and the user can select the character to use. 
The modules are checked for non-validated modules as well as the needed modules.

Resolves #

## Other information


source/MissionTestMod/TestMod.cs
```diff
@@ -72,7 +75,16 @@ protected override void OnSubModuleLoad()
               "Join Online Tavern",
               new TextObject("Join Online Tavern"),
               9991,
-              StartClientInTavern,
+              () =>
```
garrettluskey: Move this into it's own function and this is under "Join Tavern". If you could make it work for both

source/Missions/Services/Arena/View/ArenaLoadUI.cs
```diff
@@ -0,0 +1,221 @@
...
+using TaleWorlds.MountAndBlade.View;
+using TaleWorlds.SaveSystem;
+using TaleWorlds.SaveSystem.Load;
+using TaleWorlds.ScreenSystem;
+using TaleWorlds.TwoDimension;
+
+namespace Missions.Services.Arena.View
+{
+
+    class ArenaSelectedGameVM : SavedGameVM
```
garrettluskey: Could you add class level comments to all these classes

garrettluskey (changes requested):

source/MissionTestMod/TestMod.cs
```diff
@@ -117,54 +121,27 @@ protected override void OnApplicationTick(float dt)
...
-                InformationManager.DisplayMessage(new InformationMessage("Unable to find a save without Mods. Create a fresh game and try again."));
-            }
-            else
-            {
-                SandBoxSaveHelper.TryLoadSave(save, StartGameArena, null);
-            }
-        }
-
-        private static readonly HashSet<string> allowedModules = new HashSet<string>()
+        private static void SelectCharacterTavern()
```
garrettluskey: Rename this to `SelectSaveTavern` to better reflect what this method is doing

source/MissionTestMod/TestMod.cs
```diff
@@ -117,54 +121,27 @@ protected override void OnApplicationTick(float dt)
             Updateables.UpdateAll(frameTime);
         }
 
-        private static void StartClientInTavern()
+        private static void SelectCharacterArena()
```
garrettluskey: Rename this to `SelectSaveArena` to better reflect what this method is doing

source/Missions/View/MissionsLoadUI.cs
```diff
@@ -0,0 +1,321 @@
...
+using TaleWorlds.SaveSystem.Load;
+using TaleWorlds.ScreenSystem;
+using TaleWorlds.TwoDimension;
+
+namespace Missions.View
+{
+	/// <summary>
+	/// VM for the game selection for missions
+	/// </summary>
+    class MissionSelectedGameVM : SavedGameVM
```
garrettluskey, line 24: Fix tabs in this file

garrettluskey (changes requested):

garrettluskey (approved):

PR #630, opened 2023-06-10
Party Behavior Optimizations

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Party behavior was going through events over the network rather than packets.

## What is the new behaviour?

Party behavior is now sent directly through a packet rather than events increasing the performance of movement.

## Other information


deploy.ps1
```diff
@@ -18,6 +18,7 @@ $TemplateDir            = "${BaseDirWithoutQuotes}template"
 $UIMovieDir             = "${BaseDirWithoutQuotes}UIMovies"
 
 # create output directory structure
+Remove-Item $DeployDir -Recurse
```
garrettluskey, line 21: removes the temp deploy directory for any lingering files. I noticed that this would sometimes be filled with old dlls

source/GameInterface/Services/GameState/Messages/CampaignReady.cs
```diff
@@ -2,6 +2,6 @@
 
 namespace GameInterface.Services.GameState.Messages;
 
-public record CampaignLoaded : IEvent
+public record CampaignReady : IEvent
```
garrettluskey, line 5: Renamed for better representation of what this event is

source/GameInterface/Services/GameState/Patches/GameLoadedPatch.cs
```diff
@@ -5,18 +5,19 @@
...
 
-[HarmonyPatch(typeof(MapScreen))]
+[HarmonyPatch(typeof(Campaign))]
 internal class GameLoadedPatch
 {
     private static readonly ILogger Logger = LogManager.GetLogger<SandBoxGameManager>();
 
     [HarmonyPostfix]
-    [HarmonyPatch("OnActivate")]
+    [HarmonyPatch("OnSessionStart")]
```
garrettluskey, line 18: OnActivate was happening too early sometimes resulting in crashes. This hook should be more appropriate for the campaign start

orfolei (approved):
Looks good

PR #707, opened 2023-12-25
Packet Amount Optimizations on Clan Requests

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

A lot of clan influence changes are sent

## What is the new behaviour?

Resolves #

Less clan influence changes are sent (empty clan influence changes are not sent)

## Other information

### DIFF
--- a/source/GameInterface/Services/Clans/Patches/ClanAddRenownPatch.cs
+++ b/source/GameInterface/Services/Clans/Patches/ClanAddRenownPatch.cs
@@ -18,14 +18,14 @@ namespace GameInterface.Services.Clans.Patches
         {
             if (value == 0f) return false;
 
+            if (AllowedInstance.IsAllowed(__instance)) return true;
+
             if (PolicyProvider.AllowOriginalCalls) return true;
 
             // On the client if it is not the player client skip the call
             if (ModInformation.IsClient && __instance != Clan.PlayerClan) return false;
 
             CallStackValidator.Validate(__instance, AllowedInstance);
-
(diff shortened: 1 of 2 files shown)


source/GameInterface/Services/Clans/Patches/ClanChangeInfluencePatch.cs
```diff
@@ -17,8 +17,9 @@ public class ClanChangeInfluencePatch
 
         static bool Prefix(Clan clan, float amount)
         {
-            if (PolicyProvider.AllowOriginalCalls) return true;
+            if (amount == 0f) return false;
```
garrettluskey, line 22: move `if (AllowedInstance.IsAllowed(clan)) return true;` after this line. if we want to allow this call then it should be one of the first checks

source/GameInterface/Services/Clans/Patches/ClanAddRenownPatch.cs
```diff
@@ -16,6 +16,8 @@ public class ClanAddRenownPatch
 
         static bool Prefix(ref Clan __instance, float value, bool shouldNotify = true)
         {
+            if (value == 0f) return false;
```
garrettluskey, line 19: move `if (AllowedInstance.IsAllowed(clan)) return true;` after this line. if we want to allow this call then it should be one of the first checks

garrettluskey (changes requested):

garrettluskey (approved):

PR #758, opened 2024-02-08
Tranpiler for town daily tick

### LINKED ISSUE #757: [Report] Fief does not sync correctly
## Severity of issue/bug

- [ ] High
- [x] Average
- [ ] Low

``` [HarmonyPatch(typeof(Fief),"FoodStocks",MethodType.Setter)]``` are not called by the game when a food stock change
## Steps to reproduce

Let the game run for some seconds, when a fief food change should change (it change every seconds on the map). The prefix patch is not trigger

### PR DESCRIPTION
**Pull request by thomas-soutif**
handle the case of the foodStock of a fief
I should modify it because the food stack value is not exactly the one we want, but Im' in good progress

## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behaviour?

Resolves #757 

## Other information

### DIFF
--- a/source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
+++ b/source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
@@ -29,7 +29,7 @@ public static class TownDailyTickPatch
         
         foreach (var instruction in instrs)
         {
-            if (instruction.opcode == Call.opcode && instruction.operand as MethodInfo == foodStocksSetter)
+            if (instruction.opcode == Call.opcode && instruction.operand as MethodInfo == AccessTools.PropertySetter(typeof(Fief), "FoodStocks"))
             {
                 
                 yield return new CodeInstruction(OpCodes.Call, AccessTools.Method(typeof(TownDailyTickPatch), "InterceptSetFoodStock"));
--- a/source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
+++ b/source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
@@ -1,0 +1,51 @@
+﻿using HarmonyLib;
+using System.Collections.Generic;
+using System.Reflection.Emit;
+using System.Reflection;
+using System;
+using TaleWorlds.CampaignSystem.Settlements;
+using TaleWorlds.CampaignSystem.Encounters;
+using System.Linq;
+using static HarmonyLib.Code;
+using Common.Messaging;
+using GameInterface.Services.Fiefs.Messages;
+using Newtonsoft.Json.Linq;
+using Autofac.Core.Activators;
+using GameInterface;
+
+[HarmonyPatch(typeof(Town), "DailyTick")]
+public static class TownDailyTickPatch
+{
+    // Get the PropertyInfo for the FoodStocks property
+    private static PropertyInfo foodStocksProperty = typeof(Fief).GetProperty("FoodStocks");
+
+    // Get the setter method for the FoodStocks property
+    private static MethodInfo foodStocksSetter = foodStocksProperty.GetSetMethod(true);
+
+    [HarmonyPatch(typeof(Town), "DailyTick")]
+    [HarmonyTranspiler]
+    private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instrs)
+    {
+        List<CodeInstruction> instructions = instrs.ToList();
+        int setterLastIdx = instructions.FindLastIndex(i => i.opcode == Call.opcode && i.operand as MethodInfo == foodStocksSetter);
+        if (setterLastIdx == -1) return instrs;
+
+        instructions.InsertRange(setterLastIdx, new CodeInstruction[]
+        {
+            new CodeInstruction(OpCodes.Ldarg_0),
+            new CodeInstruction(OpCodes.Call, AccessTools.Method(typeof(TownDailyTickPatch), "LogFoodStock"))
+        });
+        
+        return instructions;
+        
+    }
+
+    public static void LogFoodStock(Fief fief)
+    {
+        // If it's the client, return
+        if (!ModInformation.IsServer) return;
+        
+        var message = new FiefFoodStockChanged(fief.StringId, fief.FoodStocks);
+        MessageBroker.Instance.Publish(fief, message);
+    }
+}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by thomas-soutif** [comment]
still investigating

File: source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+        int setterLastIdx = instructions.FindLastIndex(i => i.opcode == Call.opcode && i.operand as MethodInfo == foodStocksSetter);
+        if (setterLastIdx == -1) return instrs;
+
+        instructions.InsertRange(setterLastIdx, new CodeInstruction[]
+        {
+            new CodeInstruction(OpCodes.Ldarg_0),
+            new CodeInstruction(OpCodes.Call, AccessTools.Method(typeof(TownDailyTickPatch), "LogFoodStock"))
+        });
+        
+        return instructions;
```
garrettluskey: Currently this will only patch the last FoodStocks set

This also does not block setting on the client

Below will replace the call with your method

The stack will also contain the new food stocks value so you will have to set it yourself for more control

```suggestion
        foreach (var instriction in instrs)
{
    if (i.opcode == Call.opcode && i.operand as MethodInfo == foodStocksSetter)
    {
        yield return new CodeInstruction(OpCodes.Ldarg_0);
        yield return new CodeInstruction(OpCodes.Call, AccessTools.Method(typeof(TownDailyTickPatch), "LogFoodStock"));
        continue;
    }

    yield return instriction
}
```

source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
```diff
@@ -0,0 +1,51 @@
...
+        
+    }
+
+    public static void LogFoodStock(Fief fief)
```
garrettluskey: Update to match new patch

```suggestion
    public static void LogFoodStock(float newValue, Fief fief)
```

source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        // If it's the client, return
+        if (!ModInformation.IsServer) return;
```
garrettluskey: No '!' is less confusing at a glance

```suggestion
        if (ModInformation.IsClient) return;
```

source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
```diff
@@ -0,0 +1,51 @@
...
+        
+        var message = new FiefFoodStockChanged(fief.StringId, fief.FoodStocks);
```
garrettluskey:
```suggestion
        fief.FoodStocks = newValue;
        var message = new FiefFoodStockChanged(fief.StringId, fief.FoodStocks);
```

garrettluskey (changes requested):

source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
thomas-soutif: When I remplace with this code, I have this crash :

image

thomas-soutif: The crash seem to come from the yield return

source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
thomas-soutif: Because we get from the stack the Ldarg_0, that represent the "fief" instance , putting a float newValue make the program crash
we should found the value return by the previous stack (get_foodStack) in the dailyTick() method
image

source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
thomas-soutif: So I was able to make it work without crash by removing the new code insruction(OpCodes.Ldarg0)
the crash occurs because we add two new code instruction for one instruction replace
but now I have this as a result : 
image
the fief instance return here for exemple is a list of "village", maybe because 3 villages have the same food stock in same time ?

thomas-soutif: Finally make it works !
```
    foreach (var instruction in instrs)
    {
        if (instruction.opcode == Call.opcode && instruction.operand as MethodInfo == foodStocksSetter)
        {
	    yield return new CodeInstruction(OpCodes.Call, AccessTools.Method(typeof(TownDailyTickPatch), "LogFoodStock"));
            //yield return new CodeInstruction(OpCodes.Ldarg_0);
            continue;
        }
        yield return instruction;
    }
    
}

public static void LogFoodStock(Fief fief, float value)
```
so I remove the Ldarg_0 and change the order of the parameter that take LogFoodStock, the problem was there

thomas-soutif:
Bug fixed, ready for review

source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
```diff
@@ -0,0 +1,50 @@
...
+    // Get the setter method for the FoodStocks property
+    private static MethodInfo foodStocksSetter = foodStocksProperty.GetSetMethod(true);
+
+    [HarmonyTranspiler]
+    private static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instrs)
+    {
+        
+        foreach (var instruction in instrs)
+        {
+            if (instruction.opcode == Call.opcode && instruction.operand as MethodInfo == foodStocksSetter)
```
garrettluskey:
```suggestion
            if (instruction.opcode == Call.opcode && instruction.operand as MethodInfo == AccessTools.PropertySetter(typeof(Fief), "FoodStocks"))
```

source/GameInterface/Services/Towns/Patches/DailyTickTownPatch.cs
```diff
@@ -0,0 +1,50 @@
...
+using GameInterface.Services.Fiefs.Patches;
+
+[HarmonyPatch(typeof(Town), "DailyTick")]
+public static class TownDailyTickPatch
+{
+    // Get the PropertyInfo for the FoodStocks property
+    private static PropertyInfo foodStocksProperty = typeof(Fief).GetProperty("FoodStocks");
+
+    // Get the setter method for the FoodStocks property
+    private static MethodInfo foodStocksSetter = foodStocksProperty.GetSetMethod(true);
```
garrettluskey: This data is stored the entire lifetime of the program, if they are not used outside the transpiler use the "=>" operator or move it into the transpiler like suggested
```suggestion
```

garrettluskey (changes requested):

garrettluskey (approved):

PR #988, opened 2024-10-07
936 sync workshop fields

### LINKED ISSUE #936: Sync: Workshop Fields
## Description

1. Sync the values in the **Synced** list below.
2. Create an end to end (E2E) tests for all fields (1 for server changing the value and 1 for client changing the value if applicable)
3. Create a command to test changing each value in game [see similar command](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/blob/development/source/GameInterface/Services/Towns/Commands/TownDebugCommand.cs#L455)

Also see additional information at the bottom for more information and templates

**Key**
Server side only (only allow the running of the function on the server side)
Client side only (only allow the running of the function on the client side)
Client side sync (request the server to change the value, server then changes the value and replies to the client allowing the change)
Server side sync (only server allows running of the function and send to all clients the value changed and clients update the value/call original function)

**Synced**

| Method/Field Name | Sync Type | Additional Information |
|:-----------------|:---------:|:--------------|
| _customName | Server side sync | - |
| _owner | Server side sync | - |
| _settlement | Server side sync | - |
| _tag | Server side sync | - |

**Deferred**

None

**Externally Deferred**

None

**Non-Synced**

None

## Intended Design

For the patch itself please refer to https://github.com/Bannerlord-Coop-Team/BannerlordCoop/wiki/Patching-Basics#field-patching

**Field Sync Dataflow**
FieldSyncDataflow
**Integration Testing**
image
## Location

## Related Issues

Blocked by:
- #935

## Requirements

N/A

## Additional information

Using the [GameInterface Service Tempale](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/tree/development/source/GameInterface/Services/Template) create commands to test the synced values.

## Definition of Done
- [x] XUnit E2E tests exist for every property on the server (and client if applicable)
- [ ] Each field has a command getter and setter that used the correct intercept

spk-berthel:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

https://github.com/Bannerlord-Coop-Team/BannerlordCoop/issues/936

## What is the new behavior?
Syncs the fields `_owner` `_customName` `_tag` and `_settlement`

Resolves #936

## Other information
Added commands for the in-game console for testing.
Removed unsused `using` statements

The changes to `WorkshopTypeBuilder` and creation of the `ObjectBuilder` for `ItemCategory` was initially made for when patching the function `CangeOwnerOfWorkshop`.  But thinking ahead, we might have a use for it in the future.


source/GameInterface/Services/Workshops/Patches/WorkshopPropertyPatches.cs
```diff
@@ -114,5 +118,77 @@ private static bool SetInitialCapitalPrefix(Workshop __instance, int value)
...
+            {
+                Logger.Error("Client tried to set {name}\nCallstack: {callstack}", nameof(Workshop.Tag), Environment.StackTrace);
+                return false;
+            }
+
+            var message = new WorkshopPropertyChanged(PropertyType.Tag, __instance, value);
+            MessageBroker.Instance.Publish(__instance, message);
+
+            return ModInformation.IsServer;
+        }
```
georgyrudnev: You can remove this I believe since these fields are only set in the constructor and there are no seperate setter methods for those

source/GameInterface/Services/Workshops/Patches/WorkshopPropertyPatches.cs
```diff
@@ -114,5 +118,77 @@ private static bool SetInitialCapitalPrefix(Workshop __instance, int value)
 
             return ModInformation.IsServer;
         }
+
+        [HarmonyPatch(nameof(Workshop.SetCustomName), MethodType.Normal)]
+        [HarmonyPrefix]
+        private static bool SetCustomNamePrefix(Workshop __instance, TaleWorlds.Localization.TextObject value)
```
georgyrudnev: Instead of TaleWorlds.Localization.TextObject value you need to use the exact parameter name for this method in this case should be TaleWorlds.Localization.TextObject customName

georgyrudnev (changes requested):

source/GameInterface/Services/Workshops/Patches/WorkshopPropertyPatches.cs
spk-berthel: Good catch. I must have hit ctrl z a time too many

spk-berthel: Good Catch.

source/GameInterface/Services/Workshops/Patches/WorkshopPropertyPatches.cs
spk-berthel: I might have misunderstood something, but don't we need to use the `MessageBroker` here to sync these values? Because i was initially told to just sync all four fields.

If that is the case, would we still not need to sync when the workshop owner changes? There is a setter method for it called `ChangeOwnerOfWorkshop`

EgardA:
We want to patch the setters directly, instead of methods that set it, mainly in case they change method name (they do that seemingly for no reason sometimes) and if they add other methods in the future that changes it.

source/GameInterface/Services/Workshops/Patches/WorkshopPropertyPatches.cs
EgardA, line 1: Please revert this to how you had it before, patching the setters instead of the methods :) It's better for maintenance later on and will hopefully lead to less headache.

EgardA (approved):
Looks good to me!

garrettluskey (approved):

PR #1102, opened 2025-03-01
Test for other approach on how events are processed

Gerseras:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #


source/GameInterface/Services/BesiegerCamps/Messages/Collection/BesiegerPartyRemoved.cs
```diff
@@ -7,9 +9,14 @@ namespace GameInterface.Services.BesiegerCamps.Messages;
...
+public record BesiegerPartyRemoved : GenericEvent<BesiegerCamp, MobileParty>
 {
     public BesiegerPartyRemoved(BesiegerCamp instance, MobileParty value) : base(instance, value)
     {
     }
+
+    public override void HandleEvent(IObjectManager objectManager, INetwork network)
+    {
+        throw new System.NotImplementedException();
+    }
```
garrettluskey, line 21: I'm going to have to reject this idea.

There is not enough value added with these to justify changing/debugging our system
Current handlers are more generic and are meant to collect the required dependencies.
If the dependencies ever change or another is required, refactoring this would be difficult

PR #1131, opened 2026-03-31
feat: suppress safe-mode popup when using DebugAutoConnect

Gears321:
## Problem

When launching Bannerlord via the DebugAutoConnect build configuration, both the server and client processes are started and later killed hard by the debugger (or by stopping the VS debug session). Bannerlord tracks whether it shut down cleanly using a flag in engine_config.txt:

    safely_exited = 0   (written at startup)
    safely_exited = 1   (overwritten on clean shutdown)

Because the processes never reach a clean shutdown under DebugAutoConnect, the flag is left at 0 every single run. On the next launch Bannerlord detects this, assumes the previous session crashed, and shows the safe-mode recovery popup — prompting the developer to disable mods. This made every single DebugAutoConnect iteration require an extra manual dismissal step before the game would continue loading.

## Why a simple file edit on disk isn't enough

The game rewrites safely_exited=0 at startup before Coop even loads, so manually setting it to 1 before launching doesn't help — Bannerlord immediately resets it. The only reliable approach (without a command-line flag from TW, which doesn't exist) is the technique documented by TaleWorlds developer Skosnowich: set safely_exited=1 and then mark the file read-only so the game cannot overwrite it during the session.

## Solution

Added EnsureSafeExitConfig() to CoopMod, called only when the /autoconnect argument is present (i.e. exclusively under the DebugAutoConnect build config):

1. Removes the read-only attribute so the file can be edited.
2. Finds the safely_exited line and sets the value to 1 (appends the line if missing).
3. Marks engine_config.txt read-only so Bannerlord cannot reset it to 0 at startup or at shutdown.

The method is intentionally NOT called during normal Debug/Release launches, so the safe-mode popup continues to work correctly for everyday development and for end users. The file is also left entirely unmodified on normal launches — no writable-restore step is needed because EnsureSafeExitConfig() is the only place that ever locks it.

## Scope

- source/Coop/CoopMod.cs only
- Zero impact on non-DebugAutoConnect configurations
- Fully wrapped in try/catch — if the config file is missing or locked by another process the failure is logged as a warning and startup continues normally

## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

safe mode pop ups

## What is the new behavior?
no safe mode pop ups if in debugAutoConnect config

### DIFF
--- a/source/Coop/CoopMod.cs
+++ b/source/Coop/CoopMod.cs
@@ -66,6 +66,7 @@ namespace Coop
             {
                 Logger.Information("[AutoConnect] Full command line: {CommandLine}", fullCommandLine);
                 Logger.Information("[AutoConnect] isServer={IsServer} isAutoConnect={IsAutoConnect}", isServer, isAutoConnect);
+                EnsureSafeExitConfig();
             }
 
             GameLoopRunner.Instance.SetGameLoopThread();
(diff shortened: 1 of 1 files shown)


EgardA (approved):
lgtm

PR #1132, opened 2026-04-02
Devfred/fix dynamicsync export race

Gears321:
Fix DynamicSync export directory race in DebugAutoConnect

When running DebugAutoConnect both the server and client processes start
simultaneously and both called DynamicSyncBuilder.Build(), which deleted
and recreated the DynamicSyncExport directory. This caused two races:

  1. IOException ('directory is not empty') - the losing process hit
     Directory.Delete while the winner was mid-deletion.
  2. FileNotFoundException on File.SetAttributes - the winner deleted
     files between the loser's Directory.GetFiles enumeration and its
     subsequent SetAttributes call.

Root cause: both processes share the same export directory on the same
machine, but only the server's export is needed (both sides generate
identical output).

Fix:
- DynamicSyncConfiguration.ExportFiles gains a public setter (was
  get-only). Defaults to true so server behaviour is unchanged.
- DynamicSyncBuilder.Build() gates the entire directory cleanup block
  on ExportFiles, so a client with ExportFiles=false skips it entirely.
  File.SetAttributes errors are now caught and logged as warnings rather
  than swallowed silently, in case a race ever surfaces again.
- CoopartiveMultiplayerExperience.StartAsClient() sets
  DynamicSyncConfiguration.ExportFiles = false before calling PatchAll(),
  so the client never touches the export directory.
- Roslyn compilation errors in Build() are now logged via Serilog before
  throwing, making failures easier to diagnose.

This fix is safe for both deployment topologies:
- Same machine (DebugAutoConnect): race eliminated at the root because
  only one process (server) manages the shared directory.
- Separate machines (normal multiplayer): ExportPath is local to each
  machine, so no race was ever possible; client skipping its own local
  export is a minor debug inconvenience only."

## PR Checklist

## PR Type

- [X ] Bug fix (non-breaking change that fixes an issue)
- [ ] Build related changes
- [X ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #

### DIFF
--- a/source/Coop.Core/CoopartiveMultiplayerExperience.cs
+++ b/source/Coop.Core/CoopartiveMultiplayerExperience.cs
@@ -8,6 +8,7 @@ using Coop.Core.Common.Services.Connection.Messages;
 using Coop.Core.Server;
 using GameInterface;
 using GameInterface.AutoSync;
+using GameInterface.DynamicSync;
 using GameInterface.Services.GameDebug.Messages;
 using GameInterface.Services.UI.Messages;
 using HarmonyLib;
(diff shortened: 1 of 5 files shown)


EgardA (approved):
LGTM

PR #1133, opened 2026-04-02
Fix ID collision crash and add autoconnect support for 3-client sessions

Gears321:
When a 2nd player client had already joined and a 3rd client connected via autoconnect, the server would attempt to register a new Hero/MobileParty/Clan for the 3rd client. AutoRegistry.GetNewId generated IDs (e.g. Coop_Hero_1733) that were already occupied by Client 2's objects - which had been registered via NetworkCreateInstance (AddExisting) without advancing InstanceCounter past their numeric suffix. This caused AddNewObject to fail silently, leaving the new hero with null Coop IDs and vanilla StringIds ('main_hero', 'player_party'), triggering a NullReferenceException in CampaignObjectManager.GetIdParts during InitializeCampaignObjectsOnAfterLoad and making the client appear to load in as the host.

Fix: AutoRegistry.RegisterExistingObject now parses the numeric suffix from each Coop ID and advances InstanceCounter to at least that value, ensuring GetNewId always produces a fresh ID beyond all previously registered objects.

HeroInterface.UnpackHero now guards against null IDs before assigning back to StringId, preventing corruption if registration ever fails.

ClientDebug 2 project and solution updated to add the DebugAutoConnect build configuration, wiring the 3rd client launch with /autoconnect and platformId testclient2 so all three processes (host + 2 players) can be started and debugged from Visual Studio in a single solution build configuration.

## PR Checklist

## PR Type

- [x ] Bug fix (non-breaking change that fixes an issue)
- [x ] New feature (non-breaking change that adds functionality)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #

### DIFF
--- a/source/ClientDebug 2/ClientDebug2.csproj
+++ b/source/ClientDebug 2/ClientDebug2.csproj
@@ -22,6 +22,15 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'DebugAutoConnect|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>bin\Debug\</OutputPath>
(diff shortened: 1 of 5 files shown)


EgardA (approved):
Works well, great job!

PR #1137, opened 2026-04-03
Fix ID collision crash and add autoconnect support for 3-client sessions

garrettluskey:
Remade Bannerlord-Coop-Team/BannerlordCoop#1133


source/GameInterface/Registry/Auto/AutoRegistry.cs
```diff
@@ -47,7 +47,27 @@ protected override string GetNewId(T obj)
...
+                    current = InstanceCounter;
+                    if (current >= parsedNumber) break;
+                }
+                while (Interlocked.CompareExchange(ref InstanceCounter, parsedNumber, current) != current);
+            }
+            else
+            {
+                Interlocked.Increment(ref InstanceCounter);
+            }
+        }
```
garrettluskey: All Ids are assigned server side, so we should never need this

PR #1146, opened 2026-04-13
Devfred/perf fixes debugger hitching

Gears321:
perf: eliminate debugger-attached hitching and per-tick DynamicSync spam

CoopMod.cs — disable WriteTo.Debug, fix MinimumLevel
  WriteTo.Debug was writing every log event across the VS debugger IPC
  channel. At Verbose/Debug level this reached 1300+ writes/sec causing
  severe frame hitching whenever both processes ran under the debugger.
  Disabled the sink entirely; all output still goes to Coop_client.log /
  Coop_server.log which are readable live via tail. MinimumLevel changed
  from Verbose to Debug (DebugAutoConnect) / Information (Release) so the
  file sink is no longer flooded with sub-Debug noise in production.

ClanSync.cs — remove Clan.CurrentTotalStrength from DynamicSync
  KingdomManager.HourlyTickClan() calls Clan.UpdateCurrentStrength() on
  every clan each in-game hour. That method first zeros CurrentTotalStrength
  then rebuilds it with += per WarPartyComponent and per Garrison, firing
  the DynamicSync property setter on every intermediate write. This produced
  ~253 network messages per second during normal gameplay. CurrentTotalStrength
  is a derived/cached value recomputed server-side; clients run no AI or
  election logic that depends on it and do not need live sync.

MobilePartyBehaviorHandler.cs — no changes any longer - getting latest proved we didn't need to make the change.

MilitiaPartyComponentLifetimePatches.cs — downgrade null-guard logs to Debug
  PartyOwner, Name, and GetDefaultComponentBanner Harmony prefixes guard
  against NullReferenceExceptions during sync transitions where Settlement
  is transiently null. These fire dozens of times per second as a normal
  part of multiplayer operation. Previously logged at Warning, which meant
  re-enabling WriteTo.Debug with a Warning+ restriction immediately caused
  lag (same root cause as the Verbose flood). Downgraded to Debug so they
  are captured in file logs during development but will not cross the
  debugger IPC channel if WriteTo.Debug is ever re-enabled with Warning+.

## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #


EgardA (approved):
LGTM

PR #1152, opened 2026-04-18
Devfred/patrol party sync and nullref fixes

Gears321:
fix: NullReferenceException in SettlementHandler.HandleCollectNotablesToCache
Root cause:
  Protobuf omits empty repeated fields on the wire and deserializes them as
  null rather than an empty collection. SettlementHandler.HandleCollectNotablesToCache
  received a CollectNotablesToCache message whose NotablesCache list was null
  (the settlement had no notables), then iterated it with a plain foreach,
  producing a NullReferenceException at runtime.

Stack trace location:
  SettlementHandler.cs:221 - foreach (string heroStringId in obj.NotablesCache)

Fix:
  Changed the foreach to use the null-coalescing fallback:
    foreach (string heroStringId in obj.NotablesCache ?? Enumerable.Empty<string>())
  Added 'using System.Linq' for Enumerable.Empty<string>().

Behavior fixed:
  Clients no longer crash when a settlement sync message arrives for a
  settlement that has an empty notables cache. The loop simply executes
  zero iterations, matching the intended no-op behavior.

---------------

feat: full PatrolPartyComponent sync between server and client
Background:
  PatrolPartiesCampaignBehavior spawns, ticks, replenishes, and destroys
  patrol parties on behalf of settlements. Previously PatrolPartyComponent
  was completely absent from the coop sync infrastructure, causing two
  distinct failures and leaving patrol parties invisible/broken on clients.

--- NullReferenceException (server crash) ---

Root cause:
  MobilePartyDataPatches.SetPartyComponentIntercept intercepts the
  MobileParty._partyComponent field assignment inside MobileParty.CreateParty.
  It calls PartyComponentRegistry.TryGetId(value, out componentId). Because
  PatrolPartyComponent was not in PartyComponentRegistry.ManagedTypes, TryGetId
  returned false and the interceptor returned WITHOUT assigning _partyComponent.
  Back in PatrolPartyComponent.CreatePatrolParty, the next line called
  mobileParty.PatrolPartyComponent.SortRoster(). PatrolPartyComponent casts
  _partyComponent to PatrolPartyComponent - which was still null - causing
  NullReferenceException at line 62.

  Server log confirmed this with:
    'Component was not registered with PartyComponentRegistry'
  immediately before the fatal exception.

--- Sync implementation ---

Files added:

  PatrolPartyComponentLifetimePatches.cs
    Harmony prefix on all PatrolPartyComponent constructors. Fires before the
    constructor body and publishes PartyComponentCreated, registering the new
    instance in the coop ObjectManager. Constructor parameters homeSettlement
    and isNaval are captured directly from the Harmony prefix signature (by
    parameter name matching) and bundled into the message. This avoids any
    dependency on DynamicSync field-update ordering when the client
    reconstructs the component.

  PatrolPartyComponentSync.cs
    Registers _homeSettlement (field) and IsNaval (property) with DynamicSync
    so that any post-construction mutations are automatically broadcast to all
    clients, matching the pattern used by CaravanPartyComponentSync et al.

Files modified:

  PartyComponentRegistry.cs
    Added typeof(PatrolPartyComponent) to ManagedTypes so that
    SetPartyComponentIntercept can find registry IDs for patrol components,
    and RegisterAll correctly registers patrol parties loaded from save games.

  PartyComponentHandler.cs
    - Added PatrolPartyComponent at index 6 in partyTypes[] (VillagerPartyComponent
      shifted to index 7). Index is the wire-format type discriminator used by
      NetworkCreatePartyComponent.
    - In Handle(PartyComponentCreated): populates new PartyComponentData fields
      HomeSettlementId and IsNaval from the creation message.
    - In Handle(NetworkCreatePartyComponent) case 6: creates a SkipConstructor
      shell, registers it with the ObjectManager, then atomically reconstructs
      its state - sets _homeSettlement via reflection, sets IsNaval via the
      private property setter, and calls InitializePartyComponentProperties via
      reflection. InitializePartyComponentProperties sets the _name cache and
      calls Settlement.SetPatrolParty(this), which populates Settlement.PatrolParty
      on the client so patrol party UI and queries work correctly.
      If the settlement cannot be found in the ObjectManager a warning is logged
      but the component is still registered (graceful degradation).

  PartyComponentCreated.cs
    Added IsNaval property (default false) so the lifetime patch can forward
    the constructor parameter to the handler without losing information.

  PartyComponentData.cs
    Added optional ProtoMember(3) HomeSettlementId and ProtoMember(4) IsNaval.
    Null/false for all non-patrol component types; populated only for index 6.
    Bundling these in the creation packet eliminates the race condition where
    a previous implementation called InitializePartyComponentProperties in
    MobilePartyDataHandler and relied on DynamicSync messages having already
    delivered _homeSettlement - which was not guaranteed.

  DisablePatrolPartiesCampaignBehavior.cs (new file, client-side gate)
    Prefixes PatrolPartiesCampaignBehavior.RegisterEvents with
    !ModInformation.IsClient. The server runs the behavior normally (spawning,
    ticking, replenishing, destroying patrol parties). The client must not run
    it independently - doing so would cause it to spawn its own local patrol
    parties that fight the server-authoritative synced ones and re-trigger the
    PartyComponentRegistry error on the client side.

Destruction path:
  When the server destroys a patrol party, PartyLifetimeHandler sends
  NetworkDestroyParty to clients. The client calls party.RemoveParty() inside
  AllowedThread, which invokes PatrolPartyComponent.OnFinalize(), which calls
  Settlement.SetPatrolParty(null). The settlement link is correctly cleaned up.

Save-game compatibility:
  Patrol parties loaded from an existing save go through the normal
  MBObjectManager load path which calls OnInitialize() directly, setting
  Settlement.PatrolParty as usual. RegisterAll() now includes them because
  PatrolPartyComponent is in ManagedTypes.

-----------------------

## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #

### DIFF
--- a/source/GameInterface/Services/MobileParties/Patches/Disable/DisablePatrolPartiesCampaignBehavior.cs
+++ b/source/GameInterface/Services/MobileParties/Patches/Disable/DisablePatrolPartiesCampaignBehavior.cs
@@ -1,0 +1,32 @@
+using Common;
+using HarmonyLib;
+using TaleWorlds.CampaignSystem.CampaignBehaviors;
+
+namespace GameInterface.Services.MobileParties.Patches;
+
+/// <summary>
+/// Disables <see cref="PatrolPartiesCampaignBehavior"/> on the CLIENT only.
+///
+/// The server must run this behavior normally: it spawns patrol parties via
+/// <c>SpawnPatrolParty</c>, ticks their AI, replenishes their rosters, and destroys
+/// them when appropriate. All of those actions flow through the coop sync pipeline
+/// (party creation → <c>NetworkCreatePartyComponent</c> + <c>NetworkCreateParty</c>,
+/// destruction → <c>NetworkDestroyParty</c>, field mutations → DynamicSync).
+///
+/// The client must NOT run this behavior. If it did, it would independently spawn its
+/// own local patrol parties, duplicating the ones already synced from the server and
+/// triggering the same NullReferenceException that motivated this patch:
+/// <c>MobilePartyDataPatches.SetPartyComponentIntercept</c> would fire, find the
+/// newly constructed <c>PatrolPartyComponent</c> in the registry (it is now registered),
+/// publish a <c>PartyComponentChanged</c> event, and the client-side party would
+/// incorrectly broadcast as if it were a server-authoritative change.
+///
+/// Clients receive patrol parties exclusively through the sync pipeline; this patch
+/// simply prevents the behavior from subscribing to any campaign events on the client.
+/// </summary>
+[HarmonyPatch(typeof(PatrolPartiesCampaignBehavior))]
+internal class DisablePatrolPartiesCampaignBehavior
+{
+    [HarmonyPatch(nameof(PatrolPartiesCampaignBehavior.RegisterEvents))]
+    static bool Prefix() => !ModInformation.IsClient;
+}
--- a/source/GameInterface/Services/PartyComponents/Data/PartyComponentData.cs
+++ b/source/GameInterface/Services/PartyComponents/Data/PartyComponentData.cs
@@ -10,4 +10,20 @@ public record PartyComponentData(int TypeIndex, string Id)
 
     [ProtoMember(2)]
     public string Id { get; } = Id;
+
+    /// <summary>
+    /// Optional: the StringId of the home Settlement for a <c>PatrolPartyComponent</c>.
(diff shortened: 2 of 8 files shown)


Osttik (approved):
Let it ride

garrettluskey (approved):

PR #1175, opened 2026-05-15
Hero Hitpoints and banneritem synced plus debugcommands

NatteTosti69:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ x ] All new classes have class-level documentation comments, if there are any at all
- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ x ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update
- [ ] Other... Please describe:

## What is the current behavior?

adding sync for hero hitpoints and banneritem

## What is the new behavior?

Resolves # 1049

added hero.hitpoints and banneritem sync plus debugcommands

## Other information
Created EquipmentElementSerializer, a static custom serializer that transmits only the ItemObject.StringId instead of the full struct, since EquipmentElement is a struct that can't be decorated with protobuf attributes Registered the serializer in HeroSync using static method references DynamicSyncRegistry.AddSerializer signature to accept Func<byte[], TargetType> for the deserialize parameter Added IsCustomSerializer flag to DynamicSyncPropertyBuilder.GetTemplateData Updated SubscribeSetValueTemplate to use non-generic deserialize syntax for custom serializers, since they don't need a type parameter unlike RawSerializer.Deserialize<T> also added debugcommand to set hero.gold

PR #1177, opened 2026-05-16
Fix PartyBehaviorPatchTests after ObjectManager-based id lookup

AndreasOhmer:
## Summary
- Aligns `PartyBehaviorPatchTests` with the new `IObjectManager`-based id lookup introduced in commit 73723e32
- Replaces `party.Name` with `party.StringId` in the `IsPartyControlled` error log to avoid a `NullReferenceException` on partially-initialised parties

## Root cause
Commit 73723e32 (\"fixed old sting ids\") changed `PartyExtensions.IsPartyControlled` from using `party.StringId` directly to calling `objectManager.TryGetId(party, out var partyId)`. The existing `PartyBehaviorPatchTests` were not updated for the new flow:

- The test party is created via `ObjectHelper.SkipConstructor<MobileParty>()` and only has `StringId` assigned; it is never added to the `IObjectManager`.
- `TryGetId` therefore returns false and the failure branch executes `Logger.Error(\"...\", party.Name)`. The `MobileParty.Name` getter touches uninitialised internal state on a `SkipConstructor` instance, throwing `NullReferenceException`.

Both `UncontrolledParty_SkipsMethod` and `ControlledParty_RunsMethod` fail with the same NRE for this reason.

## Changes
1. `PartyBehaviorPatchTests` now register the test party with `IObjectManager.AddExisting(party.StringId, party)` and resolve the id back through `TryGetId` to feed `IControlledEntityRegistry.RegisterAsControlled`. This mirrors the production flow in `MobilePartyRegistry.RegisterAllObjects`.
2. `PartyExtensions.IsPartyControlled` logs `party.StringId` instead of `party.Name`, matching the defensive style of `MobilePartyAuditData` and `SkillHelperRobustnessPatches`.

## Test plan
- [x] `dotnet test source/GameInterface.Tests --filter PartyBehaviorPatchTests`: 0/2 -> 2/2
- [x] `dotnet test source/GameInterface.Tests` (full suite): 202/215 -> 204/215 (no regression)
- [x] `dotnet test source/Common.Tests`: 34/34 (unchanged)
- [x] `dotnet build Coop.sln`: no new errors introduced

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/source/GameInterface.Tests/Services/Parties/Patches/PartyBehaviorPatchTests.cs
+++ b/source/GameInterface.Tests/Services/Parties/Patches/PartyBehaviorPatchTests.cs
@@ -3,6 +3,7 @@ using Common.Messaging;
 using Common.Util;
 using GameInterface.Services.Entity;
 using GameInterface.Services.MapEvents.Patches;
+using GameInterface.Services.ObjectManager;
 using GameInterface.Tests.Bootstrap;
 using System;
 using TaleWorlds.CampaignSystem.Party;
@@ -29,6 +30,9 @@ public class PartyBehaviorPatchTests : IDisposable
         var party = ObjectHelper.SkipConstructor<MobileParty>();
         party.StringId = "TestEntityId";
 
+        var objectManager = Container.Resolve<IObjectManager>();
+        objectManager.AddExisting(party.StringId, party);
+
         var dt = 0.1f;
 
         // Act
@@ -48,9 +52,13 @@ public class PartyBehaviorPatchTests : IDisposable
 
         var idProvider = Container.Resolve<IControllerIdProvider>();
         var entityRegistry = Container.Resolve<IControlledEntityRegistry>();
+        var objectManager = Container.Resolve<IObjectManager>();
+
+        objectManager.AddExisting(party.StringId, party);
+        objectManager.TryGetId(party, out var partyId);
 
         idProvider.SetControllerId(controllerId);
-        entityRegistry.RegisterAsControlled(controllerId, party.StringId);
+        entityRegistry.RegisterAsControlled(controllerId, partyId);
 
         var dt = 0.1f;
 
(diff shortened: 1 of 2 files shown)


garrettluskey (dismissed):

AndreasOhmer:
Rebased onto current `development` after #1166 (`1.3-interactions`) was merged.

Conflict was a 1-line `using` change — the file under test (`EncounterManagerPatches`) moved from `GameInterface.Services.MobileParties.Patches` to `GameInterface.Services.MapEvents.Patches`. Resolution: keep the new namespace and add `using GameInterface.Services.ObjectManager;` for the test-side registration introduced by this PR.

Transparency: I could not run the tests locally on this rebase. After the `1.3-interactions` merge, `Coop.Core.csproj` references `Scriban 7.1.0` but `GameInterface` (via #1178) now requires `Scriban 7.2.0` — NuGet refuses with `NU1605` (package downgrade) so the test projects don't restore. I'll open a separate small PR bumping `Coop.Core`'s Scriban to 7.2.0 so this and other PRs can build green again.

Please re-run CI once that lands.

AndreasOhmer:
Re-rebased onto current `development` after #1175 + #720e8b1f (Scriban fix) landed.

The conflict resolution from my previous rebase still applies (uses new `MapEvents.Patches` namespace + adds `using GameInterface.Services.ObjectManager`). Verified locally:

```
GameInterface.Tests --filter PartyBehaviorPatchTests:
  Passed: 2/2
```

The test container's `PartyExtensions.IsPartyControlled` requires the party to be registered in `IObjectManager` so it can resolve a `partyId` — this PR adds the `objectManager.AddExisting()` + `TryGetId()` plumbing in both test methods.

Note: `GameInterface.Tests` as a whole has many pre-existing failures (~168) on `development` post-#1166, but the targeted `PartyBehaviorPatchTests` pass. Those other failures are orthogonal and need separate fixes.

garrettluskey (approved):

PR #1280, opened 2026-06-06
Fix Harmony global UnpatchAll crash on coop session teardown

garrettluskey:
GameInterface.UnpatchAll() called the parameterless Harmony.UnpatchAll(), which is HarmonyLib's global unpatch: it removes every patch in the process regardless of owning id -- the base game's patches, other mods', and the test bootstrap's included. Leaving a coop session therefore stripped patches the engine and other mods depend on, crashing the game. Scope the teardown to Coop's own id via harmony.UnpatchAll(harmony.Id).

Add HarmonyUnpatchAllLifetimeTests: drives the real GameInterface.UnpatchAll() while a foreign Harmony id holds a live patch and asserts the foreign patch survives while Coop's own patches are removed (fails pre-fix, passes post-fix).

Isolate GameInterfaceModuleLifetimeTests onto a dedicated Harmony id instead of the shared static GameInterfaceModule harmony. Driving a broad UnpatchAll on the shared harmony re-wrapped JIT-optimized patches left behind by AutoSyncTests/ DynamicSyncTests and threw InvalidProgramException, making the AutoSync suite order-dependent (failed ~3/4 runs on plain development). Both lifetime tests now build the real GameInterface around their own Harmony + targets; the suite is deterministic.

Also fix a stale reference in Coop.Tests (Handle_GameLoaded -> Handle_CampaignReady) that blocked the E2E build.

## PR Checklist

## PR Type
- [x] Other: also adds a runtime patch-disable guard across all patches

Leaving a coop session could crash the game two ways:
1. `GameInterface.UnpatchAll()` called the parameterless `Harmony.UnpatchAll()`, a *global* unpatch that strips every patch in the process (base game + other mods), not just Coop's.
2. `Hideout..ctor` and `GauntletMapEventVisual..ctor` were each patched by **two** prefixes (a manual `*LifetimePatches`/`Debug` class *and* the generic AutoRegistry). Unpatching a constructor with two stacked prefixes throws `InvalidProgramException` during wrapper regeneration.

Additionally, there was no reliable way to neutralize Coop's Harmony patches at runtime.

## What is the new behavior?
Resolves #

- `UnpatchAll()` is scoped to Coop's own id: `harmony.UnpatchAll(harmony.Id)`.
- Removed the redundant manual `Hideout` lifetime system (`HideoutLifetimePatches`, `HideoutLifetimeHandler`, `HideoutCreated`, `NetworkCreateHideout`) — `AutoRegistryHandler<Hideout>` already covers it (also fixes a duplicate creation packet). Removed the stray no-op `Debug` prefix on `GauntletMapEventVisual..ctor`.
- **Every** `GameInterface` prefix/postfix (~332 methods) now begins with `if (CallOriginalPolicy.IsOriginalAllowed()) return true;`/`return;`, so the `DisablePatches` flag (and existing sync gating) fully reverts the mod to vanilla at runtime. Expression-bodied `Disable*` patches were converted to block bodies.

Tests:
- `HarmonyUnpatchAllLifetimeTests` — foreign-id patches survive Coop teardown.
- `ConstructorPatchUniquenessTests` — fails if any constructor is prefixed by more than one lifetime system.
- `GameInterfaceModuleLifetimeTests` isolated onto a dedicated Harmony id (was order-flaky on the shared static harmony).

## Other information
The patch-gating is per-patch only (patches are still applied, just no-op when disabled); it does not prevent patch/unpatch-time issues. Reviewers: the guard commit (`167642bd`) is large but mechanical — one guard line per patch, build verified.


garrettluskey:
Too many changes

PR #1295, opened 2026-06-08
Battle Simulation

### LINKED ISSUE #1192: Player Battle Simulation
# For New Contributors

## Description

Sync Simulation for player battles (Player vs AI)
Essentially the Send Troops button

## Location

## Definition of Done

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #1192


source/GameInterface/Services/MapEvents/Handlers/BattleSimulationRunHandler.cs
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    /// <summary>[Client] Resolve a streamed round and queue it for playback.</summary>
+    private void Handle_NetworkBattleSimulationRound(MessagePayload<NetworkBattleSimulationRound> payload)
+    {
+        var message = payload.What;
+        if (message.Changes == null || message.Changes.Length == 0)
+            return;
+
+        var resolved = new List<BattleSimulationReplay.ResolvedChange>(message.Changes.Length);
```
ShoT-UPfps: I think we need to wrap 219-234 here with GameLoopRunner.RunOnMainThread since:

1. the code block interacts with `objectManager` 
2. code block uses `TryResolveCharacterObject`
3. line 234 enqueues to `Queue`

which are otherwise at risk of race conditions

source/GameInterface/Services/MapEvents/Handlers/BattleSimulationRunHandler.cs
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        if (resolved.Count > 0)
+            BattleSimulationReplay.EnqueueRound(resolved.ToArray());
+    }
+
+    /// <summary>[Client] Server finished simulating: end playback once the queued rounds drain.</summary>
+    private void Handle_NetworkBattleSimulationFinished(MessagePayload<NetworkBattleSimulationFinished> payload)
+    {
+        if (PlayerEncounter.CurrentBattleSimulation == null)
```
ShoT-UPfps: I think we need to wrap 240-246 here with GameLoopRunner.RunOnMainThread since:

1. line 240 could be mutated at the same time
2. line 246 updates the `finishRequested` flag

source/GameInterface/Services/MobileParties/Handlers/PartyLeaderHandler.cs
```diff
@@ -0,0 +1,71 @@
...
+        if (payload.What.NewLeader != null && !objectManager.TryGetIdWithLogging(payload.What.NewLeader, out leaderHeroId))
+            return;
+
+        network.SendAll(new NetworkChangePartyLeader(mobilePartyId, leaderHeroId));
+    }
+
+    /// <summary>[Client] Apply the leader change.</summary>
+    private void Handle_NetworkChangePartyLeader(MessagePayload<NetworkChangePartyLeader> payload)
+    {
+        if (!objectManager.TryGetObjectWithLogging<MobileParty>(payload.What.MobilePartyId, out var mobileParty))
```
ShoT-UPfps: I think we need to wrap 56-69 here with GameLoopRunner.RunOnMainThread since:

1. lines 56 + 60 reads from `objectManager` which could be getting updated at the same time
2. line 65 mutates parties which (as I understand it) is only supposed to be done by the main thread

source/GameInterface/Services/MapEvents/Handlers/BattleSimulationRunHandler.cs
```diff
@@ -0,0 +1,264 @@
...
+            {
+                rounds++;
+                sim.MapEvent.SimulatePlayerEncounterBattle();
+
+                var changes = sim.Observer.FlushRound();
+                if (changes.Length > 0)
+                    network.Send(sim.Peer, new NetworkBattleSimulationRound(mapEventId, changes));
+            }
+
+            if (sim.MapEvent.HasWinner)
```
ShoT-UPfps, line 198: If I understand correctly, most of this workflow depends on the client. The server only continues when it receives a `NetworkAdvanceBattleSimulation` and it only restores the observer and removes `activeSimulations` entry when `HasWinner` is true. 

If the client crashes or otherwise drops connection and doesn't reconnect, server doesn't receive any more `NetworkAdvanceBattleSimulations` and the resources are never cleaned up nor does the simulation advance.

I think we might need something that detects the player is disconnected and restores the observer and drops the `activeSimulations` entry. I guess the server should finish the sim on its own at that point?

source/GameInterface/Services/MapEvents/Messages/Start/BattleSimulationStarted.cs
```diff
@@ -0,0 +1,19 @@
...
+using TaleWorlds.CampaignSystem.MapEvents;
+
+namespace GameInterface.Services.MapEvents.Messages.Start;
+
+/// <summary>
+/// [Client] Raised when the local player opens the auto-resolve battle simulation screen
+/// (<c>MapState.StartBattleSimulation</c>). Forwarded to the server so the server can run the
+/// authoritative simulation for this <see cref="TaleWorlds.CampaignSystem.MapEvents.MapEvent"/>.
+/// </summary>
+internal readonly struct BattleSimulationStarted : IEvent
```
ShoT-UPfps, line 11: general question, I didn't sim much in native Bannerlord. Can parties join mid-way? If yes, do we want to / need to handle another player joining the sim in this PR? and does the current implementation properly handle when an AI party joins midway?

source/GameInterface/Services/MapEvents/Patches/PlayerDefeatCapturePatch.cs
```diff
@@ -0,0 +1,92 @@
...
+/// deactivate — not destroy — the party), then remove those parties from the list vanilla processes so it
+/// never runs its destroy path on them.
+/// </remarks>
+[HarmonyPatch(typeof(MapEvent), "CaptureDefeatedPartyMembers")]
+internal class PlayerDefeatCapturePatch
+{
+    private static readonly ILogger Logger = LogManager.GetLogger<PlayerDefeatCapturePatch>();
+
+    [HarmonyPrefix]
+    private static void Prefix(MBReadOnlyList<MapEventParty> winnerParties, ref MBReadOnlyList<MapEventParty> defeatedParties)
```
ShoT-UPfps: i think this prefix already exists at `PlayerStartCaptivityPatches.cs:Prefix_CaptureDefeatedPartyMembers`

I guess we need to merge the two of them. Also the existing one handles retreats but this new one doesn't. At the same time, the existing one doesn't do the `defeatedParties` filtering that this one does. Probably best to remove this file and merge the updates into the existing one

source/GameInterface/Services/MobileParties/Patches/ChangePartyLeaderPatch.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace GameInterface.Services.MobileParties.Patches;
+
+/// <summary>
+/// Replicates party leader changes from the server. The leader is what gives a party its nameplate and
+/// drives its map visual, and nothing else syncs it - so a server-side <c>ChangePartyLeader</c> (e.g. when
+/// a captured hero is removed, or restored on release) would otherwise leave the client's party leaderless.
+/// </summary>
+[HarmonyPatch(typeof(MobileParty))]
+internal class ChangePartyLeaderPatch
```
ShoT-UPfps, line 18: according to the docstring,  this patch is only needed so that it won't "leave the client's party leaderless."

But the `ChangePartyLeader` will happen constantly as AI parties battle, right? and we will be publishing a ton of messages for those AI battles as well with the current `Prefix` implementation.

can we add another check to make sure that we are only broadcasting when a player is impacted? or do we really need to broadcast all of them here?

ShoT-UPfps:
@garrettluskey et al., as I was reading through this PR and looking at `ObjectManager` I noticed that many of the existing methods e.g.. `ObjectManager.TryGetId` uses a `lock (_gate)` to avoid race conditions but  `TryGetObject` and `Contains(object)` do not.

I haven't been able to understand why that is, and it does seem to me that they should. This PR makes several calls to `TryGetObject` which could be getting mutated at the same time by main thread calling `Add` or `Remove`. 

do you see the same issue, or am i missing something?

source/GameInterface/Services/MapEvents/BattleSimulationReplay.cs
```diff
@@ -0,0 +1,169 @@
...
+    {
+        if (simulation == null || mapEventId == null || simulation.IsSimulationFinished)
+            return;
+
+        // Replay everything the server has sent so far onto the scoreboard.
+        while (arrivedRounds.Count > 0)
+            ApplyRound(simulation, arrivedRounds.Dequeue());
+
+        // Finish only once the server is done and all streamed rounds have been shown.
+        if (finishRequested && arrivedRounds.Count == 0)
```
ShoT-UPfps: once we reach this line, `arrivedRounds.count` is guaranteed to be 0 because the previous code block loops until `arrivedRounds.Count > 0`.

therefore line 111 is unreachable; if `finishRequested` is true, the method will always return on line 107

source/GameInterface/Services/MapEvents/Patches/PlayerHeroSimulationExclusionPatch.cs
```diff
@@ -0,0 +1,49 @@
...
+    private static void Postfix(List<(FlattenedTroopRosterElement, MapEventParty, float)> priorityList, bool includePlayer)
+    {
+        // includePlayer is true for real missions (players fight) and false for auto-resolve.
+        if (includePlayer || priorityList == null)
+            return;
+
+        priorityList.RemoveAll(entry => IsPlayerMainHero(entry.Item1));
+    }
+
+    private static bool IsPlayerMainHero(FlattenedTroopRosterElement element)
```
ShoT-UPfps: can we simplify by reusing `element.Troop?.HeroObject?.IsPlayerHero()` instead?

source/GameInterface/Services/MapEvents/Handlers/BattleSimulationRunHandler.cs
```diff
@@ -0,0 +1,264 @@
...
+        {
+            int rounds = 0;
+            while (rounds < maxRounds && rounds < MaxSimulationRounds && !sim.MapEvent.HasWinner)
+            {
+                rounds++;
+                sim.MapEvent.SimulatePlayerEncounterBattle();
+
+                var changes = sim.Observer.FlushRound();
+                if (changes.Length > 0)
+                    network.Send(sim.Peer, new NetworkBattleSimulationRound(mapEventId, changes));
```
ShoT-UPfps: if user "Skips" the round sequence, then the rounds get processed immediately and each round sends `NetworkBattleSimulationRound` + the casualty syncs.

If the user doesn't skip then those rounsd and casualty syncs get spread out more over time, but with skip it's gonna send a ton of packets in a very short interval. This will cause the server's outbound packet queue to get overloaded and the server's gonna pause the game for the queue to drain.

since skip is jumping to the end anyway, can we send a single final snapshot instead of streaming every round + per-casualty packet? that way, regardless of how many rounds or casualties, clients just consume one packet?

ShoT-UPfps (dismissed):
Mostly looks good. Some race condition concerns and other minor issues.

EgardA:
Thanks for the comments, this is now updated to work better :) There may still be some issues with more than 1 party

source/GameInterface/Services/MapEvents/Messages/Start/BattleSimulationStarted.cs
EgardA: This is not handled at the moment, might look into adding it but harder to test.

ShoT-UPfps: which part isn't handled? the AI parties joining, a real player joining, or both?

EgardA: A real player currently cannot join any player map event so that's not a problem at the moment. But AI is not specifically handled but not sure how it would look with this as it hasn't happened to me yet

ShoT-UPfps: works for me. I'm fine with moving forward without that verification, but do you think an issue should be created to look at this edge case so that it's not forgotten? or will you remember to look at it soon?

EgardA: If it's an issue we will catch it straight away in testing :)

ShoT-UPfps (approved):
looks good now

Issue #1325, opened 2026-06-11
Epic: Shared settlement/location missions (players see each other in tavern/town scenes)


ShoT-UPfps:
## Description

Location Sync makes campaign-side rosters identical across clients, but players entering the same settlement scene still cannot see each other or interact with the same live agents. Enabling shared settlement missions requires mission-level agent sync wired into the campaign settlement-entry flow: spawning other players' controlled agents into the local mission, synchronizing movement/state via the existing `NetworkAgentRegistry` infrastructure, and connecting `TavernsGameManager` (currently test-only) to the real settlement menu path.

This is the logical next step after Location Sync: once rosters agree at the campaign layer, the mission layer can use them to populate a shared scene.

## Related Issues

Related: #1312

Issue #1324, opened 2026-06-11
Keep server-side location rosters live on governor/prisoner events


ShoT-UPfps:
## Description

Vanilla refreshes hero placement in settlement locations on events such as `OnGovernorChanged`, `PrisonersChangeInSettlement`, and `HeroPrisonerTaken` (driven by `HeroAgentSpawnCampaignBehavior`). These handlers are gated on the local player being inside the settlement, which is never true on a stationary co-op server, so server-side location rosters go stale between player visits. Location Sync mitigates this with a full roster snapshot on each settlement entry, but rosters can still be wrong during an active visit if such an event fires mid-session.

Follow-up: add server-side handlers for these campaign events that update already-populated settlements' rosters live and broadcast the delta to clients currently inside the affected settlement.

## Related Issues

Related: #1312

garrettluskey:
I think this was handled by #1298

ShoT-UPfps:
i do not think so

ShoT-UPfps:
> **🤖 AI GENERATED** — static/decompilation analysis of the merged code, no live run.

## TL;DR: #1298 did not handle this — it should stay open

#1298 laid some groundwork and synced governor *data*, but the specific thing this issue asks for — **location rosters updating live, mid-session, while a player is inside the settlement** — is still missing for all three named events (`OnGovernorChangedEvent`, `PrisonersChangeInSettlement`, `HeroPrisonerTaken`).

### What #1298 actually contributed
- Flipped `DisableHeroAgentSpawnCampaignBehavior` from `Prefix() => false` (behavior fully disabled) to `Prefix() => true` ("Needs to also run on the client") — so `HeroAgentSpawnCampaignBehavior` at least *runs* on visiting clients now.
- Added governor-change **data** sync (`ChangeGovernorActionPatches` + `ChangeGovernorHandler`): the governorship assignment replicates correctly, and `OnGovernorChanged` *does* re-fire on the receiving client.

### Why it still falls short

**Governor.** On a visiting client `OnGovernorChanged` fires → `RefreshLocationOfHeroForSettlement(governor, …)` → `Location.AddCharacter` / `LocationComplex.ChangeLocation` for the governor **hero**. Both are deliberately blocked on clients (hero placement is server-driven by design):
[...]

Issue #1323, opened 2026-06-11
Propagate mission-time NPC kills/removals to server location roster


ShoT-UPfps:
## Description

When a client kills or otherwise removes a synced location character during a settlement mission, that removal is not propagated to the server. With strict server authority, no client can unilaterally remove a roster entry, so the NPC reappears on the next mission load.

Needs a client-request → server-approve removal flow: the client sends a removal request (e.g. on `OnAgentRemoved` / agent death), the server validates and updates the `Location._characterList` entry, then broadcasts the removal to all peers.

## Related Issues

Related: #1312

ShoT-UPfps:
dont think we need this or can really test it till we have issues implemented

Issue #1322, opened 2026-06-11
Consistent ambient NPCs in settlement locations (tavern keepers, townsfolk, guards)


ShoT-UPfps:
## Description

Ambient non-hero location NPCs — tavern keepers, townsfolk, guards, musicians, etc. — are spawned client-locally by ~10 SandBox campaign behaviors reacting to `LocationCharactersAreReadyToSpawn`. These behaviors use scene-derived data (`FindUnusedUsablePointCount`) that requires a loaded mission scene, which a headless co-op server cannot run. Location Sync therefore lets each client populate ambience locally, producing a divergent NPC roster per client. Several of these behaviors (e.g. `TavernEmployeesCampaignBehavior`, `SettlementMusiciansCampaignBehavior`) are disabled in co-op entirely.

Follow-up: give clients a consistent ambient NPC roster, either via server-seeded deterministic generation (same seed → same characters) or a client-population request flow where the server supplies the character list on settlement entry.

We also need "conversation locks": Only one character at a time can talk to an NPC.
## Related Issues

Related: #1312

PR #1334, opened 2026-06-12
[1312] Location Sync

### LINKED ISSUE #1312: Epic: Location Sync
## For New Contributors

## Description
- [ ] Create LocationRegistry
- [ ] Sync `_characterList`
- [ ] Sync `SpecialItems`

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Locations are not synced

## What is the new behavior?

Resolves #1312

- `LocationRegistry` registers every settlement's locations on server and clients.
- `Location._characterList` is synced 
- `Location.SpecialItems` is synced
- When a player party enters a settlement, the server populates its locations (notables, wanderers, lords, prisoners, and each visiting party's companions in the tavern), sends a roster snapshot on every entry so late joiners reconcile, and clears the rosters when the last player leaves. Characters added while another player is inside a location scene walk into the running mission.
- `coop.debug.location.*` commands cover every synced update type plus `populate` for testing without a second client.
- `CampaignTimeSyncHandler` reads the campaign tick with a volatile read and re-arms its timer after each broadcast, keeping the once-a-second broadcast off the game thread; the location E2E tests also exposed a pre-existing test-harness lock leak that could hang the suite, now fixed.

Out of scope, tracked as follow-ups: #1322 (consistent ambient NPCs), #1323 (mission-kill propagation), #1324 (governor/prisoner roster refresh), #1325 (shared settlement missions).

server+client+client are showing synced locations (note that locations/special items are only loaded when a player is actually at a settlement, thats why the values are showing zero):
server-client-client

able to see characters in taverns, **and i tested adding a character WHILE a client is in a tavern, and the character correctly appears in the tavern**
tavern

the `.populate` command populates locations without a client actually visiting, confirmed that it's synced between all clients:
populate

verified that when client 1 is already at the settlement, and client 2 joins with a companion, the companion becomes visible on client 1's hero menu in the tavern. Further, when client 1 visits the tavern, client 2's companion is there 
image

player 1 cannot recruit player 2's companion from the tavern:
image


source/Coop.Core/Server/Services/Time/Handlers/CampaignTimeSyncHandler.cs
garrettluskey, line 1: We want to avoid overloading the game thread. A better option would be to update TryGetCurrentTicks to use Volatile.Read(ref tracker._numTicks)

source/GameInterface/Services/Locations/Patches/LocationCharacterGuardPatches.cs
```diff
@@ -0,0 +1,40 @@
...
+    // Companions accompanying a player are placed by the server for every visiting party. The
+    // local spawn would duplicate them, because the mission spawn check compares agent origins by
+    // reference and cannot recognize the server-broadcast entry as the same character.
+    [HarmonyPatch(typeof(MissionLocationLogic), nameof(MissionLocationLogic.SpawnCharactersAccompanyingPlayer))]
+    [HarmonyPrefix]
+    static bool SpawnCharactersAccompanyingPlayerPrefix()
+    {
+        if (CallOriginalPolicy.IsOriginalAllowed()) return true;
+
+        return ModInformation.IsClient == false;
```
garrettluskey:
```suggestion
        return ModInformation.IsServer;
```

garrettluskey (changes requested):

source/Coop.Core/Server/Services/Time/Handlers/CampaignTimeSyncHandler.cs
ShoT-UPfps: updated

source/GameInterface/Services/Locations/Patches/LocationCharacterGuardPatches.cs
ShoT-UPfps: updated

garrettluskey (approved):

PR #1337, opened 2026-06-12
[1273] Gate client network packets while joining a session

### LINKED ISSUE #1273: [Report]Connect_To_Host_Client_Crash_Shinsha Vonkrieg
## Severity of issue/bug

- [x] High
- [ ] Average
- [ ] Low

After trying to connect to the host (bannerlordcoop.duckdns.org) during the testing session, client crashed

## Steps to reproduce

1) Run Start-client.bat
2) Join co-op-campaign (ip = bannerlordcoop.duckdns.org ; port 4200 ; no password)
3) Crash during Loading Screen (after passing please wait for transfer state)

## Additional [information]

[ShinshaVonkrieg_Coop_client.log](https://github.com/user-attachments/files/28671630/ShinshaVonkrieg_Coop_client.log)

file added :
ShinshaVonkrieg_Coop_client.log

ShoT-UPfps:
## PR Checklist

## PR Type

A joining client starts receiving packets before the server forcibly pauses the map which causes handlers to run against a campaign that isn't ready yet.

The server also sends the transferred save from a worker thread after taking the snapshot on the main thread. World changes made in that gap reach the joining client ahead of the save on the same ordered channel, where they are wrongly treated as already part of the snapshot and dropped.

## What is the new behavior?

Resolves #1273

`ClientPacketGate` withholds packets while the client is joining:

- Ordered messages received before the save snapshot (`NetworkGameSaveDataReceived`) are already contained in it and are discarded.
- Messages received after the snapshot, and all unordered packets, are held and replayed in arrival order once the campaign is entered and remote player heroes have been created.
- Replay is driven by a new `ReleaseNetworkBacklog` message published after `ClientCampaignEntered`, so deferred heroes exist before the held packets that reference them are applied.
- The join handshake (validation, save transfer, new-player hero, time/map-event locks, info messages) stays on an allowlist so it is never withheld.

`ICommonSerializer.TryPeekType` classifies a packet from the protobuf wrapper without materializing the payload, so the gate runs no surrogate object-manager lookups against the loading campaign.

The server now sends the transferred save synchronously, in the same main-thread block that takes the snapshot, so no world change made after the snapshot can reach a joining client ahead of the save — keeping the pre-save discard above correct.

One of my hopes is to eventually remove the pause the server enforces while players join. Not sure how feasible it truly is, but this is one of the pre-reqs to that end (and stops a race condition that causes crashes)

this does make the server pause for a little longer than normal when a client joins, but not meaningfully so imo


garrettluskey:
Replaced by #1349

PR #1342, opened 2026-06-12
[1338] [1339] Fix fast-forward spam crashes during client join + make sure players joining at same time get the other player registered

### LINKED ISSUE #1338: Server NRE in MapEvent.Update when a party is destroyed mid-tick under high churn
## What

Under heavy map-event churn the dedicated server dereferences a null/destroyed party while updating a map event and crashes. The log shows dozens of `[TaleWorlds.CampaignSystem.Party.MobileParty] party was null` and `[ObjectManager] Failed to get id for object of type PartyBase`, a `MobileParty` destroyed mid-tick, then an unhandled NullReferenceException:

```
   at TaleWorlds.CampaignSystem.MapEvents.MapEvent.Update_Patch3(MapEvent this)
   at TaleWorlds.CampaignSystem.MapEvents.MapEventManager.Tick()
   at TaleWorlds.CampaignSystem.Campaign.Tick_Patch1(Campaign this)
   at TaleWorlds.CampaignSystem.GameState.MapState.OnMapModeTick_Patch1(MapState this, Single dt)
   at TaleWorlds.CampaignSystem.GameState.MapState.OnTick(Single dt)
   at TaleWorlds.Core.GameStateManager.OnTick_Patch1(GameStateManager this, Single dt)
   at TaleWorlds.Core.Game.OnTick(Single dt)
   at TaleWorlds.Core.GameManagerBase.OnTick(Single dt)
   at TaleWorlds.MountAndBlade.Module.OnApplicationTick(Single dt)
   at TaleWorlds.DotNet.Managed.ApplicationTick(Single dt)
```

## Repro

Host a server with 2+ clients joining and spam fast-forward so many battles resolve at once. Parties are created and destroyed faster than they sync; a map event ticks against a party that is now null and the server crashes.

## Suggested direction

A map event retains a reference to a party that has been destroyed/desynced. Guard the map-event update against null member parties, and/or ensure party destruction removes the party from its map event before the next tick.

Surfaced during live testing of the #1273 client-join fix (PR #1337); related to #1329.

### LINKED ISSUE #1339: TeleportHeroHandler crashes all machines via PartyNameplateVM when hero/party state is incomplete
## What

When a synced hero teleport is applied, `TeleportHeroHandler.Handle_TeleportHero` calls `TeleportHeroAction.ApplyInternal`, which fires `CampaignEventDispatcher.OnHeroTeleportationRequested`. The game invokes it on every party nameplate and `PartyNameplateVM.OnHeroTeleportationRequested` throws a NullReferenceException at its first line — `if (arg1.MapFaction == Party.MapFaction)` — because the teleported hero's faction chain (`Hero -> Clan -> MapFaction`) or the nameplate's own `Party` is not fully built, crashing all connected machines.

```
   at SandBox.ViewModelCollection.Nameplate.PartyNameplateVM.OnHeroTeleportationRequested(Hero arg1, Settlement arg2, MobileParty arg3, TeleportationDetail arg4)
   at TaleWorlds.CampaignSystem.MbEvent`4.InvokeList(EventHandlerRec`4 list, T1 t1, T2 t2, T3 t3, T4 t4)
   at TaleWorlds.CampaignSystem.CampaignEvents.OnHeroTeleportationRequested(Hero hero, Settlement targetSettlement, MobileParty targetParty, TeleportationDetail detail)
   at TaleWorlds.CampaignSystem.CampaignEventDispatcher.OnHeroTeleportationRequested(Hero hero, Settlement targetSettlement, MobileParty targetParty, TeleportationDetail detail)
   at GameInterface.Services.Actions.Handlers.TeleportHeroHandler.Handle_TeleportHero(MessagePayload`1 obj) in C:\Users\Andrew\Desktop\BannerlordCoop\source\GameInterface\Services\Actions\Handlers\TeleportHeroHandler.cs:line 65
```

## Repro

With clients joining and the world fast-forwarded, the server broadcasts a flood of `TeleportHero` messages (~91/sec observed). A teleport is applied referencing a hero or party whose faction/nameplate state is not fully built, and the nameplate event handler NREs — crashing all connected machines.

## Suggested direction

`Handle_TeleportHero` (`source/GameInterface/Services/Actions/Handlers/TeleportHeroHandler.cs:65`) applies the teleport with possibly-null or incompletely-built targets. Validate that the hero's `MapFaction` and party belonging are fully resolved before invoking `TeleportHeroAction.ApplyInternal`, and review whether the receive-path apply should run under `AllowedThread`.

whatever we do, we want to make sure that when the client truly loads in, their sync state aligns with the server. so im concerned about you tlaking about packets being applied to "possibly-null or incompletely-built targets"

Surfaced during live testing of the #1273 client-join fix (PR #1337); related to #1329.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

1. while a client is joining, the server sends it the full world-state stream before that client has even received the save. the client applies those packets against a campaign that isn't loaded yet, and the flood overloads that peer's send queue. so if you spam the fast forward for the entire duration that clients join, the server and all clients will crash shortly after the clients join (or even sooner)
2. if multiple clients are overloaded at once, then it is possible for the game to not automatically resume itself once they are caught up
3. a client only learns that another player exists from a single message sent when that player finishes creating their character. a client that joins later, or the second of two players making a character at the same time, never receives it, so on that client the other player's party/hero/clan aren't registered as player-controlled

## What is the new behavior?
Resolves #1338
Resolves #1339

1. the server now withholds world broadcasts from a peer until it has sent that peer the save (everything broadcast before the save is already inside the snapshot). the joining client stops applying packets against an unloaded campaign and its send queue no longer overloads, so there is no crash
2. game will always resume itself
3. the player registry is sent to each client along with the save, so every player is registered as player-controlled on every client regardless of join order


garrettluskey (dismissed):

garrettluskey:
Would it make more sense to implement this? https://github.com/Bannerlord-Coop-Team/BannerlordCoop/issues/1365

garrettluskey:
3 is specifically handed by RemotePlayerHeroHandler (soon to be replaced with the queue from #1365)

PR #1347, opened 2026-06-12
[1328] Mark the client in-game before adding other players on join

### LINKED ISSUE #1328: ClientCampaignEntered published before ClientLogic.State is assigned
**What**

CampaignState's constructor publishes ClientCampaignEntered (and sends NetworkPlayerCampaignEntered) from inside ClientLogic.SetState<CampaignState>(), which assigns ClientLogic.State only after the constructor returns. Every subscriber that acts on ClientCampaignEntered — including the held-packet replay in ClientPacketGate and the remote-player-hero drain in RemotePlayerHeroHandler — therefore executes while ClientLogic.State still reports LoadingState. As a result, ClientSyncPolicy.AllowOriginal() returns true (treats the client as "not in campaign") during precisely the work that applies buffered server state.

**Why it matters**

Subscribers that gate behavior on the current ClientLogic state (via ClientSyncPolicy or direct State checks) will make the wrong decision when invoked from within the ClientCampaignEntered handler. This can cause buffered packets to be dropped or applied without the intended sync suppression, leading to incorrect client state after join.

**Suggested direction**

Assign ClientLogic.State before running entry side effects, or move the ClientCampaignEntered publish to after SetState completes rather than inside the new state's constructor. Surfaced during code review of the #1273 fix (client join packet gate).

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

The last step of joining a game happens in the wrong order: the other players get added to your world while the game still thinks you're loading. Because "still loading" updates are handled by different rules, the other players can come out wrong — broken or out-of-sync parties on your screen, and `Failed to get id` errors in the log.

## What is the new behavior?
Resolves #1328
Same join, right order: the game first marks you as done loading, *then* adds the other players — so they're handled by the normal in-game rules and come in correctly. Nothing else about joining changes.


ShoT-UPfps:
this is no longer valid due to issue 1349 being solved

PR #1349, opened 2026-06-12
Loading Race Condition Fix

### LINKED ISSUE #1273: [Report]Connect_To_Host_Client_Crash_Shinsha Vonkrieg

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

The server save was being sent potentially before the server was paused causing desync.

## What is the new behavior?

Resolves https://github.com/Bannerlord-Coop-Team/BannerlordCoop/issues/1273

Pausing and saving now happen synchronously.

This also adds a queue for messages while loading.


docs/BattleTroopSpawning.md
```diff
@@ -0,0 +1,291 @@
+# Bannerlord Native Architecture: MapEvents → Battles → Agents
+
```
ShoT-UPfps: did you intend on committing this?

garrettluskey: I did not, should be removed now

source/Coop.Core/Client/Network/LoadingPacketBuffer.cs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    private void Handle_ClientCampaignEntered(MessagePayload<ClientCampaignEntered> payload)
+    {
+        // Defer the drain to the poller thread (DrainIfRequested) so replayed packets stay ordered
+        // relative to live ones and the queue stays single-threaded.
+        drainRequested = true;
+    }
+
+    private void Handle_MainMenuEntered(MessagePayload<MainMenuEntered> payload)
```
ShoT-UPfps: this gets executed when a new player is joining the session and is in character creation because of `ReceivingSavedDataState.cs:Handle_NetwokrGameSaveDataReceived` and the comment below indicates that it's "left the campaign" which is not accurate. not exactly sure if that will lead to a crash immediately or just the original issue that is trying to be solved here, but im pretty sure its one of them haha

source/ServerHeadless/CoopServerLauncher.cs
ShoT-UPfps: does this still exist at this path?

ShoT-UPfps: does this still exist at this path?

ShoT-UPfps: does this still exist at this path?

source/Coop.Core/Server/Connections/States/TransferSaveState.cs
```diff
@@ -15,47 +16,41 @@ namespace Coop.Core.Server.Connections.States;
...
+
+            // Disconnect peer on failure
+            if (!saveResults.Success)
+            {
+                connectionLogic.Peer.Disconnect();
+                return;
+            }
+                
+            network.Send(ConnectionLogic.Peer, savePacket);
+        }, blocking: true);
```
ShoT-UPfps, line 49: worried about what happens if theres a blocknig timeout here. if theres a blocking timeout then the closure still stays on the `RunOnMainThread` queue, the only thing that gets release is the network thread. so the network thread will start consuming packets but the main thread is still gonna be stuck trying to load the user in. eventually the user will load in but the server won't be syncing them

source/Coop.Core/Client/Network/LoadingPacketBuffer.cs
```diff
@@ -0,0 +1,119 @@
...
+            buffering = true;
+            return false;
+        }
+
+        if (!buffering) return false;
+
+        // Infrastructure wrapper is never gameplay state — let it through.
+        if (packet.PacketType == PacketType.PacketWrapper) return false;
+
+        queue.Enqueue((peer, packet));
```
ShoT-UPfps, line 89: theres a race condition here because `buffering` could be true on line `81` but then the main thread in `LoadingPacketBuffer.cs:115` could set it to false anywhere between `LoadingPacketBuffer.cs:82-86`

repro:

network thread sees `buffering=true` on `LoadingPacketBuffer.cs:81` so it continues
main thread sets `buffering=false` on :115 and drains the queue on :117
network thread reaches :86 and enqueues a packet after the queue was already cleared
meanwhile other packets will bypass the queue because `buffering=false`

the result of this is that the player will get the packet in the wrong order

source/Coop.Core/Client/CoopClient.cs
```diff
@@ -167,6 +176,12 @@ public override void Update(TimeSpan frameTime)
     {
         netManager.PollEvents();
 
+        // Replay any packets buffered during the transfer-save load, in order, on this poller thread.
+        foreach (var (peer, packet) in loadingPacketBuffer.DrainIfRequested())
+        {
+            packetManager.HandleReceive(peer, packet);
```
ShoT-UPfps: should we add a try-catch here? if just one packet causes an exception for whatever reason then we lose all of the other packets in the buffer

source/Coop.Core/Client/States/ValidateModuleState.cs
```diff
@@ -108,7 +112,7 @@ public override void Connect()
 
     public override void Disconnect()
     {
-        messageBroker.Publish(this, new EnterMainMenu());
+        gameStateInterface.GoToMainMenu();
```
ShoT-UPfps: if module validation fails then this line runs but doesnt do anything because `Campaign.Current` is already null and `Handle_MainMenuEntered` never runs so their game isn't cleanly terminating the connection

source/Coop.Core/Client/Network/LoadingPacketBuffer.cs
garrettluskey: Now it is cleared on dispose

source/ServerHeadless/CoopServerLauncher.cs
garrettluskey: Replaced with new reflection

garrettluskey: Replaced with new reflection

garrettluskey: Replaced with new reflection

source/Coop.Core/Server/Connections/States/TransferSaveState.cs
garrettluskey: Unlikely to timeout. Will fix if it becomes an issue

source/Coop.Core/Client/Network/LoadingPacketBuffer.cs
garrettluskey: I think this is outdated

source/Coop.Core/Client/CoopClient.cs
garrettluskey: Added

source/Coop.Core/Client/Network/LoadingPacketBuffer.cs
garrettluskey: This only happens on CoopClient.Update which is single threaded

source/Coop.Core/Client/States/ValidateModuleState.cs
garrettluskey: Fixed by moving the finalizer logic into this block

source/Coop.Core/Client/Network/LoadingPacketBuffer.cs
ShoT-UPfps: ya, this is no longer valid

source/Coop.Core/Client/CoopClient.cs
ShoT-UPfps: i don't see it?

garrettluskey: I added it to OnNetworkReceive, which was also needed. Now in both spots

ShoT-UPfps (approved):
glooks

PR #1351, opened 2026-06-12
[1343] Fix phantom troop after escaping captivity

### LINKED ISSUE #1343: [Report] Phantom troop after escaping captivity: party shows 2 troops, only the hero exists
## Severity of issue/bug

- [ ] High
- [x] Average
- [ ] Low

Playing as a client, I was captured by looters while completely alone in my party (no troops). After eventually getting the "you were able to get away" escape message, my party came back wrong:

- The party nameplate/tooltip shows **Troops (2), Infantry 2** (and party speed reflects 2 men).
- Holding Alt for the detailed breakdown lists **only my character's name** — there is no second troop to see.

So the member roster apparently holds the player hero with a count of 2 (a phantom extra man) instead of 1. The extra man is invisible in the roster UI but is counted everywhere that uses `TotalManCount` (nameplate, speed, presumably food/wages/party size).

## Steps to reproduce
1. Join a co-op server as a client with no troops in your party (just your hero).
2. Get defeated/captured by looters.
3. Wait in captivity until the escape pop-up ("you were able to get away").
4. After release, hover your party / hold Alt: troop count shows 2, but only your hero is listed.

## Additional information
Player release is server-authoritative: `PlayerCaptivityServerHandler.ReleasePlayerFromCaptivity` restores the parked party with `playerParty.AddElementToMemberRoster(playerHero.CharacterObject, 1, true)`.

Two things in that area look like candidates for the double count:

- `ReleasePlayerFromCaptivity` is not idempotent and has no "already released/active" guard (unlike the capture side, where `Handle_PrisonerTaken` skips an already-parked party). It is reachable from two paths for the same captivity — the client-requested release (`Handle_NetworkEndPlayerCaptivityAttempted`) and the server-initiated one (`Handle_PlayerCaptivityEndedByServer`) — and `AddElementToMemberRoster(hero, 1)` on a roster that already contains the hero increments the element to 2 rather than failing. A doubled `EndCaptivityAction` on the escape path would do the same (#1296 was a doubled message in this same flow).
- Alternatively the count could be a roster desync between server and client: capture parks the party via `MemberRoster.Clear()` server-side; if that clear never applied on the owning client, the server's release `+1` lands on a client roster that still has the hero.

Note that the server sees the party as 1 still while the client sees 2.

Image

Related: #1191 (player captivity sync), #1296 (player captivity double message).

### LINKED ISSUE #1350: [Report] Captor party counts the captured player twice in its prisoner roster on clients; phantom prisoner persists after release
## Severity of issue/bug

- [ ] High
- [x] Average
- [ ] Low

When a player hero is captured in co-op, every **client** ends up with the hero counted **twice** in the **captor party's** prisoner roster, while the server counts 1. After the player is released, the `-1` from the server's release leaves a **phantom prisoner element (count 1) in the captor's roster on every client, permanently** — the looter/captor party keeps showing a prisoner it no longer has, and a later re-capture by the same party stacks a fresh duplicate element on top.

This is the same divergence family as #1343, one roster over: #1343 was the *player party's member roster* after escape; this is the *captor party's prison roster* during and after any player captivity.

## Steps to reproduce
1. Join a co-op server as a client and get captured by looters (e.g. attack them and surrender).
2. While captive, inspect the looter party on a client (nameplate/tooltip prisoner count) and compare with the server (`coop.debug.mobileparty.info <captorPartyId>`, id from `coop.debug.mobileparty.list`): client shows the hero counted twice in the prison roster, server once.
3. Escape captivity. The looter party on clients still shows one prisoner; the server shows none.

## Additional information
The double-add is the capture replication applying twice on clients:
1. The server runs native `TakePrisonerAction.Apply` under `AllowedThread` (`TakePrisonerActionPatches`); inside it, `AddPrisoner` → `PrisonRoster.AddToCounts` is published by the AllowedThread-aware `TroopRosterAddToCountsPatch`, so the prison-roster **+1 delta replicates** to all clients.
2. `NetworkTakePrisoner` then makes every client **replay** `TakePrisonerAction.ApplyInternal` locally (`MapEventPartyHandler.Handle_NetworkTakePrisoner`), which calls `AddPrisoner` again — native `AddToCounts` finds the replicated element and increments it to 2.
3. On release, `ReleasePlayerFromCaptivity` removes the hero from the captor's prison roster patches-live; the replicated `-1` takes clients from 2 to 1, stranding the phantom element.

A fix needs to deduplicate the two application paths (e.g. skip or compensate the `AddPrisoner` replay client-side), and tests should assert the **captor** `PrisonRoster` counts on clients after capture and after release — the captivity E2E tests added for #1343 only cover the player party's member roster.

Related: #1343, #1191, #1296.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

if you escape captivity then your party shows 2 members instead of 1. server correctly sees 1

## What is the new behavior?
Resolves #1343
Resolves #1350
it correctly shows 1


source/E2E.Tests/Services/MapEvents/MapEventTestBase.cs
```diff
@@ -419,6 +422,86 @@ protected void DefeatPlayerByBattleStateSync(string playerHeroId, string playerP
...
+    }
+
+    /// <summary>
+    /// Seeds <paramref name="count"/> of <paramref name="characterId"/> into the member roster of the
+    /// party with <paramref name="partyId"/> on the server and every client without triggering sync,
+    /// so each instance starts from the same known state.
+    /// </summary>
+    protected void SeedPartyTroopOnAll(string partyId, string characterId, int count)
+    {
+        void Seed(EnvironmentInstance instance)
```
ShoT-UPfps, line 462: i feel like this isnt idiomatic to the codebase but this pattern is already done at line 236 in this file. if you want me to replace both with a different approach lmk

garrettluskey: I actually don't mind this pattern

source/GameInterface/Services/MapEventParties/Handlers/MapEventPartyHandler.cs
```diff
@@ -218,9 +221,39 @@ private void Handle_NetworkTakePrisoner(MessagePayload<NetworkTakePrisoner> payl
...
+
+            // Park the local copy of the captured player party the same way the server does
+            // (PlayerCaptivityServerHandler.Handle_PrisonerTaken). The native ApplyInternal above only
+            // removes the hero from the member roster while hero.PartyBelongedTo is set, and the
+            // captor's replicated prison-roster add has already nulled it by this point — leaving a
+            // stale hero element that the release's re-add would double. The forfeit is absolute and
+            // idempotent, so the roster converges to the server's parked (empty) state regardless of
+            // what the replay did — including when it threw, which is why the park runs outside its
+            // try. It runs AFTER the replay: native RemoveTroop on an already-emptied roster would
+            // index out of bounds.
```
garrettluskey:
```suggestion
```

source/GameInterface/Services/PlayerCaptivityService/Handlers/PlayerCaptivityServerHandler.cs
```diff
@@ -110,9 +111,16 @@ private void Handle_PrisonerTaken(MessagePayload<PrisonerTaken> payload)
...
-        // it; the captivity-end flow reactivates it.
-        playerParty.MemberRoster.Clear();
-        playerParty.PrisonRoster.Clear();
+        // it; the captivity-end flow reactivates it. The roster forfeit runs under AllowedThread so its
+        // per-index removal deltas are NOT broadcast: every client parks its own copy when it applies
+        // NetworkTakePrisoner (MapEventPartyHandler.Handle_NetworkTakePrisoner), and the deltas — computed
+        // against a server roster that already lost its hero to the silenced native capture — would land on
+        // misaligned indices on the still-stale client rosters. Leader change and IsActive stay outside the
+        // scope (the leader change replicates; IsActive is server-only state).
+        using (new AllowedThread())
```
garrettluskey: This is server side, we do not want to use allowed thread here or the clear will not be broadcast to the clients

garrettluskey (changes requested):

source/GameInterface/Services/PlayerCaptivityService/Handlers/PlayerCaptivityServerHandler.cs
ShoT-UPfps: hehe, this pr isn't ready yet! im in the middle of removing it :D

ShoT-UPfps: removed now

source/GameInterface/Services/MapEventParties/Handlers/MapEventPartyHandler.cs
ShoT-UPfps: removed

garrettluskey (approved):

PR #1361, opened 2026-06-13
[1355] Reserve saved players' ids when allocating a new client's hero

### LINKED ISSUE #1355: Single client in a battle is double-counted: "2 player(s) are in a map event" when only 1 is
## What

When a dedicated host has exactly one client connected and that client's party enters a map event (e.g. engages looters), the fast-forward time-control check reports **2 player(s) are in a map event** instead of 1. Only one player exists in the session.

## Repro

1. Start a dedicated host (stationary server, no party of its own).
2. Connect exactly one client and load a co-op campaign.
3. Move the client's party into contact with a group of looters to start a map event. The encounter menu opens and map speed locks at 1x.
4. Attempt to fast-forward time.
5. Observe the message: **"2 player(s) are in a map event"**.

## Expected vs Actual

| | |
|---|---|
| **Expected** | "1 player(s) are in a map event" |
| **Actual** | "2 player(s) are in a map event" |

Image

## Scope

The counter returns to zero correctly after the battle ends — time control is **not** stuck. The defect is purely an inflated count (off-by-one / double-count) while the single client is in the map event.

Reproduces on the `development` branch with one connected client and an otherwise idle dedicated host.

ShoT-UPfps:
## PR Checklist

## PR Type

In a co-op campaign that already has more than one saved player, loading it and connecting a single client, then taking that client's party into a battle (e.g. looters), made the fast-forward check over-count: attempting to fast-forward reported "2 player(s) are in a map event" when only one player was actually in a battle, and 2x time stayed blocked.

## What is the new behavior?

Resolves #1355

The fast-forward check now reports the real number of players in a map event — a single client in a battle shows "1 player(s) are in a map event" — and 2x time unblocks correctly once no player is in a battle. A newly connecting player is no longer given an identity that already belongs to a previously-saved player, and a rejoining player still gets its own party.


garrettluskey:
How does this fix the fast forward check?

ShoT-UPfps:
> How does this fix the fast forward check?

this issue specifically only happens when there is a saved player that's disconnected and then a second player joins. 

With the existing implementation, there is a bug that causes the second player to get assigned the same mobile party id as the disconnected player. The fast forward check iterates over all the players in `PlayerRegistry` and then uses the player's mobile party id to check if they're in a mapevent. both the disconnected player and the connected player both have entries in `PlayerRegistry` but both of their `MobilePartyId`s are the same, so the check sees them as both being in a `MapEvent` even though its really only the second player thats in a `MapEvent`

This PR fixes that by making sure that a newly connected player gets assigned a unique mobile party id

ShoT-UPfps:
I noticed in my `TransferSave.json` that I had two playesr with the same `HeroId`, ~MobilePartyId`, `ClanId`, ...

this PR is basically a band aid over that. I deleted the save and haven't been able to recreate that issue so it might be fixed. Going to close this for now and if this happens again we have a better idea of where to look

PR #1390, opened 2026-06-14
Prevent freeze due to data corruption race condition

ShoT-UPfps:
## PR Checklist

## PR Type

[i was ignored](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/1295#issuecomment-4666212664)

## What is the new behavior?

locks in objectmanager


garrettluskey (changes requested):
Instead can we convert idObjs to a ConcurrentDictionary and remove the _gate

ShoT-UPfps:
> Instead can we convert idObjs to a ConcurrentDictionary and remove the _gate

i dont think we should remove _gate. objectmanager mutates `idObjs` and `objsIds` together in some places and if there's no lock then there will be a race condition between the two

we could remove the lock in some places though if it only touches `idObjs`

garrettluskey:
Agreed, expand _gate to encompass all mutations

garrettluskey (approved):

PR #1422, opened 2026-06-15
[1418] Run map event creation on main thread

### LINKED ISSUE #1418: Client NRE in MapEventManager.Tick while inside a settlement during battle replication
> 🤖 AI-surfaced during live manual testing of #1324 (location-roster sync); this crash is unrelated to that change.

## What

A co-op **client** crashes with a `NullReferenceException` inside vanilla `MapEventManager.Tick()` while the player is **inside a settlement scene** (loaded into the tavern). In co-op the backgrounded map keeps ticking while a settlement screen is on top (`GameStateManagerPatches.Prefix` → `MapState.OnTick` → `MapEventManager.Tick()`), and that tick walks `_mapEvents` and dereferences a null on a map event that is mid-replication.

```
  Source=TaleWorlds.CampaignSystem
   at TaleWorlds.CampaignSystem.MapEvents.MapEventManager.Tick()
   at TaleWorlds.CampaignSystem.GameState.MapState.OnTick(Single dt)
   at GameInterface.Services.Time.Patches.GameStateManagerPatches.Prefix(GameStateManager& __instance, Single dt) in source/GameInterface/Services/Time/Patches/GameStateManagerPatches.cs:line 41
```

## Evidence

`Coop_client.log` immediately before the crash is a flood of map-event replication — `MapEventParty` / `MapEventSide` / `TroopRoster` creation, `GauntletMapEventVisual` create/destroy, repeated `DynamicSync.PartyBase_Handler … _mapEventSide`, and:

```
[WRN GameInterface.Services.MobileParties.Patches.ParallelRobustnessPatches] Resetting stale HasMapEvent for party Created_2025 in ParallelTickStationaryParties: MapEvent is null
```

i.e. a battle is being synced to the client and map events are being created/destroyed faster than they fully resolve.

## Analysis

Decompiled vanilla `MapEventManager.Tick()`:

```csharp
internal void Tick()
{
    for (int num = _mapEvents.Count - 1; num >= 0; num--)
    {
        if (_mapEvents[num].IsFinalized)
            _mapEvents.RemoveAt(num);
        else if (_mapEvents[num].IsRaid || _mapEvents[num] != MobileParty.MainParty.MapEvent)
            _mapEvents[num].Update();
    }
}
```

The client already **skips** the per-event update calls, so the NRE is not in `Update()`:

- `MapEvent.Update` — `MapEventPatches.PrefixUpdate` returns `false` when `ModInformation.IsClient` (clients never tick a map event locally; the server is authoritative).
- `MapEventParty.Update` — `MapEventPartyUpdatePatch.PrefixUpdate` returns `ModInformation.IsServer` (skips on client).

That leaves the NRE in `Tick()`'s own iteration over `_mapEvents`, on one of:

- `_mapEvents[num].IsFinalized` / `_mapEvents[num].IsRaid` — a **partially-synced** `MapEvent` whose internal side/component state isn't populated yet (the list is being mutated by sync while the tick walks it), or a stale/null entry; or
- `MobileParty.MainParty.MapEvent` — `MobileParty.MainParty` momentarily null during the settlement-scene transition.

So this is the **client-side** sibling of #1338 ("Server NRE in `MapEvent.Update` when a party is destroyed mid-tick under high churn", closed via #1384). That fix guarded the **server** `MapEvent.Update` path; it does not cover the **client** `MapEventManager.Tick` iteration, which still runs every tick because the backgrounded map keeps ticking in settlement scenes.

## Repro

1. Co-op session, 2+ clients.
2. As a client, enter a settlement and take a walk (e.g. into the tavern) so a settlement scene is active and the map is backgrounded.
3. Have AI battles resolving / map events being created and destroyed (high churn, e.g. during fast-forward elsewhere) so the client receives a burst of map-event sync.
4. The client NREs in `MapEventManager.Tick()`.

## Suggested direction

- Guard the **client** map-event tick against null / not-fully-synced map events — either a prefix on `MapEventManager.Tick` that, on the client, skips entries whose required state (`InvolvedParties`/sides) isn't set, or null-guards around the `IsFinalized`/`IsRaid`/`MainParty.MapEvent` accesses — mirroring the existing server-side `MapEvent.Update` robustness.
- And/or ensure a map event isn't present in the client's `_mapEvents` until it is fully synced, and is removed on destroy before the next tick.
- Consider null-guarding `MobileParty.MainParty` in this path for the scene-transition window.

## Notes

Unrelated to #1324 (location-roster live updates); discovered while live-testing that change. Related: #1338, #1341, and the `ParallelRobustnessPatches` map-event robustness work.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Map event creation runs on network thread which can cause race conditions with main thread

## What is the new behavior?

Resolves #1418

It runs on the main thread instead


source/GameInterface/Services/MapEvents/MapEventRegistry.cs
```diff
@@ -52,9 +52,19 @@ public override void OnClientCreated(MapEvent obj, string id)
...
+        // (MapEventManager.Tick) walks every frame. This callback runs on the network thread, so defer
+        // the add to the main thread — matching OnClientDestroyed — so it can't race that iteration and
+        // leave a torn/null slot the tick dereferences.
+        GameLoopRunner.RunOnMainThread(() =>
+        {
+            using (new AllowedThread())
+            {
+                Campaign.Current.MapEventManager.OnMapEventCreated(obj);
+            }
+        });
```
ShoT-UPfps, line 67: i considered moving this into the `using` just a few LOC above but decided against it because we still would need to keep the extra `using (new AllowedThread()` since it may not be in scope by the time the main thread gets to running it

garrettluskey (approved):

PR #1423, opened 2026-06-15
[1417] Apply DynamicSync changes on the main thread

### LINKED ISSUE #1417: Apply DynamicSync changes on the game-loop thread instead of the network thread
## Description

DynamicSync applies received field/property changes **inline on the network (poller) thread** and never marshals them onto the game-loop thread. `GenericHandler.SubscribeNetwork` / `SubscribeNetworkReference` run the apply under `using (new AllowedThread())` with no `GameLoopRunner.RunOnMainThread` (`source/GameInterface/Utils/GenericHandler.cs:82-132`).

Because the game-loop thread is running vanilla campaign code at the same time, every synced setter that mutates a game collection is a cross-thread data race. This is the root cause behind a class of intermittent `Collection was modified` / `AccessViolationException` crashes.

**Confirmed manifestation (#1411):** applying a synced `Hero.StayingInSettlement` change runs the vanilla setter → `Settlement.AddHeroWithoutParty`/`RemoveHeroWithoutParty` → `CollectNotablesToCache()` → `settlement._notablesCache.Clear()/.Add()` on the poller thread, while the game-loop thread enumerates `settlement.Notables` during recruitment (`RecruitmentCampaignBehavior.RecruitVolunteersFromNotable`). Captured in a server run; ~5,500 `StayingInSettlement` messages in 45 s make the collision easy to hit. #1392 (party-visual `AccessViolationException`) is the same family.

## Intended Design

Marshal the DynamicSync apply onto the game-loop thread, matching how every other game mutation in the codebase is applied (e.g. `ArmyPatches.AddMobilePartyInArmy`, the #1392 party-visual handlers):

```csharp
GameLoopRunner.RunOnMainThread(() =>
{
    using (new AllowedThread())
    {
        messageHandler(instance, data);
    }
});
```

in both `SubscribeNetwork` and `SubscribeNetworkReference`.

Design questions to settle:
- **Blocking vs non-blocking `RunOnMainThread`.** Non-blocking preserves throughput but defers applies to the next pump, which can reorder them relative to handlers that still run inline (e.g. AutoRegistry create/destroy) → risk of apply-before-create (`Failed to get id`). Blocking preserves ordering but adds per-message latency on the poller thread at high volume.
- **Ordering vs AutoRegistry create/destroy** — confirm instance creation, destruction, and field applies remain consistently ordered.
- **Performance** at high message volume (the `StayingInSettlement` stream alone is thousands/min).

## Location
- `source/GameInterface/Utils/GenericHandler.cs` — `SubscribeNetwork`, `SubscribeNetworkReference`
- DynamicSync builders, if any apply path bypasses `GenericHandler`.

## Related Issues
- Root cause behind #1411 (settlement `Notables` crash, closed in favor of this).
- Same class as #1392 (party-visual `AccessViolationException`).

## Additional information
- `GameLoopRunner.RunOnMainThread` runs inline if already on the game-loop thread, otherwise enqueues to a FIFO drained each frame on the game-loop thread (`source/Common/GameLoopRunner.cs`).
- `AllowedThread` is `[ThreadStatic]`, so it must be entered **inside** the marshaled lambda (on the game-loop thread), not on the poller thread.
- `MessageBroker.Publish` is synchronous on the calling thread; network receive runs on the `Poller` background task (`CoopNetworkBase`).

## Definition of Done
- [ ] DynamicSync applies run on the game-loop thread.
- [ ] No new ordering regressions vs AutoRegistry create/destroy.
- [ ] XUnit/integration coverage added where feasible without the game runtime.

ShoT-UPfps:
## PR Checklist

## PR Type

DynamicSync runs on the network thread and largely mutates data that could be simultaneously accessed by the main thread

## What is the new behavior?

Resolves #1417

DynamicSync enqueues the mutations to the main thread instead of doing it directly on the network thread


source/GameInterface/Services/MobileParties/Patches/MobilePartyMapFactionPatch.cs
```diff
@@ -0,0 +1,43 @@
+using Common;
```
ShoT-UPfps, line 1: This file catches an unhandled exception that crashed my game when I was testing this. It's not actually related to this change but I fixed it here

garrettluskey (dismissed):

garrettluskey (dismissed):

EgardA (approved):

PR #1424, opened 2026-06-15
[1317] Move a lot of stuff off network thread and onto main thread

### LINKED ISSUE #1317: Network message handlers touch game state and UI off the main thread
## Description

Network messages are delivered on a background poller thread (`CoopNetworkBase` starts a `Poller` that runs `netManager.PollEvents()` on a `Task`, and `MessageBroker.Publish` invokes subscribers synchronously on the calling thread). Handlers that touch game state or UI from there race the main game loop — this is the bug class behind #1274, fixed there for `Handle_NetworkStartAttackMission` by deferring via `GameLoopRunner.RunOnMainThread`.

These call sites still do game-state/UI work directly on the network thread:

- `BattleHandler.Handle_NetworkAttackMissionAttempted` (server) — mutates `mapEvent._sides` via `side.MakeReadyForMission(null)` (`source/GameInterface/Services/MapEvents/Handlers/BattleHandler.cs`)
- `ConversationRequestHandler.Handle_NetworkAllowConversation` — `EncounterManager.RestartPlayerEncounter` / `PlayerEncounter.RestartPlayerEncounter`, which reopen the encounter menu (`source/GameInterface/Services/MapEvents/Handlers/ConversationRequestHandler.cs`)
- `GameMenuRefreshHandler.Handle_RefreshGameMenu` — `GameMenu.SwitchToMenu` (`source/GameInterface/Services/GameMenus/Handlers/GameMenuRefreshHandler.cs`); `RefreshGameMenu` arrives over the network (sent from `SellPrisonersHandler`)
- `ClanNameHandler.Handle(NetworkChangeClanName)` — reads `ScreenManager.TopScreen` and calls `RefreshValues()` (`source/GameInterface/Services/Clans/Handlers/ClanNameHandler.cs`)
- `HeroDataHandler.Handle_HeroChangeName` — reads `ScreenManager.TopScreen` (`source/GameInterface/Services/Heroes/Handlers/HeroDataHandler.cs`); `ChangeHeroName` is published synchronously from network handlers

`ScreenManager`'s layer lists are static and unsynchronized, and its screen-change methods only soft-assert the main thread (`Debug.FailedAssert` is log-only in the shipping client), so off-thread use fails intermittently instead of loudly.

The established fix shape: wrap the game-mutating body in `GameLoopRunner.RunOnMainThread(...)` and re-validate state inside the deferred action (see `BattleHandler.Handle_NetworkMapEventFinalized`). Note that handlers needing to send a network reply after the game work use `blocking: true` (see `BattleHandler.Handle_NetworkMapEventFinalizeAttempted`).

## Design constraints

- **A broker-level `SubscribeOnMainThread` wrapper would be silently garbage-collected.** `MessageBroker.Subscribe` stores subscriptions as `WeakDelegate` over `delegate.Target` (`source/Common/Messaging/WeakDelegate.cs` keeps only `new WeakReference(@delegate.Target)`). A wrapper like `Subscribe<T>(p => GameLoopRunner.RunOnMainThread(() => h(p)))` compiles to a closure referenced only by that weak reference, so after the next GC the subscription goes dead and `Publish` prunes it — the handler silently stops firing mid-session, with no log and no crash. A mechanical broker-level fix must strongly root wrapper delegates and map original→wrapper for `Unsubscribe`.
- **The marshal-then-revalidate preamble is already drifting between hand-rolled copies.** `BattleHandler` alone has three copies, and the `blocking: true` one has no `Campaign.Current` guard while the non-blocking ones do. Converting the handlers above by copy-paste risks reintroducing the deferred-action crash class; a small shared GameInterface helper (defer + `Campaign.Current` guard + structured warning log + try/catch, since `GameLoopRunner.Update` invokes queued actions unguarded) would keep the conversions consistent.

## Related Issues

Found while investigating #1274.

ShoT-UPfps:
## PR Checklist

## PR Type

Many actions that should be running on the main thread are running on network thread which causes race condition issues across the entire game

There were also some places that weren't using AllowedThread when I think they should have.

## What is the new behavior?
Resolves #1317

There's a lot more running on main thread. I didn't see any performance issues locally though

Added AllowedThread in places where I thought it was missing


source/GameInterface/Services/Companions/Handlers/HireCompanionHandler.cs
```diff
@@ -59,16 +61,30 @@ private void Handle_CompanionHired(MessagePayload<CompanionHired> obj)
...
-        if (!objectManager.TryGetObjectWithLogging<MobileParty>(obj.What.MainPartyId, out var mainParty)) return;
+        var data = obj.What;
+        var peer = obj.Who as NetPeer;
 
-        GiveGoldAction.ApplyBetweenCharacters(mainHero, oneToOneConversationHero, obj.What.HiringPrice, false);
-        AddCompanionAction.Apply(playerClan, oneToOneConversationHero);
-        AddHeroToPartyAction.Apply(oneToOneConversationHero, mainParty, true);
+        // The hire applies vanilla game actions; defer them to the game-loop thread so they
+        // run there instead of on the network (poller) thread that delivered the message.
+        GameLoopRunner.RunOnMainThread(() =>
```
garrettluskey: Add RunOnMainThreadSafe which wraps the call in a try catch with logging. Don't bother replacing but we will try to use this for manual applications

source/GameInterface/Services/PartyComponents/BanditPartyComponents/Handlers/BanditPartyComponentHandler.cs
```diff
@@ -61,26 +63,36 @@ private void Handle(MessagePayload<NetworkUpdateBanditPartyComponent> payload)
     {
         var obj = payload.What;
 
-        if (objectManager.TryGetObject<BanditPartyComponent>(obj.ComponentId, out var component) == false)
+        GameLoopRunner.RunOnMainThread(() =>
```
garrettluskey: Can we rename GameLoopRunner.RunOnMainThread to GameThread.Run

garrettluskey (changes requested):

source/GameInterface/Services/PartyComponents/BanditPartyComponents/Handlers/BanditPartyComponentHandler.cs
ShoT-UPfps: done

source/GameInterface/Services/Companions/Handlers/HireCompanionHandler.cs
ShoT-UPfps: Added `RunSafe` since it was renamed to `Run` per the comment below

source/GameInterface/Services/PartyComponents/BanditPartyComponents/Handlers/BanditPartyComponentHandler.cs
ShoT-UPfps: There was also a `SetGameLoopRunnerThread` fxn that ir enamed to `SetGameThreadThread` since we renamed `GameLoopRunner` to `GameThread` ..

open to `SetGameThread` instead but neither feels right 🤷

garrettluskey: MarkGameThread?

garrettluskey (approved):

PR #1482, opened 2026-06-19
Sync TroopRosters by character identity instead of array index

ShoT-UPfps:
## PR Type

- [x] Bug fix

we sync live TroopRoster changes with per-index deltas, but client rosters start empty, so an index-keyed delta lands past the end of the client array and vanilla throws `IndexOutOfRange` on basically every roster each frame (the `Failed to AddToCountsAtIndex` storm, #1516).

## What is the new behavior?

sync each live change as a delta keyed by the troop's character instead of its array index, and apply it on the client with vanilla `AddToCounts`/`SetElement*` found by character. a positive `AddToCounts` creates the troop if the client doesnt have it yet; the absolute `SetElement*` sets need it to already exist and skip otherwise, because a placeholder would corrupt the cached totals and get wiped by the next `RemoveZeroCounts`. no index crosses the wire so it cant land out of range, and it sends straight off the server's patches so it replicates on a headless host too. same shape as our `ItemRoster` sync. the patch only fires for a roster the registry knows, so a battle's scratch tally rosters (the per-party died/wounded/routed dummies) dont churn the game thread or spam `Failed to get id`.

this only swaps the live per-op sync. the rest of the TroopRoster work thats on development stays as-is: the owner-party sync, the dummy-roster registration that fixes the encounter/tooltip `Failed to get id` (#1501), and the party-screen done-logic delta (`PackTroopRosterDelta`/`ApplyTroopRosterDeltas`), which runs on its own path. `TroopRosterInterface` keeps both that done-logic delta and the snapshot pack/unpack, so nothing on development loses its sync. the only change i make to the interface is scoping the `UpdateWithData` companion-preserve to a party's own member roster, so a server-freed companion no longer stays stuck as a prisoner on the client.

## Other information

how to test: a server + a client in one co-op campaign. the `coop.debug.mobileparty.*` cheats drive the delta paths (`addtroops`/`addtroopxp`/`addprisoners`/`removeprisoners` on the server, `characterids` to read either side), plus party-management, selling prisoners and recruiting in-game for the screen flows.


garrettluskey:
Moving to draft state until clan management is merged

ShoT-UPfps:
> Moving to draft state until clan management is merged

wasn't this already in draft? 🫪

garrettluskey:
Solved with #1515

ShoT-UPfps:
not solved. reopened

source/GameInterface/Services/TroopRosters/Interfaces/TroopRosterInterface.cs
```diff
@@ -108,18 +108,33 @@ public IEnumerable<TroopRosterElement> UnpackTroopRosterData(TroopRosterData tro
...
-        // Clear without removing MainHero (causes issues if MainHero is removed)
+        // Preserve the local main hero and the player's companions across a whole-roster replace, but only
+        // when replacing a party's own MEMBER roster: removing the main hero breaks the player party, and clan
+        // equipment/inventory work relies on companions persisting there. Any other roster, a prison roster or
+        // an ownerless prisoner-management roster, must mirror the server exactly, otherwise a companion the
+        // server freed, ransomed, or sold stays stuck as a prisoner on the client. Detecting the member roster
+        // affirmatively (rather than "not the prison roster") also fails safe for an ownerless roster, where
+        // OwnerParty is null. The mainHero != null guard also stops a null mainHero matching a basic troop's
+        // null HeroObject and wrongly preserving it.
+        bool preserveHeroes = mainHero != null && targetTroopRoster.OwnerParty?.MemberRoster == targetTroopRoster;
```
jordanbrymora: Would you be okay with me moving this fix to #1537 to replace what I already have there? I saw this after I created my own but this solution is better.

ShoT-UPfps: sounds good to me

garrettluskey (approved):

PR #1526, opened 2026-06-21
[1491] Show a loading screen while the host applies patches

### LINKED ISSUE #1491: Show Applying Patches when Server is Loading
# For New Contributors

## Description

The server looks like it freezes while loading, add relevant loading screens

## Location

## Definition of Done

ShoT-UPfps:
## PR Checklist

## PR Type

Hosting freezes the menu for ~30s with no notice while the coop Harmony patches and the
one-time sync-patch compile run, so it looks like the game hung.

## What is the new behavior?
Resolves #1491

The host now applies its Harmony patches off the main thread before loading the save, so the
game thread keeps rendering instead of freezing. During that window a coop loading screen shows
"Applying patches..." over the menu, then the normal save load takes over with
"Loading campaign save..." into the map. The engine's own loading window doesn't paint over the
main menu, so the patch-phase screen is a small mod-owned layer. Patches run before the load, so
the save deserializes with them live.

Headless is unchanged: no loading window, patches applied synchronously.

https://github.com/user-attachments/assets/f18c4abd-034a-43d2-b1fe-4c7126dce8a1


source/Coop.Core/CoopartiveMultiplayerExperience.cs
garrettluskey, line 1: These changes add complexity, please revert

source/GameInterface/Services/LoadingScreen/Interfaces/CoopTextLoadingScreen.cs
garrettluskey, line 1: Any reason we can't just use the default loading screen?

garrettluskey (changes requested):

source/GameInterface/Services/LoadingScreen/Interfaces/CoopTextLoadingScreen.cs
ShoT-UPfps: I tried to make it work with that but for some reason the ScreenManager was refusing to render the loading screen as is prior to applying patches. The option I was seeing was making it sleep for X frames before starting to apply the patches which seemed like a worse alternative

source/Coop.Core/CoopartiveMultiplayerExperience.cs
ShoT-UPfps: I wasn't able to find any good way to do it without handling it on a separate thread. Will close this PR and see if someone else can find a different path

source/GameInterface/Services/LoadingScreen/Interfaces/CoopTextLoadingScreen.cs
garrettluskey: Ah I see, we ran into that issue as well. Not sure if the issue this is attached to is worth pursuing in that case. Thanks for the help

PR #1529, opened 2026-06-22
[1450] Stop client re-applying battle results; replicate clan renown

### LINKED ISSUE #1450: Player-battle renown and recent-events morale desync between host and client
## Description

In co-op, after a player wins a field battle, that player's clan **renown** and party **recent-events morale** (`MobileParty.RecentEventsMorale`) don't match between the host and that player's client. Troop xp and clan influence are unaffected (they stay in sync).

Two things cause it:

1. **The client re-applies the battle result locally.** The host is authoritative for battle resolution, but the client's `PlayerEncounter` post-battle path is unpatched and re-runs `MapEvent.CalculateAndCommitMapEventResults` on the client. That re-applies the result commit (recent-events morale, renown, etc.) on the client's own copy, on top of what the host computed, so the values drift.

2. **Clan renown never replicates to clients.** `Clan.Renown` is a trivial auto-property; its setter gets JIT-inlined into its writers (`Clan.AddRenown` / `ResetClanRenown`), so the AutoSync setter hook never fires and the host's renown changes are never sent. The client's renown is then only whatever it computed itself, which diverges from the host. (Influence stays in sync because it's synced as a field, and recent-events morale is also sent during the round simulation; those don't hit the renown replication gap, but recent-events morale still diverges because of cause 1.)

Observed: in one run the client's renown read about half the host's, and recent-events morale differed between the two by a varying amount.

## Suggested direction

- Stop the client re-applying the battle-result commit; it should keep the host's replicated result.
- Replicate clan renown from the host's writer methods (`Clan.AddRenown` / `ResetClanRenown`) instead of the inlined property setter.

## Location

- `source/GameInterface/Services/MapEvents/Patches/MapEventPatches.cs` — the client now skips the battle-result commit (`CommitCalculatedMapEventResults`) so the host stays authoritative; the loot-staging steps that run before it are left intact.
- `source/GameInterface/Services/Clans/Patches/ClanRenownPatch.cs` — publishes the host's new renown from the writer methods (`Clan.AddRenown` / `ResetClanRenown`), replacing the AutoSync property hook that never fired.
- `source/GameInterface/Services/Clans/ClanSync.cs` — `Clan.Renown` was registered as a synced property, but its inlined setter means the hook never ran; that registration is removed.

## Related Issues

Found while investigating #1318.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

After a player wins a battle, that player's clan renown and party recent-events morale don't match between the host and that player's client. The client's renown ends up wrong and recent events morale differs. Troop xp and influence are unaffected.
image

## What is the new behavior?

Resolves #1450

The host stays authoritative for battle rewards. The client no longer re-applies them locally and the clan renown the host awards now replicates to clients. After a battle the player's renown and recent-events morale match the host.

desync fixed

Also adds a debug console command used to test this: `coop.debug.clan.economy [clanId]` (prints a clan's renown, influence, recent-events morale, and troop xp).

### DIFF
--- a/source/GameInterface/Services/Clans/Handlers/ClanRenownHandler.cs
+++ b/source/GameInterface/Services/Clans/Handlers/ClanRenownHandler.cs
@@ -6,7 +6,6 @@ using Common.Util;
 using GameInterface.Services.Clans.Messages;
 using GameInterface.Services.ObjectManager;
 using Serilog;
-using System;
 using TaleWorlds.CampaignSystem;
 
 namespace GameInterface.Services.Clans.Handlers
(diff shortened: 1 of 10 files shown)


source/GameInterface/Services/MapEvents/Patches/MapEventPatches.cs
```diff
@@ -114,6 +114,23 @@ private static bool Prefix_OnBattleWon(MapEvent __instance)
...
+            return true;
+
+        // The server commits the battle economy (troop xp, renown, influence, morale, gold) during OnBattleWon
+        // and replicates it. The client's encounter result path reaches this commit locally too (both when a
+        // battle ends and when a multi-round battle continues) and re-applies it on top of the server's
+        // already-received values with no later resync, drifting the player's clan renown/influence, the party's
+        // morale, and surviving troops' xp above the server. (Hero gold and prisoner capture are already blocked
+        // on the client by the GiveGoldAction/TakePrisonerAction patches.) Skip the commit on the client so the
+        // economy stays the server's; the loot-staging steps that run before it are left intact.
+        return !ModInformation.IsClient;
```
garrettluskey:
```suggestion
        return ModInformation.IsServer;
```

source/GameInterface/Services/Clans/Handlers/ClanRenownHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            var payload = obj.What;
+
+            GameThread.RunSafe(() =>
+            {
+                if (!objectManager.TryGetObject<Clan>(payload.ClanId, out var clan))
+                {
+                    Logger.Error("Unable to find clan ({clanId}) for renown change", payload.ClanId);
+                    return;
+                }
```
garrettluskey:
```suggestion
                if (!objectManager.TryGetObjectWithLogging<Clan>(payload.ClanId, out var clan))
                    return;
```

garrettluskey (changes requested):

garrettluskey (approved):

Issue #1539, opened 2026-06-24
[ISSUE] Server GC deadlock from AutoSync detouring tiny methods and corrupting x64 unwind info


ShoT-UPfps:
# For New Contributors
[How to contribute](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/wiki/Getting-Started-as-a-Contributor)

## Description

The co-op **server** can hard-freeze after a large battle. The game thread wedges permanently, the network heartbeat keeps running, and Visual Studio can't even break in ("debugger timed out trying to pause, process is in a broken state"). It surfaced while live-testing the TroopRoster delta PR (#1482) but is not caused by it, this is a pre-existing fault in how our AutoSync Harmony patches are applied.

Root cause, in plain terms:

- Harmony patches a method by overwriting the **first 5 bytes** of its compiled code with a `jmp` to our patched version. That's fine for normal-sized methods.
- Some methods are tiny. `Clan`'s `IFaction.get_IsKingdomFaction` is literally `=> false`, which compiles to a **3-byte** stub. For a stub that small, the CPU's stack-unwind metadata (x64 `.xdata` / unwind info, the table the runtime uses to walk the stack) sits **inline, right after those 3 bytes**.
- So our 5-byte `jmp` writes 3 bytes of code **plus 2 bytes of that adjacent unwind info**, corrupting it.
- The workstation GC suspends every thread and walks each stack. The campaign-map nameplate render (`PartyNameplatesVM.UpdateNameplatesInRange`) runs a native parallel-for over the map's parties every frame (this save had ~1548 mobile parties on the **map**, measured from the dump, not in the battle), so a worker thread is routinely sitting at the entry of one of these detoured accessors. If a GC lands in that window, the unwinder reads the garbage unwind info on the corrupted method, raises an exception, and the exception machinery has to **allocate** an object, mid-GC. Allocating mid-GC blocks waiting for the GC to finish, but the GC is the thread that's blocked. Self-deadlock, un-interruptible. A big battle just raises the odds by driving more allocation and therefore more frequent GCs; it doesn't add parties to the loop.

This is fully grounded in the crash dump (cdb/SOS) and a standalone static harness, both attached. The corrupt method's `.fnent` shows garbage ("version 6", a flood of unknown unwind opcodes); a normal method shows clean "version 1". The harness independently measures `get_IsKingdomFaction` at code length 3 with its unwind info 4 bytes in, inside the 5-byte write.

[...]

PR #1542, opened 2026-06-24
Reduce server freezes (GC deadlocks) by skipping detours that corrupt method unwind info

ShoT-UPfps:
## Testing

influence syncs correctly
influence syncing

renown syncs correctly
renown syncing
renown syncing clan page

gold syncs correctly
gold syncing

hp syncs correctly
hp syncing

party wage limit syncs correctly
party wage syncing

recruiting and xp syncs correctly
recruit and xp syncing

queued buildings on settlements syncs correctly
queued buildings syncing

buying workshops syncs correctly
buying workshop syncing

governor + prosperity + security + loyalty syncs correctly
governor syncing (og governor is Honoratus)

## PR Type

- [x] Bug fix

co-op detours a lot of methods on synced types, including tiny stubs the patch never actually rewrites. the generated `*_DynamicPatches` classes (autosync) blanket-detour every declared method of a synced type, and the hand-written collection patches (HeroCollectionPatches etc.) do the same.

that's what froze the server. harmony's inline detour overwrites the first 5 bytes of a method, and for a method whose native body is shorter than that (a constant getter like clan's `IFaction.IsKingdomFaction => false`, a 3-byte stub with its inline x64 unwind info right after the code) the jmp spills past the code into the unwind info and corrupts it. when the GC suspends a thread parked at that entry and `RtlVirtualUnwind` reads the garbage, the CLR allocates an exception object mid-GC and self-deadlocks in `gc_heap::wait_for_gc_done`. un-interruptible, VS cant even break in. root-caused from a server dump in #1539.

## What is the new behavior?

a single harmony prefix on `PatchTools.DetourMethod`, the one chokepoint every patch funnels through to write the 5-byte detour. it skips that write only for a method that is all three of:
- **fragile**: its detour would overwrite its own inline x64 unwind info (`.xdata` within 5 bytes of the entry, measured with `RtlLookupFunctionEntry`), and
- a **no-op**: it does no field store, so no field transpiler can actually rewrite it, and
- has no prefix/postfix/finalizer.

skipping that detour leaves the method running its original code, which is identical behavior since the detour rewrote nothing, but with its unwind info intact. the tiny stubs that corrupted the unwind info (constant getters, `=> false`, forwarders) stop being detoured; nothing else does.

because it sits at the byte-write chokepoint it covers every patch the same way, the autosync-generated ones and the hand-written collection patches both, with no per-class wiring. windows x64 only: the fragility check returns false where the unwind API isnt available (linux/CI), so off that platform the guard is inert.

fail-safe: if a method body cant be cleanly decoded, or it stores a field, or it has a prefix/postfix, the detour is kept. so the guard only ever skips a detour it has positively proven is a fragile no-op, it never drops a load-bearing one.

## Other information

related to #1539 (full root-cause writeup + dump). this also covers the hand-written collection patches, so the follow-up that was split out as #1544 is no longer needed.

verified:
- GameInterface compiles clean against the real game refs (0 errors).
- guard unit tests (`GameInterface.Tests/DynamicSync/FragileDetourGuardTests.cs`, 10 cases) green: the skip decision keeps a fragile field-storer and a fragile method that has a prefix, skips a fragile no-op stub, and never skips a non-fragile method; `DoesFieldStore` and the windows/non-windows contract are pinned.
- full E2E green. the guard patches harmony's own `DetourMethod` first, then every co-op patch flows through it, so a green run confirms it doesnt break patch application. the guard is inert on linux so CI exercises the safe path.
- checked against the real game DLLs: across every synced type the guard skips the freeze stubs (incl `IFaction.get_IsKingdomFaction`) and drops zero load-bearing methods.


source/GameInterface/AutoSync/Templates/Patches/DynamicPatchTemplate.txt
```diff
@@ -27,18 +27,22 @@ public class {{DeclaringTypeName}}_DynamicPatches : GenericPatches<{{DeclaringTy
...
+        // deadlocking the GC. SelectTargetMethods narrows to the referencing methods, plus the explicit
+        // targets below, and falls back to the declared methods if the narrowing matched nothing.
+        var syncedMembers = GenericPatchHelpers.GetSyncedMemberTokens(
+            typeof({{DeclaringType}}),
+            new string[] { {{ for fieldName in SyncedFieldNames }}"{{ fieldName }}", {{ end }} },
+            new string[] { {{ for fieldName in SyncedCollectionFieldNames }}"{{ fieldName }}", {{ end }} },
+            new string[] { {{ for propertyName in SyncedPropertyNames }}"{{ propertyName }}", {{ end }} });
 
-        foreach (var method in AccessTools.GetDeclaredMethods(typeof({{DeclaringType}})).Where(m => !m.IsAbstract))
+        return GenericPatchHelpers.SelectTargetMethods(typeof({{DeclaringType}}), syncedMembers, new MethodBase[]
```
garrettluskey: Why not just pass the selected methods as TargetMethods?

garrettluskey (dismissed):

source/GameInterface/AutoSync/Templates/Patches/DynamicPatchTemplate.txt
ShoT-UPfps: I reverted all updates to Autosync as the root cause isn't really to do with auto sync but only to do with Harmony patching.

The tl;dr is that some setters/getters compile into <5 bytes of assembly instructions. Harmony patching replaces it with a `jmp` instruction that is 5 bytes. Since the setter/getter is <5bytes and `jmp` is 5 bytes, the `jmp` takes more space up than the original instructions did, and overwrites memory that is used for exception handling and generating stack traces. 

The garbage collector uses that memory to safely suspend threads in order to perform garbage collection. But if theres a thread that's on one of the instructions with messed up exception handling and stack traces, it throws an exception and GC never completes, so the server freezes.

The new approach prevents Harmony from patching methods that compile to <5 bytes _as long as there is no prefix/postfix/finalizer/field store_. If there IS a prefix/postfix/finalizer/field store then harmony is allowed to do the detour and corrupt the additional memory. But my hope here is that this is enough to substantially reduce the frequency of the server freezing because this is only an issue when GC happens while a thread is on a corrupted instruction, and with this new approach there should be substantially lesser instances of corrupted instructions

This should also only be an issue on Windows because Linux apparently pads the memory that the OG function was at and the exception+stack trace memory is elsewhere

source/GameInterface/Utils/FragileDetourGuard.cs
```diff
@@ -0,0 +1,226 @@
...
+
+            var patchTools = typeof(Harmony).Assembly.GetType("HarmonyLib.PatchTools");
+            var detourMethod = patchTools == null ? null : AccessTools.Method(patchTools, "DetourMethod");
+            if (detourMethod == null)
+            {
+                Logger.Error("FragileDetourGuard: HarmonyLib.PatchTools.DetourMethod not found");
+                return;
+            }
+
+            harmony.Patch(detourMethod, prefix: new HarmonyMethod(typeof(FragileDetourGuard), nameof(DetourMethodPrefix)));
```
garrettluskey, line 64: I am hesitant to patch harmony methods directly. I have a solution in mind to scan all BL dlls and patch only the required methods. That may not fix the root issue here but adding something in that idea may work

garrettluskey (dismissed):

source/GameInterface/Utils/FragileDetourGuard.cs
ShoT-UPfps: Agreed that this approach isn't great but we are stuck in a pretty shitty situation with this one already.

I already attempted to "patch only the required methods" without patching Harmony by updating Autosync itself but reverted it. When I stopped Autosync from overpatching, the JIT started inlining those methods now that nothing was patching them and we had custom patches break because they rely on the methods not being inlined. It's hard to tell how many custom patches we have which rely on Autosync's current side effect of overpatching.

That's why I ended up on this approach where it only skips patching on functions that are verifiably unused by any of our patching, be it autosync or custom patches. 

I think this is the least invasive thing we can do right now to minimize the issue. My 2 cents would be to merge this and then we can revisit it later when we have more of an appetite to break things and have testers catch it (and perhaps also disable inlining)

garrettluskey (approved):

PR #1545, opened 2026-06-24
[1449] Fix race condition with campaign map battle sound 

### LINKED ISSUE #1449: [ISSUE] Populate a client map event's sides before its battle visual initializes (correct ambient battle-size)
# For New Contributors

## Description

On a client, a map event is created **before** its sides are wired up. `MapEventRegistry.OnClientCreated` sets `_sides = new MapEventSide[2]` (both elements null), and the sides are populated later by a **separate** message (`NetworkAssignMapEventSide` → `mapEvent._sides[side] = …`). The replicated battle-visual init (`NetworkGauntletMapEventVisualInitialized`) is **not** gated on the sides being populated, so it can run first.

When it does, vanilla `GauntletMapEventVisual.Initialize` reaches its field-battle / sally-out branch and calls `GetBattleSizeValue()` → `MapEvent.GetNumberOfInvolvedMen()`, which reads the **real headcount** of both sides to choose the field-battle ambient sound's `battle_size` parameter.

The stopgap shipped for #1426 guards `GetBattleSizeValue` so that, while the sides aren't populated yet, it returns the smallest size (`0`) instead of dereferencing the null sides. That stops the `NullReferenceException` and lets the ambient sound play — but `Initialize` is **one-shot**, so the size is **stuck at `0`** for the life of that map event and never reflects the real battle once the sides finish syncing. A large battle that lost the race gets a too-quiet ambient for its whole duration.

This issue tracks the **durable fix**: ensure a client map event's sides are populated before its visual initializes (or refresh the ambient sound once they are), so the ambient battle-size is correct rather than stuck at the stopgap default.

## Intended Design

Candidate approaches (to be weighed during design):

1. **Client-side readiness gate + deferred init (recommended).** When `NetworkGauntletMapEventVisualInitialized` is handled and the map event's sides aren't both assigned, hold the pending init (keyed by map-event id) instead of running it immediately, and flush it once both sides are assigned (trigger from where `NetworkAssignMapEventSide` completes the second side). Keeps the change client-local and yields a correct size without re-implementing vanilla sound setup.
2. **Decouple the icon from the sound.** Let `Initialize` run immediately for the battle icon (which already works), and (re)build the ambient sound separately once the sides are populated. Keeps the icon instant but is more invasive — it re-implements the vanilla sound setup outside the sealed `Initialize`.
3. **Server-side ordering.** Ensure the side-assignment messages are sent and applied before the visual-init message, and preserve that order through the client game-loop FIFO. Fragile on its own; would likely still need a client-side readiness gate as a backstop.

Design tradeoff to settle: the battle **icon** currently appears *before* the failing line in `Initialize`, so deferring the **whole** init until the sides are ready would delay the icon (a regression from current behavior). The fix should keep the icon instant while getting the sound's size right — e.g. option 1 with the icon handled immediately, or option 2.

## Location

- `source/GameInterface/Services/GuantletMapEventVisuals/Handlers/GauntletMapEventVisaulHandler.cs` — the replicated visual-init handler (`Handle_NetworkGauntletMapEventVisualInitialized`).
- `source/GameInterface/Services/GuantletMapEventVisuals/Patches/GauntletMapEventVisualPatches.cs` — the stopgap `GetBattleSizeValue` guard that defaults the size to `0`.
- `source/GameInterface/Services/MapEventSides/Handlers/MapEventSideDataHandler.cs` — `NetworkAssignMapEventSide` apply, where a side becomes ready (the natural flush trigger).
- vanilla `SandBox.GauntletUI.Map.GauntletMapEventVisual.Initialize` / `GetBattleSizeValue`, `TaleWorlds.CampaignSystem.MapEvents.MapEvent.GetNumberOfInvolvedMen`.

## Related Issues
Follow-up to #1426 (the stopgap that defaults the ambient size to `0` during the race).

## Requirements
- After a client joins a field battle while the map event's sides are still syncing, the field-battle ambient sound reflects the **real** battle size (not stuck at the smallest), with no `NullReferenceException`.
- The battle icon still appears immediately (no regression vs. the current behavior).

## Additional information
- The codebase already documents this same not-yet-populated-sides window for the client in `MapEventVisibilityClientPatch` (guards `mapEvent?.AttackerSide == null || mapEvent.DefenderSide == null`).
- The race is more likely now that DynamicSync applies run on the game-loop thread, which can move the side-population after the visual init in the game-loop queue.

## Definition of Done

ShoT-UPfps:
## PR Checklist

## PR Type

when you are near a battle on the map its supposed to emit a noise of swords clashing and people yelling. the volume is supposed to be scaled by the number of troops in the battle but right now based on a race condition (mapevent sides not loaded yet) it could be the lowest volume when it should not be

## What is the new behavior?

Resolves #1449
 volume of battle on campaign map always scales with number of troops in the battle


source/GameInterface/Services/GuantletMapEventVisuals/MapEventBattleSizeCorrection.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace GameInterface.Services.GuantletMapEventVisuals;
+
+/// <summary>
+/// Client-side: re-applies a field battle's ambient <c>battle_size</c> when its visual initialized before
+/// the sides/parties synced. Vanilla <see cref="GauntletMapEventVisual"/> bakes the size into the sound once
+/// (the #1426 guard defaults it to 0 while un-synced and never updates it), so without this a large battle
+/// stays too quiet for its whole life. We re-apply the real size to the existing sound as the parties arrive.
+/// </summary>
+internal static class MapEventBattleSizeCorrection
```
garrettluskey: This may conflict with the current battles arch. Maybe we should hold off

garrettluskey (changes requested):

source/GameInterface/Services/GuantletMapEventVisuals/MapEventBattleSizeCorrection.cs
ShoT-UPfps: I'll mark this as draft and leave it for now, once the battles arch is completed lmk

source/GameInterface/Services/GuantletMapEventVisuals/MapEventBattleSizeCorrection.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace GameInterface.Services.GuantletMapEventVisuals;
+
+/// <summary>
+/// Client-side: re-applies a field battle's ambient <c>battle_size</c> when its visual initialized before
+/// the sides/parties synced. Vanilla <see cref="GauntletMapEventVisual"/> bakes the size into the sound once
+/// (the guard defaults it to 0 while un-synced and never updates it), so without this a large battle
+/// stays too quiet for its whole life. We re-apply the real size to the existing sound as the parties arrive.
+/// </summary>
+internal static class MapEventBattleSizeCorrection
```
garrettluskey: Can we make this not static and implement an interface?

garrettluskey (changes requested):

source/GameInterface/Services/GuantletMapEventVisuals/MapEventBattleSizeCorrection.cs
ShoT-UPfps: updated

garrettluskey (approved):

PR #1549, opened 2026-06-25
[26] Fix freed prisoner unable to move after a save/reload, and parties stuck chasing a removed target

ShoT-UPfps:
## PR Checklist

## PR Type

If the game saves when a client is prisoner and the server reloads, then when the client reconnects and frees itself from the imprisoning party, it is unable to move

https://github.com/user-attachments/assets/a1ca9f8e-21bc-490f-b13d-4aeca6815d13

This is one symptom of a wider problem: after a save/reload (and on a fresh client join) a party can come back still set to chase a party that no longer exists. So as well as the freed prisoner being stuck, any party that was mid-engagement when the save happened (the looters/bandits that caught you, nearby AI lords) stops moving and spams GetTargetCampaignPosition errors in the log.

## What is the new behavior?

Resolves Bannerlord-Coop-Team/Bannerlord-Coop-Issues#26

The prisoner client is able to move after freeing itself. Parties left chasing a party that no longer exists now just stop and re-decide instead of getting stuck or spamming errors, so the log stays clean after a reload or a client join.

https://github.com/user-attachments/assets/e1f43d18-067b-4997-8a55-e1fa3cce3b0e

### DIFF
--- a/source/GameInterface/Services/MobileParties/Extensions/MobilePartyExtensions.cs
+++ b/source/GameInterface/Services/MobileParties/Extensions/MobilePartyExtensions.cs
@@ -58,17 +58,12 @@ public static class MobilePartyExtensions
     }
 
     /// <summary>
-    /// Clears a party's navigation orders so it stops at its current position (and is no longer classified
-    /// as moving), then asks the AI to re-decide at the next hourly tick. This is the navigation half of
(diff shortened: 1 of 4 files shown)


EgardA (approved):
Looks fine to me!

Issue #1465, opened 2026-06-18
Replicate the backstreet-menu mercenary hire (buy_mercenaries_on_consequence)


ShoT-UPfps:
## Context

PR #1463 (issue #1309) replicates the tavern **dialogue** mercenary hire — `RecruitmentCampaignBehavior.BuyMercenaries`, reached via `conversation_mercenary_recruit_accept(_some)_on_consequence`.

Vanilla has a **second, independent** mercenary-hire path that PR #1463 does not cover: the town **backstreet** game menu option **"Recruit N {troop}"** (`recruit_mercenaries`, added in `RecruitmentCampaignBehavior.AddGameMenus`) → `RecruitmentCampaignBehavior.buy_mercenaries_on_consequence`. This method does **not** call `BuyMercenaries`; it directly does `MobileParty.MainParty.MemberRoster.AddToCounts(mercenaryData.TroopType, n)`, `GiveGoldAction.ApplyBetweenCharacters(null, Hero.MainHero, -(n * cost))`, and `mercenaryData.ChangeMercenaryCount(-n)`.

The backstreet menu is reachable in co-op (it is entered from the town menu's "Go to the tavern district" game-menu transition, which `DisablePlayerTownVisitCampaignBehavior` does not gate).

## Problem

A mercenary hire made through the backstreet menu mutates the conversing client's party roster and gold **locally only** — no `MercenariesHired` event fires, so the server and other clients never see the troops or the gold change (desync).

## Suggested fix

Patch `RecruitmentCampaignBehavior.buy_mercenaries_on_consequence` with the same client→server forward PR #1463 added for `BuyMercenaries` (capture troop type / count / gold on the client, suppress the local apply, have the server apply `AddElementToMemberRoster` + `GiveGoldAction` with patches live). Note this path adds `mercenaryData.TroopType` (not `OneToOneConversationCharacter`).

## Relevant code
- Vanilla `RecruitmentCampaignBehavior.buy_mercenaries_on_consequence` / `recruit_mercenaries` menu
- `source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs` (the dialogue-path patch added by #1463 — mirror it)

zzzzzzzzott:
Resolved by https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/1558

PR #1558, opened 2026-06-26
Synced mercenary stock and restock and prevents double purchase.

### LINKED ISSUE #1464: Sync town mercenary stock across clients (display divergence + double-sell)
## Context

PR #1463 (issue #1309) replicates a tavern mercenary hire's troops and gold to all clients, but it does **not** sync the town's mercenary stock (`RecruitmentCampaignBehavior.TownMercenaryData` — the `TroopType` + `Number` shown in a town's tavern recruit dialogue).

`TownMercenaryData` is rolled independently on each machine: `RecruitmentCampaignBehavior.DailyTickTown` → `UpdateCurrentMercenaryTroopAndCount` uses `MBRandom`, and the per-machine RNG streams are not seed-synced, so each client's stock drifts apart over a campaign. A grep of the source shows `GetMercenaryData` / `ChangeMercenaryCount` are referenced only by the hire path; nothing replicates the stock.

## Problems this causes

1. **Display divergence** — each client's tavern shows a different number/type of mercenaries available.
2. **Double-sell window** — two clients in the same town's tavern can each hire "all" the mercenaries from their own (divergent) stock view. The server applies both hires (it trusts each client's computed count), producing phantom troops and gold deductions for mercenaries that were already sold, and a town stock that can go negative on the server.

PR #1463 deliberately did **not** add a server-side stock rejection, because validating against the server's own (equally divergent) stock would wrongly reject legitimate solo hires — a proper fix needs the stock to actually be synced first.

## Suggested fix

Make `TownMercenaryData` server-authoritative and replicated (sync `TroopType` + `Number` server→client, e.g. via AutoSync or a dedicated message). Then a client hire can be validated against the synced stock, the server can reject over-hires, and every client's tavern shows the same availability.

## Relevant code
- `source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs` (the hire patch added by #1463)
- `source/GameInterface/Services/MobileParties/Handlers/MercenaryHireHandler.cs`
- Vanilla `RecruitmentCampaignBehavior.TownMercenaryData` / `GetMercenaryData` / `ChangeMercenaryCount` / `DailyTickTown`

zzzzzzzzott:
This PR syncs mercenary stocks and handling purchasing mercenaries from the campaign city menu. It also prevents double buying mercs from two clients.

Here are recommended commands to help test this -
`coop.debug.town.refresh_mercenary_stocks <townName|ALL>`
`coop.debug.town.request_mercenary_stock <townName>`

## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Mercenary stocks are not synced

## What is the new behavior?

Resolves #1464 
Resolves maybe #1465? (Can you confirm @ShoT-UPfps?)

Mercenary stocks are now synced


source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
```diff
@@ -44,33 +46,131 @@ private static bool BuyMercenariesPrefix(RecruitmentCampaignBehavior __instance)
...
+            return false;
+
+        int count = GetMercenaryHireCount(0, mercenaryData.Number, Hero.MainHero.Gold, unitPrice);
+        if (count <= 0)
+            return false;
+
         int goldAmount = count * unitPrice;
 
-        mercenaryData.ChangeMercenaryCount(-count);
+        PublishMercenariesHired(__instance, town, mercenaryData.TroopType, count, goldAmount);
```
ShoT-UPfps: we dont do `mercenaryData.ChangeMercenaryCount` anymore so it seems like the stock count wont be updated till server responds. is the client able to purchase extra mercs by spamming the button b4 the server responds?

source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
```diff
@@ -44,33 +46,131 @@ private static bool BuyMercenariesPrefix(RecruitmentCampaignBehavior __instance)
...
+        PublishMercenaryStock(__instance, town);
+    }
+
+    [HarmonyPostfix]
+    [HarmonyPatch("DailyTickTown")]
+    private static void DailyTickTownPostfix(RecruitmentCampaignBehavior __instance, Town town)
+    {
+        if (!ModInformation.IsServer) return;
+
+        PublishMercenaryStock(__instance, town);
```
ShoT-UPfps: this publishes merc stock for every town all the time. but most towns will not change per tick. can we do something to only publish if there has actually been a change

source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
```diff
@@ -44,33 +46,131 @@ private static bool BuyMercenariesPrefix(RecruitmentCampaignBehavior __instance)
...
+        int goldAmount = count * unitPrice;
+
+        PublishMercenariesHired(__instance, town, mercenaryTroop, count, goldAmount);
+
+        return false;
+    }
+
+    [HarmonyPostfix]
+    [HarmonyPatch("BuyMercenaries")]
+    private static void BuyMercenariesPostfix(RecruitmentCampaignBehavior __instance)
```
ShoT-UPfps: is this actually needed? it returns if it's a client and the server will enver buy mercenaries

source/GameInterface/Services/MobileParties/Handlers/MercenaryHireHandler.cs
```diff
@@ -76,13 +78,59 @@ private void Handle_HireMercenaries(MessagePayload<HireMercenaries> obj)
...
+                var mercenaryData = recruitmentBehavior?.GetMercenaryData(town);
+                int unitPrice = mercenaryData?.TroopType == null
+                    ? 0
+                    : Campaign.Current.Models.PartyWageModel.GetTroopRecruitmentCost(mercenaryData.TroopType, mainHero).RoundedResultNumber;
+                if (mercenaryData == null ||
+                    data.Count <= 0 ||
+                    data.GoldAmount <= 0 ||
+                    data.HeroGold < data.GoldAmount ||
+                    unitPrice <= 0 ||
+                    mercenaryData.TroopType != mercenaryTroop ||
```
ShoT-UPfps: is it possible that a client can be mid-conversation with a merc while the settlement ticks and removes that merc from the stock, then the client tries to buy the merc? what happens in that case?

source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
```diff
@@ -44,33 +46,131 @@ private static bool BuyMercenariesPrefix(RecruitmentCampaignBehavior __instance)
         if (ModInformation.IsServer)
             return true;
 
-        var town = PlayerEncounter.EncounterSettlement?.Town;
-        if (town == null)
+        if (!TryGetCurrentMercenaryTown(out var town))
             return true;
```
ShoT-UPfps: shouldnt this be `false` so that it doesn't run?

source/GameInterface/Services/MobileParties/Handlers/MercenaryHireHandler.cs
```diff
@@ -76,13 +78,59 @@ private void Handle_HireMercenaries(MessagePayload<HireMercenaries> obj)
 
         // The hire applies vanilla game actions; defer them to the game-loop thread so they run there
         // instead of on the network (poller) thread that delivered the message.
-        GameThread.Run(() =>
+        GameThread.RunSafe(() =>
```
ShoT-UPfps: since you changed this to `RunSafe`, remove the try-catch that starts on line 83

source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
```diff
@@ -44,33 +46,131 @@ private static bool BuyMercenariesPrefix(RecruitmentCampaignBehavior __instance)
...
 
-        int count = __instance._selectedMercenaryCount;
+        var mercenaryData = __instance.GetMercenaryData(town);
+        if (mercenaryData.TroopType == null || mercenaryData.Number <= 0)
+            return false;
+
+        var mercenaryTroop = CharacterObject.OneToOneConversationCharacter;
+        int unitPrice = Campaign.Current.Models.PartyWageModel.GetTroopRecruitmentCost(mercenaryData.TroopType, Hero.MainHero).RoundedResultNumber;
```
ShoT-UPfps: seems to be duped code between here and BuyMercenariesOnConsequencePrefix, can we use a shared helper

source/GameInterface/Services/MobileParties/Handlers/MercenaryStockHandler.cs
```diff
@@ -0,0 +1,225 @@
...
+
+        Campaign.Current.CurrentMenuContext.Refresh();
+    }
+
+    private static IEnumerable<Town> GetMercenaryTowns()
+    {
+        var campaign = Campaign.Current;
+        if (campaign == null) return Enumerable.Empty<Town>();
+
+        return campaign.CampaignObjectManager.Settlements
```
ShoT-UPfps, line 220: use the IsMercenaryTown helper

source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
```diff
@@ -44,33 +46,131 @@ private static bool BuyMercenariesPrefix(RecruitmentCampaignBehavior __instance)
...
+    private static void DailyTickTownPostfix(RecruitmentCampaignBehavior __instance, Town town)
+    {
+        if (!ModInformation.IsServer) return;
+
+        PublishMercenaryStock(__instance, town);
+    }
+
+    internal static void PublishMercenaryStock(RecruitmentCampaignBehavior behavior, Town town)
+    {
+        if (town?.Settlement == null || !town.Settlement.IsTown) return;
```
ShoT-UPfps: use the IsMercenaryTown helper

source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
```diff
@@ -44,33 +46,131 @@ private static bool BuyMercenariesPrefix(RecruitmentCampaignBehavior __instance)
...
+        int affordableCount = heroGold / unitPrice;
+        int requestedCount = selectedMercenaryCount <= 0 ? affordableCount : selectedMercenaryCount;
+
+        return TaleWorlds.Library.MathF.Min(availableMercenaries, TaleWorlds.Library.MathF.Min(requestedCount, affordableCount));
+    }
+
+    internal static bool TryGetCurrentMercenaryTown(out Town town)
+    {
+        town = PlayerEncounter.EncounterSettlement?.Town;
+        if (town?.Settlement != null && town.Settlement.IsTown)
```
ShoT-UPfps: use the IsMercenaryTown helper

source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
```diff
@@ -44,33 +46,131 @@ private static bool BuyMercenariesPrefix(RecruitmentCampaignBehavior __instance)
...
+            return true;
+
+        town = MobileParty.MainParty?.CurrentSettlement?.Town;
+        return town?.Settlement != null && town.Settlement.IsTown;
```
ShoT-UPfps: isnt this guaranteed to be false because you have the same condition on line 169 which would cause it to `return true`

source/GameInterface/Services/MobileParties/Handlers/MercenaryHireHandler.cs
```diff
@@ -76,13 +78,59 @@ private void Handle_HireMercenaries(MessagePayload<HireMercenaries> obj)
...
+
+                    if (recruitmentBehavior != null)
+                    {
+                        RecruitmentCampaignBehaviorPatch.PublishMercenaryStock(recruitmentBehavior, town);
+                    }
+
+                    return;
+                }
+
+                if (mainHero.Gold != data.HeroGold)
```
ShoT-UPfps: i think if the user spams the button before the server responds it will allow them to buy more mercs than they can actually afford

source/GameInterface/Services/MobileParties/Handlers/MercenaryStockHandler.cs
```diff
@@ -0,0 +1,225 @@
...
+
+        network.SendAll(new NetworkRequestMercenaryStockSync(true));
+    }
+
+    internal void Handle_MercenaryStockChanged(MessagePayload<MercenaryStockChanged> payload)
+    {
+        if (!ModInformation.IsServer) return;
+
+        if (!objectManager.TryGetIdWithLogging(payload.What.Town, out var townId)) return;
+        if (!objectManager.TryGetIdWithLogging(payload.What.TroopType, out var troopTypeId)) return;
```
ShoT-UPfps: i dont think we should be using `WithLogging` here. Isn't `null` expected when the town no longer has any mercs? which would cause this to log an error because it wont be found in objectmanager

source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
ShoT-UPfps: we need to implement this, otherwise AI parties could buy mercs and it wont be recognized on the client until the next DailyTickTownPostfix

ShoT-UPfps (changes requested):

source/GameInterface/Services/MobileParties/Handlers/MercenaryHireHandler.cs
zzzzzzzzott: Changed it so the server will verify requests itself then reject if the stock no longer matches

source/GameInterface/Services/MobileParties/Handlers/MercenaryHireHandler.cs
zzzzzzzzott: Changed it so the server computes the hire cost server side then validates it against the servers record for the users gold. That should prevent button spamming

source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
zzzzzzzzott: Changed so the server is checking against its own stock before allowing purchases to go through, should prevent spam.

source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
zzzzzzzzott: Server on every tick will now compare the stock before and after and only publish if a change was actually made.

I dont think this should be too expensive on the process

source/GameInterface/Services/MobileParties/Patches/RecruitmentCampaignBehavior.cs
zzzzzzzzott: 99 percent of the time yea, but one is checking against `PlayerEncounter.EncounterSettlement?.Town` while the other is checking against `MobileParty.MainParty?.CurrentSettlement?.Town`, and I just wanted to second check to make sure all is good in case of some wierd interactions/states

source/GameInterface/Services/MobileParties/Handlers/MercenaryHireHandler.cs
```diff
@@ -76,51 +77,116 @@ private void Handle_HireMercenaries(MessagePayload<HireMercenaries> obj)
...
+                availableCount,
+                data.HeroGold,
+                mainHero.Gold,
+                data.GoldAmount,
+                goldAmount);
+
+            if (recruitmentBehavior != null)
             {
-                logger.Error(e, "Failed to apply {Message}", nameof(HireMercenaries));
+                RecruitmentCampaignBehaviorPatch.PublishMercenaryStock(recruitmentBehavior, town);
```
ShoT-UPfps: on a reject this republishes through `PublishMercenaryStock` which `SendAll`s to every client, but only the requesting peer is out of sync here. could we `network.Send(peer, ...)` the current stock back to just them instead of broadcasting to everyone?

source/GameInterface/Services/MobileParties/Handlers/MercenaryHireHandler.cs
```diff
@@ -76,51 +77,116 @@ private void Handle_HireMercenaries(MessagePayload<HireMercenaries> obj)
...
+        int goldAmount,
+        int clientHeroGold,
+        int serverHeroGold,
+        int unitPrice,
+        bool availableTroopMatches,
+        int availableMercenaries)
+    {
+        return count > 0 &&
+               goldAmount > 0 &&
+               clientHeroGold >= goldAmount &&
```
ShoT-UPfps: is this still needed now that we check `serverHeroGold >= goldAmount` on the next line? `goldAmount` is computed server-side and `clientHeroGold` is just the client's snapshot, so i dont think this gates anything real anymore.

source/GameInterface/Services/MobileParties/Messages/MercenaryStockMessages.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        TownId = townId;
+    }
+}
+
+[ProtoContract(SkipConstructor = true)]
+internal readonly struct NetworkRequestMercenaryStockSync : ICommand
+{
+    [ProtoMember(1)]
+    public readonly bool Request;
```
ShoT-UPfps: `Request` is never read, the handler ignores it and we always pass `true`. can we drop it and make this a parameterless marker?

source/GameInterface/Services/MobileParties/Handlers/MercenaryStockHandler.cs
```diff
@@ -0,0 +1,229 @@
...
+    internal void Handle_CampaignReady(MessagePayload<CampaignReady> payload)
+    {
+        if (ModInformation.IsServer) return;
+
+        network.SendAll(new NetworkRequestMercenaryStockSync(true));
+    }
+
+    internal void Handle_MercenaryStockChanged(MessagePayload<MercenaryStockChanged> payload)
+    {
+        if (!ModInformation.IsServer) return;
```
ShoT-UPfps: can we use `ModInformation.IsClient` instead of `!ModInformation.IsServer` here, reads cleaner. same at `:98`, `:124`, `MercenaryHireHandler.cs:73`, and `RecruitmentCampaignBehavior.cs:25/98` (and the `!ModInformation.IsServer` part of `RecruitmentCampaignBehavior.cs:39`, keep the `__runOriginal` check).

ShoT-UPfps (approved):

PR #1563, opened 2026-06-26
Initial Caravans Sync

### LINKED ISSUE #1161: Caravan/Ally Party Trade
# For New Contributors

## Intended Design

N/A

## Location

## Related Issues

N/A

## Requirements

N/A

## Additional information

N/A

## Definition of Done

### LINKED ISSUE #1474: Epic: Caravans
## For New Contributors

## Description
Sync Everything Caravan
- Trading (should work out of the box)
- Dialogue Effects
- Disable Hostile Actions (until battles are ready)

jordanbrymora:
### Synced
- Forming, changing leader of and disbanding player caravans
- Managing player caravans (prohibiting kingdoms for trade, setting home settlement)
- AI caravans doing their thing on the world map
- Trading with caravans
- Looting caravans (taking gold, items and prisoners)
- Saving players' interactions with caravans
- Getting trade rumours from caravans
- Updating party visuals & names when leader changes or when disbanding on clients.

## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #1474 
Resolves #1161


source/GameInterface/Services/PartyComponents/Handlers/PartyComponentHandler.cs
```diff
@@ -54,12 +55,18 @@ public PartyComponentHandler(IMessageBroker messageBroker, INetwork network, IOb
...
+        messageBroker.Subscribe<PartyComponentLeaderChanged>(Handle_PartyComponentLeaderChanged);
+        messageBroker.Subscribe<NetworkPartyComponentLeaderChanged>(Handle_NetworkPartyComponentLeaderChanged);
     }
 
     public void Dispose()
     {
         messageBroker.Unsubscribe<PartyComponentCreated>(Handle);
         messageBroker.Unsubscribe<NetworkCreatePartyComponent>(Handle);
+
+        messageBroker.Subscribe<PartyComponentLeaderChanged>(Handle_PartyComponentLeaderChanged);
```
ShoT-UPfps: should be Unsubscribe

source/GameInterface/Services/Caravans/Handlers/FormPlayerClanCaravanHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+            GiveGoldAction.ApplyForCharacterToSettlement(mainHero, currentSettlement, obj.What.GoldCost, false);
+
+            // Send message to all clients to fade out the caravan leader character if present in their current mission
+            network.SendAll(new NetworkFadeOutNewCaravanLeader(obj.What.CaravanLeaderId));
+        });
+    }
+    
+    private void Handle_NetworkFadeOutNewCaravanLeader(MessagePayload<NetworkFadeOutNewCaravanLeader> obj)
+    {
+        if (!objectManager.TryGetObjectWithLogging<Hero>(obj.What.CaravanLeaderId, out var caravanLeader)) return;
```
ShoT-UPfps: this function should be wrapped in `GameThread.RunSafe` since it queries `objectManager` and subsequently updates state that the game thread uses

source/GameInterface/Services/Caravans/Interfaces/SessionCaravansPlayerDataInterface.cs
```diff
@@ -0,0 +1,176 @@
...
+
+    public SessionCaravansPlayerDataInterface(ICoopSessionProvider coopSessionProvider, IObjectManager objectManager, IPlayerManager playerManager)
+    {
+        this.coopSessionProvider = coopSessionProvider;
+        this.objectManager = objectManager;
+        this.playerManager = playerManager;
+    }
+
+    public void AddProhibitedKingdom(string playerHeroId, string kingdomId)
+    {
```
ShoT-UPfps, line 46: these functions need to be wrapped with `GameThread.RunSafe` - we can't query objectmanager outside of game thread because the game thread queue could be sitting on packets whihc would change the result of the objectmanager lookup and we cant update game state outside of the game thread because the game thread could concurrently be accessing the same game state

source/GameInterface/Services/Caravans/Interfaces/SessionCaravansPlayerDataInterface.cs
```diff
@@ -0,0 +1,176 @@
...
+
+        CaravansPlayerData.PlayerProhibitedKingdomsForPlayerCaravans[playerHeroId].Remove(kingdomId);
+    }
+
+    public void RemoveProhibitedKingdomForAllPlayers(string kingdomId)
+    {
+        foreach (var player in playerManager.Players)
+        {
+            string playerHeroId = player.HeroId;
+            if (CaravansPlayerData.PlayerProhibitedKingdomsForPlayerCaravans[playerHeroId].Contains(kingdomId))
```
ShoT-UPfps, line 72: is it possible that `KeyNotFoundException`  gets thrown here? if it is then the for-loop might not iterate over all the players that it should

same for the for-loop in `RemoveInteractedCaravanForAllPlayers`

source/GameInterface/Services/Caravans/Handlers/CaravansConversationsHandler.cs
```diff
@@ -0,0 +1,320 @@
...
+    private void Handle_ChangeCaravanHomeSettlement(MessagePayload<ChangeCaravanHomeSettlement> obj)
+    {
+        if (!objectManager.TryGetIdWithLogging(obj.What.ConversationParty, out var conversationPartyId)) return;
+        if (!objectManager.TryGetIdWithLogging(obj.What.Settlement, out var settlementId)) return;
+
+        var message = new NetworkChangeCaravanHomeSettlement(conversationPartyId, settlementId);
+        network.SendAll(message);
+    }
+
+    private void Handle_NetworkChangeCaravanHomeSettlement(MessagePayload<NetworkChangeCaravanHomeSettlement> obj)
```
ShoT-UPfps, line 102: we need to move the object manager lookups  and game state mutations in this class into `GameThread.RunSafe` (if they're not already running on game thread)

source/GameInterface/Services/Caravans/Handlers/CaravansCampaignBehaviorHandler.cs
```diff
@@ -0,0 +1,298 @@
...
+        network.SendAll(message);
+    }
+
+    private void Handle_NetworkCaravansKingdomDestroyed(MessagePayload<NetworkCaravansKingdomDestroyed> obj)
+    {
+        // Update data on all clients
+        GameThread.Run(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Kingdom>(obj.What.DestroyedKingdomId, out var destroyedKingdom)) return;
+            GetCaravansBehavior().OnKingdomDestroyed(destroyedKingdom);
```
ShoT-UPfps: isnt this disabled https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/1563/changes#diff-6688b06f2de84045d83e30e0fe8e464bf33a7cc698c1ced5f116f4aa693c4e3bR38

source/GameInterface/Services/Caravans/Patches/CaravansConversationsPatches.cs
```diff
@@ -0,0 +1,222 @@
...
+
+        return false;
+    }
+
+    [HarmonyPatch(nameof(CaravansCampaignBehavior.caravan_start_talk_on_condition))]
+    [HarmonyPostfix]
+    public static void CaravanStartTalkOnConditionPostfix(ref CaravansCampaignBehavior __instance)
+    {
+        // Local update is needed so separately publish message to server in postfix to store change in CoopSession
+        var message = new SetPlayerCaravanInteraction(Hero.MainHero, MobileParty.ConversationParty, CaravansCampaignBehavior.PlayerInteraction.Friendly);
```
ShoT-UPfps, line 82: `CaravansCampaignBehavior.PlayerInteraction.Friendly`

does this only fire the first time the player ever talks to the caravan? otherwise what about if the caravan is owned by someone who's in a kingdom that's at war with the client

source/GameInterface/Services/MobileParties/MobilePartyRegistry.cs
```diff
@@ -66,6 +67,7 @@ public override void OnClientCreated(MobileParty obj, string id)
             obj.InitMembers();
             obj.InitCached();
             obj.Initialize();
+            obj.HasLandNavigationCapability = true;
```
ShoT-UPfps, line 70: i recognize that this is a war sails thing, but is there any reason we can't just properly sync it that way we don't have to update it when we do get around to supporting war sails?

source/GameInterface/Services/PartyComponents/Patches/CaravanPartyComponentPatches.cs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+[HarmonyPatch(typeof(CaravanPartyComponent.InitializationArgs))]
+internal class CaravanPartyComponentInitializationArgsPatches
+{
+    [HarmonyPatch(nameof(CaravanPartyComponent.InitializationArgs.InitializeCaravanOnCreation))]
+    [HarmonyPrefix]
+    public static bool InitializeLordPartyPropertiesPrefix(ref LordPartyComponent.InitializationArgs __instance, MobileParty mobileParty, Settlement settlement)
+    {
+        // Shouldn't be needed?
```
ShoT-UPfps: can we remove this?

source/GameInterface/Services/Caravans/Handlers/CaravansConversationsHandler.cs
```diff
@@ -0,0 +1,320 @@
...
+                conversationParty.ItemRoster.Clear();
+            }
+            int num = MathF.Max(conversationParty.PartyTradeGold, 0);
+            if (num > 0)
+            {
+                GiveGoldAction.ApplyForPartyToCharacter(conversationParty.Party, mainHero, num, false);
+            }
+            BeHostileAction.ApplyMajorCoercionHostileAction(mainParty.Party, conversationParty.Party);
+            GetCaravansBehavior()._lootedCaravans.Add(conversationParty, CampaignTime.Now);
+            SkillLevelingManager.OnLoot(mainParty, conversationParty, conversationParty.ItemRoster, false);
```
ShoT-UPfps: does this do anything? `conversation.ItemRoster` is cleared on line 265 but we pass it again here and it would be empty

same pattern happens on line 310

source/GameInterface/Services/Caravans/Patches/CaravansCampaignBehaviorPatches.cs
```diff
@@ -0,0 +1,176 @@
...
+
+        return false;
+    }
+
+    /// <summary>
+    /// Only need to update _tradeActionLogs for clients when a caravan leaves a settlement
+    /// because clients only need _tradeActionLogs for dialogue with caravans to obtain trade rumours
+    /// </summary>
+    [HarmonyPatch(nameof(CaravansCampaignBehavior.OnSettlementLeft))]
+    [HarmonyPrefix]
```
ShoT-UPfps, line 87: shouldn't this prevent the original from running as well? it seems like this is a replacement for the original

ShoT-UPfps (changes requested):

source/GameInterface/Services/Caravans/Patches/CaravansConversationsPatches.cs
jordanbrymora: You would think that it should work like that. Weirdly TaleWorlds' own implementation assumes interacting with a caravan at all is immediately a "friendly" interaction and uses the interaction before that to determine the greeting message. This only gets changed to hostile if the player actually takes a hostile action against the caravan. If you talk to a caravan, take a hostile action, leave and come back they are hostile with you. If you then leave and come back again they are "friendly" with you. This seems to be somewhat of a half-baked vanilla mechanic. Some examples of this is that `CaravansCampaignBehavior.PlayerInteraction.TradedWith` is never even used and not all hostile actions actually set a hostile interaction.

You pointing out this function did let me spot that I should add the same guard against non-caravan parties that vanilla uses though. Thanks.

source/GameInterface/Services/MobileParties/MobilePartyRegistry.cs
jordanbrymora: This just implements the default value for the property on clients. Dynamic sync won't set this, and without it clients think that new caravans are sailing because it defaults to false instead of true.

source/GameInterface/Services/Caravans/Patches/CaravansCampaignBehaviorPatches.cs
jordanbrymora: I missed this, thanks for pointing it out. Fixed with latest commit.

source/GameInterface/Services/PartyComponents/Patches/CaravanPartyComponentPatches.cs
jordanbrymora: I missed this, thanks for pointing it out. Fixed with latest commit.

source/GameInterface/Services/Caravans/Handlers/CaravansCampaignBehaviorHandler.cs
jordanbrymora: I missed this, thanks for pointing it out. Fixed with latest commit.

source/GameInterface/Services/Caravans/Handlers/CaravansConversationsHandler.cs
jordanbrymora: Should be fixed with latest commit.

source/GameInterface/Services/Caravans/Interfaces/SessionCaravansPlayerDataInterface.cs
jordanbrymora: Should be fixed with latest commit.

source/GameInterface/Services/Caravans/Handlers/FormPlayerClanCaravanHandler.cs
jordanbrymora: Should be fixed with latest commit.

source/GameInterface/Services/PartyComponents/Handlers/PartyComponentHandler.cs
jordanbrymora: I missed this, thanks for pointing it out. Fixed with latest commit.

source/GameInterface/Services/Caravans/Interfaces/SessionCaravansPlayerDataInterface.cs
jordanbrymora: It shouldn't ever happen but I added a guard to both just in case. Thanks for pointing it out.

source/GameInterface/Services/Caravans/Handlers/CaravansConversationsHandler.cs
jordanbrymora: Thanks for pointing this out, I missed the point of the client-side ItemRoster copy. Changed to send item roster elements to update skills with latest commit.

source/GameInterface/Services/Caravans/Handlers/FormPlayerClanCaravanHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+            GiveGoldAction.ApplyForCharacterToSettlement(mainHero, currentSettlement, obj.What.GoldCost, false);
+
+            // Send message to all clients to fade out the caravan leader character if present in their current mission
+            network.SendAll(new NetworkFadeOutNewCaravanLeader(obj.What.CaravanLeaderId));
+        });
+    }
+    
+    private void Handle_NetworkFadeOutNewCaravanLeader(MessagePayload<NetworkFadeOutNewCaravanLeader> obj)
+    {
+        GameThread.Run(() =>
```
ShoT-UPfps: use `RunSafe`

source/GameInterface/Services/Caravans/Handlers/CaravansConversationsHandler.cs
```diff
@@ -0,0 +1,330 @@
...
+
+    private void Handle_NetworkCaravanLootedLeaveOnConsequence(MessagePayload<NetworkCaravanLootedLeaveOnConsequence> obj)
+    {
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Hero>(obj.What.MainHeroId, out var mainHero)) return;
+            if (!objectManager.TryGetObjectWithLogging<MobileParty>(obj.What.MainPartyId, out var mainParty)) return;
+            if (!objectManager.TryGetObjectWithLogging<MobileParty>(obj.What.ConversationPartyId, out var conversationParty)) return;
+
+            ItemRoster itemRoster = new();
```
ShoT-UPfps, line 220: Can two players "loot the caravan" at the same time or is there something enforcing it so that only one player can loot at a time?

I'm not sure what "loot the caravan" means really, is that any different from battling them and winning?

The reason I ask is I'm wondering if two players happen "loot the caravan" at the exact same time, what's stopping them both from getting a full payout? is there a race condition we need to be aware of?

jordanbrymora: Two players can't talk to a mobile party at the same time so they shouldn't be able to be looted at the same time. Its all done through dialogue. Battles ending won't matter because the caravan ether gets destroyed (and can't be looted again) or wins the battle (doesn't get looted).

_lootedCaravans stores the caravans that have been looted after taking a hostile action with or without battle. It just means that if one player loots a caravan, other players also have to wait for the vanilla 10 day cooldown before another hostile action can be taken against it.

source/GameInterface/Services/Caravans/Handlers/FormPlayerClanCaravanHandler.cs
jordanbrymora: Fixed

source/GameInterface/Services/Caravans/Handlers/CaravansConversationsHandler.cs
```diff
@@ -0,0 +1,330 @@
...
+    {
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Hero>(obj.What.MainHeroId, out var mainHero)) return;
+            if (!objectManager.TryGetObjectWithLogging<MobileParty>(obj.What.MainPartyId, out var mainParty)) return;
+            if (!objectManager.TryGetObjectWithLogging<MobileParty>(obj.What.ConversationPartyId, out var conversationParty)) return;
+
+            ItemRoster itemRoster = new()
+            {
+                _data = obj.What.ItemRosterElements
```
ShoT-UPfps: I think we need to assign more than just `_data`, here's vanilla `OnLoot`

Image

I don't think these fields (`TradeGoodsTotalValue`, `NumberOfMounts`, ...) get derived when you just assign `_data`

Same thing for :307

jordanbrymora: Changed to use Add() instead of assigning _data directly in latest commit.

source/Coop.Core/Server/Services/Save/Handlers/SaveGameHandler.cs
```diff
@@ -68,7 +72,12 @@ private void Handle_GameLoaded(MessagePayload<GameLoaded> obj)
         savedSession = saveManager.LoadCoopSession(obj.What.SaveName);
         if(savedSession == null)
         {
```
ShoT-UPfps: should we do something to support saves that existed before caravans were supported?

I added something for this in the alley sync PR
image

jordanbrymora: I can see what you mean although I could see this solution getting messy as we add more player data structures.

jordanbrymora: I'm also not sure its strictly necessary right now to add support for features that are being added before the v0.1 release. We aren't aiming to support old saves from before the initial release are we?

ShoT-UPfps: Idts. it would just help cut back on issues that are hard to attribute and it's not much to add. it's always a bit of a pain to have everyone delete their save 

agree it can be better esp once the list grows tho, there is def a more elegant way to do it

ShoT-UPfps: maybe once we actually release 0.1 we can delete all these backwards compatibility handlers and just tell all devs to delete saves

ShoT-UPfps: this approach seems better:
image

jordanbrymora: I like this solution, its clean and expandable. Updated with latest commit.

ShoT-UPfps (approved):

PR #1587, opened 2026-06-28
[1586] Only show Coop saves in Host Co-op Sandbox

### LINKED ISSUE #1586: Prevent loading vanilla save through "Host co-op sandbox"
Loading vanilla saves is unsupported. One example of an issue is that active map events wont have a network ID.

ShoT-UPfps:
## PR Checklist

## PR Type

`Host Co-op Sandbox` lists saves that were never saved with Coop enabled, so they can be selected for hosting.

This can cause a cascade of issues. One example is that the vanilla save could have active MapEvents which won't register  network IDs when the save gets loaded.

## What is the new behavior?

Resolves #1586

`Host Co-op Sandbox` only lists saves that were saved with Coop enabled. A vanilla campaign can be prepared by loading and saving it once with Coop enabled.


ShoT-UPfps:
closing this, the module metadata check doesnt prove the save was created with co-op lifetime patches active

Issue #1597, opened 2026-06-29
Rescuer's own battle side loads empty when attacking bandits holding a player prisoner


ShoT-UPfps:
Split out from #1584 (the rescue-an-imprisoned-player report), which is three separate bugs. This is the core one and the only one the #1584 PR doesn't fix.

## Symptom

When a client (player B) attacks a bandit party that's holding another player (A) prisoner, B loads the encounter but B's own side is empty: the banner shows e.g. "Mountain Bandits 7 vs 0" and the only menu option is "try to escape", there is no "Attack". B's party is right there on the map with its troops, they just don't make it into B's side of the map event. If B leaves and rejoins the side populates ("7 vs 11", "Attack!" available) and B can fight, so a full reload fixes it.

(screenshots are on #1584)

## What we found

B's own MapEventParty isn't in its `MapEventSide._battleParties` when the encounter menu is evaluated, so vanilla `EncounterAttackCondition` drops the Attack option (`HasTroopsOnBothSides()` is false). It's missing membership, not an empty roster, B's hero isn't even counted as 1.

Side membership (`_sides` / `_battleParties`) is deliberately excluded from auto-sync and rebuilt from separate id-only messages (`NetworkAssignMapEventSide` / `NetworkAddBattleParty`) applied on the game thread. `PartyBase._mapEventSide` IS auto-synced as a field, so B "knows" it's in the event (`B.MapEvent != null`) without ever being added to its side's party list. `MapEventCreationCoordinator.RequestBlocking` returns as soon as the MapEvent object id resolves, it doesn't wait for B's own membership to apply.

One odd thing: it's specifically B's own side (the attacker, added first) that's empty while the bandit side (added second) is fine. A uniform "membership hasn't streamed in yet" delay would empty both, so this looks like a real race or a player-party-specific drop, not just lateness.

## Candidate fix

Make `RequestBlocking` block until B's own MapEventParty is actually in its `MapEventSide._battleParties` (evaluated on the game thread), not just until the MapEvent object resolves. Risk: if the real cause is a dropped/raced membership message rather than lateness, the wait would time out, the battle aborts and the client freezes up to `ObjectCreationTimeout`, which is worse. So we want a DEBUG-build repro log of the rescue first (`MapEventPartyRegistry` / `MapEventSideRegistry` debug on) to tell "applied late" from "dropped" before committing to the barrier.

[...]

Issue #1584, opened 2026-06-27
[Report] Player is unable to rescue another player that is imprisoned by bandits


ShoT-UPfps:
## Severity of issue/bug

- [x] High
- [ ] Average
- [ ] Low

see https://discord.com/channels/709516043332354119/1520545878765731860

the imprsisoned player can see the map event with the swords

<img width="1909" height="1079" alt="Image" src="https://github.com/user-attachments/assets/ef04866f-9153-4fac-a572-b7435e28136a" />

 the second player that is attempting to rescue loads a map event where their side isnt loaded
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/60e8103f-2058-40ee-867d-185d7b0bd2ff" />

when client leaves and rejoins the map event loads properly with his side populated and he is able to attack

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/9f4e3c85-7872-4bd1-94f1-52b80ea72dd2" />

after the client 2 defeats the bandits, the imprisoned player (clinet 1) can correctly see himself on the map  and that another set of looters engaaged client 2 with a map event initiated

<img width="1906" height="1079" alt="Image" src="https://github.com/user-attachments/assets/eec7035a-5795-49a4-bfa3-62ae59d319d1" />

but on the client 2 PoV he doesn't see the freed client 1 , and also he doesnt see the map event, but he does see the looters on top of him on the map

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/d2b871e9-f0a2-483a-b505-cfe6bce03025" />

when client 2 leaves and rejoins again, its fine, he can see the freed player and the mapevent against him with the looters

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e62cf6f2-b078-4eca-bd9f-de21eb635f50" />

PR #1598, opened 2026-06-29
[1584] Show freed prisoners and new battle icons without reconnecting

ShoT-UPfps:
## PR Checklist

## PR Type

As reported in #1584 (to confirm against a baseline run): after a player held prisoner by bandits is freed by another player defeating the bandits, the rescuer doesn't see the freed player's party on the map until they reconnect (the freed player sees themselves fine). And a battle that starts as a player is leaving another fight shows the enemy party on the map but not its crossed-swords battle icon, again until they reconnect.

## What is the new behavior?

Intended, pending manual verification: the freed prisoner's party position is replicated on release, so it should appear on other players' maps at the release spot without a reconnect; and a newly created battle's map icon is registered, so it should appear after a player returns to the map without a reconnect.

Covers two of the three problems in #1584. The remaining one — the rescuer's own side loading empty so they can't attack without rejoining — is tracked separately in #1597.


ShoT-UPfps:
closing will be fixed by https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/1599

PR #1615, opened 2026-06-30
[1614] Stop clients hanging on a loading screen entering a town with an owned alley

### LINKED ISSUE #1614: Client hangs on a loading screen entering a town center where its hero owns an alley
When a client loads into a town and clicks to walk around the town center, it just hangs on a loading screen. nothing recovers it, leaving the settlement and coming back doesnt help.

`Coop_client.log` shows the mission load throw once, then `Failed to run Game.OnTick` spamming every tick after it:

```
System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   at SandBox.CampaignBehaviors.AlleyCampaignBehavior.AddPlayerAlleyCharacters(Alley alley)
   at SandBox.CampaignBehaviors.AlleyCampaignBehavior.LocationCharactersAreReadyToSpawn_Patch1(...)
   at SandBox.Missions.MissionLogics.Towns.TownCenterMissionController.AfterStart()
   at TaleWorlds.MountAndBlade.MissionState.FinishMissionLoading()
```

`AfterStart` throwing means `FinishMissionLoading` never completes, so the mission is stuck in `TickLoading` forever, which is the loading screen. the repeated `MissionMainAgentInteractionComponent.FocusTick()` NREs after it are just downstream of the half-loaded mission, fixing the alley throw makes them go away too. `OnTickRobustnessPatches` swallows each one so it doesnt crash to desktop, it just hangs.

root cause: on join, `AlleyInitializationHandler` (PlayerHeroChanged) sets `Alley.State = OccupiedByPlayer` for every alley the client's hero owns, but only adds the matching `_playerOwnedCommonAreaData` entry for alleys that came with transferred management data. the co-op session alley data starts empty (`SaveGameHandler.cs:84`) and is never seeded from the loaded save, so an alley the hero already owned from the save gets State=2 with no entry. vanilla's `AddPlayerAlleyCharacters` then does an unguarded `_playerOwnedCommonAreaData.First(x => x.Alley == alley)` and throws because nothing matches.

regression from the alley sync work (#1477 / PR #1554).

repro: host a save where a client's hero owns an alley in some town, join as that client already in that town, walk around the town center.

fix direction: keep `State == OccupiedByPlayer` and the `_playerOwnedCommonAreaData` entry in lockstep (only set State when the entry is added), and seed the session's per-alley management data from the host's loaded `_playerOwnedCommonAreaData` so the owned alley stays manageable instead of just non-crashing.

ShoT-UPfps:
## PR Type

Some alley data is lost between saves and if a client loads into a save where he owns an alley then tries to walk around in that settlement, the client will get stuck on the loading screen

## What is the new behavior?

Resolves #1614.

`State = OccupiedByPlayer` is now set only when the `_playerOwnedCommonAreaData` entry is added, so the two can't diverge and vanilla's `.First` always matches. an owned alley with no entry stays gang-occupied (no crash) instead of throwing.

the host also seeds the session's management data for the save's owned alleys, read from `Alley.Owner` (gang-occupied ones excluded) once its objects register, so they transfer to the client that adopts the owner and stay manageable instead of just non-crashing.

### DIFF
--- a/source/GameInterface/Services/Alleys/Interfaces/AlleyCampaignBehaviorInterface.cs
+++ b/source/GameInterface/Services/Alleys/Interfaces/AlleyCampaignBehaviorInterface.cs
@@ -6,6 +6,7 @@ using SandBox.CampaignBehaviors;
 using Serilog;
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.Reflection;
 using TaleWorlds.CampaignSystem;
 using TaleWorlds.CampaignSystem.Roster;
@@ -14,16 +15,22 @@ using TaleWorlds.CampaignSystem.Settlements;
 namespace GameInterface.Services.Alleys.Interfaces;
 
 /// <summary>
-/// Client-side access to the owning player's alley management state, which lives in the
-/// (internal, nested) <c>AlleyCampaignBehavior.PlayerAlleyData</c> entries of its private
-/// <c>_playerOwnedCommonAreaData</c> list. Reached by reflection because that type is internal.
-/// Populating this list is what lights up the in-game manage-alley menus on the owning client.
+/// Access to the owning player's alley management state, which lives in the (internal, nested)
+/// <c>AlleyCampaignBehavior.PlayerAlleyData</c> entries of its private <c>_playerOwnedCommonAreaData</c>
+/// list. Reached by reflection because that type is internal. Populating this list is what lights up the
+/// in-game manage-alley menus on the owning client; the host reads it to seed the authoritative session.
 /// </summary>
 public interface IAlleyCampaignBehaviorInterface : IGameAbstraction
 {
     void AddOrUpdatePlayerAlleyData(Alley alley, Hero overseer, TroopRoster garrison);
     void RemovePlayerAlleyData(Alley alley);
     bool TryGetCurrentSettlementAlley(out Alley alley);
+
+    /// <summary>
+    /// Reads the loaded behavior's player-owned alleys (alley, overseer, garrison) so the host can seed
+    /// the authoritative CoopSession management data from a save that was never played in co-op.
+    /// </summary>
+    IEnumerable<(Alley alley, Hero overseer, TroopRoster garrison)> GetPlayerOwnedAlleys();
 }
 
 /// <inheritdoc cref="IAlleyCampaignBehaviorInterface"/>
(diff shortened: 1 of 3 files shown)


source/GameInterface/Services/Alleys/Interfaces/AlleyCampaignBehaviorInterface.cs
```diff
@@ -79,6 +91,12 @@ public void AddOrUpdatePlayerAlleyData(Alley alley, Hero overseer, TroopRoster g
                 var data = PlayerAlleyDataCtor.Invoke(new object[] { alley, garrison });
                 if (overseer != null) AssignedClanMemberField.SetValue(data, overseer);
                 list.Add(data);
+
+                // Promote State in lockstep with the entry. Vanilla's AddPlayerAlleyCharacters does an
+                // unguarded _playerOwnedCommonAreaData.First(x => x.Alley == alley), so an OccupiedByPlayer
+                // alley with no entry throws; setting State only here means OccupiedByPlayer always has a
+                // matching entry, and an alley with no entry stays at its loaded gang-occupied State.
+                StateSetter.Invoke(alley, new object[] { Alley.AreaState.OccupiedByPlayer });
```
garrettluskey: This can be set directly

source/GameInterface/Services/Alleys/Interfaces/AlleyCampaignBehaviorInterface.cs
garrettluskey: This should work with the publicizer no?

garrettluskey (dismissed):

source/GameInterface/Services/Alleys/Interfaces/AlleyCampaignBehaviorInterface.cs
ShoT-UPfps: you right, updated

source/GameInterface/Services/Alleys/Interfaces/AlleyCampaignBehaviorInterface.cs
ShoT-UPfps: you right, updated

garrettluskey (approved):

PR #1617, opened 2026-06-30
Recruitment Index Fix

### LINKED ISSUE #1564: Volunteer recruit slots desync on clients — sync drops empty slots and shifts/leaves stale entries
the volunteer-slot sync is lossy, so client recruit slots dont match the server.

`Hero.VolunteerTypes` is a fixed-length-6 array where null slots are meaningful (empty recruit slots), but the sync compacts them away:

- send `VolunteerTypesHandler.Handle_VolunteersUpdated` (`GameInterface/Services/MobileParties/Handlers/VolunteerTypesHandler.cs:85-92`): starts `new string[]{}` and `AddItem`s each entry, but skips nulls (`if (character is null) continue;`) and any character whose id wont resolve, so gaps are dropped and everything after a gap shifts left.
- apply `Handle_UpdateVolunteers` (`VolunteerTypesHandler.cs:113-121`): writes the compacted ids into slots `0..n-1` and never clears the remaining slots (and on an unresolved id it `continue`s, leaving that slot stale), so clients end up with shifted volunteers plus leftover stale entries.

this matters because the slot index is meaningful: `RemoveVolunteer` keys off `bitCode` = the slot index (`VolunteerTypesHandler.cs:68` does `VolunteerTypes[bitCode] = null`), and recruitment reads per-slot, so a client recruiting or removing a volunteer can act on the wrong troop.

git blame says this wasnt a deliberate choice: introduced in `4bdc73c8` (jordanbrymora, "Initial Party Management Sync", 2026-06-04), with no comment explaining the null-drop.

fix: make it a fixed-length-6, index-preserving snapshot (carry nulls / explicit empty markers) so client slot positions match the server, and null-clear any slot not present in the update. rests on permanent invariants (server-authoritative volunteer updates + the fixed-length array), not on any disabled feature.

(found while profiling #1557 but its a correctness bug, not a bandwidth one.)

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #1564


source/GameInterface/Services/MobileParties/Handlers/VolunteerTypesHandler.cs
```diff
@@ -105,18 +111,23 @@ private void Handle_UpdateVolunteers(MessagePayload<UpdateVolunteers> obj)
         {
             try
```
jordanbrymora: Can we change all methods to use `GameThread.RunSafe` in this handler and remove the try catch blocks to clean this up a bit more?

jordanbrymora (changes requested):

ShoT-UPfps:
now that we added tihs and the server is now getting the length 6 array, do we need an `IsClient` return check in `UpdateVolunteersOfNotablesInSettlementPostfix` ?

jordanbrymora (approved):

PR #1621, opened 2026-06-30
[1595] Flush the send coalescer before the join save snapshot

### LINKED ISSUE #1595: Flush the per-tick send coalescer before the join save snapshot (double-apply on join)
came out of the #1572 ItemRoster coalescing review

the #1576 `SendCoalescer` defers sends to a per-tick flush on the game thread (`CoopServer.Update` -> `GameThread.RunSafe(coalescer.Flush)`). the join handshake assumes broadcasts only come from the poll thread, which is parked during the save snapshot, see the comments in `TransferSaveState.cs:50-54` and `ConnectionMessageQueue.cs:44-50` ("the poller is parked... cannot broadcast a received delta that races the snapshot"). the coalescer's game-thread flush breaks that assumption.

so a coalesced delta enqueued before a peer joins is captured in the save snapshot (the roster already has it) AND flushed after `BeginQueueing`, queued for that peer and replayed on campaign entry. the joining peer double-applies it and is permanently desynced. this affects every coalescer consumer (#1570/#1571/#1572/#1574), not just ItemRoster.

fix: flush the coalescer before taking the join snapshot in `TransferSaveState` (the peer is still `Dropping`, so the flush is dropped for it and reflected for everyone else, and the snapshot stays consistent). needs `ISendCoalescer` threaded into the connection context and `TransferSaveState`.

rare today (ItemRoster updates are rare while AI trade is disabled) but a real latent desync, and more likely once trade/economy behaviors re-enable.

ShoT-UPfps:
## PR Checklist

## PR Type

A client that joins while the server has world updates buffered for the current tick can end up permanently out of sync. The same change gets applied to that client twice, once from the save it loads on join and once from a buffered update replayed right after, so its world silently diverges from everyone else's and never recovers. This path isn't reachable in normal play yet (the systems that buffer per-tick updates aren't enabled), but it's a real latent desync that would bite once trade/economy updates turn on.

## What is the new behavior?

Resolves #1595

The server now sends out any buffered per-tick updates before it snapshots the save for a joining client, so an update already baked into that save isn't also queued and replayed to the client. Joining clients stay in sync and the double-apply can't happen. No change for anyone already in the game.


source/Coop.Core/Server/Connections/States/TransferSaveState.cs
```diff
@@ -31,6 +38,18 @@ public TransferSaveState(
...
+            // delta would otherwise be both captured in the snapshot and replayed to this peer after BeginQueueing
+            // (double-apply). Guarded so a throwing send can't strand this blocking GameThread.Run.
+            try
+            {
+                coalescer.Flush(network);
+            }
+            catch (Exception ex)
+            {
+                Logger.Error(ex, "Failed to flush coalesced sends before the join save snapshot");
+            }
```
garrettluskey, line 51: Anyway we can make the flush not throw an exception?

garrettluskey (changes requested):

source/Coop.Core/Server/Connections/States/TransferSaveState.cs
ShoT-UPfps: thoughts on updating line 33 to `RunSafe` instead? otherwise I think we'll just move the try-catch inside of `Flush`

ShoT-UPfps: or add `FlushSafe` 🤪

garrettluskey: This is probably fine

garrettluskey (approved):

PR #1626, opened 2026-06-30
[1605] Reconcile live-created party attachment ids on rejoin

### LINKED ISSUE #1605: Rejoined client gives a live-created party's attachments divergent ids
Split out of #1599

## Symptom

A party created on the server while a client is connected resolves correctly on that client. But after the client exits to the main menu and rejoins, the same party's non-`MBObjectBase` attachments (`PartyBase`, member/prison `TroopRoster`, `ItemRoster`, the party component, the visual) now carry **different ids than the server holds**, so every server to client AutoSync update for that party fails `Failed to get id` and the party desyncs (wrong/zero data on the client).

left side is server, right side is client
Image

## Root cause

Two id-derivation schemes that never reconcile:

- **Live create** (server): AutoRegistry mints a runtime counter id, e.g. `PartyBase_Created_3328`, `LordPartyComponent_2688`, `TroopRoster_Created_4531`, `ItemRoster_3797`.
- **Join / `RegisterAllObjects`** (rejoining client): derives the id from the owning party's StringId, e.g. `PartyBase_Created_2835` (where `Created_2835` is the MobileParty StringId).

`MobileParty` matches because it is `MBObjectBase`: its runtime `Created_N` id is persisted into StringId, so both sides re-derive it. `MapEvent` matches for the same reason (StringId is a `SaveableProperty` and round-trips). The non-`MBObjectBase` attachments are not persisted that way, so a rejoining client invents ids the server never uses.

Reproduced live with `coop.debug.mobileparty.attachment_ids` (added in #1599): spawn a party on the server, run `attachment_ids` on the server and a present client (every id matches), then disconnect and rejoin the client and run it again (`PartyBase`/rosters/component/visual now differ).

## Scope

- #1599 fixes the present-at-create `ItemRoster` case (first screenshot there) and the reconnect-breaks-registration case (third screenshot), but **not** this id divergence (second screenshot).
- Same registration family as #1582 / #1584.

## Fix

Designed and mostly implemented as "Fix A": derive the live-create id from the owner's already-frozen StringId at first registration so it equals what `RegisterAllObjects` mints on rejoin (`PartyBase`/`TroopRoster`/`ItemRoster`/`PartyComponent`/visual, plus `MapEventSide`). It was split out of #1599 because it broke 72 E2E tests: `E2ETestEnvironment.CreateRegisteredObject` creates objects standalone and asserts immediate registration, but the fix defers/relocates registration (a component's MobileParty is null at construction). Needs E2E-harness reconciliation plus live validation in its own PR.

ShoT-UPfps:
## PR Checklist

## PR Type

A party created while a client is connected works fine for that client. But after the client exits to the main menu and rejoins, that party desyncs on the rejoined client: its troop count, inventory, party type, and AI behavior stop matching the host, and the client log fills with `Failed to get id` for it. A brand-new player who joins after the party was created hits the same desync. Parties that were already in the save when the host loaded it are unaffected.

When the client exits to main menu and rejoins,  parties that were created when it was connected now have the wrong StringIds
image

## What is the new behavior?

Resolves #1605

A client that  rejoins after a party was created now has that party fully in sync with the host: its troops, inventory, and party type line up with no `Failed to get id` errors for it.

On a client that rejoined:
image


garrettluskey:
Instead of this we must update how these ids get initially registered. On registration it should assign the same postfix as the party

garrettluskey:
Essentially the creation id and the id assigned during register all are diverging. The register all id seems more stable so maybe we can use the same assignment on the creation id

garrettluskey (dismissed):

garrettluskey (approved):

PR #1636, opened 2026-07-01
updated SpawnNewHeroesForNewCompanionClan and ApplyPlayerRelation

### LINKED ISSUE #1531: CompanionRolesHandler: promoting a companion to a new clan gives relation to Hero.MainHero instead of the requesting player
same `Hero.MainHero` bug as #1433 but in `CompanionRolesHandler.Handle_DoClanNameSelection` (the server apply when you promote a companion to their own clan via the "create new clan" conversation).

the handler already resolves the requesting player as `mainHero` (`CompanionRolesHandler.cs:94`) and uses it correctly for the gold and influence applies right below (`CompanionRolesHandler.cs:125-126`). but `SpawnNewHeroesForNewCompanionClan` gives relation with `Hero.MainHero` internally, which is null on the dedicated server (the host has no hero) so the relation gift breaks, and is the wrong hero otherwise. the author already flagged it inline:

`CompanionRolesHandler.cs:124`
```
companionRolesCampaignBehavior.SpawnNewHeroesForNewCompanionClan(oneToOneConversationHero, clan, selectedFief); // Gives relation with Hero.MainHero, need to expand this to work for mainHero
```

unlike #1433 this isnt a one-line swap because `SpawnNewHeroesForNewCompanionClan` is a vanilla `CompanionRolesCampaignBehavior` method that references `Hero.MainHero` inside it, so we need to reimplement/patch that relation step to use the resolved `mainHero` (like the other `//Replacement for...` reimplementations we already have in these handlers).

probably worth checking `ChangeRelationAction.ApplyPlayerRelation(oneToOneConversationHero, 50, ...)` on `CompanionRolesHandler.cs:127` too since that also targets the player.

NatteTosti69:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #1531

Changed `SpawnNewHeroesForNewCompanionClan` and `ApplyPlayerRelation` to use `clientMainHero` instead of `Hero.MainHero` which is null on the server.

### DIFF
--- a/source/GameInterface/Services/Companions/Handlers/CompanionRolesHandler.cs
+++ b/source/GameInterface/Services/Companions/Handlers/CompanionRolesHandler.cs
@@ -2,7 +2,9 @@
 using Common.Logging;
 using Common.Messaging;
 using Common.Network;
+using GameInterface.Services.Actions.Patches;
 using GameInterface.Services.Companions.Messages;
+using GameInterface.Services.Companions.Patches;
 using GameInterface.Services.ObjectManager;
 using Serilog;
 using System;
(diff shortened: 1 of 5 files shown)


source/GameInterface/Services/Companions/Handlers/CompanionRolesHandler.cs
```diff
@@ -87,48 +89,55 @@ private void Handle_DoClanNameSelection(MessagePayload<DoClanNameSelection> obj)
...
+            if (!objectManager.TryGetObjectWithLogging<MobileParty>(data.MainPartyId, out var mainParty)) return;
+
+            var companionRolesCampaignBehavior = Campaign.Current.GetCampaignBehavior<CompanionRolesCampaignBehavior>();
+
+            RemoveCompanionAction.ApplyByByTurningToLord(mainHero.Clan, oneToOneConversationHero);
+            oneToOneConversationHero.SetNewOccupation(Occupation.Lord);
+            TextObject textObject = GameTexts.FindText("str_generic_clan_name", null);
+            textObject.SetTextVariable("CLAN_NAME", new TextObject(data.ClanName, null));
+            int randomBannerIdForNewClan = companionRolesCampaignBehavior.GetRandomBannerIdForNewClan();
+            Clan clan = Clan.CreateCompanionToLordClan(oneToOneConversationHero, selectedFief, textObject, randomBannerIdForNewClan);
```
ShoT-UPfps: does this actually get past here on the server? GetRandomBannerIdForNewClan and Clan.CreateCompanionToLordClan both read Hero.MainHero which might NRE on server? (not sure)

ShoT-UPfps (changes requested):

source/GameInterface/Services/Companions/Handlers/CompanionRolesHandler.cs
NatteTosti69: You're right, `GetRandomBannerIdForNewClan` reads the culture from the mainhero which leads to different potential bannerclans. `Clan.CreateCompanionToLordClan` also sets `clan.Kingdom = Mainhero` which is also problematic.

source/GameInterface/Services/Companions/Handlers/CompanionRolesHandler.cs
```diff
@@ -121,14 +127,20 @@ private void Handle_DoClanNameSelection(MessagePayload<DoClanNameSelection> obj)
                     partyBelongedTo.Party.SetVisualAsDirty();
                 }
                 companionRolesCampaignBehavior.AdjustCompanionsEquipment(oneToOneConversationHero);
-                companionRolesCampaignBehavior.SpawnNewHeroesForNewCompanionClan(oneToOneConversationHero, clan, selectedFief); // Gives relation with Hero.MainHero, need to expand this to work for mainHero
+                // Set ResolvedMainHero for the duration of these calls so ChangeRelationActionPatches
```
ShoT-UPfps: is this comment in the right place ?

source/GameInterface/Services/Companions/Handlers/CompanionRolesHandler.cs
ShoT-UPfps: yep that covers it. one small thing while we're here, `ResolvedMainHero` is now a separate `[ThreadStatic]` in `CompanionRolesPatches`, `ChangeRelationActionPatches` and `ClanPatches`, all set and cleared to the same value together. could we make it one shared field the prefixes read instead? otherwise each new prefix that needs the hero adds another copy to set and clear and its easy to miss one.

source/GameInterface/Services/Companions/Handlers/CompanionRolesHandler.cs
NatteTosti69: updated

source/GameInterface/Services/Companions/Handlers/CompanionRolesHandler.cs
NatteTosti69: updated

ShoT-UPfps (approved):
lgtm, thanks for the fixes

PR #1639, opened 2026-07-01
[1624] Trim redundant party-icon rebuilds on client leader changes

### LINKED ISSUE #1624: Trim redundant party-icon visual rebuilds on the client
Some client leader-change handlers mark a party's map icon dirty unconditionally, forcing a full agent-visual teardown+rebuild (skeleton + weapon meshes) even when the leader didn't actually change. The crash this could trigger is fixed separately by #1552 (forcing synchronous prep), but the redundant rebuilds are still wasteful and add to the recurring caught `RefreshPartyIcon`/`ValidateIsDirty` NRE spam in the client log.

The party map figure is built from the leader (`GetVisualPartyLeader` → `LeaderHero.CharacterObject`) and the clan banner, and nothing in vanilla re-marks the visual dirty in reaction to a leader change, so the explicit `SetVisualAsDirty()` is needed on a real change but redundant on a no-op.

Offenders:
- `MobileParties/Handlers/PartyLeaderHandler.cs:74` — marks dirty even when the leader didn't actually change.
- `PartyComponents/Handlers/PartyComponentHandler.cs:270` — same, and for a player party both `PartyLeaderChanged` and `PartyComponentLeaderChanged` arrive, so the client rebuilds twice for one change.

Gate each on an actual leader change (mirroring vanilla's own `if (leader != Leader)`): only call `SetVisualAsDirty()` when the leader actually changed. That drops the no-op rebuilds and dedups the player-party double-rebuild.

Not in scope (checked, genuinely necessary and left as-is):
- `Banners/Handlers/PlayerBannerHandler.cs:95` — the map icon renders the clan banner, so every war-party in the clan genuinely changes on a banner edit; the rebuild is required and there is no lighter API than `SetVisualAsDirty()`.
- `Settlements/Handlers/SettlementOwnershipHandler.cs:97/:100` — settlements render through `SettlementVisual` (siege/wall/raid FX, no character agent-visuals), the owner banner/color is nameplate-driven and self-refreshes with no dirty flag, and these calls mirror `ChangeOwnerOfSettlementAction`, which the client deliberately skips replaying. They are not the agent-visual churn and are needed for parity.

Related: #1552.

ShoT-UPfps:
## PR Checklist

## PR Type

When a party's leader changes on the client, the party's whole map icon is torn down and rebuilt. That rebuild also runs when an update arrives that doesn't actually change the leader, and a player's own party rebuilds its icon twice for a single leader change. These extra rebuilds are wasteful and add to the recurring caught icon-refresh error spam in the client log.

## What is the new behavior?

Resolves #1624

The client rebuilds a party's map icon only when the leader actually changed, and only once per change. A genuine leader change still updates the icon to the new leader; redundant or no-op updates no longer trigger a rebuild.

### DIFF
--- a/source/GameInterface/Services/MobileParties/Commands/MobilePartyDebugCommand.cs
+++ b/source/GameInterface/Services/MobileParties/Commands/MobilePartyDebugCommand.cs
@@ -83,6 +83,44 @@ internal class MobilePartyDebugCommand
         return stringBuilder.ToString();
     }
 
+    // coop.debug.mobileparty.change_leader <PartyStringID> <HeroStringID>
+    // Run on the server: drives ChangePartyLeader so the leader-change sync fires to clients. The new
+    // leader must already be a member of the party's roster (vanilla no-ops otherwise).
+    [CommandLineArgumentFunction("change_leader", "coop.debug.mobileparty")]
+    public static string ChangeLeader(List<string> args)
+    {
+        if (ModInformation.IsClient)
+        {
+            return "Change leader is only to be called on the server";
+        }
+
+        if (args.Count < 2)
+        {
+            return "Usage: coop.debug.mobileparty.change_leader <PartyStringID> <HeroStringID>";
+        }
+
+        MobileParty mobileParty = Campaign.Current.CampaignObjectManager.Find<MobileParty>(args[0]);
+        if (mobileParty == null)
+        {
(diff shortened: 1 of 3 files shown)


ShoT-UPfps:
closing wont-do. the redundant rebuild this guards is low value on its own since player-party leader changes are rare, and the real fix is deduping the double leader-change send, tracked in #1642.

PR #1649, opened 2026-07-01
[1645] Stop the movement poller at mission end to fix a battle-teardown crash

### LINKED ISSUE #1645: Client AVE in AgentMovementHandler.PollAgents on battle teardown (Agent.IsActive on a freed agent)
client crashed with an `AccessViolationException` (read/write protected memory) after a coop field battle completed. the movement poller `AgentMovementHandler.PollAgents` runs on the `Poller` background thread and calls `agent.IsActive()` on an agent whose native object was already freed during mission teardown, so the guard call itself reads freed native memory and faults.

## stack
```
System.AccessViolationException: Attempted to read or write protected memory...
  Source=TaleWorlds.MountAndBlade
   at TaleWorlds.MountAndBlade.Agent.IsActive()
   at Missions.Agents.Handlers.AgentMovementHandler.PollAgents(TimeSpan delta) in AgentMovementHandler.cs:line 114
   at Common.Util.Poller.<PollAsync>d__9.MoveNext() in Poller.cs:line 89
   ...
   at System.Threading.Tasks.Task.DelayPromise.Complete()
   at System.Threading.TimerQueueTimer.Fire()
   at System.Threading.TimerQueue.FireNextTimers()
```

## why
`PollAgents` (`AgentMovementHandler.cs:114`) already tries to skip freed agents with `if (agent != null && agent.Mission != null && agent.IsActive())` — but at teardown the native `Agent` is freed while the registry still holds the reference, so `IsActive()` (native, `TaleWorlds.MountAndBlade`) is exactly the call that reads the freed memory. the guard can't protect itself. it's a race between the background poll thread and the game thread freeing the agents at mission end, so it's timing dependent (doesnt always crash).

## suggested direction
stop the poller (or clear the agent registry) at mission end before the agents are freed, instead of guarding from inside the poll. catching the AVE isnt reliable since it's real native corruption. same class as the defer-create/destroy-to-main-thread agent fixes.

related: #617 / #468 (AgentMovementHandler), sibling AVE #1552. surfaced during the #1633 manual test but unrelated to it (battle-end teardown race, not start-of-battle sizing).

ShoT-UPfps:
## PR Checklist

## PR Type

Finishing a field battle sometimes crashes a client to the main menu with `System.AccessViolationException: Attempted to read or write protected memory` (source `TaleWorlds.MountAndBlade`, thrown from `Agent.IsActive()`) as the battle screen closes back to the map.

The client broadcasts agent movement on a background poll thread (`AgentMovementHandler.PollAgents`, ~10ms cadence). At mission end the game thread frees the mission's native agents, but that poller keeps running: it was only ever stopped by the handler's finalizer, which the game doesn't run at teardown. So the poll thread reaches into a native agent (`agent.IsActive()` / `agent.Mission`) whose memory the game thread just freed, and the read faults. The existing per-agent guard can't prevent it, because the guard call is itself the native read that touches the freed memory.

It's a race between the poll thread and the game thread freeing agents, so it's timing dependent, the same battle doesn't always crash.

## What is the new behavior?

Resolves #1645

Battles now tear down cleanly and clients stay connected, so finishing a battle no longer crashes anyone back to the menu.

### DIFF
--- a/source/Common.Tests/Utils/PollerTests.cs
+++ b/source/Common.Tests/Utils/PollerTests.cs
@@ -1,0 +1,54 @@
+﻿using Common.Util;
+
+namespace Common.Tests.Utils
+{
+    public class PollerTests
+    {
+        [Fact]
+        public void StopAndWait_BlocksUntilInFlightPollCompletes()
+        {
+            // The poll enters, then parks until the test releases it — simulating a poll iteration that is
+            // still running when teardown asks the poller to stop.
+            var pollEntered = new ManualResetEventSlim(false);
+            var releasePoll = new ManualResetEventSlim(false);
+            var pollCompleted = false;
+
+            var poller = new Poller(_ =>
+            {
+                pollEntered.Set();
+                releasePoll.Wait();
+                Volatile.Write(ref pollCompleted, true);
+            }, TimeSpan.FromMilliseconds(5));
+            poller.Start();
+
+            Assert.True(pollEntered.Wait(TimeSpan.FromSeconds(5)), "poll never started");
+
+            // StopAndWait must not return while the poll iteration is still in flight.
+            var stopTask = Task.Run(() => poller.StopAndWait(TimeSpan.FromSeconds(5)));
+            Assert.False(stopTask.Wait(TimeSpan.FromMilliseconds(200)), "StopAndWait returned before the in-flight poll finished");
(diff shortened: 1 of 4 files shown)


ShoT-UPfps:
closing this, #1652 does the same poller teardown fix and more. going with that one.

Issue #1368, opened 2026-06-13
Integrate Steam for hosting and joining co-op sessions


ShoT-UPfps:
## Summary
Integrate Steam so players can host and join co-op sessions directly through Steam, instead of manually exchanging IP addresses and ports. A player should be able to launch the game, host a session backed by Steam, and invite friends through the Steam overlay / friends list — invitees join with one click, no IP entry.

## Motivation
Today, connecting to a co-op server requires manual network setup (sharing an IP/port, port forwarding, etc.), which is a real barrier for casual play. Steam provides lobbies, relay / NAT punch-through networking, friend invites, and the in-game overlay that together make "host and invite" frictionless and remove the need for port forwarding.

## Proposed user flow
1. Player launches Bannerlord with the mod.
2. Player chooses "Host" — a Steam lobby is created and the co-op server is started.
3. Player invites friends via the Steam overlay / friends list (or sends a Steam invite).
4. Friends accept and connect automatically through Steam networking — no manual IP entry.

## Scope / things to figure out
- Steamworks wrapper choice (Steamworks.NET vs Facepunch.Steamworks) and its licensing / compatibility with the mod and the .NET versions in use.
- Steam lobby creation/management and mapping a lobby to a running server instance.
- Transport: route the existing networking (LiteNetLib) over Steam's relay / P2P sockets, or keep the current transport and use Steam only for discovery + invites.
- Reconcile with the dedicated-server model (the host has no party/hero on the map): clarify whether "host" launches an embedded server that the hosting player then joins as a client, or connects to a co-located server process.
- Invite / join handshake: handle Steam lobby join requests and pass the connection details through to the client.
- Steam App ID requirement (Bannerlord's own app id vs the spacewar/dev app id for testing).
- Fallback: keep the direct-IP connection path available for the standalone ServerConsole / dedicated hosting.

## Out of scope (initial)
- Matchmaking / public server browser.
- Steam authentication / anti-cheat (VAC).

PR #1667, opened 2026-07-02
[1368] Steam integration phase 1: Steam invites working

ShoT-UPfps:
**host still has to port forward 4200/4201** the port forward requirement will be removed by #1668 
## PR Checklist

## PR Type

There is no Steam integration

## What is the new behavior?

Part of #1368

User starts the server the same way they normally have with `start-server.bat` and clicking `Host Co-Op Sandbox`. 

When a user click "Join Co-op Sandbox": 
image
image

Once user starts removing "localhost" from the Server IP Address, the "IP Address for Friends" box disappears, since they are ostensibly not the host and will not be sharing an IP Address with their friends. This is how users can connect directly without a Steam invite
image

When the user hosting the server joins, they are able to invite friends through steam
image

Checking the logs on the host's client we can see that a steam lobby was created
image

Running `coop.debug.steam.status` shows `advertising=True` indicating that the lobby is being advertised to user's steam friends
image

Testing invites thru the actual steam flow requires a friend with the mod on it. I don't have that so I open a second client and use this cheat to simulate accepting an invite through steam: `coop.debug.steam.join <lobby id from logs>` while in the main menu
image

And since I'm testing on the same steam account both players load into the same hero. But I verified that client 1 moving the party also moves it on client 2, and that other parties are moving, so syncing is working.


ShoT-UPfps:
closing, folding into https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/1703

PR #1678, opened 2026-07-02
[1571-PART2] Coalesce per-party AI behavior updates

### LINKED ISSUE #1571: De-duplicate and coalesce UpdatePartyBehaviorPacket per party
this issue came out of the investigation into #1557 (sustained campaign-map AI-sync bandwidth, ~250-380 KB/s, almost all world-wide AI): we profiled the server packet stream, traced the worst offenders, and split the durable fixes into separate issues. `UpdatePartyBehaviorPacket` is the #1 profiler line (~40 KB/s, ReliableUnordered). its a separate Coop.Core packet path (not AutoSync), carrying an AI party's behavior+target+position bundle, broadcast on every behavior change for every world party.

phase 1 (focused, ship first): remove the 2x double-send. today for server-controlled AI parties both emit sites fire per change: send #1 pre-apply at `ServerMobilePartyBehaviorHandler.cs:34`, then the local apply at `MobilePartyBehaviorHandler.cs:131` triggers send #2 post-apply at `RequestMobilePartyBehaviorPacketHandler.cs:58`. keep the post-apply one (it re-reads PartyPosition so its the more-correct copy) and drop the pre-apply duplicate. ~half this line, low risk.

phase 2: per-party latest-wins coalescing buffer keyed by MobilePartyId, covering both emit sites. the client apply (`MobilePartyBehaviorHandler.cs:92-139`) is a pure latest-state overwrite that doesnt accumulate, so coalescing is information-preserving (the client genuinely needs the latest target to move the party, so we coalesce, not drop). dropping the position correction is safe because Position is dead-reckoned client-side, not autosynced.

durable: no dependence on any disabled feature. could eventually merge into the same movement-snapshot channel as the AutoSync behavior fields, but ship standalone first.

depends on #1576 (the reusable per-tick coalescer): phase 2 should plug this packet path into the shared coalescer (per-party latest-wins key) rather than a bespoke buffer. phase 1 (drop the double-send) is independent and can land first.

ShoT-UPfps:
## PR Checklist

## PR Type

- [x] Other... Please describe: Performance / network-bandwidth optimization

Every time an AI party changes its behavior or target on the campaign map, the server sends a separate network update to clients, even when a party retargets several times in the same instant.

## What is the new behavior?

Resolves #1571.

The server now sends at most one behavior update per party per server tick, carrying that party's latest behavior and target, and drops the redundant intermediate sends.

| Metric (per 10s window, AI parties moving) | Current | New (this PR) |
| --- | --- | --- |
| party-behavior packets sent | ~1,763 | ~1,231 (-30%) |
| party-behavior bytes sent | ~168 KB | ~114 KB (-32%) |


source/Coop.Core/Client/Services/MobileParties/Packets/UpdatePartyBehaviorPacket.cs.cs
ShoT-UPfps, line 1: no longer needed because we swapped the packet broadcast for the coalesced `NetworkUpdatePartyBehavior` message, so nothing sends this anymore

source/Coop.Core/Client/Services/MobileParties/PacketHandlers/RequestMobilePartyBehaviorPacketHandler.cs
ShoT-UPfps, line 1: no longer needed because the receive side is now `NetworkPartyBehaviorMessageHandler`, which handles the new `NetworkUpdatePartyBehavior` message instead of this packet

garrettluskey (approved):

PR #1681, opened 2026-07-02
City Notifications

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

No city notifications

## What is the new behavior?

City notifications


source/GameInterface/Services/UI/Notifications/Handlers/SettlementNotificationHandler.cs
```diff
@@ -0,0 +1,229 @@
...
+            if (!objectManager.TryGetObjectWithLogging<Hero>(data.RecruiterHeroId, out var recruiterHero)) return;
+            if (!objectManager.TryGetObjectWithLogging<Settlement>(data.SettlementId, out var settlement)) return;
+            if (!objectManager.TryGetObjectWithLogging<CharacterObject>(data.TroopId, out var troop)) return;
+
+            Hero troopSource = null;
+            if (data.TroopSourceHeroId != null && !objectManager.TryGetObjectWithLogging(data.TroopSourceHeroId, out troopSource)) return;
+
+            using (new AllowedThread())
+            {
+                CampaignEventDispatcher.Instance.OnTroopRecruited(recruiterHero, settlement, troopSource, troop, data.Amount);
```
ShoT-UPfps: doesnt this also trigger other side effects than the city notification pop-up? i think this might be doing too much

EgardA: The vanilla recruitment from what I can tell also calls this method, so it should not impact anything as they all have their own conditions. However, this would probably cover more than just city notifications like army notifications etc.

ShoT-UPfps: hmm ok. Still, I wonder if it makes more sense to call the VM directly? SettlementNameplateNotificationsVM.OnTroopRecruited

garrettluskey: I have to agree with @ShoT-UPfps, what if Taleworlds (or some other mod) hooks into that event

ShoT-UPfps (approved):

PR #1690, opened 2026-07-03
[1662] Fix AgentData mount NRE race on the poll thread

### LINKED ISSUE #1662: AgentData can NRE on the poll thread
from a static review of #1647, this basically never fires so it should be validated first (might not be worth chasing).

`AgentData` ctor snapshots `Agent mount = agent.MountAgent` (`AgentData.cs:32`) then guards with `agent.HasMount && mount.IsActive()` (`AgentData.cs:33`). `HasMount` re-reads `MountAgent`, so the guard and the deref read the field twice. `AgentData` is built on the background poll thread (`AgentMovementHandler.PollAgents`) while mount/dismount happens on the game thread, so if an agent mounts in the gap between the two reads, `HasMount` is true but the `mount` snapshot is null and `mount.IsActive()` NREs. the poller swallows it, so we just drop that agent's movement for that tick.

ShoT-UPfps:
## PR Checklist

## PR Type

## What is the new behavior?
Resolves #1662


EgardA (approved):
LGTM

garrettluskey (approved):

PR #1694, opened 2026-07-03
[1668] Steam integration phase 2: Tunnel sessions over Steam P2P so friends can join without host port forwarding

### LINKED ISSUE #1668: Steam P2P transport tunnel: join co-op sessions without port forwarding
## Summary
Phase 2 of #1368. With Steam lobby discovery and invites in place (#1667), the remaining barrier is the host's router: friends still connect to the host's public IP over plain UDP, so the host must port-forward UDP 4200-4201 and know their public address. Route the existing LiteNetLib traffic over Steam's peer-to-peer networking (automatic NAT punch-through + Valve relay fallback) so a hosted session is joinable with zero router configuration.

## Design (already partly worked out during #1368 phase 1)
- This requirement was added after the below points were generated: Ideally, starting the server also starts the steam lobby and etcetera, visible to the player's own client and friends. The player can then open the client, click "Join Co-op Campaign" and see a list of steam lobbies as well as directly connect to a server via IP.
- Keep LiteNetLib end-to-end; tunnel its UDP datagrams through `SteamNetworkingSockets` P2P connections via a small localhost datagram pump on each side. LiteNetLib datagrams are ≤1432 bytes and map 1:1 onto unreliable Steam messages (≤512 KB cap), so LiteNetLib's own reliability/ordering keeps working unchanged. `INetwork`, `NetPeer` identity, the join handshake, and every handler stay untouched.
- The `/server` process gets its own Steam identity via the GameServer API (`GameServer.Init` + `LogOnAnonymous`), pumped by `GameServer.RunCallbacks()` from the mod's update loop — this avoids the same-account identity collision with the host player's client. Fallback topology if GameServer init misbehaves next to the game's own user-mode `SteamAPI.Init`: host the pump in the hosting player's client instead (Barotrauma's owner-relay pattern); the phase-1 interfaces cover both.
- Plugs into the phase-1 seams: a new `IJoinEndpointPreparer` implementation starts the client pump and returns `127.0.0.1:<pumpPort>`; a new `ISessionJoinInfoSource` implementation waits for the server's Steam id (new server→client message) and the lobby data gains a `ServerSteamId` key (the lobby-data codec is already version-gated). The `Public Address` field then becomes unnecessary for Steam sessions.
- Mission mesh: under the tunnel, client↔client NAT punch cannot succeed (the rendezvous observes loopback endpoints), so the existing `RelayPacket` server relay becomes the primary mission transport for tunneled sessions. Phase 2 must suppress punch attempts for tunneled peers (don't introduce loopback endpoints) and measure a relayed battle's latency/bandwidth to decide whether direct client↔client Steam sessions (both are real distinct Steam users) need to be pulled forward.

## Verification notes
- Needs two machines/Steam accounts for the P2P legs; `SteamNetworkingSockets.CreateSocketPair` can drive single-machine pump tests.
- Runtime-verify early: GameServer anonymous login inside a Bannerlord `/server` process where TaleWorlds already initialized user-mode SteamAPI.

ShoT-UPfps:
### 
**This is a stacked PR**

## PR Checklist

## PR Type

The user must port forward 4200/4201.

## What is the new behavior?

Resolves #1668

Steam lobbies now connect players through Steam's P2P networking with a Valve relay fallback, so the host doesn't need to port forward.

### DIFF
--- a/source/Common.Tests/Network/Session/TunnelAdvertisementTests.cs
+++ b/source/Common.Tests/Network/Session/TunnelAdvertisementTests.cs
@@ -1,0 +1,90 @@
+﻿using Common.Network;
+using Common.Network.Session;
+using Moq;
+
+namespace Common.Tests.Network.Session;
+
+public class TunnelAdvertisementTests
+{
+    private sealed class FakeTunnelHost : ISessionTunnelHost
+    {
+        public bool StartThrows;
+        public bool Listening;
+        public int? StartedPort;
+
+        public bool IsListening => Listening;
+        public int PeerCount => 0;
+
+        public void Start(int serverPort)
+        {
+            if (StartThrows) throw new InvalidOperationException("no relay");
+
+            StartedPort = serverPort;
+            Listening = true;
+        }
+
+        public void Stop() => Listening = false;
+
+        public void Dispose()
+        {
+        }
+    }
+
+    private static INetworkConfig Config(string address, bool tunneled = false)
+    {
+        var config = new Mock<INetworkConfig>();
+        config.Setup(c => c.Address).Returns(address);
+        config.Setup(c => c.Port).Returns(4200);
+        config.Setup(c => c.IsTunneled).Returns(tunneled);
+        return config.Object;
+    }
+
+    [Theory]
+    [InlineData("localhost")]
+    [InlineData("127.0.0.1")]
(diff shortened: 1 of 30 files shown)


ShoT-UPfps:
closing, folding into https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/1703

PR #1716, opened 2026-07-04
[1704] Fix field battle scoreboard troop count desync for alive/dead/wounded/routed

### LINKED ISSUE #1704: [Report] Field Battle Troop Scoreboard incorrect tracking
see https://discord.com/channels/709516043332354119/1522699739077017620/1522699739077017620

ShoT-UPfps:
## PR Checklist

## PR Type

In a field battle with multiple players both players will see different values for alive/dead/wounded/routed

Resolves #1704

## What is the new behavior?

Alive/dead/wounded/routed stays synced between both


source/Coop/CoopMod.cs
```diff
@@ -88,16 +88,43 @@ public override void NoHarmonyInit()
...
 
+        // Held open for the whole process lifetime (never disposed) so the claim can't race with a second
+        // instance's own attempt: a check-then-release probe would leave a window, between our own close and
+        // Serilog's later open of the real log file, where a second process's identical probe could also
+        // succeed. A sidecar lock file kept open under FileShare.None the entire time closes that window, and
+        // is independent of whatever sharing mode Serilog itself uses to open the actual log file.
+        private static FileStream logLockHandle;
+
+        // True if filePath was free (no other live process holds its lock).
+        private static bool TryClaimExclusive(string filePath)
```
ShoT-UPfps, line 99: This is making it so that we can have logs for both clients if we're running 2 clients. One client will have the normal `Coop_client.log` the other client will have `Coop_client_{pid}.log`. This is more important now that we are testing field battles where 2 client testing is extremely common

ShoT-UPfps:
Codefactor recommendation is bad, project wo't compile with it

source/Coop/CoopMod.cs
```diff
@@ -88,16 +88,43 @@ public override void NoHarmonyInit()
...
 
             var filePostfix = isServer ? "server" : "client";
             var filePath = $"Coop_{filePostfix}.log";
 
+            // File.Delete alone can't detect another live instance: it succeeds even on a file another
+            // process still has open, as long as that handle allows shared delete (Serilog's file sink
+            // does), so two same-install clients would silently keep fighting over one file. An exclusive
+            // open is the only check that actually fails when someone else already has the file open.
+            if (!TryClaimExclusive(filePath))
+                filePath = $"Coop_{filePostfix}_{System.Diagnostics.Process.GetCurrentProcess().Id}.log";
```
garrettluskey, line 150: Since our log files are currently not truncated and get to the MBs, unless we delete all log files before starting a game, I see a future where PCs run out of space

garrettluskey (dismissed):

source/Coop/CoopMod.cs
ShoT-UPfps: Updated to keep the last 5 logfiles from client 2 (sidenote, i think we should keep more for the main client too but didnt add that here)

garrettluskey (approved):

PR #1753, opened 2026-07-05
[1730] Fix client flickering loading screen after field battle

ShoT-UPfps:
this is solving for a race condition. When a field battle ends:

1. the client ends up nulling its `party.MapEventSide` (happens from server packet that gets queued on gamethread)
2. the client's scoreboard tries to fetch the rewards (i.e. renown) which relies on `party.mapEventSide` (happens from `Game.OnTick`)

these 2 can happen in any order but if (1) happens before (2) then this soft lock happens

The fix in this PR is to cache the rewards and use the cache when scoreboard tries to fetch the rewards if the mapEventSide was already nulled


ShoT-UPfps:
waiting till https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/1716 gets merged to test - need the second client's logs

source/GameInterface/Services/MapEvents/MainPartyBattleRewardsCache.cs
```diff
@@ -0,0 +1,56 @@
+﻿using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.MapEvents;
+
+namespace GameInterface.Services.MapEvents;
+
+// Snapshot of MainParty's battle-reward figures, captured by MapEventRegistry.OnClientDestroyed right before it
+// nulls MainParty's MapEventSide. PlayerEncounterPatches.GetBattleRewardsPrefix falls back to this when that
+// teardown already removed MainParty from both MapEvent sides before the client's own scoreboard could read it
+// live.
+internal static class MainPartyBattleRewardsCache
```
garrettluskey: Kindly make transient with interface

garrettluskey (dismissed):

source/GameInterface/Services/MapEvents/MainPartyBattleRewardsCache.cs
ShoT-UPfps: updated to use DI but didnt make transient because other components have to share it

source/GameInterface/Services/MapEvents/Patches/PlayerEncounterPatches.cs
```diff
@@ -115,6 +119,44 @@ private static bool PrefixCheckNearbyPartiesToJoinPlayerMapEvent()
         return false;
     }
 
+    // Vanilla looks up MainParty's MapEventParty via PartiesOnSide(PlayerSide), which indexes _sides by
+    // (int)PlayerSide and throws once MapEventRegistry.OnClientDestroyed nulls MainParty's MapEventSide first
+    // (PlayerSide becomes BattleSideEnum.None). Search both sides for MainParty directly instead; if teardown
+    // already removed it from both, fall back to the snapshot OnClientDestroyed captured before nulling it.
+    [HarmonyPatch(nameof(PlayerEncounter.GetBattleRewards))]
+    [HarmonyPrefix]
+    private static bool GetBattleRewardsPrefix(PlayerEncounter __instance, out ExplainedNumber renownChange,
```
garrettluskey, line 139: This may conflict with #1778

garrettluskey (approved):

Issue #1750, opened 2026-07-05
Give mounts a stable cross-client identity for routed damage


ShoT-UPfps:
Follow-up from #1728.

`BattleDamageRouter.Handle_NetworkApplyBattleDamage` resolves a routed mount hit as `info.Agent.MountAgent` at apply time, keyed off the rider's registry id (mounts aren't registered themselves). This works for the common case but has a narrow race: if the rider dismounts and remounts a different horse inside the network round-trip between the hit being intercepted and the routed message being applied, the blow lands on the new horse instead of the one actually struck.

Low severity/likelihood: requires two mount changes on the same rider within a single routed message's flight time (tens of ms).

**Second consequence of the same gap (found by codex-review on #1728):** when a routed mount hit is lethal, `BattleAgentDiedPatch` publishes `BattleAgentDied` for the horse, but `AgentDeathReporter` silently drops it because only the rider is registered, not the mount. So the horse's death is never broadcast to peers. The rider's dismount still replicates correctly (via the existing movement-sync path: the owner's `MountData` drops out of its next `MovementPacket` once `HasMount` goes false, and `SyncMountState` unmounts the puppet everywhere), so #1728's actual bug (rider stuck at horse speed) is fixed regardless. But the horse itself is left standing as a riderless "zombie" agent on every client except its true owner's, since it never dies/ragdolls anywhere else.

Fixing either of these properly means giving mounts their own stable identity in the agent registry (like riders have), so a routed hit resolves by the mount's own id instead of the rider's current live state, and so a mount death can be broadcast the same way a rider's is. That's a bigger change than #1728's fix (registering mounts, deciding how their deaths get broadcast, deciding whether a masterless horse should also become puppet-gated) and was intentionally scoped out of #1728.

PR #1754, opened 2026-07-05
[1728] Route mount hits through the rider's owner

### LINKED ISSUE #1728: [ISSUE] After a client kills a horse with a rider, the rider still moves at horse speed
Observed in my own gameplay. Check multiple clients in same battle

ShoT-UPfps:
## PR Checklist

## PR Type

After a client kills the horse under an enemy rider, the horse dies only on that client's own screen. Every other machine — including the rider's real owner — still shows the owner reporting itself mounted, so the rider keeps sliding around at horse speed (or snaps back onto the dead horse) instead of dismounting, on every machine but the killer's.

## What is the new behavior?

Resolves #1728

A hit against a horse is now routed to the horse's rider's real owner the same way a hit against a human puppet already is, so the kill is decided once, by the owner, instead of independently on whichever client's local swing happens to land it. Once the owner's own copy of the rider properly dismounts, every other client's existing movement sync picks that up and dismounts its copy too — the rider now walks at normal foot speed everywhere, not just on the client that struck the blow.

A horse with no rider, or one carrying your own troop, still resolves the same as before (no behavior change there).

## Other information

Filed #1750 to track two known follow-ups from this fix's review, both from the same root cause (mounts have no cross-client identity of their own — routing is keyed off the rider instead): a narrow remount-timing race that could misapply damage to the wrong horse, and lethal mount deaths not being broadcast to peers (so a killed horse can be left standing, riderless, on non-owner clients even though the rider correctly dismounts).

### DIFF
--- a/source/GameInterface/Services/MapEvents/Messages/BattlePuppetHit.cs
+++ b/source/GameInterface/Services/MapEvents/Messages/BattlePuppetHit.cs
@@ -13,18 +13,22 @@ namespace GameInterface.Services.MapEvents.Messages;
 /// </summary>
 public record BattlePuppetHit : IEvent
 {
-    /// <summary>The puppet that was hit (owned by another client).</summary>
+    /// <summary>The puppet that was hit, or (for a mount hit) the puppet riding it — mounts are never
+    /// registered/puppet-gated themselves, so ownership is always resolved through a rider.</summary>
     public Agent Victim { get; }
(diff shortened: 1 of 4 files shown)


ShoT-UPfps:
waiting till https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/1716 gets merged to test - need the second client's logs

garrettluskey:
Instead, I'm thinking we treat the mount the same as a troop. Registration, battle damage, ownership, etc... I'm planning on working this

PR #1762, opened 2026-07-06
[1750] Mount refactor

### LINKED ISSUE #1728: [ISSUE] After a client kills a horse with a rider, the rider still moves at horse speed

### LINKED ISSUE #1750: Give mounts a stable cross-client identity for routed damage

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #1750
Resolves #1728

### DIFF
--- a/source/Missions/Agents/Handlers/MountMovementApplier.cs
+++ b/source/Missions/Agents/Handlers/MountMovementApplier.cs
@@ -1,0 +1,79 @@
+using Common;
+using Common.PacketHandlers;
+using Common.Util;
+using LiteNetLib;
+using Missions.Agents.Packets;
+using System.Collections.Generic;
+using TaleWorlds.MountAndBlade;
+
+namespace Missions.Agents.Handlers;
+
+/// <summary>
+/// Receive side for <see cref="MountMovementPacket"/>: applies another owner's MASTERLESS-horse snapshots to
+/// our local copies of those horses, through the same <see cref="AgentMountData.ApplyMount"/> path a ridden
+/// horse's pose takes, and feeds the shared position interpolator. Send side lives in
+/// <see cref="AgentMovementHandler"/>'s poll (one registry pass partitions riders and riderless horses), which
+/// also owns this applier's packet-manager registration so the two movement handlers share one lifecycle —
+/// this class has no threads or subscriptions of its own.
+/// </summary>
+public class MountMovementApplier : IPacketHandler
+{
+    private readonly INetworkAgentRegistry agentRegistry;
+    private readonly IAgentPositionInterpolator interpolator;
+
+    public MountMovementApplier(INetworkAgentRegistry agentRegistry, IAgentPositionInterpolator interpolator)
+    {
+        this.agentRegistry = agentRegistry;
+        this.interpolator = interpolator;
+    }
+
+    public PacketType PacketType => PacketType.MountMovement;
+
+    // Nothing to tear down: no threads, no subscriptions; the owning AgentMovementHandler removes this
+    // handler's packet-manager registration in its own Dispose.
+    public void Dispose() { }
+
+    public void HandlePacket(NetPeer peer, IPacket packet)
+    {
+        var movement = (MountMovementPacket)packet;
+        if (movement.MountIds == null) return;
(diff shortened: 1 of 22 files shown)


EgardA (approved):
Game launches and no virus attached, LGTM

Code factors are fake too

source/Missions/Agents/Handlers/MountMovementApplier.cs
```diff
@@ -0,0 +1,79 @@
...
+        if (movement.MountIds == null) return;
+
+        // Resolve the horses to apply (skipping our own) on the network thread, then apply the whole batch
+        // in ONE game-thread action — matching AgentMovementHandler.HandlePacket.
+        var toApply = new List<(Agent horse, AgentMountData data)>();
+        for (int i = 0; i < movement.MountIds.Length; i++)
+        {
+            var mountId = movement.MountIds[i];
+            if (agentRegistry.IsLocallyControlled(mountId)) continue;
+            if (!agentRegistry.TryGetAgentInfo(mountId, out var mountInfo)) continue;
```
ShoT-UPfps, line 48: I think these lookups need to be in the `RunSafe`

ShoT-UPfps: If the game thread queue currently has a callback that is supposed to spawn a horse then this lookup misses it otherwise

ShoT-UPfps (changes requested):

source/Missions/Battles/BattleAuthorityMigrator.cs
```diff
@@ -188,24 +194,62 @@ private void DespawnControllerTroops(string controllerId)
...
+
+            var rider = agent.RiderAgent;
+            bool riderWithdrew = rider != null && despawnedRiders.Contains(rider);
+            if (!allMounts && !riderWithdrew) continue;
+
+            bool ridden = rider != null && rider.IsActive() && !riderWithdrew;
+            if (!ridden && agent.IsActive())
+                agent.FadeOut(false, true);
+
+            registry.RemoveAgent(info.AgentId);
```
ShoT-UPfps, line 245: isnt it possible that a different client than the one that spawned it in is on this horse and we unregister the horse?

ex: 
non-host client A is retreating
client B is currently riding a horse originally spawned and registered under A
allMounts is true, so the for loop does not `continue`
B is active and did not withdraw, so the horse does not fade
the horse is still removed from the registry and B keeps riding it

PR #1776, opened 2026-07-07
[1698] Fix battle agents sliding and snapping

### LINKED ISSUE #1698: Troops on foot in battles appear to glide/slide along the ground
# For New Contributors

## Description

Some troops (both foot troops and horses) in a field battle will not play their movement animations and instead appear to glide/slide along the ground.

## Location

## Definition of Done

ShoT-UPfps:
## PR Checklist

## PR Type

Troops and mounts owned by other clients slide across the ground with no movement animation. Mounted ones also keep trotting forward and snap back to the owner when the owner slows down or gets stopped by colliding with a crowd.

## What is the new behavior?

Resolves #1698

mounted puppets now animate their gait and track the owner smoothly, no more sliding or snapping.

why it didnt work out of the box: foot and mounted puppets were driven the same way, apply the owner's `MovementInputVector` and ease the position toward the owner with `SetTargetPositionAndDirection`. that works for foot because human walk/run is procedural, the engine builds the walk cycle from the synced input with no controller needed. a horse is different. its gait (walk/trot/canter/gallop) is a discrete action on channel 0 that the native locomotion controller picks from the mount's speed, and our puppets are `Controller.None` so nothing picks it and the horse just slides. and `SetTargetPositionAndDirection` physically seeks the horse toward the target at its own gallop speed and collides with the local crowd, so when the owner slowed or got stopped by troops the puppet couldnt match its speed or reach its spot, it wedged and snapped back.

two changes fix it:
- sync the mount's channel 0 gait action. we already synced channel 1 but never captured channel 0 (the gait), so the horse now animates and re-syncs as the owner shifts walk/trot/canter/gallop/stand.
- position mounted puppets by easing the horse straight onto the owner's reported position each frame instead of physically seeking it, so it tracks the owner exactly, theres no seek to clobber the gait, and it doesnt fight the local crowd. tradeoff is the mounted puppet is a visual proxy at the owner's position now, so it can clip through local units.


source/Missions/Agents/AgentPositionInterpolator.cs
```diff
@@ -128,5 +145,124 @@ public void Tick(float dt)
...
         }
+
+        diagnosticElapsed += dt;
+        if (diagnosticElapsed >= DiagnosticInterval)
+        {
+            diagnosticElapsed = 0f;
+            int tracked = foot.Tracked + mounted.Tracked + mounts.Tracked;
+            if (tracked > 0)
+            {
+                Logger.Debug(
```
ShoT-UPfps: open to removing this if desired but it was a pain in the ass to debug this so i'd be in favor of keeping it in

ShoT-UPfps:
the codefactors are fake

garrettluskey (approved):

PR #1777, opened 2026-07-07
Sieges

ShoT-UPfps:
## PR Type

no sieges synced

## What is the new behavior?

full siege sync, reusing the field-battle P2P infra:

- campaign layer: besieger camp, siege event, engine containers + construction progress, strategy and wall hp all sync; `SiegeEventManager.Tick` and the siege campaign behaviors run server-authoritative
- entry/exit: besiege / join / break go through the server, and a player party inside a besieged town gets pulled into the defender encounter
- the walls assault runs as a coop mission (`CoopSiegeBattleLauncher`) on the same mesh/puppet/owner-damage/host-commit path as field battles; engines deploy and machine state (rams, towers, ladders, gates, hp) replicate from the mission host
- capture happens at the walls like a vanilla AI assault; aftermath is server-applied, and a player-led devastate/pillage/mercy choice routes through the server
- the production popup build/remove engine requests are server-authoritative

not synced: the lords-hall stage (capture at the walls instead, same as vanilla AI assaults)

known follow-ups:

- ammo counts can diverge when a machine changes simulator
- the lead besieger is correctly in the settlement after winning the siege but the locations list is missing all the notables until a player enters the settlement at which point it loads correctly on the lead besieger's screen 
image

- the server reuses the cached mission snapshot when re-assaulting the same siege, so engine damage from a failed assault doesnt show in the next assault
- when the battle host leaves with no successors the cleared assignment isnt broadcast, so clients can keep a stale host between assaults of the same siege
- a local melee hit on a machine can race the replicated destruction state and throw IndexOutOfRange in ReplaceEntityWithBrokenEntity
- sometimes on a client the troops that were moving the battering ram will continue in the "Swinging the ram" animation even after the door is broken and they have moved into the subsequent hallway
- (unsure if happens only to host or nonhost or both) one of the clients will say "Defenders ran away" and get the victory screen prematurely while the other client has to wait for them to run to the keep before they get the victory screen (Which does not then say "defenders ran away")
- (probably unrelated) if its a player defender vs player attacker in siege and the attacker retreats while defender is in deployment phase, the defender's screen doesn't update with victory. Defender can open scoreboard to also retreat but then the players are soft locked as the campaign map says 2 palyers are still in a map event


ShoT-UPfps:
all codefactors are complex methods

source/E2E.Tests/Services/Villages/VillageHostileActionTests.cs
garrettluskey, line 1: What?

garrettluskey (changes requested):

source/GameInterface/Services/SiegeEvents/Patches/SiegeDebugPinPatch.cs
```diff
@@ -0,0 +1,24 @@
...
+
+/// <summary>
+/// Debug-only: for a party pinned by coop.debug.siege.start, skips AiMilitaryBehavior's hourly military
+/// think so the campaign AI never re-tasks it off the forced siege. Without this the think re-scores the
+/// lord, flips its DefaultBehavior off BesiegeSettlement, and the besieger camp ejects it. The siege
+/// still advances to an assault because that is driven by the siege/encounter system, not this think.
+/// No-op unless the cheat has pinned a party.
+/// </summary>
+[HarmonyPatch(typeof(AiMilitaryBehavior), "AiHourlyTick")]
+internal static class SiegeDebugPinPatch
```
garrettluskey: Can we remove this?

garrettluskey (changes requested):

garrettluskey (approved):

PR #1792, opened 2026-07-09
Skill xp properly assigned from focus points after character creation

### LINKED ISSUE #1769: [Report] Focus point assignment during character creation should result in +10 lvls of the skill but does not
Verify the  asserted vanilla behavior before "fixing" it in co-op. See https://discord.com/channels/709516043332354119/1523455402044559360/1523455402044559360

jordanbrymora:
## PR Checklist

## PR Type

Post character creation is calling patched `HeroDeveloper` methods, resulting in errors because objects can't be resolved on the network at this stage of a client connecting.

## What is the new behavior?

Resolves #1769 

Clients now get their starting skill levels and perk points correctly.

Fixed by putting post character creation methods that use `Hero.MainHero.HeroDeveloper` into `AllowedThreads` so local modifications are allowed before sending the created hero data to the server.

### DIFF
--- a/source/GameInterface/Services/GameState/Patches/AllowCharacterCreationContentPatches.cs
+++ b/source/GameInterface/Services/GameState/Patches/AllowCharacterCreationContentPatches.cs
@@ -1,0 +1,32 @@
+﻿using Common.Util;
+using HarmonyLib;
+using System.Collections.Generic;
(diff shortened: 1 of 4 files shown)


EgardA (approved):
LGTM 👍

PR #1826, opened 2026-07-11
Tournaments

zzzzzzzzott:
image

Mostly reuses battle infrastructure, but needed some key differences - 
* Spawning is custom for obvious reasons.
* Damage/knockouts are attacker authoritative rather than victim authoritative because every client must apply the same damage to keep native tournament state synchronized. Puppet health desync is also more obvious in tournaments.
* Late joining is different because players need the current match state, including health, scores, bracket state, equipment, and dropped items. I just made an entirely different handler to do this rather than edit the existing.
* Host migration restores different state. Tournament hosts resume the current bracket/match from its spawn manifest and runtime snapshot, while other mission battle hosts take over MapEvent party and troop state.
* Lifecycle handling is different for obvious reasons.

There is potentially some other stuff that **could** have been reused, but I didn't really want to fuck with other areas for this.

There is an argument to be made that tournament puppet syncing could take a much more aggressive approach to syncing compared to typical battles since there is much less to sync and its more important to the game state that tiny minuet details are synced on puppets.

Like for one, projectiles should def be hard synced in tourneys, but that can come later. (Could be because of local testing, but it didn't always look like projectiles were in sync)

But I did pretty much fully reuse network agent registry/ownership tracking, puppet movement/interpolation, animations, mount movement/state, projectile spawning, weapon pickup/drop, shield damage, along with general networking

UI FLOW - 
image
Standard bannerlord ui

image
After clicking join

image
After second player clicks join (both clients have the same name)

image
Players must all vote on an option before the tournament proceeds

image
After a tournament starts, its locked in that city. Players may still join to spectate.

Leaving the tournaments forfeits that players position and replaces their spot with a recruit of that towns culture.
If all players leave, the tournament immediately resolves and completes.

## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

no tournaments

## What is the new behavior?
tournaments


ShoT-UPfps:
What's the UX like? How do multiple players join tournaments together?

source/GameInterface/Services/Tournaments/TournamentBettingMath.cs
```diff
@@ -0,0 +1,48 @@
...
+        float ownPower = heroPower + playerTeamPower;
+        float currentMatchPower = ownPower + currentMatchOpponentPower;
+        float otherMatchPower = Math.Max(
+            0f,
+            totalRoundPower - playerTeamPower - currentMatchOpponentPower);
+        float adjustedRoundPower = ownPower + (0.5f * otherMatchPower);
+        if (ownPower <= 0f || currentMatchPower <= 0f || adjustedRoundPower <= 0f)
+            return 1.1f;
+
+        float winFactor = (ownPower / currentMatchPower) / (ownPower / adjustedRoundPower);
```
ShoT-UPfps: i think the math here is different than vanilla? ownPower vs heroPower?

source/GameInterface/Services/Tournaments/TournamentLifecycleCoordinator.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    public TournamentLifecycleCoordinator(
+        IMessageBroker messageBroker,
+        ITournamentSessionRegistry sessionRegistry)
+    {
+        this.messageBroker = messageBroker;
+        this.sessionRegistry = sessionRegistry;
+    }
+
+    public bool CompleteActiveSessionsAndReleasePreparation()
```
ShoT-UPfps: This doesn't seem to be called? but based on what it's doing i think it should be?

source/GameInterface/Services/Tournaments/TournamentSessionRegistry.cs
```diff
@@ -0,0 +1,1063 @@
...
+                ? TournamentMutationStatus.StaleRevision
+                : TournamentMutationStatus.Rejected;
+            return false;
+        }
+
+        status = TournamentMutationStatus.Applied;
+        return true;
+    }
+
+    private void RemoveSession(TournamentSessionState session)
```
ShoT-UPfps, line 742: Is this ever used? (it should be - otherwise there's a memory leak)

source/GameInterface/Services/Tournaments/TournamentGameInterface.cs
```diff
@@ -0,0 +1,633 @@
...
+                        if (!participants.TryGetValue(slotId, out var participant))
+                            return false;
+                        team.AddParticipant(participant);
+                        matchParticipants.Add(participant);
+                    }
+                    teams[teamIndex] = team;
+                }
+
+                typeof(TournamentMatch)
+                    .GetField("_teams", System.Reflection.BindingFlags.Instance | System.Reflection.BindingFlags.NonPublic)
```
ShoT-UPfps: Publicizer?

source/Missions/Tournaments/TournamentNativeBracketHydrator.cs
```diff
@@ -0,0 +1,158 @@
...
+namespace Missions.Tournaments;
+
+public interface ITournamentNativeBracketHydrator
+{
+    void Apply(CoopTournamentBehavior behavior, TournamentSessionSnapshot snapshot);
+}
+
+public class TournamentNativeBracketHydrator : ITournamentNativeBracketHydrator
+{
+    private static readonly FieldInfo MatchTeams = typeof(TournamentMatch).GetField("_teams", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public);
```
ShoT-UPfps: Publicizer?

source/GameInterface/Services/Tournaments/Patches/DisableTournamentCampaignBehavior.cs
```diff
@@ -1,11 +1,12 @@
...
+using Common;
 
 namespace GameInterface.Services.Tournaments.Patches;
 
 [HarmonyPatch(typeof(TournamentCampaignBehavior))]
 internal class DisableTournamentCampaignBehavior
 {
     [HarmonyPatch(nameof(TournamentCampaignBehavior.RegisterEvents))]
-    static bool Prefix() => false;
+    static bool Prefix() => ModInformation.IsServer;
```
ShoT-UPfps, line 25: please patch the individual handlers instead of RegisterEvents

source/GameInterface/Services/Tournaments/Handlers/TournamentHitProgressionHandler.cs
```diff
@@ -0,0 +1,139 @@
...
+                CombatXpModel.MissionTypeEnum.Tournament,
+                data.AttackerMounted,
+                data.SameTeam,
+                false,
+                data.DamageAmount,
+                data.Fatal,
+                false,
+                data.Charging,
+                data.SneakAttack);
+            acceptedHitProgression.Add(dedupeKey);
```
ShoT-UPfps, line 78: does anything ever clear this or does stuff only ever get added to it? (memory leak)

source/GameInterface/Services/Tournaments/UI/TournamentUIController.cs
```diff
@@ -0,0 +1,440 @@
...
+    private void SendLeaveActive(TournamentSessionSnapshot snapshot)
+    {
+        network.SendAll(new NetworkRequestLeaveActiveTournament(
+            snapshot.SessionId,
+            snapshot.Revision));
+    }
+
+    private void RouteRemovedSession(string townId)
+    {
+        if (!ModInformation.IsClient)
```
ShoT-UPfps:
```suggestion
        if (ModInformation.IsServer)
```

source/GameInterface/Services/Tournaments/UI/TournamentUIController.cs
```diff
@@ -0,0 +1,440 @@
...
+        return isCurrentTown &&
+               (menuId == CoopTournamentCampaignBehavior.PreparationMenuId ||
+                menuId == CoopTournamentCampaignBehavior.ActiveMenuId)
+            ? MenuRoute.TownCenter
+            : MenuRoute.Refresh;
+    }
+
+    private void RouteMenu(TournamentSessionSnapshot snapshot)
+    {
+        if (!ModInformation.IsClient)
```
ShoT-UPfps:
```suggestion
        if (ModInformation.IsServer)
```

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -0,0 +1,1557 @@
...
+            }
+        }
+        if (!matchEnded) return;
+
+        TournamentMatchResultData result;
+        if (pendingResult == null)
+        {
+            TournamentMatchData matchData = snapshot.Rounds
+                .SelectMany(round => round.Matches)
+                .Single(match => match.MatchId == snapshot.CurrentMatchId);
```
ShoT-UPfps: if this throws whats catching it?

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -0,0 +1,1557 @@
...
+            .Where(entry => aliveSlots.Contains(entry.contestant.SlotId))
+            .Select(entry => tournamentBehavior._participants[entry.index])
+            .ToList();
+    }
+
+    private System.Collections.Generic.List<TournamentTeam> ResolveAliveTeams()
+    {
+        TournamentMatchData match = snapshot.Rounds
+            .SelectMany(round => round.Matches)
+            .First(data => data.MatchId == snapshot.CurrentMatchId);
```
ShoT-UPfps: same as the .Single above, .First throws if the match isnt found.

ShoT-UPfps:
can we swap all the `new T[0]` for `Array.Empty<T>()`? theres ~44 across the tournament files

source/GameInterface/Services/Tournaments/TournamentLifecycleCoordinator.cs
zzzzzzzzott: Sorry this was an older file I had for the first attempt, its logic is handled in `TournamentSessionHandler` and `TournamentSessionRegistry` now

source/GameInterface/Services/Tournaments/TournamentSessionRegistry.cs
zzzzzzzzott: Yea used in line 312 and 739 as of writing this

source/Missions/Tournaments/TournamentSpawnIdentityResolver.cs
```diff
@@ -0,0 +1,31 @@
+using GameInterface.Services.Tournaments.Data;
+using System.Linq;
+
+namespace Missions.Tournaments;
+
+public static class TournamentSpawnIdentityResolver
+{
+    public static bool TryResolve(
```
ShoT-UPfps: this used anywhere outside tests?

source/GameInterface/Services/Tournaments/TournamentSimulationInterface.cs
```diff
@@ -0,0 +1,265 @@
...
+using TaleWorlds.CampaignSystem.Settlements;
+using TaleWorlds.CampaignSystem.TournamentGames;
+using TaleWorlds.CampaignSystem.CharacterDevelopment;
+using TaleWorlds.Core;
+
+namespace GameInterface.Services.Tournaments;
+
+public sealed partial class TournamentGameInterface
+{
+    public bool TrySimulateCurrentMatch(
```
ShoT-UPfps: this used anywhere?

source/GameInterface/Services/Tournaments/TournamentGameInterface.cs
```diff
@@ -0,0 +1,633 @@
...
+        var prizeGame = ObjectHelper.SkipConstructor<FrozenRosterFightTournamentGame>();
+        prizeGame.Town = town;
+        prizeGame.Mode = tournamentGame.Mode;
+        prizeGame.Prize = null;
+        prizeGame._lastRecordedLordCountForTournamentPrize = -1;
+        prizeGame.FrozenCharacters = frozenCharacters;
+        return prizeGame.GetTournamentPrize(false, -1);
+    }
+
+    private static int CountLords(IEnumerable<CharacterObject> characters)
```
ShoT-UPfps: still used?

source/GameInterface/Services/Tournaments/TournamentBettingMath.cs
```diff
@@ -0,0 +1,48 @@
...
+        if (winFactor <= 0f)
```
ShoT-UPfps: how coudl this ever be <=0?

source/GameInterface/Services/Tournaments/Data/TournamentSessionData.cs
```diff
@@ -0,0 +1,295 @@
...
+
+    public TournamentPlayerChoiceData(string controllerId, TournamentPlayerChoice choice)
+    {
+        ControllerId = controllerId;
+        Choice = choice;
+    }
+}
+
+[ProtoContract(SkipConstructor = true)]
+public sealed partial class TournamentContestantData
```
ShoT-UPfps: why `partial`?

ShoT-UPfps (changes requested):

source/GameInterface/Services/Tournaments/TournamentSimulationInterface.cs
zzzzzzzzott: No, removed

source/GameInterface/Services/Tournaments/Handlers/TournamentHitProgressionHandler.cs
zzzzzzzzott: ooo good catch, just fixed it

source/GameInterface/Services/Tournaments/TournamentGameInterface.cs
zzzzzzzzott: Gone

source/GameInterface/Services/Tournaments/TournamentBettingMath.cs
zzzzzzzzott: Gone already

source/GameInterface/Services/Tournaments/Data/TournamentSessionData.cs
zzzzzzzzott: Removed from here and TournamentMatchData

source/GameInterface/Services/Tournaments/TournamentBettingMath.cs
ShoT-UPfps: i think this still needs to multiply the two factors, not divide them? vanilla does `(ownPower / currentMatchPower) * (heroPower / adjustedRoundPower)` before calculating the odd

source/GameInterface/Services/Tournaments/TournamentSaveDeferral.cs
```diff
@@ -0,0 +1,61 @@
+using System.Collections.Generic;
+using TaleWorlds.CampaignSystem;
+
+namespace GameInterface.Services.Tournaments;
+
+public interface ITournamentSaveDeferral : IGameAbstraction
```
garrettluskey: Why do we have to defer saves?

source/Missions/Tournaments/Messages/NetworkApplyTournamentDamage.cs
```diff
@@ -0,0 +1,51 @@
+using Common.Messaging;
+using ProtoBuf;
+using System;
+using TaleWorlds.MountAndBlade;
+
+namespace Missions.Tournaments.Messages;
+
+[ProtoContract(SkipConstructor = true)]
+public sealed class NetworkApplyTournamentDamage : IEvent
```
garrettluskey, line 9: Why not use the existing damage in Missions?

source/Missions/Tournaments/Patches/TournamentBlowInterceptPatch.cs
```diff
@@ -0,0 +1,141 @@
...
+using Missions.Tournaments.Spectators;
+using SandBox.Tournaments.MissionLogics;
+using System.Linq;
+using TaleWorlds.MountAndBlade;
+
+namespace Missions.Tournaments.Patches;
+
+internal static class TournamentCombatPatchInstaller
+{
+    private const string HarmonyId = "Coop.TournamentCombat";
```
garrettluskey: Why do we need a standalone harmony instance here?

source/GameInterface/Services/Tournaments/Data/TournamentMissionData.cs
```diff
@@ -0,0 +1,270 @@
+using ProtoBuf;
+using System;
+
+namespace GameInterface.Services.Tournaments.Data;
+
+[ProtoContract(SkipConstructor = true)]
+public sealed class TournamentEquipmentElementData
```
garrettluskey: I think we already have a serializer for EquipmentElementData

source/GameInterface/Services/Tournaments/Data/TournamentMissionData.cs
```diff
@@ -0,0 +1,270 @@
...
+    [ProtoMember(10)]
+    public readonly float PositionX;
+    [ProtoMember(11)]
+    public readonly float PositionY;
+    [ProtoMember(12)]
+    public readonly float PositionZ;
+    [ProtoMember(13)]
+    public readonly float DirectionX;
+    [ProtoMember(14)]
+    public readonly float DirectionY;
```
garrettluskey: Vec2 and Vec3 are serializable

source/GameInterface/Services/Tournaments/Data/TournamentMissionData.cs
garrettluskey, line 1: Can we split these into separate files?

garrettluskey (changes requested):

source/GameInterface/Services/Tournaments/TournamentSaveDeferral.cs
zzzzzzzzott: Mb this was was from an attempt to make player disconnections safer but it was not working I need to remove

source/Missions/Tournaments/Patches/TournamentBlowInterceptPatch.cs
zzzzzzzzott: didnt

source/GameInterface/Services/Tournaments/Data/TournamentMissionData.cs
zzzzzzzzott: Changed to use  EquipmentElementSurrogate

source/Missions/Tournaments/Messages/NetworkApplyTournamentDamage.cs
zzzzzzzzott: This damage applier applys damage immedietly on the local client. The goal was to make combat feel a little more response in networked arena combat where details are easier to see.

So instead of waiting for the puppets owner to apply the damage and relay the puppets death message, the local client applies the damage immedietly (and kills the puppet if needed), then broadcasts that out to other clients. This makes the death feel more responsive

garrettluskey (approved):

source/GameInterface/Services/Tournaments/TournamentSessionRegistry.cs
```diff
@@ -0,0 +1,1051 @@
...
+                .FirstOrDefault(state => state.MissionInstanceId == missionInstanceId);
+            if (session == null || !session.IsActive) return false;
+            return session.TryGetActiveContestant(controllerId, out _) ||
+                session.PendingSpectators.Contains(controllerId) ||
+                session.Spectators.Contains(controllerId) ||
+                session.Entrants.Contains(controllerId);
+        }
+    }
+
+    public bool HasActiveSessions
```
ShoT-UPfps: is this used? are we supposed to be using this for loading tournaments from a save?

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -0,0 +1,1597 @@
...
+    }
+
+    private void WakeTransferredAgent(Agent agent)
+    {
+        if (agent == null || agent.IsMount || !agent.IsActive() || !session.IsLocalHost) return;
+
+        coopMissionComponent.AgentMovementHandler.Interpolator.Forget(agent);
+        if (agent.MountAgent != null)
+            coopMissionComponent.AgentMovementHandler.Interpolator.Forget(agent.MountAgent);
+        agent.Controller = AgentControllerType.AI;
```
ShoT-UPfps, line 1621: mount needs an AI controller too or?

UIMovies/CoopTournamentScreen.xml
```diff
@@ -0,0 +1,442 @@
...
+                                HorizontalAlignment="Center" Sprite="SPGeneral\SPOptions\standart_slider_frame" />
+                        <ImageWidget Id="SliderHandle" DoNotAcceptEvents="true" WidthSizePolicy="Fixed" HeightSizePolicy="Fixed"
+                                     SuggestedWidth="14" SuggestedHeight="38" HorizontalAlignment="Left"
+                                     Brush="SPOptions.Slider.Handle" />
+                      </Children>
+                    </SliderWidget>
+                    <ListPanel WidthSizePolicy="CoverChildren" HeightSizePolicy="CoverChildren" HorizontalAlignment="Center"
+                               MarginTop="20" StackLayout.LayoutMethod="HorizontalLeftToRight">
+                      <Children>
+                        <ToggleButtonWidget DoNotPassEventsToChildren="true" WidthSizePolicy="Fixed" HeightSizePolicy="Fixed"
```
ShoT-UPfps: needs to set `IsBetWindowEnabled=false`?

source/GameInterface/Services/Tournaments/TournamentGameInterface.cs
```diff
@@ -0,0 +1,628 @@
...
+        string townId,
+        string sceneName,
+        string prizeId,
+        string replacementId,
+        IReadOnlyList<CharacterObject> sortedCharacters,
+        out TournamentSessionSeed seed)
+    {
+        seed = null;
+        var identity = new TournamentSessionIdentity();
+        if (!objectManager.AddNewObject(identity, out var sessionId))
```
ShoT-UPfps: do we ever clear out these tournament session identities or is it a memroy leak?>

source/GameInterface/Services/Tournaments/Handlers/TournamentLifecycleHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+        if (!sessionRegistry.Remove(snapshot.SessionId))
+        {
+            if (sessionRegistry.TryGet(snapshot.SessionId, out _))
+                return false;
+
+            RemoveAcceptedHitProgression(acceptedHitProgression, snapshot.SessionId);
+            return true;
+        }
+
+        RemoveAcceptedHitProgression(acceptedHitProgression, snapshot.SessionId);
```
ShoT-UPfps: what about `completionTransactions` and `liveCombatSessions`? need to clear those too?

source/GameInterface/Services/Tournaments/Data/TournamentAgentSpawnData.cs
```diff
@@ -0,0 +1,127 @@
...
+    public readonly string TeamBannerCode;
+    [ProtoMember(8)]
+    public readonly string ControllerId;
+    [ProtoMember(9)]
+    public readonly EquipmentElement[] Equipment;
+    [ProtoMember(10)]
+    public readonly Vec3 Position;
+    [ProtoMember(11)]
+    public readonly Vec2 Direction;
+    [ProtoMember(15)]
```
ShoT-UPfps: this should start from 12 not 15

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -0,0 +1,1597 @@
...
+            blow.WeaponRecord._isMissile = false;
+            blow.WeaponRecord.AffectorWeaponSlotOrMissileIndex = -1;
+        }
+
+        applyingDamage.Add(message.VictimAgentId);
+        NetworkApplyTournamentDamage previousDamageMessage = activeDamageMessage;
+        activeDamageMessage = message;
+        try
+        {
+            victim.RegisterBlow(blow, in collisionData);
```
ShoT-UPfps: can we use `RegisterBlowPatch.RunOriginalRegisterBlow` here? `RegisterBlowPatch` returns false for puppet victims otherwise, so only the victim owner applies the damage and the host/other clients keep stale health.

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -0,0 +1,1597 @@
...
+            ApplyTournamentDamage(message);
+        }
+        return false;
+    }
+
+    private void Handle_ApplyTournamentDamage(MessagePayload<NetworkApplyTournamentDamage> payload)
+    {
+        NetworkApplyTournamentDamage message = payload.What;
+        if (snapshot == null || message.SessionId != snapshot.SessionId) return;
+        if (message.MatchId != snapshot.CurrentMatchId || message.Revision > snapshot.Revision) return;
```
ShoT-UPfps: these P2P messages can beat the campaign snapshot or manifest. can we buffer future-revision and pre-manifest damage/runtime/knockout packets and drain them after `ApplySessionUpdate` and `ApplyPendingManifest`? `TryAccept` also needs to happen after the required agents resolve, otherwise the packet consumes its sequence and cant be retried.

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -0,0 +1,1597 @@
...
+            "[Tournament] Applying remote knockout in {MatchId}: agent={AgentId}, authority={Authority}, host={IsHost}",
+            snapshot.CurrentMatchId,
+            message.AgentId,
+            message.OriginControllerId,
+            session.IsLocalHost);
+
+        applyingRuntimeState = true;
+        try
+        {
+            if (agent != null && agent.IsActive()) agent.FadeOut(false, true);
```
ShoT-UPfps, line 669: damage comes from the attacker peer but knockout comes from the victim peer. if the elected host gets the knockout first, `ApplyKnockout` removes the victim from the registry, then the damage returns at `TryGetAgentInfo` and `CaptureHitProgression` never records the finishing hit. i think knockout needs to carry the causing damage origin/sequence so the host can apply that damage before cleanup.

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -0,0 +1,1597 @@
...
+                pair.Value.StopPhysicsAndSetFrameForClient(frame, null);
+                continue;
+            }
+
+            RemoveRuntimeWorldItem(pair.Key, pair.Value);
+        }
+
+        foreach (var pair in canonical)
+        {
+            if (registry.TryGet(pair.Key, out _)) continue;
```
ShoT-UPfps, line 1460: `WeaponDropHandler` already calls `DropItem` on every peer, but those local `SpawnedItemEntity`s arent in `worldItemRegistry`. when the host runtime state arrives this lookup misses and `SpawnRuntimeWorldItem` creates another copy, leaving the original untracked. i think the drop message needs a world item id so every peer registers the item it already dropped.

source/GameInterface/Services/Tournaments/TournamentSessionRegistry.cs
zzzzzzzzott: No removed along with some other non used things i found

source/Missions/Tournaments/CoopTournamentController.cs
zzzzzzzzott: added

UIMovies/CoopTournamentScreen.xml
zzzzzzzzott: native code does this alr

ShoT-UPfps:
i see thumbsup etc to my comments but doesnt seem like any(?) of them are addressed

UIMovies/CoopTournamentScreen.xml
ShoT-UPfps: where?

zzzzzzzzott:
> i see thumbsup etc to my comments but doesnt seem like any(?) of them are addressed

Just pushed, was testing some stuff before I pushed lol

UIMovies/CoopTournamentScreen.xml
zzzzzzzzott: `WidgetToClose` on the cancel button already sets `IsBetWindowEnabled` to false through the binding. Same as native

source/GameInterface/Services/Tournaments/UI/CoopTournamentCampaignBehavior.cs
```diff
@@ -0,0 +1,293 @@
...
+
+    internal const string TownCenterMenuId = "town";
+    internal const string TownArenaMenuId = "town_arena";
+
+    public override void RegisterEvents()
+    {
+        CampaignEvents.OnSessionLaunchedEvent.AddNonSerializedListener(this, AddGameMenus);
+    }
+
+    public override void SyncData(IDataStore dataStore)
```
ShoT-UPfps, line 26: removing `HasActiveSessions` doesnt address the save issue. `SyncData` is empty, so shutting down during an active tournament saves the native campaign without our bracket or accepted bet state.

source/Missions/Agents/Handlers/WeaponDropHandler.cs
```diff
@@ -59,29 +63,36 @@ private void WeaponDropSend(MessagePayload<WeaponDropped> obj)
...
+                Logger.Error($"Tried to drop a weapon from an empty slot ({obj.What.EquipmentIndex})");
+                return;
+            }
 
-        GameThread.Run(() =>
-        {
+            HashSet<SpawnedItemEntity> existingItems = WeaponDropItemTracker.Capture();
             agent.DropItem(obj.What.EquipmentIndex);
+            SpawnedItemEntity droppedItem = WeaponDropItemTracker.FindDroppedItem(existingItems);
+            worldItemRegistry.Register(obj.What.WorldItemId, droppedItem);
```
ShoT-UPfps, line 95: this still duplicates the item if runtime state arrives before the direct drop. `SpawnRuntimeWorldItem` creates the item first, then this drops another one and `Register` only replaces the registry mapping without removing the first scene entity.

source/GameInterface/Services/Tournaments/Handlers/TournamentBettingHandler.cs
```diff
@@ -0,0 +1,163 @@
...
+    private void ProcessBetRequest(NetPeer peer, Player player, NetworkRequestTournamentBet request)
+    {
+        if (!TryGetBetSession(peer, request, out var snapshot))
+            return;
+
+        BetLedgerEntry ledger = GetOrCreateBetLedger(request.SessionId, player.ControllerId);
+        ledger.MatchAmounts.TryGetValue(request.MatchId, out var currentBet);
+        if (TryReplayBetResponse(peer, request, ledger))
+            return;
+        if (!TryResolvePlayer(player, out var hero, out _) ||
```
ShoT-UPfps, line 20: we still need the `IsBettingOpen` checks here. otherwise a stale request can arrive after the snapshot advances to `LiveMatch` or a different match and still deduct gold.

source/GameInterface/Services/Tournaments/UI/CoopTournamentCampaignBehavior.cs
zzzzzzzzott: I first planned on accounting for saving in tournaments which is why stuff like this is here, I do want to expand on it in the near future. But I decided to push tournament saving related logic to outside of this PR. 

Native bannerlord doesnt save anything while in a tournament (best you can do is save before starting a tournament in vanilla, and if the game closed you would resume before the tournament started).

Obviously that can be more complex for our case. But for now I wanted it to not save the tournament state at all in the event that the campaign saves while players are in a tournament. This way, if a client/server crashes while players are in a tournament, when they resume, they would be able to just start the tournament over. 

I'm sure there could be some bugs here that I don't foresee but you see what I mean?

source/GameInterface/Services/Tournaments/Handlers/TournamentBettingHandler.cs
zzzzzzzzott: But do we actually care though? Like if a bet request lags for whatever reason and arrives after the round started, we can still treat it as valid?

The only time this is an issue is if the bet message was so laggy it arrived AFTER the match ended, in which case it would apply itself to the next match lmao. And at that point the players internet mustve shit itself.

source/Missions/Agents/Handlers/WeaponDropHandler.cs
zzzzzzzzott: Maybe im misunderstanding this, but the ordering doesnt duplicate the item. The runtime update already removed the weapon, so when the drop message arrives later the slot is empty and it does nothing.

I had codex make a test to prove this and it looks good to me lol - `RuntimeWorldItemBeforeDirectDrop_SkipsSecondDropAndPreservesRuntimeItem`

ShoT-UPfps (approved):
lgtm, thanks for the fixes

PR #1849, opened 2026-07-11
[1844] Sync ranged weapon projectiles across clients

### LINKED ISSUE #1844: Projectiles shot by weapons not visible by nonhost client
During field battles/sieges/etc the projectiles shot by weapons like arrows are mostly (maybe all) missing on the nonhost client. It needs to be synced

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Ranged weapon projectiles fired by other players and AI troops are mostly or entirely invisible to a nonhost client during field battles and sieges. Archers play their firing animations, but arrows and similar projectiles do not appear.

## What is the new behavior?

Resolves #1844

Ranged weapon projectiles fired by players and AI troops now appear on nonhost clients and remain visible through impact. The same replacement path runs in tracked co-op location missions, while standalone tournaments, arenas, and other vanilla missions retain their native projectiles. Siege engines still produce one projectile per launch, without duplicate stones or bolts.


source/Missions/MissionModule.cs
```diff
@@ -17,6 +20,14 @@ namespace Missions;
 /// </summary>
 public class MissionModule : Module
 {
+    internal const string MissilePatchCategory = "CoopMissilePatches";
+
+    static MissionModule()
+    {
+        new Harmony(GameInterfaceModule.HarmonyId + ".Missiles")
```
garrettluskey: Why not use the DI harmony?

garrettluskey (dismissed):

source/Missions/MissionModule.cs
ShoT-UPfps: updated, lmk if you were thinking of something else https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/1849/changes/8db74cec201ae65e1db67d63cfa5eaca73f9daa5

source/Missions/MissionModule.cs
```diff
@@ -17,10 +19,16 @@ namespace Missions;
 /// </summary>
 public class MissionModule : Module
 {
+    internal const string MissilePatchCategory = "CoopMissilePatches";
```
garrettluskey, line 22: Can we just patch this without the category?

garrettluskey (dismissed):

source/Missions/MissionModule.cs
ShoT-UPfps: It seems not because these changes reuse stuff that's in the `Missions` module which is currently not being scanned for patches and if we start uncategorized patching for the `Missions` module then we'll incidentally activate a bunch of other patches that are currently not being applied.

ShoT-UPfps: One option would be to move these changes into `GameInterface` as well as move all the `Mission` types that these changes use into `GameInterface`

garrettluskey (approved):

PR #1894, opened 2026-07-12
[1755] Add standalone Steam server hosting

ShoT-UPfps:
Server  can be joined and discovered even without steam invitess

Able to set visibility when hosting server
image

Able to set a password on server
image

Client now has a server lobbies button and list
image

This is what happens if a server has a different version
image


ShoT-UPfps:
All codefactors are complex methods

source/Coop.Core/Client/Services/Session/ConfiguredSessionJoinInfoSource.cs
```diff
@@ -25,6 +25,8 @@ public SessionJoinInfo Get()
         {
             Address = advertisementConfig.PublicAddress,
             Port = networkConfig.Port,
+            ModVersion = global::Common.ModInformation.BuildVersion,
```
garrettluskey:
```suggestion
            ModVersion = ModInformation.BuildVersion,
```

source/Directory.Build.props
garrettluskey, line 1: Is this required?

garrettluskey (dismissed):

source/Coop.Core/Client/Services/Session/ConfiguredSessionJoinInfoSource.cs
ShoT-UPfps: updated

source/Directory.Build.props
ShoT-UPfps: Yes, for the current build setup. There is not another shared Coop version source: `SubModule.xml` is generated from `$(ModVersion)`, while the existing assembly versions are unrelated legacy values. `CoopVersion` is the single source used for both the numeric Bannerlord module version and the exact `Common.dll` build version used by Steam compatibility checks. The exact file could be replaced by another shared MSBuild props file, but removing the shared property would duplicate the version and allow the two outputs to drift.

garrettluskey (approved):

PR #1911, opened 2026-07-13
Fix flaky tests

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Tests that change the process-wide server/client role use separate collections, so they can run at the same time and overwrite the role another test is using.

## What is the new behavior?

All role-sensitive tests share one nonparallel collection, so one test cannot change the server/client role while another test is using it.

### DIFF
--- a/source/GameInterface.Tests/Services/Actions/ChangeGovernorActionPatchesTests.cs
+++ b/source/GameInterface.Tests/Services/Actions/ChangeGovernorActionPatchesTests.cs
@@ -14,6 +14,7 @@ namespace GameInterface.Tests.Services.Actions
     /// <summary>
     /// Tests for the governor removal prefix in <see cref="ChangeGovernorActionPatches"/>.
     /// </summary>
+    [Collection(global::GameInterface.Tests.ModInformationRoleCollection.Name)]
     public class ChangeGovernorActionPatchesTests
     {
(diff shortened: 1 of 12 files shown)


source/GameInterface.Tests/Services/Actions/ChangeGovernorActionPatchesTests.cs
```diff
@@ -14,6 +14,7 @@ namespace GameInterface.Tests.Services.Actions
     /// <summary>
     /// Tests for the governor removal prefix in <see cref="ChangeGovernorActionPatches"/>.
     /// </summary>
+    [Collection(global::GameInterface.Tests.ModInformationRoleCollection.Name)]
```
garrettluskey: add a using for the namespace

Same with all below

```suggestion
    [Collection(ModInformationRoleCollection.Name)]
```

garrettluskey (dismissed):

source/GameInterface.Tests/Services/Actions/ChangeGovernorActionPatchesTests.cs
ShoT-UPfps: updated

garrettluskey (approved):

PR #1923, opened 2026-07-13
[1879] Close zero-enemy map events after victory

### LINKED ISSUE #1879: [ISSUE] Client soft locked on a map event with zero looters
See https://discord.com/channels/709516043332354119/1525642905790845018/1525642905790845018
Image

The user sees themselves as being a part of a mapevent against enemy looters with an allied AI noble on the player side. The enemy loots show as having zero troops but the mapevent doesn't close down. The player only has the "Capture the Enemy" option on their menu which doesn't work.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Winning a field battle can leave the player who exits stuck in the encounter after the enemy reaches zero troops if the elected battle host remains in the mission. The only remaining option is `Capture the Enemy`, which does not close the battle.

## What is the new behavior?

Resolves #1879

After a shared victory, each player can leave the field battle when they choose. The zero-enemy encounter closes when the first player leaves, while players who remain in the mission are not forced out.

### DIFF
--- a/source/E2E.Tests/Services/MapEvents/CoopBattleFinalizeTests.cs
+++ b/source/E2E.Tests/Services/MapEvents/CoopBattleFinalizeTests.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using Common.Messaging;
+using Common.Network;
 using Common.Util;
 using E2E.Tests.Environment.Instance;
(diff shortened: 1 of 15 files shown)


ShoT-UPfps:
Closing; unable to repro. Assumed fixed for now.

PR #1977, opened 2026-07-14
Toggle unit markers option

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Not able to toggle unit markers, making "surprise" attacks hard against players.

## What is the new behavior?

Resolves #


ShoT-UPfps:
If we're adding server-wide configurations now, can we make it settable only by the actual server instance?

ShoT-UPfps (dismissed):
design disagreement

EgardA:
Changed it to a server-only debug command for now, until we decide on how we want to setup server side configs

source/GameInterface/Services/UI/Commands/TacticalUnitSymbolsDebugCommand.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        if (!CommandHelpers.IsServerOnlyCommand(out var error, CommandName)) return error;
+        if (args.Count != 1) return Usage;
+
+        switch (args[0].ToLowerInvariant())
+        {
+            case "on":
+            case "true":
+            case "1":
+                return Apply(true);
```
ShoT-UPfps: this feels weird, tactical symbols on makes it sounds like they would be shown. but Apply(true) hides them?

EgardA: I agree, are you able to fix this? As I'm away until Sunday

ShoT-UPfps: done

garrettluskey (approved):

PR #2036, opened 2026-07-15
Battle Hardening

### LINKED ISSUE #1630: [BUG] Reserves not available to host when a player disconnects
# For New Contributors

## Description

When a player disconnects their reserve troops will never spawn.

Update the implementation so reserve troops can spawn by the host

## Location

## Related Issues

need to check if https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/1641 fixes this at all

## Definition of Done

### LINKED ISSUE #1838: Sieges - Clients can retain a stale battle host after the host leaves with no successor
## Description

When the battle host leaves and no successor remains in the mission, clients can keep the departed player as the host assignment.

This can affect a later assault of the same siege.

## Steps to reproduce

1. Start a siege assault with one player in the mission.
2. Let that player become the battle host.
3. Have the player leave the mission while there are no successors.
4. Start another assault in the same siege.
5. Inspect the host assignment on the clients.

## Expected behavior

All clients clear the host assignment when the host leaves with no successor.

## Actual behavior

Clients can retain the departed host assignment between assaults.

### LINKED ISSUE #1989: Spawn missing army reserves when the battle host crashes before catch-up
## What happened

I joined a field battle between armies. The battle host crashed before my client received its join handshake or agent replay. My client was promoted to host and received the full army reserves, but the mission stayed populated with only my own party.

## Reproduction

1. Client A enters a field battle between armies and becomes battle host.
2. Client B joins before the army agents finish catching up from A.
3. Crash client A.
4. Client B is promoted to host.

Client B loads or remains in a battle with only its own party. The allied and enemy army parties never spawn.

## Log evidence

From Coop_client.log for client B:

```text
22:42:41 Attacker: SetReserve 0 parties / 0 troops
22:42:41 Defender: SetReserve 1 parties / 158 troops
22:42:42 Coop spawn sized on start: Defender=158, Attacker=0
22:43:15 Became host of MapEvent_Created_3037 via migration
22:43:15 Attacker: SetReserve 14 parties / 1336 troops
22:43:15 Defender: SetReserve 12 parties / 996 troops
22:45:42 Replayed 158 of our agent(s) to a joining client
```

There was no Received join info entry from the crashed host for this map event, so there were no host-owned puppets for BattleAuthorityMigrator to adopt.

## Expected

When the successor receives its expanded reserves after host migration, it should field any remaining parties that have no live agents. CoopBattleMissionSpawnHandler currently ignores reserve changes after _sized is set, while the migration path assumes existing agents can be adopted.

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #1630
Resolves #1989
Resolves #1838


source/Missions/Battles/BattleHostHandler.cs
```diff
@@ -164,25 +235,96 @@ private void Handle_NetworkRequestBattleReserves(MessagePayload<NetworkRequestBa
...
+        if (!hostPeers.TryGetValue(mapEventId, out var host) || host.ControllerId != assignment.HostControllerId)
+        {
+            // A promoted host that has not re-requested yet has no live peer recorded — and needs no refresh:
+            // its own upcoming adoption re-request is already served the post-return scope.
+            Logger.Information("[BattleHost] No live peer recorded for host {Host} of {MapEventId}; skipping reserve refresh (its next request is already re-scoped)",
+                assignment.HostControllerId, mapEventId);
+            return;
+        }
+
+        SendOwnedReserves(mapEventId, mapEvent, host.Peer, host.ControllerId, includeEmptySides: true);
```
ShoT-UPfps: we need to flush the host's supplied pointer before replacing its reserve here. otherwise it can spawn a troop after its last 1s progress report, lose that party on this refresh, and the returner gets the stale pointer and spawns the same descriptor again

source/Missions/Battles/BattleHostHandler.cs
```diff
@@ -219,29 +398,35 @@ private void Handle_MissionMemberDeparted(MessagePayload<MissionMemberDeparted>
...
-                    // The battle is fully abandoned (no one left to continue it). Forget the WHOLE map event's
-                    // reserves — not just the leaver's own party (ForgetController above) — so restarting the
-                    // SAME map event re-flattens the AI/enemy parties the host had been fielding. Otherwise their
-                    // supplied pointers stay at the end and only the host's own re-flattened party re-spawns.
+                    // Fully abandoned even though the membership signal did not mark the instance empty (e.g.
+                    // a member still on the loading screen was never mission-ready): forget the WHOLE map
+                    // event's reserves for the same reason as the empty-instance branch above.
                     if (objectManager.TryGetObject<MapEvent>(mapEventId, out var abandonedEvent))
                         reserveBuilder.ForgetMapEvent(abandonedEvent);
+                    ClearBattleRecords(mapEventId);
```
ShoT-UPfps: we can't clear the battle records here while `IsInstanceEmpty` is false. a player can still be loading and absent from `successors`, so if the ready host disconnects this drops the absent marker and the eventual host never gets the disconnected host's party

source/Missions/Battles/BattleHostHandler.cs
```diff
@@ -200,14 +342,51 @@ private void Handle_MissionMemberDeparted(MessagePayload<MissionMemberDeparted>
...
+            // the reserve (the host adopts the troops and reinforces them from the existing pointer) — mark
+            // the dropped member ABSENT so its parties resolve into the HOST's reserve scope until it returns.
+            if (payload.What.WasRetreat)
+            {
+                if (objectManager.TryGetObject<MapEvent>(mapEventId, out var mapEvent))
+                    reserveBuilder.ForgetController(mapEvent, controllerId);
+            }
+            else
+            {
+                MarkAbsent(mapEventId, controllerId);
```
ShoT-UPfps, line 657: this needs to mark the controller absent even when there isn't a host assignment yet. otherwise a player who disconnects while everyone is still loading returns above, and the eventual host never gets that still-registered player's reserves

source/Missions/Battles/HostEpochPolicy.cs
```diff
@@ -0,0 +1,33 @@
...
+/// </list>
+/// </para>
+/// </summary>
+public static class HostEpochPolicy
+{
+    public static bool IsStale(int messageEpoch, int localEpoch)
+    {
+        return messageEpoch != 0
+            && localEpoch != 0
+            && messageEpoch < localEpoch;
```
ShoT-UPfps: this also needs to track the highest message epoch we've already accepted. otherwise a client still on epoch 1 can accept epoch 3, then accept a delayed epoch 2 from another peer because both are greater than its local assignment, and apply the older siege state last

source/GameInterface/Services/PlayerCaptivityService/Handlers/PlayerCaptivityServerHandler.cs
```diff
@@ -122,20 +124,104 @@ private void Handle_PrisonerTaken(MessagePayload<PrisonerTaken> payload)
...
+        for (int i = 0; i < memberRoster.Count; i++)
+        {
+            var element = memberRoster.GetElementCopyAtIndex(i);
+            if (element.Character?.IsHero != true) continue;
+
+            // A depleted hero element (e.g. the captured leader's leftover) is not a live member.
+            if (element.Number <= 0) continue;
+
+            var companion = element.Character.HeroObject;
+            if (companion == null || !companion.IsAlive || companion.IsPrisoner) continue;
```
ShoT-UPfps: this needs to check `DeathMark` before capturing. battle deaths only set `DeathMark.DiedInBattle` while the map event is active, so `IsAlive` is still true here and we turn the dead companion into a prisoner

source/Missions/Battles/BattleAuthorityMigrator.cs
```diff
@@ -288,12 +303,44 @@ private void Handle_BattleHostMigrated(MessagePayload<BattleHostMigrated> payloa
...
     }
 
+    // [New host] Adopt the agents of every controller that holds registry entries here but is no longer in
+    // the mission — orphans left behind by earlier holder-local adoptions of the host line that just
+    // departed. The retreat record is consumed per swept controller for the same reason as the migration
+    // adoption above: a host that retreated earlier in the chain had its own-party troops despawned locally,
+    // and only the AI it ran should be (re-)adopted.
+    private void SweepAgentsOfAbsentControllers(string previousHost)
+    {
+        var present = new HashSet<string>(missionContext.ControllersInMission);
```
ShoT-UPfps, line 346: we need to clear the membership when our local battle mission exits. BR-054 lets a retreated player re-enter the persisting instance, but `BattleInstanceLifecycle.Leave` leaves `MissionContext` populated, so controllers who drop while we're away still look present and this sweep skips their orphaned agents

source/GameInterface/Services/MapEvents/BattleDeploymentTimer.cs
```diff
@@ -0,0 +1,92 @@
...
+    public bool Tick(float dt)
+    {
+        if (!running) return false;         // before mission-ready, after a finish, or after the one firing
+        if (limitSeconds <= 0f) return false; // the configured limit disables the gate
+
+        elapsedSeconds += dt;
+        if (elapsedSeconds < limitSeconds) return false;
+
+        // Fire exactly once: the caller auto-finishes the deployment; from then on the gate stays quiet.
+        disarmed = true;
```
ShoT-UPfps, line 141: dont disarm this until the auto-finish succeeds. if a 5s limit expires while reserves are still inside the 15s hold, `TeamSetupOver` is false and `FinishNativeDeployment` returns, then this never fires again once setup completes

source/Missions/Battles/HostEpochPolicy.cs
```diff
@@ -0,0 +1,33 @@
...
+/// this is the documented self-healing convergence window.</item>
+/// <item><c>messageEpoch == localEpoch</c> — the current hosting generation. Rejecting it would
+/// drop the live host's traffic.</item>
+/// <item><c>messageEpoch &gt; localEpoch</c> — the sender heard about a migration before this
+/// receiver did (server broadcasts race the P2P mesh). Dropping would silence the NEW host until
+/// the assignment broadcast lands here; accept and let the assignment catch up.</item>
+/// </list>
+/// </para>
+/// </summary>
+public static class HostEpochPolicy
```
ShoT-UPfps: please make this an injected transient instead of a static helper. add an interface and `InstancePerDependency` registration, inject it into `CoopBattleController`, and pass it to both siege replicators

doc/BattleRequirementsTestPlan.md
```diff
@@ -0,0 +1,426 @@
...
+| **BR-025** | Deployment Time Limit | **No** — No per-player deployment time limit exists. | **None** | unit+manual | L | — | First implement a per-player deployment timer that starts when the client becomes mission-ready and, on expiry, invokes the native FinishDeployment (leaving troops at current positions -> reveal per BR-023 and count as a deployment finish per BR-024), with the duration read from a game-configuration value. Then unit-test the pure timer/expiry decision (BattleDeploymentActivator/Coordinator style), and add a manual/E2E test that an expired player's deployment auto-finishes (broadcasts own troops at current positions + drives activation).<br>**Missing:** Blocked by implementation of the deployment-time-limit feature; needs a mission-ready timer, a configurable duration value, and (for full verification) a live client to observe the native FinishDeployment on expiry.<br>*Note:* Confirmed NOT implemented (matches project-history hint). Depends on BR-013 mission-ready observation (timer start) and on BR-023/BR-024 (expiry must trigger the visibility reveal and count as a finish for NPC activation). |
```
ShoT-UPfps: are these docs supposed to be committed? this matrix needs to be updated from the alignment results above before merging. BR-025, BR-033/034, and BR-102 are still marked unimplemented, and BR-101 still says election happens at entry

ShoT-UPfps (changes requested):
please address the comments above

doc/BattleRequirementsTestPlan.md
garrettluskey: Removed

source/Missions/Battles/BattleHostHandler.cs
garrettluskey: Fixed in 4cea756d3. The shrink refresh now carries a FlushRequested flag; the host acks with an IsFlush supply-progress message whose dropped-party pointers are captured atomically inside the SetReserve REPLACE lock, and the server defers the returner's grant until the acks land in the ledger (per-side ack counting, FIFO per battle on the host's reliable stream) or a 2.5s tick-driven deadline. Fallbacks cover no-host-peer, host departure mid-handshake, lost acks, and duplicate returns — the returner is never stranded, and unflagged shrinks behave exactly as before.

source/Missions/Battles/BattleHostHandler.cs
garrettluskey: Fixed in 1cb9b5ccf — the no-successor host departure no longer clears the battle records: it removes the hostless assignment and re-flattens reserves but preserves the absent markers, so a still-loading player's eventual election inherits the departed host's party. ForgetMapEvent is kept deliberately: with no ready successor, none of the departed host's on-field troops persist anywhere, so the eventual host must re-spawn the casualty-reduced roster from a reset pointer.

source/Missions/Battles/BattleHostHandler.cs
garrettluskey: Fixed in 1cb9b5ccf — the drop/retreat bookkeeping now runs before the no-assignment early-return, so a player disconnecting while everyone is still loading is absent-marked and the eventual first-ready host is granted their party (a pre-election retreat now also re-flattens on rejoin).

source/Missions/Battles/HostEpochPolicy.cs
garrettluskey: Fixed in c19eef3fb — the policy now tracks the highest accepted message epoch per battle instance and rejects anything below it, so a delayed epoch-2 after an accepted epoch-3 is dropped even though both clear the local assignment. The existing rules (unstamped accepted, ahead-of-assignment accepted) are unchanged.

source/GameInterface/Services/PlayerCaptivityService/Handlers/PlayerCaptivityServerHandler.cs
garrettluskey: Fixed in 83a81e984 — companion capture now treats a pending Hero.DeathMark as dead (battle deaths keep IsAlive true while the map event is active), with a regression test for exactly that state.

source/Missions/Battles/BattleAuthorityMigrator.cs
garrettluskey: Fixed in c19eef3fb — leaving the battle mission now ends the local MissionContext membership, so after a BR-054 re-engagement the absent-controller sweep sees peers who dropped while we were away and adopts their orphaned agents.

source/GameInterface/Services/MapEvents/BattleDeploymentTimer.cs
garrettluskey: Fixed in 5b3937f26 — the timer now disarms only when the native finish actually runs: while TeamSetupOver is false it keeps firing on later ticks and auto-finishes once setup completes. A manual finish still disarms permanently, as does a mission with no deployment phase at all.

source/Missions/Battles/HostEpochPolicy.cs
garrettluskey: Fixed in c19eef3fb — IHostEpochPolicy, registered InstancePerDependency, injected into CoopBattleController and passed as one shared instance to both siege replicators, so the accepted-epoch watermark is per battle and resets with the controller's lifetime.

ShoT-UPfps (approved):
lgtm, thanks for the fixes

PR #2048, opened 2026-07-16
Replace unsafe binary package deserialization

ShoT-UPfps:
**Most of the LOC is just a rename**

Client-provided hero data is deserialized with `BinaryFormatter` on the server, then the same bytes are sent to and deserialized by other clients. A malicious payload can use arbitrary object deserialization.

## What is the new behavior?

Replace `BinaryFormatter` with a closed serializer that only accepts our `IBinaryPackage` types. Primitive values are encoded as inert strings, and serialized collection type names can only resolve to already loaded assemblies and the supported collection shapes. Unknown, malformed, and oversized package payloads are rejected.


garrettluskey (approved):
I trust this with my life, eventually we will want to replace the BinaryPackageFactory with protobuf

PR #2088, opened 2026-07-17
Fix clan banners not showing in battles

jordanbrymora:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] All new classes have class-level documentation comments, if there are any at all
- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update
- [ ] Other... Please describe:

## What is the new behavior?

Resolves #2061

PR #2100, opened 2026-07-17
Fix E2E Cross Test State Leaks

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Two E2E tests fail in CI depending on xUnit test-class order (which shifts between builds), while always passing when run solo or per-class. Both were failing on the `FixClanBannersNotShowingInBattles` PR (#2088) through no fault of that branch:

1. **`VillageHostileActionTests.ActiveSlowRaidMapEvent_StartSettlementEncounterRequestsJoinConversation`** — fails with `Sequence contains no elements`. `PlayerPartyInteractionDialogState` is a process-wide static; several `PlayerPartyInteractionFlowTests` end mid-dialog and leave `HasActiveState = true` after their environment is torn down. `ConversationRequestHandler.Handle_ConversationRequested` silently drops every conversation request while that flag is set, so the village test's `NetworkRequestConversation` never gets sent when the flow tests run earlier in the same process. This is also a real game bug: leaving a coop session mid-dialog left the flag set for the rest of the process, silently blocking all later encounter conversations.

2. **`TournamentMissionRulesTests.MountedManifest_RoundTripsExactMountIdentityAndRejectsNullAgentArrays`** — fails with `Expected: "horse-item" / Actual: null`. The class runs without a test environment, but the process-wide Harmony patches from earlier environment-based tests stay applied. `MBObjectBasePatches` gates the `MBObjectBase.StringId` **setter** on the ambient sync policy, and `new ItemObject("...")` routes through it. When a previous teardown leaks a live container into `ContainerProvider` (E2E's `TestPolicy` denies all originals), the setter is silently skipped and the item's `StringId` stays null. `E2ETestEnvironment.Dispose` could leak exactly that way: any throw during teardown skipped `ContainerProvider.Clear()`.

## What is the new behavior?

- `PlayerPartyInteractionHandler.Dispose()` clears `PlayerPartyInteractionDialogState`, so the dialog state cannot outlive its session (fixes the game-side leak) or leak across test environments (fixes the village test).
- `TournamentMissionRulesTests` builds its spawn manifest under an `AllowedThread`, making game-object construction immune to whatever container/policy state a previously-run test left behind.
- `E2ETestEnvironment.Dispose` runs `ContainerProvider.Clear()` (and the log-callback removal) in a `finally`, so a throwing disposal can no longer poison later environment-less tests.

Both mechanisms were reproduced deterministically with a forced-order harness (running the polluting test and the victim test in one process): the village test fails after a mid-dialog flow test and passes with the fix; `new ItemObject("probe").StringId` is null under a leaked deny-policy container and the hardened tournament test passes regardless.

Verification: full E2E suite 694 passed / 0 failed / 4 skipped; GameInterface.Tests 576 passed; Coop.Tests 431 passed; Coop.IntegrationTests 114 passed.

## Bot Changelog Entry

- Fixed flakey test

## Other information

Cherry-picked from the two fix commits originally pushed to `FixClanBannersNotShowingInBattles`; that branch is being reset to its own work so it no longer carries these commits.

### DIFF
--- a/source/E2E.Tests/Environment/E2ETestEnvironment.cs
+++ b/source/E2E.Tests/Environment/E2ETestEnvironment.cs
@@ -75,24 +75,33 @@ public class E2ETestEnvironment : IDisposable
             if (disposed) return;
             disposed = true;
 
-            if (AutoSyncConfiguration.Enabled)
+            try
             {
-                Server.Resolve<AutoSyncHandler>().Dispose();
+                if (AutoSyncConfiguration.Enabled)
+                {
+                    Server.Resolve<AutoSyncHandler>().Dispose();
+                    foreach (var client in Clients)
+                    {
+                        client.Resolve<AutoSyncHandler>().Dispose();
+                    }
+                }
+
+                Server.Dispose();
+
                 foreach (var client in Clients)
                 {
-                    client.Resolve<AutoSyncHandler>().Dispose();
+                    client.Dispose();
                 }
             }
-
-            Server.Dispose();
-
-            foreach (var client in Clients)
+            finally
             {
-                client.Dispose();
+                // Must run even when a disposal above throws: a live container left in ContainerProvider
+                // makes CallOriginalPolicy deny originals for every later environment-less test in the
+                // process (the shared Harmony patches stay applied), silently corrupting game-object
+                // construction there.
+                OutputSinkManager.RemoveLogCallback(TestOutputCallback);
+                ContainerProvider.Clear();
             }
-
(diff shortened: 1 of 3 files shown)


ShoT-UPfps:
bot changelog entry in the pr desc is "Fixed conversations and player interactions being silently blocked after a player left a session in the middle of a dialog with another player" is that right?

ShoT-UPfps (changes requested):
bot changelog entry

ShoT-UPfps (approved):

PR #2233, opened 2026-07-21
[1949] Make battle size server authoritative

### LINKED ISSUE #1949: [ISSUE] Server authoritative battle size
Field battles and sieges should use one server-authoritative battle-size budget shared across every controller. Cap the total number of human agents at 1,000 for now; higher counts can crash vanilla, and we still need to investigate how battle-size mods handle them.

Synchronizing the battle-size setting by itself is not enough. Each client currently caps its own supplier at 1,000, then spawns remote puppets outside that cap. Every owner's local agents and remote puppets need to count against the same global budget.

If someone loads in while it's full then they need to be a spectator. The game shouldn't spawn more troops from the existing players and it should prioritize spawning in that player when slots become available

## Evidence

- #1990: [Coop_client_37844.log](https://gist.github.com/ShoT-UPfps/c37083c07bd112a69cd75617875b9fdb) (1.8 MB, SHA-256 `61d5e179f13af71b965d8e43f33b00986d9835b2221f7606c9960322962ed4bc`). The `Coop_client.log` attached to #1990 belongs to pid 45136, not the process that crashed. Windows Application Error records pid 37844 crashing in `TaleWorlds.Native.dll` with `0xc0000005` at `0x74b34a`. Right before the crash, `BattleTeamDiagnostics` logged 498 defender player-team agents, 833 defender ally agents, and 502 attackers, 1,833 humans total.
- #1995 reproduces this in both [a siege](https://github.com/user-attachments/files/30032986/Coop_client.log) and [a field battle](https://github.com/user-attachments/files/30033137/Coop_client.log). The siege spawned 1,000 owned troops plus 236 remote agents, and the field battle spawned 1,000 owned troops plus 200 remote agents. Both logs end abruptly without a managed exception.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Each client applies its own battle-size limit only to the troops it controls, while other players' troops are added on top. Large field battles and sieges can therefore exceed 1,000 human agents and crash.

## What is the new behavior?

Resolves #1949

The server has a Battle Size slider under Server Options, with choices from 200 to 1,000 and a default of 1,000; the choice is remembered and sent to clients when they join or whenever it changes. The vanilla battle-size option is hidden during co-op, and field battles and sieges keep every player's troops within one shared limit.

## Bot Changelog Entry

- Add a server-controlled battle-size slider and one shared troop limit for field battles and sieges.


ShoT-UPfps:
not worth it

PR #2266, opened 2026-07-22
Let the Steam lobby password prompt reopen after canceling

### LINKED ISSUE #2238: [UI] Cannot enter a password after canceling
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

After canceling the password prompt, the player is unable to enter the password again.

## Steps to reproduce

Not provided

## Additional information

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1529148910667300884/1529148910667300884

## Images and videos

- [Video: Mount__Blade_II_Bannerlord_2026.07.21_-_23.31.25.02.mp4](<https://cdn.discordapp.com/attachments/1529148910667300884/1529148911434989609/Mount__Blade_II_Bannerlord_2026.07.21_-_23.31.25.02.mp4?ex=6aa4c7a6&is=6aa37626&hm=c546a6f564b21e08628b0437efb35624b1bc7556edb9e9252cda8ba46121d9df&>) ([Discord message](https://discord.com/channels/709516043332354119/1529148910667300884/1529148910667300884))

## Logs

Client log: Not provided

Server log: Not provided

garrettluskey:
Fixes #2238

## Root cause

Clicking Join on a password-protected server makes `SteamJoinListener` join the Steam lobby first (membership is required to read the lobby's connection data), and only then does the password prompt open. Canceling the prompt only reset the inquiry flag — nothing left the lobby. Every later Join click on the same server then hit `BeginLobbyJoin`'s already-in-this-lobby early return and silently did nothing, so the prompt could never be opened again. Joining a *different* server still worked, which matches the video in the issue. Side effect: the canceled player kept holding one of the host's limited lobby member slots.

## Fix

- New `SessionJoinAbandoned` message, published when the password prompt is canceled.
- `SteamJoinListener` handles it by leaving the lobby (via the existing `LeaveSessionLobby` path, which already covers the join-in-flight and own-advertised-lobby cases), so the next Join starts a fresh attempt and the ghost member slot is released.
- Safety net: a join request for a lobby the player is somehow still a member of now re-resolves the join info instead of returning silently, so the Join button can never permanently dead-end on stale membership (e.g. any other abandoned-attempt path).

## Testing

- New tests: `AbandonedJoin_LeavesLobbyAndAllowsRejoin` (cancel -> leave -> rejoin loop) and `RejoinRequestWhileStillMember_ResolvesAgain` (re-resolve safety net).
- Full Coop.Tests suite green: 446 passed / 0 failed / 1 pre-existing skip; E2E.Tests compiles clean (nothing in it touches these types).

### DIFF
--- a/source/Coop.Steam/SteamJoinListener.cs
+++ b/source/Coop.Steam/SteamJoinListener.cs
@@ -189,35 +189,47 @@ public class SteamJoinListener : IDisposable, ISteamLobbyMembership
 
     private void ResolveJoinInfo(ulong lobbyId)
     {
-        bool decoded = LobbyDataCodec.TryDecode(key => lobbyApi.GetLobbyData(lobbyId, key), out var info, out var error);
+        try
+        {
+            bool decoded = LobbyDataCodec.TryDecode(key => lobbyApi.GetLobbyData(lobbyId, key), out var info, out var error);
 
-        // A standalone server advertises its own game-server identity to tunnel to; otherwise the
-        // lobby owner runs the tunnel. Either is only readable while still a member of the lobby.
-        if (decoded && info.HasServerSteamId)
-        {
-            info.HostSteamId = info.ServerSteamId;
-        }
-        else if (decoded && info.Version >= SessionJoinInfo.MinTunnelVersion)
-        {
-            info.HostSteamId = lobbyApi.GetLobbyOwner(lobbyId);
-        }
+            // A standalone server advertises its own game-server identity to tunnel to; otherwise the
+            // lobby owner runs the tunnel. Either is only readable while still a member of the lobby.
+            if (decoded && info.HasServerSteamId)
+            {
+                info.HostSteamId = info.ServerSteamId;
+            }
+            else if (decoded && info.Version >= SessionJoinInfo.MinTunnelVersion)
+            {
+                info.HostSteamId = lobbyApi.GetLobbyOwner(lobbyId);
+            }
 
-        if (!decoded)
+            if (!decoded)
+            {
+                LeaveActiveLobby();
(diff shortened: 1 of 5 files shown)


source/Coop.Steam/SteamJoinListener.cs
```diff
@@ -83,7 +91,19 @@ public void JoinSessionLobby(ulong lobbyId)
...
+        if (lobbyId == 0) return;
+
+        if (activeLobbyId == lobbyId)
+        {
+            // Still a member from an earlier attempt whose join never produced a session
+            // (e.g. an abandoned password prompt). The lobby data is readable while a
+            // member, so restart the attempt from the resolve step instead of no-oping.
+            if (resolveJoinInfo)
+            {
+                ResolveJoinInfo(lobbyId);
```
ShoT-UPfps, line 103: shouldnt this use the same error handling as `OnLobbyEntered`? otherwise an exception from `GetLobbyData` or `GetLobbyOwner` leaves us in the lobby and never publishes `SessionJoinFailed`

ShoT-UPfps (changes requested):

ShoT-UPfps (approved):
lgtm

source/Coop.Steam/SteamJoinListener.cs
garrettluskey: Good catch — that path was reachable without the guard. Rather than duplicate it at the new call site, I moved the same leave-and-report handling into `ResolveJoinInfo` itself so both callers (and any future ones) are covered. Added a test that throws from the lobby read on re-resolve and asserts we leave and publish `SessionJoinFailed`.

Issue #2263, opened 2026-07-22
Leaving Siege does not remove player from map event


garrettluskey:
# For New Contributors
[How to contribute](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/wiki/Getting-Started-as-a-Contributor)

## Description

Client perspective
<img width="1028" height="733" alt="Image" src="https://github.com/user-attachments/assets/89e823c8-f00c-4f20-b144-cdba9d5b65d9" />

[Coop_client.log](https://github.com/user-attachments/files/30248559/Coop_client.log)

Server perspective
<img width="924" height="656" alt="Image" src="https://github.com/user-attachments/assets/95d03c55-b7fe-4a1e-a8b8-560641fda6db" />

## Location

Create a branch based from [development](https://github.com/Bannerlord-Coop-Team/BannerlordCoop)

## Definition of Done
- [ ] Class level comments exist for all new classes.
- [ ] XUnit tests exist for every method that does not require the game to be ran.

PR #2272, opened 2026-07-22
Route the remaining native besieger-camp clears through the server

garrettluskey:
Stacked on #2269 (base is `2263/route-siege-leave-through-server`; retargets to `development` when that merges).

## Problem

Same bug class as #2263: native flows that clear `MobileParty.MainParty.BesiegerCamp` client-locally. The AutoSync property prefix on clients only error-logs and applies the write locally without forwarding, so the client walks away while the server keeps the party besieging (ghost tent, party still in the siege event). #2269 routed the leave menus; five writers remained (decompiled from TaleWorlds.CampaignSystem v1.4.7).

## Reachability audit

| Native writer | Reachable while besieging in coop? |
|---|---|
| `join_encounter` leave lambda (in `AddGameMenus`) | Yes — a besieger prompted into `join_encounter` (nearby battle) declines involvement |
| `EncounterGameMenuBehavior.break_in_leave_consequence` | Edge — the `join_siege_event` init mostly redirects camp members away, but the write is guarded and cheap to route |
| `game_menu_encounter_leave_your_soldiers_behind_accept_on_consequence` | Yes — a besieger attacked by a relief force takes try-to-get-away |
| `PlayerEncounter.DoPlayerDefeat` | Yes — verified coop does **not** converge this: a real-time mission defeat unwinds through the native defeat state machine (`PlayerEncounter.Update` still sees a live `PlayerMapEvent`; the forced-`End` in `MapEventResultsHandler` only covers the replicated-commit path), and neither the captivity flow nor the field-battle finalize touches the camp. Retreat-ended defeats are not even captured, so nothing else would ever clear it |
| `SafePassageBarterable.Apply` | Yes — NPC barters run `Apply` locally on the client (`PlayerPartyTradeContext` intercepts only player-to-player barters) |

## Design

Two shapes, matching what each native body does:

- **Suppressed leave menus** (lambda, `break_in_leave_consequence`): prefix publishes `BreakSiegeAttempted` and skips native, approval runs `FinishLocalPlayerSiegeLeave` — the #2269 pattern. The lambda is compiler-generated, so its patch resolves the target by scanning `AddGameMenus` lambdas for the one referencing the `BesiegerCamp` setter. `break_in_leave` also replicates native's army-follower exit through the army flow (`MobileParty.Army` is not auto-synced).
- **Embedded camp writes** (try-to-get-away accept, `DoPlayerDefeat`, `SafePassageBarterable.Apply`): the native body must keep running (troop sacrifice + debrief menu, `Finish()` + taken-prisoner menu, the barter's AI holds and `LeaveEncounter`). The prefix publishes the break, pre-clears the camp under `AllowedThread` (applies the local cascade with no forwarding and no client-write error log; the guarded native write then sees null and skips itself), and returns true. A new `FinishLocalMenus` flag threaded through `BreakSiegeAttempted` → `NetworkRequestBreakSiege` → `NetworkBreakSiegeApproved` stops the approval from running `PlayerEncounter.Finish()`/`ExitToLast` over the menu the native flow just put up.

The safe-passage prefix re-evaluates native's sally-out branch first (the besieger *granting* passage to beaten defenders keeps the siege) so it never routes a break for that case — checked before the pre-null because the branch reads `OriginalParty.SiegeEvent`, which derives from the camp when the original party is the main party.

Known adjacent gaps left out of scope: the rest of the try-to-get-away flow is unsynced (client-local troop/item losses, `try_to_get_away_end` clearing other defender-side parties' camps), and a besieging player defeated via the server-committed path (client never runs `DoPlayerDefeat`) still needs a server-side camp clear in the captivity flow.

## Tests

`SiegeLeaveMenuTests` extended from 4 to 11 tests, all driving the real Harmony-patched methods (the lambda through the same IL resolution the patch uses):

- round trips for all five writers (camp cleared on server and every client, exactly one request from the leaver);
- `FinishLocalMenus` asserted true for the suppressed-menu shape, false end-to-end (request and approval) for the embedded shape;
- no-camp negative for `DoPlayerDefeat`; host passthrough (server routes nothing, camp stays until native handles it).

Suites run locally (Debug): SiegeLeaveMenuTests 11/11, SiegeEvents+BesiegerCamps 19/19, MapEvents+Armies 145/145 (1 pre-existing skip), GameInterface.Tests 636/636 (11 pre-existing skips), Coop.Tests 444/444 (1 pre-existing skip).

Part of the #2263 bug class.


source/GameInterface/Services/SiegeEvents/Patches/SiegeEntryFlowPatches.cs
```diff
@@ -162,4 +166,154 @@ private static bool LeaveSiegeAfterAttackPrefix()
...
+        if (army != null && army.LeaderParty != mainParty)
+        {
+            MessageBroker.Instance.Publish(army, new MobilePartyInArmyRemoved(army, mainParty, mainParty));
+            using (new AllowedThread())
+            {
+                ArmyPatches.RemoveMobilePartyInArmy(mainParty, army, mainParty);
+            }
+        }
+
+        MessageBroker.Instance.Publish(null, new BreakSiegeAttempted(mainParty));
```
ShoT-UPfps: this also drops the unconditional `SetMoveModeHold` at the end of `break_in_leave_consequence`. the approval runs `PlayerEncounter.Finish` under `AllowedThread`, so nothing else restores the hold.

ShoT-UPfps: we still need to route the `MapEventSide = null` branch here. otherwise a party that already has a map-event side stays in that battle after clicking "dont get involved."

source/GameInterface/Services/SiegeEvents/Patches/SiegeEntryFlowPatches.cs
```diff
@@ -162,4 +166,154 @@ private static bool LeaveSiegeAfterAttackPrefix()
...
+    private static MethodBase TargetMethod() => ResolveJoinEncounterLeaveConsequence();
+
+    [HarmonyPrefix]
+    private static bool JoinEncounterLeavePrefix()
+    {
+        if (CallOriginalPolicy.IsOriginalAllowed()) return true;
+        if (ModInformation.IsServer) return true;
+        if (MobileParty.MainParty.BesiegerCamp == null) return true;
+
+        MessageBroker.Instance.Publish(null, new BreakSiegeAttempted(MobileParty.MainParty));
```
ShoT-UPfps: this also needs the native `SetMoveModeHold` behavior. the approval runs `PlayerEncounter.Finish` under `AllowedThread`, so our finish postfix skips its hold and the party keeps its previous movement order.

ShoT-UPfps (changes requested):

source/GameInterface/Services/SiegeEvents/Patches/SiegeEntryFlowPatches.cs
garrettluskey: Fixed in cd078bc83. Reissued the hold from the prefix via a shared `HoldAfterSiegeLeave` helper: `SetMoveModeHold()` locally plus `PartyBehaviorChangeAttempted` through the gated AI-behavior channel, so the server clears the stale besiege order instead of a bare local write that dynamic sync drops (same idiom as `FinishPostfix`/`ClearEngageOrder`). Kept out of the shared approval because the other four routed leave menus (`LeaveSiege`, passive-wait, `leave_siege_after_attack`, `EncounterLeaveConsequence`) do not hold in the v1.4.7 decompile.

garrettluskey: Fixed in cd078bc83. Routed the `MapEventSide = null` clear through `PlayerLeaveBattleAttempted(mainParty.Party)` (the single-party-removal message `PlayerEncounterPatches.EncounterLeavePrefix` already uses), guarded on `MapEventSide != null` and ordered before the camp break to match native. Covered by the new `ClientBreakInLeave_WhileOnMapEventSide_RoutesBattleLeave` test.

source/GameInterface/Services/SiegeEvents/Patches/SiegeEntryFlowPatches.cs
garrettluskey: Fixed in cd078bc83. Same `HoldAfterSiegeLeave` reissue: the lambda `SetMoveModeHold()` after `Finish()` is now replicated (local hold plus `PartyBehaviorChangeAttempted`), so FinishPostfix skipping under `AllowedThread` no longer leaves the party on its previous order. Asserted via `DefaultBehavior`/`ShortTermBehavior == Hold` on the leaver.

source/GameInterface/Services/SiegeEvents/Patches/SiegeEntryFlowPatches.cs
ShoT-UPfps: the map-event removal is routed now, but the `NetworkPartyLeftBattle` handler already calls `PlayerEncounter.Finish(false)`. this break still has `FinishLocalMenus=true`, so its approval calls `GameMenu.ExitToLast()` a second time. can we send the break with `FinishLocalMenus=false` when this branch also routes the battle leave?

ShoT-UPfps (changes requested):
still 1 comment outstanding https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/2272#discussion_r3639710222

source/GameInterface/Services/SiegeEvents/Patches/SiegeEntryFlowPatches.cs
garrettluskey: Good catch. Fixed in 13075cec9 - the break now goes out with `FinishLocalMenus=false` whenever this branch also routes the battle leave, so the only menu exit is the one from the `NetworkPartyLeftBattle` finish. When there's no map-event side the break still sends true, since the approval is the only thing finishing the menu in that case. Also updated ClientBreakInLeave_WhileOnMapEventSide_RoutesBattleLeave to pin the flag on both the request and the approval.

ShoT-UPfps (approved):
lgtm

PR #2273, opened 2026-07-22
Clear the besieger camp when the server captures a defeated besieging player

garrettluskey:
## Problem

Part of the #2263 besieger-camp bug family, server side. A besieging player defeated through the **server-committed battle path** was never removed from its siege camp:

- The defeated client's encounter is staged straight to `End` by `MapEventResultsHandler`, so native `PlayerEncounter.DoPlayerDefeat` — which clears `MobileParty.BesiegerCamp` — never runs for these defeats.
- Server side, `CaptureDefeatedPlayerHeroes` applies `TakePrisonerAction`, which never touches the camp, and the coop park (`PlayerCaptivityServerHandler`) doesn't either.

Net: the captured player's party stayed parked in the besieger camp on the server while its hero sat in captivity, and no client ever saw the camp cleared. (#2272 covers the client-side real-time-mission defeat path; this covers the server-committed one.)

## Fix

`CaptureDefeatedPlayerHeroes` now schedules a server-authoritative camp clear after each capture (`ScheduleBesiegerCampClear`). Both capture call sites funnel through it (the native-path prefix and the `MapEventResultsInterface` reimplementation), and both are retreat-guarded upstream, so a retreating besieger is unaffected.

**Why the clear is deferred (`GameThread.EnqueueSafe`) instead of inline:** the capture runs inside the battle result commit, and when the captured party is the *last* besieger the clear cascades `RemoveSiegePartyInternal → SiegeEvent.FinalizeSiegeEvent → BesiegedSettlement.Party.MapEvent.FinalizeEvent()` — re-finalizing the very map event being committed when the defeat is a wall assault (`FinalizeEventAux`'s `IsFinalized` guard only trips after it completes). Native sequences camp clears outside the commit stack too: the player's runs in `DoPlayerDefeat` after `Finish()`, a defeated AI party's in `RemoveParty` on a later tick. The deferred write runs with patches live, so the cleared auto-synced `BesiegerCamp` replicates to every client, and it re-checks for null so it composes with #2272's client-side routing in either order.

## Tests

New `BesiegerCaptureTests` E2E drives the real route — a client commits the `BattleState`, the server applies it authoritatively and captures there — with the player party field-wired into a synced besieger camp, then drains the deferred clear the way the live game loop does. Asserts the clear is not applied inline (guarding the mid-commit re-finalize hazard), then that captivity synced and the camp is gone on the server and every client. Verified red without the production change.

`DefeatPlayerByBattleStateSync` is split into `CreateBattleStateDefeatMapEvent` + `CommitBattleStateDefeat` (behavior unchanged for existing callers): a besieging defender makes `MapEvent.Initialize` walk the siege graph's involved parties, which the headless siege fixture (disabled side initializers) cannot satisfy, so the camp is wired between the halves.

Full E2E suite: 795 passed / 4 skipped / 0 failed.


source/GameInterface/Services/PlayerCaptivityService/Patches/PlayerStartCaptivityPatches.cs
```diff
@@ -124,9 +124,43 @@ internal static void CaptureDefeatedPlayerHeroes(MBReadOnlyList<MapEventParty> w
...
+    /// </summary>
+    private static void ScheduleBesiegerCampClear(MobileParty defeatedParty)
+    {
+        if (defeatedParty.BesiegerCamp == null) return;
+
+        Logger.Debug("Scheduling besieger camp clear for captured player party {PartyId}", defeatedParty.StringId);
+
+        GameThread.EnqueueSafe(() =>
+        {
+            // The camp can already be gone when the queue drains (the host's own native defeat cleared it,
```
ShoT-UPfps: the host doesnt have a party or encounter, should this say the client's native defeat clear may have reached the server first?

ShoT-UPfps (changes requested):

ShoT-UPfps (approved):

ShoT-UPfps (changes requested):
failing to build

source/GameInterface/Services/PlayerCaptivityService/Patches/PlayerStartCaptivityPatches.cs
garrettluskey: Good catch — right, the parties captured here are client player parties (resolved via the player registry), so there's no host defeat on this path. What can empty the camp first is the defeated client's native defeat clear (routed to the server by the #2272 patches) landing before the deferred clear drains. Reworded in f74ee6f.

ShoT-UPfps (approved):
lgtm

PR #2293, opened 2026-07-23
Reduce CI test time by sharding them

ShoT-UPfps:
Split the E2E suite across 64 isolated jobs and run the non-E2E tests from a focused solution filter.

The E2E launcher uses POSIX sh because the CI container does not include bash.

Also fixed flaky test


ShoT-UPfps:
fake codefactors, the script needs that formatting

garrettluskey (approved):

PR #2300, opened 2026-07-23
Synchronize field battle horse animations

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Cavalry horses can stop moving at their destination on one client while continuing to play a walking animation on another.

## What is the new behavior?

Cavalry horse movement animations now settle consistently on every client when horses stop, and resume at the authoritative playback speed when they move again.

Battle diagnostics can also focus the camera on a synchronized mount and capture its rendered head-pose timeline for desync evidence.

## Bot Changelog Entry

- Fixed cavalry horses continuing to walk in place for non-host players after stopping in field battles.


source/Missions/Agents/Handlers/AgentMovementHandler.cs
```diff
@@ -370,6 +476,286 @@ private void RemoveStaleLocalState(HashSet<Guid> broadcastAgentIds)
...
+                0,
+                new ActionIndexCache(syntheticTurn.ActionIndex),
+                syntheticTurn.Progress,
+                UseActionBlendPeriod);
+        }
+        foreach (Agent mount in _completedSyntheticMountTurns)
+            ClearSyntheticMountTurn(mount);
+
+        _completedRemoteSyntheticMountTurns.Clear();
+        foreach (var pair in _remoteSyntheticMountTurns)
```
garrettluskey, line 517: This loop never checks authority, so a stale entry survives ownership migration. If this client adopts the mount mid-turn (owner disconnect / host migration), nothing can ever clear the entry: `ClearRequested` is only set by incoming packets, and both packet paths `continue` on locally-controlled agents before reaching `UpdateRemoteSyntheticMountTurn`. Result: we stomp channel 0 of our own AI-driven horse with a frozen turn pose every frame until mission end.

Mirror the local loop above — treat `agentRegistry.IsLocallyControlled(...)` (mount or rider) as completed and drop the entry.

source/Missions/Agents/Handlers/AgentMovementHandler.cs
```diff
@@ -370,6 +476,286 @@ private void RemoveStaleLocalState(HashSet<Guid> broadcastAgentIds)
...
+        {
+            _remoteSyntheticMountTurns[mount] =
+                new RemoteSyntheticMountTurnState(
+                    mountData.MountAction0TurnDirection,
+                    mountData.MountAction0TurnActionIndex,
+                    mountData.MountAction0Progress);
+            return;
+        }
+
+        if (mountData.MountAction0Progress > activeTurn.Progress)
```
garrettluskey: The ratchet holds stale progress across back-to-back turns. Sequence: turn A hits the poll limit → owner sends NoTurn → `ClearRequested` is set → owner detects another facing delta on the next 25ms poll and a new same-direction turn arrives inside the 3-frame grace → this branch resumes the old state with `Progress` still at ~0.99, and the new turn's rising progress loses to the `>` check for its entire 2 seconds. The puppet just sits on the end pose of the previous turn.

When resuming a state that had `ClearRequested` set, take the incoming progress instead of maxing.

source/Missions/Agents/Packets/AgentMountData.cs
```diff
@@ -45,17 +91,46 @@ public void ApplyMount(Agent mountAgent)
...
+                    if (currentAction0 != ActionIndexCache.act_none)
+                        mountAgent.SetActionChannel(0, ActionIndexCache.act_none);
+                }
+                else if (currentAction0 == ActionIndexCache.act_none
+                    || currentAction0.Index != desiredAction0Index
+                    || nativeTurnFlagsChanged)
+                {
+                    mountAgent.SetActionChannel(
+                        0,
+                        new ActionIndexCache(desiredAction0Index),
```
garrettluskey: This drops the safety net the old gait path had. Before, the wire index went through `GetActionNameWithCode` and a null name meant "unknown index, skip". Now an out-of-range/corrupt index from an unreliable packet goes straight into native `SetActionChannel` as a raw `ActionIndexCache`. Channel 1 below still does the name round-trip, so channel 0 is now the odd one out.

Keep the name lookup (or at least validate the index resolves to a known action) before handing it to the engine.

source/Missions/Agents/Packets/AgentMountData.cs
```diff
@@ -113,5 +455,20 @@ public void ApplyMount(Agent mountAgent)
...
+        public float MountAction0Speed { get; }
+        /// <summary>Whether the owner's rendered channel-zero animation is a locomotion gait.</summary>
+        [ProtoMember(17)]
+        public bool MountAction0IsLocomotion { get; }
+        /// <summary>The owner's rendered stationary turn direction: -1 left, 0 none, 1 right.</summary>
+        [ProtoMember(18)]
+        public int MountAction0TurnDirection { get; }
+        /// <summary>The native movement action for the owner's mount type and stationary turn direction.</summary>
+        [ProtoMember(19)]
+        public int MountAction0TurnActionIndex { get; }
```
garrettluskey: `NoActionIndex` is -1, and protobuf only skips zero defaults — a negative varint is 10 bytes, plus the 2-byte tag. So every mount entry in the common no-turn case now carries ~12 bytes of "no turn", plus ~6 for `MountAction0Speed` and ~3 for `IsLocomotion` — roughly 20 bytes per mount per send on the 40 Hz unreliable stream we've already had MTU trouble with, and it shrinks every batch.

Encode "no turn action" as 0 (store index+1, or only populate the field while a turn is active).

source/Missions/Agents/Packets/AgentMountData.cs
```diff
@@ -8,18 +10,49 @@ namespace Missions.Agents.Packets
...
+                    && mountAction0TurnActionIndex.HasValue
+                    && MountAction0Index != mountAction0TurnActionIndex.Value);
+            MountAction0IsSyntheticTurn = syntheticStationaryTurn;
+            MountAction0Flag = syntheticStationaryTurn
+                ? 0UL
+                : (ulong)mountAgent.GetCurrentAnimationFlag(0);
+            MountAction0Progress = syntheticStationaryTurn
+                ? mountAction0TurnProgress ?? 0f
+                : mountAgent.GetCurrentActionProgress(0);
+            GetRenderedAction0State(
```
garrettluskey: This now runs for every locally-owned mount on every 40 Hz poll, whether or not the snapshot gets sent. `GetSkeleton` + `GetAnimationAtChannel` are native calls and the animation name marshals a string each time — in a big cavalry battle that's thousands of native transitions and string allocs per second where the old ctor had none.

The rendered state only matters for the stationary logic. Skip the skeleton read when the mount is above `StationarySpeedThreshold` (default the speed to 1f) and this gets a lot cheaper.

source/Missions/Agents/Handlers/AgentVisualActionAccessor.cs
```diff
@@ -77,6 +84,73 @@ public bool HasVisibleAction(
...
+                    action.Index,
+                    channelParameter:
+                        Math.Max(0f, Math.Min(1f, progress)),
+                    blendPeriodOverride: blendPeriodOverride,
+                    forceFaceMorphRestart: false);
+            }
+
+            return true;
+        }
+        catch
```
garrettluskey: Bare `catch` on a per-frame path — a real bug in here just disappears silently. `GetRenderedAction0State` narrows to `NullReferenceException`; do the same here.

source/Missions/CoopMissionController.cs
```diff
@@ -96,6 +96,10 @@ public override void OnPreDisplayMissionTick(float dt)
         // Diff again here so peers receive that exact action instead of retaining the earlier pose.
         coopMissionComponent.AgentActionHandler.PollActionsAfterNativeTick();
 
+        coopMissionComponent.AgentMovementHandler
+            .ReplaySyntheticMountTurnAnimationsAfterNativeTick();
+        BattleDebugCommands.CaptureMountPoseFrame();
```
garrettluskey: Production tick calling into a debug class every frame. It early-outs cheaply so no perf complaint, but the mission controller shouldn't need to know `BattleDebugCommands` exists. Have the debug class hook a mission event / register its own behavior when a capture starts, so this line can go away.

garrettluskey (dismissed):
The core fix is sound — zeroing puppet input when the owner is stationary and syncing the gait playback speed is the right call for the walk-in-place desync, and the test coverage is thorough. But there are two correctness holes in the remote synthetic-turn lifecycle and some wire/perf costs on the movement hot path I want addressed before this merges. Details inline.

Also flagging scope: BattleDebugCommands grew by ~645 lines of camera/pose-capture tooling that the PR description doesn't mention. I'm fine keeping the tooling, but it's more than half the production diff of a "bug fix" PR — give it a sentence in the description so the changelog and reviewers aren't surprised.

source/Missions/Agents/Handlers/AgentMovementHandler.cs
ShoT-UPfps: fixed, remote turn replay now drops the state as soon as the mount or its rider becomes locally controlled

source/Missions/Agents/Handlers/AgentMovementHandler.cs
ShoT-UPfps: fixed, a same-direction turn that resumes during the clear grace now resets to the incoming progress

source/Missions/Agents/Packets/AgentMountData.cs
ShoT-UPfps: fixed, channel 0 now resolves the action name before calling SetActionChannel

source/Missions/Agents/Packets/AgentMountData.cs
ShoT-UPfps: fixed, field 19 now writes index+1 so the common no-turn value is protobuf's zero default

source/Missions/Agents/Packets/AgentMountData.cs
ShoT-UPfps: fixed, moving mounts now skip the skeleton and animation-name read and use the 1f default

source/Missions/Agents/Handlers/AgentVisualActionAccessor.cs
ShoT-UPfps: fixed, this only catches NullReferenceException now

source/Missions/CoopMissionController.cs
ShoT-UPfps: fixed, the debug tick is now owned by a mission behavior registered only when the debug camera or capture starts

source/Missions/Agents/Handlers/AgentMovementHandler.cs
```diff
@@ -370,6 +476,297 @@ private void RemoveStaleLocalState(HashSet<Guid> broadcastAgentIds)
...
+            Agent rider = mount.RiderAgent;
+            if (agentRegistry.IsLocallyControlled(mount)
+                || (rider != null
+                    && agentRegistry.IsLocallyControlled(rider)))
+            {
+                _completedRemoteSyntheticMountTurns.Add(mount);
+                continue;
+            }
+
+            visualActionAccessor.TrySetAction(
```
garrettluskey, line 538: Non-blocking, follow-up material: this is the one spot still handing a raw wire index to the engine — `RemoteSyntheticMountTurnState.ActionIndex` comes straight off the packet and goes into `SetAgentActionChannel` every frame, so it's the same class of issue as the gait path you just fixed. Validating per-frame here would be wasteful; the cheap place is `UpdateRemoteSyntheticMountTurn`, once at store time. It'd need gating on name resolution actually being available (in the E2E harness every name is null, which would kill the replay tests) — same tension `TryResolveActionTransition` already handles with its narrow fallback. Fine to land as-is and clean up separately.

source/Missions/Agents/Packets/AgentMountData.cs
```diff
@@ -45,17 +97,52 @@ public void ApplyMount(Agent mountAgent)
...
+            // A Controller.None puppet cannot select its channel-zero stand, turn, or gait action.
+            bool stationaryTurn = MountSpeed <= StationarySpeedThreshold
+                && MountAction0TurnDirection != NoTurn;
+            int desiredAction0Index = ResolveAction0Index(
+                MountAction0Index,
+                MountSpeed,
+                MountAction0IsLocomotion,
+                MountAction0TurnDirection,
+                MountAction0TurnActionIndex);
+            string desiredAction0Name = desiredAction0Index == NoActionIndex
```
garrettluskey: Tiny one: this lookup runs on every `ApplyMount`, but it's only consumed in the action-change branch. Steady state (index matches, just progress/speed updates) and synthetic turns pay a reflection `Invoke` plus allocations per mount per packet for nothing — and the old code only did the lookup on a gait change. Move it inside the `else if` and the common receive path goes back to allocation-free.

garrettluskey (approved):
Re-reviewed e00cc43ea..50a37ec2f. All seven requested changes are in and each got a regression test — verified in code, not just the replies:

- Remote replay now drops turn state once the mount or rider is locally controlled, with an authority-transfer test covering it.
- Resume-during-grace takes the incoming progress; test pins it at 0.1 after a resumed turn.
- Channel 0 goes back through `GetActionNameWithCode` and fails closed — the unknown-index test asserting zero `SetActionChannel` calls is the right trade, since the harness can't resolve names and the positive path was only E2E-testable while the code cheated with raw indices. Matches what `TryResolveActionTransition` already does.
- Field 19 ships as index+1 with the private wire property, and the `DeepClone` round-trip proves the encoding.
- Moving mounts skip the skeleton read entirely.
- `TrySetAction` narrows to `NullReferenceException`.
- The debug tick lives in a self-registering `MissionBehavior`; the controller no longer references `BattleDebugCommands`.

Build, unit, and all eight E2E shards are green. CodeFactor is red — glance at it before merging, but I'm not blocking on a linter.

Two take-it-or-leave-it notes inline; neither needs another round.

source/Missions/Agents/Handlers/AgentMovementHandler.cs
ShoT-UPfps: fixed in 8765f7c05, it validates the action when the remote turn state is stored and only uses the raw index when the animation resolver is unavailable.

source/Missions/Agents/Packets/AgentMountData.cs
ShoT-UPfps: fixed in 8765f7c05, the action-name lookup is now inside the action-change branch.

PR #2303, opened 2026-07-23
Fixed remaining barters

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Some barters not synced correctly

## What is the new behavior?

Resolves #

## Bot Changelog Entry

Remaining barters synced


EgardA:
fake codefactors

source/GameInterface/Services/Barters/Patches/LordBarterPatch.cs
```diff
@@ -0,0 +1,326 @@
...
+                PlayerEncounter.Current != null &&
+                barter.OtherParty == MobileParty.ConversationParty?.Party)
+            {
+                using (new AllowedThread())
+                {
+                    var faction = barter.OtherParty.MapFaction;
+                    if (faction != null)
+                        faction.NotAttackableByPlayerUntilTime = CampaignTime.DaysFromNow(5f);
+                }
+                PlayerEncounter.LeaveEncounter = true;
```
ShoT-UPfps: this also needs the siege safe passage branch. vanilla calls `PlayerSiege.FinalizePlayerSiege` here, otherwise the player can pay for passage and stay involved in the siege.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
```diff
@@ -0,0 +1,556 @@
...
+    }
+
+    private bool MatchesPrisoner(Hero prisoner, PeaceBarterTerm term) => prisoner?.CharacterObject != null && objectManager.TryGetId(prisoner.CharacterObject, out var id) && id == term.ObjectId;
+
+    private static void ApplySafePassage(MobileParty targetParty, MobileParty playerParty)
+    {
+        if (targetParty == null || playerParty == null) return;
+        var attackProtectionEnds = CampaignTime.HoursFromNow(32f);
+        var factionProtectionEnds = CampaignTime.DaysFromNow(5f);
+        var protectedParties = new HashSet<MobileParty> { targetParty };
```
ShoT-UPfps: `SafePassageBarterable.Apply` protects every party returned by `FindAllNpcPartiesWhoWillJoinEvent`, including nearby non-attached parties. this only protects the target's army, so another lord on the same encounter side can immediately attack the player after the barter.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
```diff
@@ -0,0 +1,556 @@
...
+            initializer = BarterManager.Instance.InitializeSafePassageBarterContext;
+            baseBarterables.Add(new SafePassageBarterable(targetHero, playerHero, targetParty, playerParty));
+            baseBarterables.Add(new NoAttackBarterable(playerHero, targetHero, playerParty, targetParty, CampaignTime.Days(5f)));
+        }
+        else if (kind == LordBarterKind.JoinKingdomAsClan)
+        {
+            initializer = BarterManager.Instance.InitializeJoinFactionBarterContext;
+            baseBarterables.Add(new JoinKingdomAsClanBarterable(
+                targetHero,
+                playerHero.Clan.Kingdom,
```
ShoT-UPfps: shouldnt the authorization include the destination from `JoinKingdomAsClanBarterable.TargetKingdom`? this rebuilds it from `playerHero.Clan.Kingdom`, so if the player's clan changes kingdoms while the barter is open the target defects to a different kingdom than the UI showed.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
```diff
@@ -0,0 +1,556 @@
...
+            }
+            if ((LordBarterKind)request.Kind == LordBarterKind.SafePassage)
+            {
+                ApplySafePassage(targetParty.MobileParty, playerParty.MobileParty);
+                ConversationPartyHold.EndEngagement(conversationPartyTracker, peer);
+            }
+
+            FlushGold(playerHero);
+            FlushGold(targetHero);
+            SendAccepted(peer, request, playerHero.Gold);
```
ShoT-UPfps, line 239: we need to apply the overpay relation bonus before sending success. vanilla does this after accepted barters, otherwise overpaying a lord here gives no relation.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
```diff
@@ -0,0 +1,556 @@
...
+            var offerValue = manager.GetOfferValueForFaction(barter, targetHero.Clan);
+            if (offerValue < -0.01f)
+            {
+                Reject(peer, request, playerHero.Gold, "The lord will not accept this offer.");
+                return;
+            }
+
+            authorizations.Remove(peer);
+            mutationStarted = true;
+            var offered = barter.GetOfferedBarterables();
```
ShoT-UPfps, line 215: `BarterManager.ApplyBarterOffer` calls `CampaignEventDispatcher.Instance.OnBarterAccepted` after applying the barterables. we need to dispatch that here too, otherwise campaign behaviors never see these accepted barters.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
```diff
@@ -0,0 +1,669 @@
...
+                    request,
+                    targetKingdom,
+                    out var barter,
+                    out reason))
+            {
+                Reject(peer, request, playerHero.Gold, reason);
+                return;
+            }
+            var manager = BarterManager.Instance;
+            var offerValue = manager.GetOfferValueForFaction(barter, targetHero.Clan);
```
ShoT-UPfps: `SafePassageBarterable.GetUnitValueForFaction` calls `PlayerEncounter.Current.FindAllNpcPartiesWhoWillJoinEvent`, but the dedicated server never has a `PlayerEncounter.Current`, so this throws here and every lord safe passage request gets rejected. we need to calculate the safe passage value without `PlayerEncounter` on the server.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
ShoT-UPfps: this still depends on `PlayerEncounter.Current`, which the dedicated server never has. `LordBarterSyncTests.SafePassage_ProtectsEveryEncounterOpponent` creates a mock server encounter, so production still only adds `targetParty` here.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
```diff
@@ -0,0 +1,669 @@
...
+    private void FlushGold(Hero hero)
+    {
+        if (sendCoalescer != null && hero != null && objectManager.TryGetId(hero, out var id)) sendCoalescer.FlushInstance(id, network);
+    }
+
+    private void Reject(NetPeer peer, NetworkRequestLordBarter request, int gold, string reason)
+    {
+        Logger.Warning("Rejected lord barter with {TargetHeroId}: {Reason}", request.TargetHeroId, reason);
+        var result = new NetworkLordBarterResult(request.ContextId, false, gold, reason, request.RequestId);
+        completedResults[peer] = result;
```
ShoT-UPfps: we shouldnt cache rejected results here. the client keeps the same request id after rejection so it can resend an edited offer, but this makes every retry replay the first rejection without reevaluating it.

source/GameInterface/Services/Barters/Patches/LordBarterPatch.cs
```diff
@@ -0,0 +1,355 @@
...
+                barter.OtherParty == MobileParty.ConversationParty?.Party)
+            {
+                var siegeEvent = barter.OtherParty?.SiegeEvent;
+                using (new AllowedThread())
+                {
+                    var faction = barter.OtherParty.MapFaction;
+                    if (faction != null)
+                        faction.NotAttackableByPlayerUntilTime = CampaignTime.DaysFromNow(5f);
+                }
+                if (siegeEvent != null &&
```
ShoT-UPfps, line 148: this also needs `GainKingdomInfluenceAction.ApplyForSiegeSafePassageBarter(MobileParty.MainParty, -10f)` when the besieged settlement belongs to the players faction. vanilla charges that here, otherwise the player keeps the influence.

source/GameInterface/Services/Barters/Patches/LordBarterPatch.cs
```diff
@@ -0,0 +1,355 @@
...
+                    siegeEvent.BesiegerCamp.HasInvolvedPartyForEventType(barter.OtherParty) &&
+                    siegeEvent.BesiegedSettlement.HasInvolvedPartyForEventType(PartyBase.MainParty))
+                {
+                    using (new AllowedThread())
+                    {
+                        Campaign.Current.GameMenuManager.SetNextMenu("menu_siege_safe_passage_accepted");
+                        PlayerSiege.FinalizePlayerSiege();
+                    }
+                }
+                else
```
ShoT-UPfps, line 158: the non-defender branch also needs to clear `MobileParty.MainParty.BesiegerCamp` when the player is involved as a besieger. vanilla clears it after `LeaveEncounter`, otherwise the player stays attached to the siege.

ShoT-UPfps (changes requested):
cmt

source/GameInterface/Services/Barters/Patches/LordBarterPatch.cs
ShoT-UPfps: this also needs to be cleared on the server. `SiegeEntryFlowPatches.cs:133` already calls out that a client-local `BesiegerCamp` write is never seen by the server, so the authoritative party stays attached here.

source/GameInterface/Services/Barters/Patches/LordBarterPatch.cs
ShoT-UPfps: this still needs to run on the server. this is only in the client result handler and it runs under `AllowedThread`, so the authoritative `Clan._influence` never changes.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.SafePassage.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        return -(int)(price / (relationFactor * relationFactor));
+    }
+
+    private static float CalculateStrengthRatio(
+        MobileParty playerParty,
+        IEnumerable<MobileParty> playerSide,
+        IEnumerable<MobileParty> opponentSide)
+    {
+        var strengthContext = playerParty.IsCurrentlyAtSea
```
ShoT-UPfps: shouldnt this use `PlainBattle` here? vanilla only uses `SeaBattle` when `PlayerEncounter.IsNavalEncounter()` sees a naval map event, but `LordBarterHandler.cs:377-378` rejects requests with a map event, so an eligible sea barter is priced differently on the client and server.

source/GameInterface/Services/Barters/SafePassagePartyResolver.cs
```diff
@@ -0,0 +1,135 @@
+using System.Collections.Generic;
+using System.Linq;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.Party;
+
+namespace GameInterface.Services.Barters;
+
+internal static class SafePassagePartyResolver
```
ShoT-UPfps: please make this a transient DI class and inject it into both handlers. we dont add new shared logic as static helpers.

ShoT-UPfps (approved):
lgtm

PR #2330, opened 2026-07-24
[2279] Improve battle animations

ShoT-UPfps:
## Bot Changelog Entry

- Improved block syncing


ShoT-UPfps:
Fake codefactas

garrettluskey:
## 🚫 Blocker 1: Weapon-pickup sync never activates in a live game

`AgentPickupPatch` is the only interceptor of `Agent.OnItemPickup` in the repo, and it has no `[HarmonyPatchCategory]` — so it is never applied outside the test harness.

### Why the patch never applies

`GameInterface.PatchAll()` applies exactly two groups of patches:

1. **Uncategorized** patch classes — but only from the **GameInterface** assembly (`PatchAllUncategorized(assembly)`), and
2. categories explicitly registered via `HarmonyPatchCategoryRegistration`.

https://github.com/Bannerlord-Coop-Team/BannerlordCoop/blob/2fc5711642b3691122384eacfb773306522cab64/source/GameInterface/GameInterface.cs#L60-L74

A patch class in the **Missions** assembly is therefore live only if it carries a registered category. This PR wires that up correctly for the two patches it newly enables:

https://github.com/Bannerlord-Coop-Team/BannerlordCoop/blob/2fc5711642b3691122384eacfb773306522cab64/source/Missions/MissionModule.cs#L187-L200

…but the pickup patch never got one:

https://github.com/Bannerlord-Coop-Team/BannerlordCoop/blob/2fc5711642b3691122384eacfb773306522cab64/source/Missions/Agents/Patches/AgentPickupPatch.cs#L13-L15

### What breaks

The whole feature chain is dead in a real session:

`Agent.OnItemPickup` runs undetoured → `WeaponPickedup` is never published → `WeaponPickupHandler` never runs → `NetworkWeaponPickedup` (including the new `CurrentEquipment` wield snapshot) is never sent → **peers never see the pickup**, and the "shield stays holstered while the remote agent is blocking" fix this PR builds never executes.

### Why the tests still pass

`WeaponPickupSyncTests` installs its own Harmony patches and drives `WeaponPickupHandler` directly, so the E2E suite exercises the handler while the production entry point stays dead. Nothing covers "is the patch actually applied by `PatchAll`".

### Suggested fix

```csharp
[HarmonyPatch(typeof(Agent), "OnItemPickup")]
[HarmonyPatchCategory(MissionModule.WeaponPickupPatchCategory)] // new
public class AgentPickupPatch
```

plus the matching constant and `yield return` in `MissionModule.CreatePatchCategoryRegistrations()`. A small E2E/unit test asserting every `[HarmonyPatch]` class in the Missions assembly carries a registered category would prevent the next one of these (this same mechanism already silently killed `ShieldDamagePatch` and `AgentDamagePatch` before this PR).

garrettluskey:
## 🚫 Blocker 2: `GuardedHitWindow` can cancel a *different* blow than the one that was blocked

The window exists to stop a blocked swing from also replaying raw damage — but the way tokens are minted and consumed lets it swallow a **legitimate follow-up hit**, silently.

### Three properties that combine badly

1. **Tokens aren't tied to a specific blow.** The key is just the `(victim, attacker)` agent pair plus a queue count — nothing correlates a token with the swing that was actually blocked:

   https://github.com/Bannerlord-Coop-Team/BannerlordCoop/blob/2fc5711642b3691122384eacfb773306522cab64/source/Missions/Agents/GuardedHitWindow.cs#L105-L128

2. **Tokens outlive the deferral.** A pending damage candidate waits 1 epoch (`ReadyEpoch = Epoch + 1`), but tokens live 2 (`RetainedEpochs = 2`) — one full tick longer than anything they could legitimately cancel.

3. **Every blocked hit mints a token, including missiles.** The `LocalAgentGuardedHit` publish has no `isMissile` filter — while the `guardReactionHandler.ObserveBlockedHit(...)` call immediately below it *does* reject missiles:

   https://github.com/Bannerlord-Coop-Team/BannerlordCoop/blob/2fc5711642b3691122384eacfb773306522cab64/source/Missions/Agents/Handlers/AgentActionHandler.cs#L495-L521

### Failure walk-through

- **Tick N** — a swing (or arrow) is cleanly blocked → `OnScoreHit(isBlocked: true)` → token minted, valid through N+2. A clean block never reaches `Agent.RegisterBlow` (native builds it with `cancelDamage: true`), so **no damage candidate exists to consume this token**.
- **Tick N+1** — the follow-up hit *lands* → `RegisterBlow` intercepted → queued with `ReadyEpoch = N+2`.
- **Tick N+2** — `DrainPendingLocalDamage` finds the stale token, consumes it, and `continue`s → the landing hit is never routed and never applied. There is no log line, so the dropped hit is undiagnosable in the field:

  https://github.com/Bannerlord-Coop-Team/BannerlordCoop/blob/2fc5711642b3691122384eacfb773306522cab64/source/Missions/Battles/BattleDamageRouter.cs#L281-L303

Same pattern in the tournament path (https://github.com/Bannerlord-Coop-Team/BannerlordCoop/blob/2fc5711642b3691122384eacfb773306522cab64/source/Missions/Tournaments/CoopTournamentController.cs#L1017).

### Concrete repros

- **Arrow blocked on a shield**, then a melee hit from the same attacker lands within the ~2-tick window → melee damage vanishes.
- **Tournament: blocked swing, then horse trample.** The tournament recorder folds the mount into the rider, so both hits share the same `(victim, attacker)` key → trample damage vanishes.

### Suggested fixes (any subset helps)

1. Don't mint tokens for missile blocks — mirror the `isMissile` check `GuardReactionHandler.ObserveBlockedHit` already does (`GuardReactionHandler.cs:143`).
2. Shrink `RetainedEpochs` to match the 1-epoch deferral, or better, correlate the token with the specific blow it should cancel.
3. Log when a pending damage candidate is consumed so a dropped hit is at least visible in the logs.

### One design question

Crush-through blows are *blocked and damaging* in the same swing — as written, the window intentionally suppresses their chip damage on puppets. Worth confirming that's the wanted gameplay outcome rather than a side effect.

garrettluskey (approved):

PR #2388, opened 2026-07-26
Reduced network usage in battles

### LINKED ISSUE #2319: Batch mission movement updates by serialized packet size
## Description

Mission movement updates currently use a fixed three-agent batch so unreliable packets stay below the 1,000-byte ceiling. An approximately 800-agent field battle produced roughly 4,500-5,100 movement packets per second per client. Payload profiling also showed that encoded size depends on agent state, LZ4 compressibility, mounted data, and identity scope, so no larger fixed agent count is guaranteed safe.

## Intended Design

Build movement batches using the actual serialized and compressed packet size. Add snapshots until the next candidate would exceed `LiteNetP2PClient.SafeSinglePacketBytes`, then send the last safe batch and continue with the remaining snapshots.

Apply the same rule to `MovementPacket` and `MountMovementPacket`. Bound identity-scope size, or explicitly handle a single snapshot that cannot fit, so oversized unreliable movement is never silently discarded.

## Location

- `source/Missions/Agents/Handlers/AgentMovementHandler.cs`
- `source/Missions/Services/Network/MovementPacketCompressor.cs`
- `source/Missions/Services/Network/LiteNetP2PClient.cs`
- `source/E2E.Tests/Services/Missions/MovementTrafficTests.cs`

Create a branch based from [development](https://github.com/Bannerlord-Coop-Team/BannerlordCoop).

## Related Issues

Follow-up to #2309.

## Requirements

- Batch on-foot and mounted movement by actual wire size instead of a fixed agent count.
- Keep every direct and server-relayed unreliable datagram within the 1,000-byte ceiling.
- Preserve compact movement IDs and legacy GUID fallback behavior.
- Do not drop the final snapshot when a candidate batch is split.
- Define safe behavior for a single snapshot that exceeds the ceiling.
- Avoid repeated serialization work that would offset the packet-count savings at large battle sizes.
- Add coverage for incompressible payloads, mounted snapshots, relay framing, and exact boundary sizes.

## Definition of Done

- [ ] Movement batching cannot emit an oversized unreliable packet.
- [ ] Large battles use fewer movement packets than the current three-agent batching.
- [ ] XUnit coverage exists for the batching and boundary rules.

EgardA:
This reverts commit a97121275b05a587fdff5a125e11f48142871e58.

## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Resolves #2319 

Please let me know what you think, it does reduce the total network usage in a 250 vs 250 by quite a bit.

## Bot Changelog Entry

Reduce network usage in battles


ShoT-UPfps:
fix codefactas

ShoT-UPfps (changes requested):
codefacta

garrettluskey (changes requested):
Can we refactor this code to be less complex and more readable

ShoT-UPfps:
is this resolving this issue? https://github.com/Bannerlord-Coop-Team/BannerlordCoop/issues/2319

EgardA:
It should @ShoT-UPfps

source/Missions/Agents/Handlers/AgentMovementHandler.cs
```diff
@@ -89,6 +89,7 @@ public AgentMovementHandler(
         this.agentRegistry = agentRegistry;
         this.controllerIdProvider = controllerIdProvider;
         this.equipmentApplier = equipmentApplier;
+        this.movementBatchSender = new MovementBatchSender(client, movementPacketCompressor);
```
ShoT-UPfps: can we inject this through an interface and register it `InstancePerDependency()` in `MissionModule`? otherwise we cant fake the batch sender in `AgentMovementHandler` tests.

ShoT-UPfps (changes requested):
minor

ShoT-UPfps (approved):
lgtm

garrettluskey (approved):
Looks Me To Good

PR #2398, opened 2026-07-26
Remove w from vec3 surrogate and use custom surrogate for colors

Gerseras:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Taleworlds is using an own implementation of vec3 and is reusing it to for colors. So they added a w property which would be a Vec4 and use it for the alpha channel

## What is the new behavior?
No longer sync w of Vec3 and use its default value of -1 instead. Add a custom Surrogate for colors which actually contains the w of Vec3. Will reduce networkload per Agent by atleast 8 bytes as its atleast 2 floats less to transmit


ShoT-UPfps (approved):

ShoT-UPfps (changes requested):
failing test

edit: just looks like flakes

ShoT-UPfps (approved):

PR #2420, opened 2026-07-27
Resolve hero field applies in queue order with hero creation

### LINKED ISSUE #2396: [ENCYCLOPEDIA] Encyclopedia does not open after returning to the game from Alt+Tab
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

The encyclopedia fails to open after returning to the game from an Alt+Tab.

## Steps to reproduce

Not provided

## Additional information

Reconnecting is a temporary workaround that fixes the issue.

## Discord source

https://discord.com/channels/709516043332354119/1530809134524530808/1530809134524530808

## Logs

Client log: [0.0.3+956d7da115b55fbae5d3a7ef5ea19e887d27b517 - Download Coop_client.log](<https://cdn.discordapp.com/attachments/1530809134524530808/1530809336509763716/Coop_client.log?ex=6aa4e34a&is=6aa391ca&hm=2ad907be7350653da2ba939b1ce27a18bb4c8f22c015fa5e50fee46caff6d91d&>)

Server log: Not provided

garrettluskey:
Fixes #2396

## Root cause

The encyclopedia was not failing to open — it was throwing a `NullReferenceException` inside the native `DefaultEncyclopediaHeroPage.InitializeListItems()` on every open, and `OnTickRobustnessPatches`/`ScreenManagerRobustnessPatches` swallowed the crash, so the UI silently did nothing. The NRE came from `hero.Name.ToString()` on a hero whose name was never applied.

Client-side hero creation is marshaled onto the game-loop thread (`AutoRegistryHandler.Handle_NetworkCreateInstance` -> `GameThread.RunSafe`), but `HeroDataHandler.Handle_HeroChangeName` and every handler in `HeroFieldsHandler` resolved the hero id **immediately on the network (poller) thread** and dropped the message when the lookup failed. A field apply arriving while the creation was still sitting in the game-thread queue — near-certain while the game loop is starved during an alt-tab — was lost forever. `ChangeHeroName` is one-shot (published from the `Hero.SetName` prefix, never re-sent), so the skeleton hero created by `HeroRegistry.OnClientCreated` stayed name-less in `CampaignObjectManager.AliveHeroes` until reconnect, which matches the reported workaround.

The attached client log shows the exact sequence: six dropped messages for `Created_3276` at 14:26:58 ("Unable to find/get Hero"), first encyclopedia NRE at 14:26:59, and no further errors for that id (the creation landed moments later).

## Fix

Resolve the hero id inside `GameThread.RunSafe`, so lookups run in queue order behind the marshaled creation — the same contract `GenericHandler.MarshalApply` documents for AutoSync applies, and the same shape the `ChangeCharacterObject` handler and `Handle_NetworkDestroyInstance` already use. This also moves the field writes themselves onto the game-loop thread instead of mutating game state from the poller thread. In passing, `Dispose` now unsubscribes `ChangeHomeSettlement`, which was missing.

## Testing

- New regression test `HeroFieldsHandlerThreadingTests` parks the test game-loop pump, queues the "creation", publishes `ChangeName` from the poller-stand-in thread, and asserts the name applies once the queue drains. Verified red on the old code (name dropped) and green on the fix.
- GameInterface.Tests: 672 passed, 11 skipped (pre-existing skips)
- Coop.IntegrationTests: 128 passed, 2 skipped
- E2E.Tests `Services.Heroes` filter: 21 passed

### DIFF
--- a/source/GameInterface.Tests/Services/Heroes/HeroFieldsHandlerThreadingTests.cs
+++ b/source/GameInterface.Tests/Services/Heroes/HeroFieldsHandlerThreadingTests.cs
@@ -1,0 +1,90 @@
+using Common;
+using Common.Messaging;
+using GameInterface.Services.Heroes.Handlers;
+using GameInterface.Services.Heroes.Messages;
+using GameInterface.Services.ObjectManager;
+using Moq;
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.Serialization;
+using System.Threading;
+using TaleWorlds.CampaignSystem;
+using Xunit;
+
+namespace GameInterface.Tests.Services.Heroes;
+
+/// <summary>
+/// Verifies that <see cref="HeroFieldsHandler"/> resolves the target hero on the game-loop thread,
+/// in queue order with the marshaled hero creation, instead of on the network (poller) thread that
+/// publishes the message. A poller-thread lookup races a creation still waiting in the game-thread
+/// queue and permanently drops the one-shot apply, leaving a partially-initialized hero — a
+/// name-less hero in the campaign's hero list breaks every encyclopedia open.
+/// </summary>
+public class HeroFieldsHandlerThreadingTests
+{
+    static HeroFieldsHandlerThreadingTests()
+    {
+        // Coop.Tests starts and continuously pumps a dedicated game-loop thread from a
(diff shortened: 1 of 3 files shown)


ShoT-UPfps (approved):
lgtm

PR #2424, opened 2026-07-27
Sync map tracks

### LINKED ISSUE #2421: [ISSUE] Enable and sync map tracks
# For New Contributors

## Location

## Definition of Done

jordanbrymora:
## PR Checklist

## PR Type

Map tracks are currently enabled but only managed client side. This means the skill xp gained from spotting tracks isn't gained and players all see different tracks across the map.

## What is the new behavior?

Resolves #2421 

Manage all tracks on the server and update visibility of tracks for clients directly. This also means the track data persists in the save game.

## Bot Changelog Entry

Map tracks are now synced and scouting can be levelled from spotting them.


source/GameInterface/Services/MapTracks/Patches/MapTracksCampaignBehaviorPatches.cs
```diff
@@ -0,0 +1,78 @@
...
+
+[HarmonyPatch(typeof(MapTracksCampaignBehavior.TrackPool))]
+internal class TrackPoolPatches
+{
+    [HarmonyPatch(nameof(MapTracksCampaignBehavior.TrackPool.RequestTrack))]
+    [HarmonyPostfix]
+    public static void RequestTrackPostfix(Track __result)
+    {
+        __result.IsEnemy = false; // Server doesn't have enemies
+        __result.IsPointer = false; // Doesn't seem to be used in vanilla
```
garrettluskey: this runs on the client too, not just the server, so a track requested through AddMapArrow (quest pointers) would get IsPointer zeroed here. is that path reachable on clients? also worth a null check on __result in case the pool ever hands back null.

garrettluskey (dismissed):
Verified the reimplemented formulas against decompiled vanilla: the detection loop, difficulty calculation, spotting distance, and expiry walk all match, with MainParty swapped for the per-player party and IsTrackDropped using the nearest player party as the comment says. Server owning _allTracks and the per-player detected sets means expiry ordering stays consistent, the join path rekeys and resends the full visible set so reconnects don't leak stale tracks, and the surrogate round-trips exactly the fields TracksMatch compares. Scouting xp now grants on the server where the detection actually happens, which fixes the leveling gap.

source/GameInterface/Services/MapTracks/Patches/MapTracksCampaignBehaviorPatches.cs
jordanbrymora: Added

source/GameInterface/Services/MapTracks/Patches/MapTracksCampaignBehaviorPatches.cs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+[HarmonyPatch(typeof(MapTracksCampaignBehavior.TrackPool))]
+internal class TrackPoolPatches
+{
+    [HarmonyPatch(nameof(MapTracksCampaignBehavior.TrackPool.RequestTrack))]
+    [HarmonyPostfix]
+    public static void RequestTrackPostfix(Track __result)
+    {
+        if (ModInformation.IsClient || __result == null) return;
```
garrettluskey: this postfix runs after vanilla `RequestTrack`. That body calls `GetTrackLife`, which dereferences `MobileParty.MainParty`, and then reads `Hero.MainHero.MapFaction`; both are null on the dedicated host. The first party that tries to leave a track will throw before reaching these assignments, so we need to replace or patch the server path before those dereferences instead of repairing the result in a postfix.

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
```diff
@@ -0,0 +1,342 @@
...
+        if (closestParty == null) return false;
+
+        float num2 = closestParty.IsActive ? (closestParty._lastCalculatedSpeed * Campaign.Current.Models.MapTrackModel.MaxTrackLife) : 0f;
+        return num2 * num2 > closestPlayerPartyDistance;
+    }
+
+    public void AddPlayerPartyKeys(string playerPartyId)
+    {
+        // If a player rejoins, their visible tracks will be lost. Reset server side detected tracks
+        playerDetectedTracks[playerPartyId] = new HashSet<Track>();
```
garrettluskey, line 363: this resets the existing set on every join, then `DetectTracksForPlayerParty` treats every nearby saved track as new and grants XP again. A normal reconnect can repeatedly award scouting XP for the same tracks, and a server restart has the same problem because this dictionary isn't saved. Keep or restore each player's detected state and send the full set on join without treating it as new detections.

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
```diff
@@ -0,0 +1,342 @@
...
+        {
+            return false;
+        }
+
+        // Find and use the closet player party to determine if a track should be dropped.
+        MobileParty closestParty = null;
+        float closestPlayerPartyDistance = float.MaxValue;
+        foreach (var playerParty in GetPlayerParties())
+        {
+            float playerPartyDistance = mobileParty.Position.DistanceSquared(playerParty.Position);
```
garrettluskey, line 349: the geometrically closest party isn't enough because the radius below depends on that party's speed. A stationary party can be closest while a faster party slightly farther away still satisfies vanilla's distance check, and then we won't create a track that player should be able to see. Check each player party's own speed/radius and return true when any one qualifies.

source/GameInterface/Surrogates/TrackSurrogate.cs
```diff
@@ -0,0 +1,104 @@
...
+    [ProtoMember(11)]
+    public int NumberOfPrisoners { get; set; }
+
+    [ProtoMember(12)]
+    public CampaignTime CreationTime { get; set; }
+
+    [ProtoMember(13)]
+    public float Life { get; set; }
+
+    [ProtoMember(14)]
```
garrettluskey, line 50: we also need `IsPointer` in this wire format. Vanilla `AddMapArrow` creates an already-detected pointer with no culture, but this path never adds it to `playerDetectedTracks` and `ApplyVisibleTrackChanges` drops culture-less tracks, so map/quest arrows never appear on clients. Send those arrows explicitly and preserve pointer state instead of routing them through the normal skill-gated scan.

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
```diff
@@ -0,0 +1,342 @@
...
+        {
+            visibleTrackChanges[playerPartyId] = new();
+        }
+        return visibleTrackChanges;
+    }
+
+    private void GrantTrackDetectionXp(MobileParty playerParty, Track track)
+    {
+        float scoutingXp = Campaign.Current.Models.MapTrackModel.GetSkillFromTrackDetected(track);
+        playerParty.EffectiveScout?.AddSkillXp(DefaultSkills.Scouting, scoutingXp);
```
garrettluskey: `IsEnemy = false` isn't only presentation here. `DefaultMapTrackModel.GetSkillFromTrackDetected` multiplies enemy-track XP by 2x to 10x, so this always awards the base amount and also removes enemy colors for high-scouting players. Hostility is relative to the observing player's faction, so keep enough source-party/faction data to calculate it per player instead of clearing it globally.

garrettluskey (changes requested):
The server-owned track list and per-player delta delivery are headed the right way, but the dedicated-server creation path still dereferences the missing host hero/party. There are also correctness gaps around reconnect XP, multi-party range checks, map arrows, and enemy-track XP that need fixing before this merges.

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
jordanbrymora: Changed

source/GameInterface/Surrogates/TrackSurrogate.cs
jordanbrymora: Added

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
jordanbrymora: Changed

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
jordanbrymora: Changed to not grant xp when initialising tracks

source/GameInterface/Services/MapTracks/Patches/MapTracksCampaignBehaviorPatches.cs
jordanbrymora: Changed to use interface method replacing `MapTracksCampaignBehavior.AddTrack` instead of a postfix on the track pool.

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
```diff
@@ -0,0 +1,498 @@
...
+    void ClearVisibleTracks(MapTracksCampaignBehavior behavior);
+}
+
+public class MapTracksCampaignBehaviorInterface : IMapTracksCampaignBehaviorInterface
+{
+    private readonly Dictionary<string, HashSet<Track>> playerDetectedTracks = new();
+
+    // When tracks are dropped the original party is lost.
+    // Keep track of the faction for computing IsEnemy on clients.
+    private readonly Dictionary<Track, IFaction> trackMapFactions = new();
```
garrettluskey, line 103: `trackMapFactions` only covers tracks created in the current process. vanilla `SyncData` saves `_allTracks`, but nothing saves or rebuilds this dictionary, so after a server restart every loaded track is treated as friendly and enemy-track xp/colors are lost. persist the source faction alongside the saved tracks and restore it on load.

source/GameInterface/Services/MapTracks/Data/MapTrackData.cs
```diff
@@ -0,0 +1,20 @@
+using ProtoBuf;
+using TaleWorlds.CampaignSystem;
+
+namespace GameInterface.Services.MapTracks.Messages;
+
+[ProtoContract(SkipConstructor = true)]
+public class MapTrackData
```
garrettluskey, line 7: please add tests for the new track surrogate/message round-trip and reconnect initialization. #2421 requires xUnit coverage for the parts that dont need a live game, but this adds no tests.

source/GameInterface/Services/MapTracks/Data/MapTrackData.cs
```diff
@@ -0,0 +1,20 @@
+using ProtoBuf;
```
garrettluskey, line 1: nit: please keep the new source files UTF-8 with a BOM. this file, `MapTracksCampaignBehaviorInterface.cs`, and the three update-message files currently omit it.

garrettluskey (changes requested):
the host-safe creation path and per-player range checks look right. source faction still disappears on save/load, and the new wire/state paths need coverage before this merges.

source/GameInterface/Services/MapTracks/Data/MapTrackData.cs
jordanbrymora: Added in `3a87101`

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
jordanbrymora: Added in `ad041fa`

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
```diff
@@ -0,0 +1,553 @@
...
+    /// <summary>
+    /// Client method
+    /// Removes every visible track, used when the server re-sends a full set of visible tracks
+    /// </summary>
+    void ClearVisibleTracks(MapTracksCampaignBehavior behavior);
+}
+
+public class MapTracksCampaignBehaviorInterface : IMapTracksCampaignBehaviorInterface
+{
+    private readonly Dictionary<string, HashSet<Track>> playerDetectedTracks = new();
```
garrettluskey, line 99: `playerDetectedTracks` needs save data too. after a server restart each party starts empty, so live tracks it already spotted outside its current radius disappear on reconnect.

source/GameInterface/Services/MapTracks/Handlers/MapTracksHandler.cs
```diff
@@ -0,0 +1,166 @@
...
+    private void Handle_NetworkInitializePlayerTracksKeys(MessagePayload<NetworkInitializePlayerTracksKeys> obj)
+    {
+        if (ModInformation.IsClient) return;
+
+        GameThread.RunSafe(() =>
+        {
+            if (!TryGetMapTracksBehavior(out var mapTracksBehavior)) return;
+            if (!objectManager.TryGetObjectWithLogging<MobileParty>(obj.What.PlayerPartyId, out var playerParty)) return;
+
+            mapTracksCampaignBehaviorInterface.AddPlayerPartyKeys(obj.What.PlayerPartyId);
```
garrettluskey: use the return value here. a first-time player currently marks every nearby track detected with xp disabled, so those tracks can never grant scouting xp later.

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
```diff
@@ -0,0 +1,553 @@
...
+    {
+        RemoveExpiredTracks(behavior);
+    }
+
+    public void QuarterHourlyTick(MapTracksCampaignBehavior behavior)
+    {
+        var visibleTrackChanges = InitializeVisibleTrackChanges();
+
+        // Run for all player parties instead of just one
+        foreach (var playerParty in GetPlayerParties())
```
garrettluskey, line 140: only run detection for connected players. `Players` keeps disconnected registrations, so offline parties currently spot tracks and gain scouting xp.

source/GameInterface/Surrogates/TrackSurrogate.cs
```diff
@@ -0,0 +1,112 @@
...
+    public static implicit operator TrackSurrogate(Track track)
+    {
+        if (track is null)
+            return null!;
+
+        return new TrackSurrogate
+        {
+            Position = track.Position,
+            Direction = track.Direction,
+            PartyName = track.PartyName.ToString(),
```
garrettluskey: use the existing `TextObject` surrogate here instead of `ToString()`, otherwise party names are rendered in the server's language and cant localize on clients.

garrettluskey (changes requested):
the source faction now survives save/load and the new wire tests cover the fields used for delivery and removal. the per-player detected state and join/offline handling still lose tracks or award scouting xp at the wrong time.

source/GameInterface/Surrogates/TrackSurrogate.cs
jordanbrymora: I agree this is a good idea for localisation but we will probably need a way to compress the packets. TextObjects are very large to send over the network and one on every map track increases the size exponentially. Also, the track packets are few in number but larger individually already.

Using `TextObjectSurrogate` increases the size of packets drastically:
image

Without using 'TextObjectSurrogate' the packets are much smaller but still quite large:
image

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
jordanbrymora: Fixed

source/GameInterface/Services/MapTracks/Handlers/MapTracksHandler.cs
jordanbrymora: Fixed

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
jordanbrymora: Fixed

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
```diff
@@ -0,0 +1,604 @@
...
+        if (ModInformation.IsClient || savedPlayerDetectedTracks == null) return;
+
+        // Load detected tracks back into playerDetectedTracks
+        foreach (var savedPlayerDetectedTrack in savedPlayerDetectedTracks)
+        {
+            if (savedPlayerDetectedTrack?.PlayerId == null || savedPlayerDetectedTrack.DetectedTracks == null) continue;
+
+            foreach (var detectedTrack in savedPlayerDetectedTrack.DetectedTracks)
+            {
+                playerDetectedTracks[savedPlayerDetectedTrack.PlayerId].Add(detectedTrack);
```
garrettluskey, line 263: `playerDetectedTracks` was cleared above and no entry is created for this `PlayerId`, so the first saved record throws `KeyNotFoundException` on load. create the `HashSet` before adding its tracks.

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
```diff
@@ -0,0 +1,604 @@
...
+            }
+        }
+
+        return visibleTrackChanges;
+    }
+
+    public List<MapTrackData> InitializePlayerVisibleTracks(MapTracksCampaignBehavior behavior, MobileParty playerParty)
+    {
+        // Don't grant scouting xp when initializing tracks on join
+        DetectTracksForPlayerParty(behavior, playerParty, false);
```
garrettluskey: this detects tracks created while the player was offline with xp disabled. they are added to the detected set here, so the first connected quarter tick can never award their scouting xp. resend the existing set without scanning, then let normal connected detection find new tracks.

garrettluskey (changes requested):
the host-safe track creation, faction persistence, wire fields, and connected-player detection are in place. detected-state load and reconnect still break the saved/rejoining path.

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
jordanbrymora: Fixed. Removed the `DetectTracksForPlayerParty` call in `InitializePlayerVisibleTracks` so that only detected tracks from `playerDetectedTracks` can be used to initialise detected tracks.

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
jordanbrymora: Fixed

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
```diff
@@ -0,0 +1,604 @@
...
+
+        // Load detected tracks back into playerDetectedTracks
+        foreach (var savedPlayerDetectedTrack in savedPlayerDetectedTracks)
+        {
+            if (savedPlayerDetectedTrack?.PlayerId == null || savedPlayerDetectedTrack.DetectedTracks == null) continue;
+
+            playerDetectedTracks[savedPlayerDetectedTrack.PlayerId] = new();
+            foreach (var detectedTrack in savedPlayerDetectedTrack.DetectedTracks)
+            {
+                playerDetectedTracks[savedPlayerDetectedTrack.PlayerId].Add(detectedTrack);
```
garrettluskey, line 263: `GameLoadFinished` removes expired tracks from `_allTracks`, but this restores them into `playerDetectedTracks` without the same check. reconnect then resends the expired track, and `RemoveExpiredTracks` can never send its removal because it is no longer in `_allTracks`. skip null or expired tracks here.

garrettluskey (changes requested):
the saved detected sets, reconnect-only resend, and source-faction data fix the prior XP and restore gaps. changes are still needed because expired saved tracks can be restored into a player's visible set after vanilla removes them from the live track list.

source/GameInterface/Services/MapTracks/Interfaces/MapTracksCampaignBehaviorInterface.cs
jordanbrymora: Fixed

garrettluskey (approved):
the saved detected-track restore now drops null and expired entries before reconnect initialization, so every resent track still belongs to the live server list and can receive normal expiry removal. the previous restore blocker is fixed and all checks pass.

PR #2439, opened 2026-07-28
[2334] Show saving overlay during server saves

### LINKED ISSUE #2334: [PERFORMANCE] Random stuttering freezes all co-op players for several seconds
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

When all three co-op players were marching near each other, the game froze for everyone for a few seconds. The freeze cleared, then happened again about ten minutes later.

## Steps to reproduce

Not provided

## Additional information

Observed in v1.4.7 in either Western or Southern Empire territory across the water from Aserai. No event was identified as the trigger; autosave was suggested as a possible cause. Screenshots were not available.

## Discord source

https://discord.com/channels/709516043332354119/1530245933789483168/1530245933789483168

## Logs

Client log: [0.0.3+318e72caca9b49fe54b8abfef8267823d861ab15 - Download Coop_client_Jay.log](<https://cdn.discordapp.com/attachments/1530245933789483168/1530246142330409071/Coop_client_Jay.log?ex=6aa4d106&is=6aa37f86&hm=881f97068af9094561aaee1bae43e1f92927d68fb212e20ea273e5a7ee99b568&>) · [0.0.3+318e72caca9b49fe54b8abfef8267823d861ab15 - Download Coop_client_Korr.log](<https://cdn.discordapp.com/attachments/1530245933789483168/1530246142594515168/Coop_client_Korr.log?ex=6aa4d106&is=6aa37f86&hm=85dc5f45e37337e4263fa85e54cd622e1112a9c6861d19119f8ee8d203b6affd&>) · [0.0.3+318e72caca9b49fe54b8abfef8267823d861ab15 - Download Coop_client_Paddon.log](<https://cdn.discordapp.com/attachments/1530245933789483168/1530246142862823626/Coop_client_Paddon.log?ex=6aa4d107&is=6aa37f87&hm=83a83c2c06c01eee8f198eb27ca52e0b725d5a77c5aea52b79be1821c72c6167&>)

Server log: [Download coop-server-20260724-125625.log](<https://cdn.discordapp.com/attachments/1530245933789483168/1530246143290638568/coop-server-20260724-125625.log?ex=6aa4d107&is=6aa37f87&hm=67a5eb27af53ddb648c5e06a62273e23978da2892a32a05c45617af8e16c9f9e&>)

## Additional report details

- A concrete workaround suggestion is to show a "Saving" overlay on all clients whenever the server saves, matching vanilla save-overlay behavior. ([Discord message](https://discord.com/channels/709516043332354119/1530245933789483168/1530246276028039318))

- Two stutter/freeze events were observed at 13:37 and 13:47. ([Discord message](https://discord.com/channels/709516043332354119/1530245933789483168/1530246282134950129))

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

When the dedicated server saves, every player can appear frozen for several seconds without any indication that a save is in progress.

## What is the new behavior?

Resolves #2334

Every client now sees Bannerlord's Saving overlay for the full duration of a dedicated-server save, and the overlay clears when saving finishes.

## Bot Changelog Entry

- Show the Saving overlay to every player while the server saves.


source/Coop.Core/Server/Services/Save/Handlers/SaveGameHandler.cs
```diff
@@ -39,10 +46,16 @@ public void Dispose()
...
         messageBroker.Unsubscribe<GameSaved>(Handle_GameSaved);
         messageBroker.Unsubscribe<GameLoaded>(Handle_GameLoaded);
+        messageBroker.Unsubscribe<GameSaveStateChanged>(Handle_GameSaveStateChanged);
 
         messageBroker.Unsubscribe<AllGameObjectsRegistered>(Handle_AllGameObjectsRegistered);
     }
 
+    private void Handle_GameSaveStateChanged(MessagePayload<GameSaveStateChanged> payload)
+    {
+        network.SendAll(new NetworkGameSaveStateChanged(payload.What.IsSaving));
```
garrettluskey: The client half of this got a test but the server relay didn't. A mirror test is cheap with the existing pieces: `TestMessageBroker` plus `Mock<INetwork>`, publish `GameSaveStateChanged`, verify `SendAll` got a `NetworkGameSaveStateChanged` carrying the same flag. `CampaignTimeSyncHandlerTests` already does the `Mock<INetwork>` setup if you want a reference.

source/GameInterface/Services/Save/Interfaces/SaveNotificationInterface.cs
```diff
@@ -0,0 +1,45 @@
...
+                ._dataSource;
+            if (dataSource == null) return;
+
+            if (isSaving)
+            {
+                dataSource.OnSaveStarted();
+            }
+            else
+            {
+                dataSource.OnSaveOver(true, null);
```
garrettluskey, line 39: I decompiled `MapSaveVM` to check this call: `OnSaveOver` ignores both arguments and just flips `IsActive` off, so the hardcoded `(true, null)` is safe.

Two things worth knowing though:

- Showing the overlay is not purely visual. The VM's active-state callback makes `GauntletMapSaveView` grab layer focus and set input restrictions, so clients are input-locked while it's up, exactly like a native save. Presumably that's what you want since the server is blocked anyway, but it should be a conscious choice.
- If the clear never arrives (server crashes or hangs mid-save), the client sits on an input-locked overlay until disconnect teardown replaces the map screen, which can take the full disconnect timeout. A `SetSaving(false)` in the client's disconnect/teardown path would be cheap insurance. Not a blocker.

source/GameInterface/Services/Save/Patches/SavePatches.cs
```diff
@@ -19,3 +22,28 @@ static bool Prefix(Game __instance, ref string saveName)
         return true;
     }
 }
+
+[HarmonyPatch(typeof(SaveHandler), "OnSaveStarted")]
```
garrettluskey, line 26: Verified in the decompile that `OnSaveEnded` is called from `OnSaveCompleted` for both outcomes (the success bool is just `result.Item1 == SaveResult.Success`), so the clear goes out on failed saves too. Good hook choice.

One scope question: the join-time transfer save (`SaveInterface.SaveCurrentGame`) calls `Game.Current.Save` directly and never goes through `SaveHandler`, so a joining player still blocks the server with no overlay for everyone else. Intentionally out of scope for #2334? Fine if so, but then it probably deserves a follow-up issue, since from the other players' point of view the symptom is identical.

source/GameInterface/Services/MapEvents/Commands/MapEventDebugCommands.cs
```diff
@@ -1376,8 +1376,9 @@ public static string FinishPlayerEncounter(List<string> args)
             return $"Unable to resolve PartyBase for player {args[0]}.";
         }
 
+        playerParty.Party.MapEventSide = null;
```
garrettluskey: This changes the existing command's behavior in two ways the description doesn't mention: `requireReady` flipped to false, and the server now nulls `MapEventSide` directly instead of only broadcasting the leave. What needed it, finishing encounters for parties that weren't ready yet while testing the overlay? I'm fine with the new semantics and the doc comment matches, I'd just like the reason on record, since debug commands tend to get copied as reference for the real flows.

source/Coop.Core/Server/Services/Save/Messages/NetworkGameSaveStateChanged.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Common.Messaging;
+using ProtoBuf;
+
+namespace Coop.Core.Server.Services.Save.Messages;
+
+[ProtoContract(SkipConstructor = true)]
+public readonly struct NetworkGameSaveStateChanged : IEvent
```
garrettluskey, line 8: Nit: the checklist claims class-level docs, but this type, `GameSaveStateChanged`, and `ISaveNotificationInterface` all ship without a summary. One line each is enough.

garrettluskey (commented):
I read this with the decompiler and the network internals open, since the real risks here are timing and native coupling rather than anything visible in the diff. The mechanism holds up:

- The broadcast is enqueued on the game thread, but the aggregation buffer is drained by the network poll thread (`FlushPendingMessages` in `CoopServer.Update`), so the saving-started packet actually leaves while the game thread is blocked in the save. That's the property this whole feature depends on, and it holds.
- Start and end ride the same reliable-ordered message stream, so they can't arrive swapped.
- A client that leaves the map mid-save can't get a stale overlay: the VM dies with the map screen, and `ShouldApplySavingState` letting the clear through unconditionally covers the rest. The truth-table test pins that helper down nicely.
- `SaveHandler.OnSaveEnded` fires on failed saves too, so the overlay clears either way.

The evidence hold on `force_autosave` is well contained: DEBUG-only, server-only, bounded, one-shot. The location and map-event debug command additions read like the tooling used to produce the test evidence, which is fine to keep; note that `coop.debug.location.enter` starts a client-local settlement encounter, which is a desync hazard by design, so a warning in its usage string wouldn't hurt.

Actionable items are inline: a missing server-side test for the relay, optional clear-on-disconnect hardening, a scope question about the join-time save, and a docs nit.

source/Coop.Core/Server/Services/Save/Handlers/SaveGameHandler.cs
ShoT-UPfps: added `SaveGameHandlerTests` to verify the first save start and last save end are relayed with the matching state.

source/GameInterface/Services/Save/Interfaces/SaveNotificationInterface.cs
ShoT-UPfps: added the clear in `SaveDataHandler.Dispose` so teardown releases the input-locked overlay even if the server end message never arrives.

source/GameInterface/Services/Save/Patches/SavePatches.cs
ShoT-UPfps: included the join snapshot path. it publishes through the same source-aware save-state aggregator and flushes the start before the blocking save, so overlapping native saves cannot clear it early.

source/GameInterface/Services/MapEvents/Commands/MapEventDebugCommands.cs
ShoT-UPfps: this was needed because the live fixture had to clean up a party already in an active map event. latest `development` now handles that with `allowActiveMapEvent: true` and `PlayerLeaveBattleAttempted`, so i took that path instead of mutating `MapEventSide` directly.

source/Coop.Core/Server/Services/Save/Messages/NetworkGameSaveStateChanged.cs
ShoT-UPfps: added summaries to `NetworkGameSaveStateChanged`, `GameSaveStateChanged`, and `ISaveNotificationInterface`.

garrettluskey (approved):
Re-reviewed at 18cf9b78a. All five threads from the first pass are properly resolved, and the new code that came out of them holds up:

- **Source-aware aggregator** (`SaveGameHandler.activeSaveSources`): I re-decompiled `SaveHandler` to check the edge cases — `OnSaveStarted`/`OnSaveEnded` only fire for the handler's own queued saves, never for the join path's direct `Game.Current.Save`, so the join path genuinely needed its own publish, and an overlapping native autosave + join save can't clear each other's overlay early. Back-to-back queued autosaves stay balanced since each cycle re-runs PreSave on the same publisher instance.
- **Game-thread `FlushPendingMessages`**: safe. `pendingMessages` is a `ConcurrentDictionary` with per-peer locks and `NetPeer.Send` is a LiteNetLib enqueue drained by its own logic thread, so flushing while the poll thread is parked on the blocking join-save action is sound, and the start packet reaches the wire before the save blocks.
- **Failure paths**: publish/flush failures are caught, the `finally` always emits the clear, failed snapshots included; a stray clear is idempotent on the client. The flush-failure test pins this.
- **Plumbing**: `SaveNotificationInterface` rides the `IGameAbstraction` auto-registration; the client handler receives on the poller thread and marshals via `GameThread.RunSafe`; the new proto message matches existing readonly-struct wire types and is additive (wire-safe under the FNV-1a type ids). The joining peer gets both broadcasts on the loading screen where `ShouldApplySavingState` correctly no-ops.

Ran the 16 affected Coop.Tests locally at head — all pass. CI is green; the CodeFactor failure is pre-existing complexity debt in `MapEventDebugCommands.cs`, which this PR no longer touches.

Non-blocking nits: `SaveGameHandlerTests` passes `null!` for three unused ctor deps (Moq stubs would match house style), and clients inside battle missions still get no indication during a server save — deliberate campaign-map-only scope, but worth remembering if #2334-style reports continue from mid-battle players.

PR #2456, opened 2026-07-29
Fix not being able to capture lords after battle

### LINKED ISSUE #2453: [ISSUE] Can't capture lords after battles
# For New Contributors

## Description

After pressing the option to take a lord as a prisoner after a battle, the lord is not actually taken as a prisoner.

## Location

## Definition of Done

jordanbrymora:
## PR Checklist

## PR Type

## What is the new behavior?

Resolves #2453

## Bot Changelog Entry

Fixed not being able to capture lords after battles.


source/GameInterface/Services/Heroes/Handlers/LordConversationsCampaignBehaviorHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            ChangeRelationAction.ApplyRelationChangeBetweenHeroes(playerHero, conversationHero, prisonerLiberationRelationReward);
+
+            if (conversationHero.IsPrisoner)
+            {
+                EndCaptivityAction.ApplyByReleasedAfterBattle(conversationHero);
+            }
+        },
+        context: nameof(Handle_NetworkLiberateLordPrisoner));
+    }
```
garrettluskey, line 94: **The server applies client-supplied ids without validating them against the sending peer** — this applies to all five `Handle_Network*` handlers.

The `Handle_NetworkPrisonerLiberationAttempted` this PR deletes did two things these handlers no longer do:

1. **Resolved the acting hero from the authenticated peer** (`payload.Who` → `playerManager.TryGetPlayer`) instead of trusting a wire id.
2. **Checked `prisoner.IsPrisoner` before granting the relation reward**, so a replayed or stale message was a no-op.

As written here:

- Any client can send `NetworkTakeLordPrisoner` with arbitrary ids and the server will capture any lord into any party.
- Replaying `NetworkLiberateLordPrisoner` farms +10 relation per message — the relation applies unconditionally; only the release is `IsPrisoner`-guarded.

Suggested shape, mirroring the deleted handler (needs `IPlayerManager` injected, as `PlayerCaptivityServerHandler` does):

```csharp
private void Handle_NetworkLiberateLordPrisoner(MessagePayload<NetworkLiberateLordPrisoner> obj)
{
    if (ModInformation.IsClient) return;

    GameThread.RunSafe(() =>
    {
        if (!(obj.Who is NetPeer peer) || !playerManager.TryGetPlayer(peer, out var player))
        {
            Logger.Error("Received {Message} without a registered player peer", nameof(NetworkLiberateLordPrisoner));
            return;
        }

        if (!objectManager.TryGetObjectWithLogging<Hero>(player.HeroId, out var playerHero)) return;
        if (!objectManager.TryGetObjectWithLogging<Hero>(obj.What.ConversationHeroId, out var conversationHero)) return;
        if (!conversationHero.IsPrisoner) return; // replayed/stale request: no reward, no release

        ChangeRelationAction.ApplyRelationChangeBetweenHeroes(playerHero, conversationHero, prisonerLiberationRelationReward);
        EndCaptivityAction.ApplyByReleasedAfterBattle(conversationHero);
    },
    context: nameof(Handle_NetworkLiberateLordPrisoner));
}
```

With the peer-derived hero, `MainHeroId` can drop out of the messages entirely. For `NetworkTakeLordPrisoner` (where the lord isn't a prisoner yet), the equivalent check is that `MainPartyId` resolves to the sending player's own party.

source/GameInterface/Services/Heroes/Patches/LordConversationsCampaignBehaviorPatches.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    [HarmonyPatch(nameof(LordConversationsCampaignBehavior.conversation_ally_thanks_meet_after_helping_in_battle_2_on_consequence))]
+    [HarmonyPrefix]
+    public static bool ConversationAllyThanksMeetAfterHelpingInBattle2OnConsequencePrefix(
+        LordConversationsCampaignBehavior __instance)
+    {
+        // TODO: PlayerMapEvent will be null here. Need to get the relation change without it
+        //int playerGainedRelationAmount = Campaign.Current.Models.BattleRewardModel.GetPlayerGainedRelationAmount(MapEvent.PlayerMapEvent, Hero.OneToOneConversationHero);
+
+        var message = new LordHelpedInBattle(Hero.MainHero, Hero.OneToOneConversationHero); // playerGainedRelationAmount
```
garrettluskey, line 43: **This dropped relation reward has a clean fix: compute the amount here in the prefix and carry it in the message.**

`MapEvent.PlayerMapEvent` is null on the *server* — but this prefix runs on the acting client, which is exactly where native computes the amount. The ally-thanks dialog is only reachable while `MapEvent.PlayerMapEvent != null` (the on_condition checks it), so it should still be valid here:

```csharp
int relationAmount = Campaign.Current.Models.BattleRewardModel
    .GetPlayerGainedRelationAmount(MapEvent.PlayerMapEvent, Hero.OneToOneConversationHero);

var message = new LordHelpedInBattle(Hero.MainHero, Hero.OneToOneConversationHero, relationAmount);
```

`NetworkLordHelpedInBattle` gains an int field, and `Handle_NetworkLordHelpedInBattle` applies the received amount (clamped server-side, since it's client-supplied). That restores the vanilla reward and removes both commented-out blocks.

If it turns out coop's battle teardown clears `PlayerMapEvent` before the consequence fires, capture the amount when the dialog condition runs and stash it for the prefix instead.

source/GameInterface/Services/Heroes/Handlers/LordConversationsCampaignBehaviorHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+namespace GameInterface.Services.Heroes.Handlers;
+
+internal class LordConversationsCampaignBehaviorHandler : IHandler
+{
+    private static readonly ILogger Logger = LogManager.GetLogger<LordConversationsCampaignBehaviorHandler>();
+
+    private readonly IObjectManager objectManager;
+    private readonly INetwork network;
+    private readonly IMessageBroker messageBroker;
+    private readonly ISendCoalescer sendCoalescer;
```
garrettluskey: `sendCoalescer` is injected and stored but never used (looks carried over from the barter handlers), and `Logger` is never used either. Removing them also clears out the now-unneeded usings:

- `using Common.Network.Coalescing;`
- `using Serilog;` / `using Common.Logging;` (only if `Logger` goes — worth keeping it if the peer-validation logging lands)
- `using TaleWorlds.CampaignSystem.Roster;` (nothing in the file uses it)
- `using static GameInterface.Services.ObjectManager.ObjectManager;` (`TryGetIdWithLogging`/`TryGetObjectWithLogging` are `IObjectManager` members, so the static import isn't needed)

source/GameInterface/Services/Heroes/Patches/LordConversationsCampaignBehaviorPatches.cs
```diff
@@ -0,0 +1,101 @@
...
+        var message = new TakeLordPrisoner(Campaign.Current.MainParty.Party, CharacterObject.OneToOneConversationCharacter.HeroObject);
+        MessageBroker.Instance.Publish(__instance, message);
+
+        return false;
+    }
+
+    [HarmonyPatch(nameof(LordConversationsCampaignBehavior.conversation_talk_lord_defeat_to_lord_release_on_consequence))]
+    [HarmonyPrefix]
+    public static bool ConversationTalkLordDefeatToLordReleaseOnConsequencePrefix(
+        LordConversationsCampaignBehavior __instance)
```
garrettluskey: `conversation_talk_lord_defeat_to_lord_release_on_consequence` is **static** in the game assembly, so Harmony injects `null` for this `__instance` (Harmony 2.4.2 emits `Ldnull` when the original is static). It doesn't crash, but the publish source is silently null and the signature implies an instance exists.

Clearer to drop the parameter and make the null explicit:

```csharp
[HarmonyPatch(nameof(LordConversationsCampaignBehavior.conversation_talk_lord_defeat_to_lord_release_on_consequence))]
[HarmonyPrefix]
public static bool ConversationTalkLordDefeatToLordReleaseOnConsequencePrefix()
{
    DialogHelper.SetDialogString("DEFEAT_LORD_ANSWER", "str_prisoner_released");

    var message = new LordDefeatToRelease(Hero.MainHero, Hero.OneToOneConversationHero);
    MessageBroker.Instance.Publish(null, message);

    return false;
}
```

source/GameInterface/Services/Heroes/Handlers/LordConversationsCampaignBehaviorHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+
+    private const int prisonerLiberationRelationReward = 10;
+    private const int letLordGoReward = 4;
```
garrettluskey: Nit: constants in this codebase are PascalCase — the deleted `PlayerCaptivityServerHandler` named this exact value `PrisonerLiberationRelationReward`:

```csharp
private const int PrisonerLiberationRelationReward = 10;
private const int LetLordGoReward = 4;
```

(the `ChangeRelationAction` call sites below need the rename too)

garrettluskey (commented):
Part of a fuller review pass. Overall the suppress-and-route approach is the right fix for #2453 (the old flow ran the native consequence on the client, where `TakePrisonerAction.ApplyInternal` is blocked, so the capture silently did nothing). Inline comments below cover server-side validation of the new network handlers, an easy win on the dropped ally relation reward, and a few cleanups.

source/GameInterface/Services/Heroes/Patches/LordConversationsCampaignBehaviorPatches.cs
jordanbrymora: `PlayerMapEvent` will be null in the prefix so this is not possible. Battles are ended immediately. When a client reaches this dialogue option, `PlayerMapEvent` will have already been destroyed including when the dialog condition runs.

source/GameInterface/Services/Heroes/Handlers/LordConversationsCampaignBehaviorHandler.cs
jordanbrymora: 1. Not necessary
2. New implementation better reflects vanilla's implementation where `IsPrisoner` check only guards the `EndCaptivityAction` not the relation change.

source/GameInterface/Services/Heroes/Handlers/LordConversationsCampaignBehaviorHandler.cs
jordanbrymora: Removed unused `sendCoalescer` and imports.

source/GameInterface/Services/Heroes/Patches/LordConversationsCampaignBehaviorPatches.cs
jordanbrymora: Changed

source/GameInterface/Services/Heroes/Handlers/LordConversationsCampaignBehaviorHandler.cs
jordanbrymora: Fixed

source/GameInterface/Services/Heroes/Patches/LordConversationsCampaignBehaviorPatches.cs
jordanbrymora: Thinking a potential fix for this (since other post battle handling also needs the map event) would be to locally store a copy of the map event the client was involved in just before it ends that can be cleared at the end of the `PlayerEncounter.UpdateInternal` cycle.

garrettluskey (approved):

PR #2469, opened 2026-07-30
[2252][2406] Restore siege encounter transitions

### LINKED ISSUE #2252: [SIEGES] Unable to sally out from a siege
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

Players cannot sally out.

## Steps to reproduce

Not provided

## Additional information

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1529239786802385159/1529239786802385159

## Logs

Client log: [0.0.2+7c268b87327107942945be13aef2888fb79e93bb - Download Coop_client.log](<https://cdn.discordapp.com/attachments/1529239786802385159/1529239869593751742/Coop_client.log?ex=6aa4739c&is=6aa3221c&hm=039aa28b39519174c12be441023fe0bc60483feacaf9b200352283826dd0257c&>)

Server log: [Download coop-server-20260721-232513.log](<https://cdn.discordapp.com/attachments/1529239786802385159/1529239924761694369/coop-server-20260721-232513.log?ex=6aa473a9&is=6aa32229&hm=5e49ede604128573b682ad37c81c57cec9fc6ededd6d383dd0549bd6cf9b689d&>)

## Additional report details

- The failure occurs specifically when a player is defending in a siege: attempting to sally out in that state does not work. ([Discord message](https://discord.com/channels/709516043332354119/1529239786802385159/1529239960413012148))

### LINKED ISSUE #2406: [SIEGES] Player gets stuck after breaking in to help defenders when Continue does nothing
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

After breaking in to help the defenders, the player becomes stuck. Clicking Continue has no effect.

## Steps to reproduce

1. Break in to help the defenders.
2. Click Continue.
3. Continue does nothing, leaving the player stuck.

## Additional information

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1530969458347212840/1530969458347212840

## Images and videos

[](<https://cdn.discordapp.com/attachments/1530969458347212840/1530969459425022155/image.png?ex=6aa4cfab&is=6aa37e2b&hm=f86e7d8b5da16e7c8143d7566fdbec793688f500daa492933720e3458b758b45&>) ([Discord message](https://discord.com/channels/709516043332354119/1530969458347212840/1530969458347212840))

## Logs

Client log: [0.0.3+7003060dde3e44abc1d0e243c45ba9a6627516cb - Download Coop_client_-_.log](<https://cdn.discordapp.com/attachments/1530969458347212840/1530969460280791100/Coop_client_-_.log?ex=6aa4cfab&is=6aa37e2b&hm=1d4829cab99bce6ae02d8876d82f3b1103be921824f6686b77ba9a4bf4f560af&>)

Server log: Not provided

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Sallying out, continuing after breaking into a siege, or choosing Attack during a personal siege can leave the client stuck in the siege menu or make the action do nothing.

## What is the new behavior?

Resolves #2252
Resolves #2406

Sally-out and break-in continuations now resume the correct active encounter, while personal siege attacks use the replicated battle and advance normally instead of leaving the player in a nonresponsive menu.

## Bot Changelog Entry

- Fixed siege actions that could leave players stuck when sallying out, continuing after breaking in, or attacking from a personal siege.


source/E2E.Tests/Services/MapEvents/PlayerPartyInteractionFlowTests.cs
```diff
@@ -1422,6 +1456,672 @@ public void AiPartyConversation_AllowedThreadFinish_ReleasesServerEngagement()
...
+            client.NetworkSentMessages.GetMessages<NetworkRequestConversation>());
+        ResetConversationRequestCooldown(client);
+
+        await System.Threading.Tasks.Task.Run(() =>
+            client.SimulateMessage(
+                Server.NetPeer,
+                new NetworkConversationDenied(
+                    ConversationDeniedReason.PartyEngaged,
+                    firstRequest.RequestId)));
+        Common.GameThread.Instance.MarkGameThread();
```
garrettluskey, line 1651: This mark never gets undone. After `await Task.Run(...)` the continuation usually resumes on a threadpool/xunit worker, so the mark lands on a transient thread and only gets corrected when the next test's `E2ETestEnvironment` ctor re-marks. `UnmarkGameThread`'s own doc comment spells out the hazard: recycled managed thread ids can silently promote an unrelated thread to "game thread" and flip `Run` from queueing to inline. We've chased exactly this class of CI-only order flake before.

Can you wrap the marked section in try/finally with `UnmarkGameThread()`, or pull the pattern into a small disposable scope? (Second occurrence at line 1649.)

source/E2E.Tests/Services/MapEvents/PlayerPartyInteractionFlowTests.cs
```diff
@@ -1422,6 +1456,672 @@ public void AiPartyConversation_AllowedThreadFinish_ReleasesServerEngagement()
...
+        await System.Threading.Tasks.Task.Run(() =>
+            client.SimulateMessage(
+                Server.NetPeer,
+                new NetworkAllowConversation(
+                    aiPartyId,
+                    playerPartyId,
+                    forcePlayerOutFromSettlement: false,
+                    ConversationRestartSource.EncounterManager,
+                    secondRequestId)));
+        Common.GameThread.Instance.MarkGameThread();
```
garrettluskey, line 1742: Same as the earlier occurrence — needs the matching `UnmarkGameThread()`.

source/Coop.Core/Client/Services/SiegeEvents/Handlers/ClientSiegeEntryHandler.cs
```diff
@@ -44,6 +48,94 @@ public ClientSiegeEntryHandler(IMessageBroker messageBroker, INetwork network, I
         messageBroker.Subscribe<AssaultSiegeAttempted>(HandleAssaultAttempt);
         messageBroker.Subscribe<NetworkPromptSiegeAssault>(HandleAssaultPrompt);
         messageBroker.Subscribe<NetworkSnapSiegeCampPartyPosition>(HandleCampPositionSnap);
+        messageBroker.Subscribe<BreakInContinuationAttempted>(HandleBreakInContinuationAttempt);
+        messageBroker.Subscribe<NetworkBreakInContinuationApproved>(HandleBreakInContinuationApproved);
+    }
+
+    private void HandleBreakInContinuationAttempt(MessagePayload<BreakInContinuationAttempted> payload)
+    {
+        if (pendingBreakInContinuation != null)
```
garrettluskey: Once a request is in flight this drops every further click, and the prefix keeps suppressing the native consequence — so if the approval never arrives, break-in stays inert for the rest of the session with a staged `LocationEncounter` left behind. The server always replies (even when applying throws), so the realistic loss case is a disconnect mid-flight.

Not blocking, but clearing `pendingBreakInContinuation` on disconnect (or giving it a simple expiry) would close it. Fine as a follow-up issue if you'd rather not grow this PR.

source/GameInterface/Services/MapEvents/Handlers/ConversationRequestHandler.cs
```diff
@@ -44,28 +44,50 @@ internal class ConversationRequestHandler : IHandler
     private readonly INetwork network;
     private readonly IObjectManager objectManager;
     private readonly ConversationPartyTracker conversationPartyTracker;
+    private readonly IConversationRestartContextTracker restartContextTracker;
     private readonly PlayerPartyInteractionHandler playerPartyInteractionHandler;
     private readonly IPlayerManager playerManager;
+    private readonly object pvpInteractionSync = new object();
 
     private DateTime lastRequestSentUtc = DateTime.MinValue;
+    private string pendingConversationRequestId;
```
garrettluskey, line 55: These fields are only safe because every touch point runs on the game thread — I checked: the request path publishes from patch prefixes, and the network handlers all hop through `GameThread.RunSafe`. Worth a one-line invariant comment above the block so the next person doesn't have to re-derive that.

source/GameInterface/Services/MapEvents/Handlers/ConversationRequestHandler.cs
```diff
@@ -462,73 +521,111 @@ private void Handle_NetworkAllowConversation(MessagePayload<NetworkAllowConversa
...
                 Logger.Error(e, "Failed to restart approved conversation encounter; releasing the server-side party hold");
-                SendConversationEndedToServer();
+                ClearPendingConversationRequest(message.RequestId);
+                SendConversationEndedToServer(message.RequestId);
             }
         }, context: nameof(Handle_NetworkAllowConversation));
     }
 
     /// <summary>[Client] This player's encounter finished; tell the server to release the held party.</summary>
     private void Handle_ConversationEnded(MessagePayload<ConversationEnded> payload)
```
garrettluskey, line 660: This three-way send is correct — I traced pending/active/legacy-null against the server-side release contract — but it took a full read to reconstruct why each branch exists. A short comment along the lines of "one ConversationEnded per request id the server may still be holding; null only for legacy/server-detected engagements" would save the next reader that trace.

garrettluskey (commented):
Went through this in depth — notes are split into inline comments below so each one is individually actionable. Overall this is solid work: the request-id correlation held up under every interleaving I could construct (same-party retry → engagement refresh + Duplicate activation; different-party retry → server denial; approval pumped mid-activation → version guard; stale denial → id-scoped cleanup), the break-in flow's ordering assumption is real (entry broadcast and approval reply are both ReliableOrdered on the same channel, and both client handlers funnel through the same `GameThread.RunSafe` FIFO), and the server re-derives eligibility from its own state instead of trusting the request. The E2E choreography here is some of the most thorough we have. `TestNetworkRouter.Deliver` suspending the sender's allowance is a suite-wide semantic change, but it models separate processes more faithfully and all 8 shards passing shows nothing depended on the leak.

Two things that don't fit a line comment:

- CodeFactor flags 6 new issues (grade still A) — probably method length on the new tests, worth a quick look before merge.
- Risk notes, no action needed: (1) `StartBattleInternalPrefix` losing the `IsOriginalAllowed` escape routes *every* client MapEvent creation through the coordinator now, not just the two fixed flows — adoption paths short-circuit via `_mapEvent != null` and CI is green, but this is where a live regression would surface if some rarely-exercised AllowedThread battle-start path exists. (2) Approved restarts now run even while a party is behind the map-event init barrier — intended, and the sally-out test covers it, but worth one eye during the next 2-client manual pass. (3) `break_in_debrief_continue_on_consequence(null)` relies on the native consequence tolerating null args — fine today, same brittleness class as our other named-menu-consequence patches, just noting it for the next game-version bump.

source/E2E.Tests/Services/MapEvents/PlayerPartyInteractionFlowTests.cs
ShoT-UPfps: fixed both by clearing the game-thread mark in `E2ETestEnvironment.Dispose()` after server/client teardown. async continuations stay marked for the fixture lifetime and every mark is still paired.

source/Coop.Core/Client/Services/SiegeEvents/Handlers/ClientSiegeEntryHandler.cs
ShoT-UPfps: fixed this in `ClientSiegeEntryHandler.Dispose()`, which restores the staged `LocationEncounter` during disconnect teardown. added a regression test for disposal before approval.

source/GameInterface/Services/MapEvents/Handlers/ConversationRequestHandler.cs
ShoT-UPfps: added the request-id release comment here.

source/GameInterface/Services/MapEvents/Handlers/ConversationRequestHandler.cs
ShoT-UPfps: added the game-thread-only invariant above the client request state.

source/E2E.Tests/Services/MapEvents/PlayerPartyInteractionFlowTests.cs
ShoT-UPfps: fixed with the same fixture-lifetime cleanup.

ShoT-UPfps:
resolved all review threads and the merge conflict. the CodeFactor issue in this PR is fixed; the four remaining notices are in `source/GameInterface/Services/MapEvents/Commands/MapEventDebugCommands.cs`, which has no diff against `development`. github build, unit tests, and all 8 E2E shards are green.

source/Coop.Core/Client/Services/SiegeEvents/Handlers/ClientSiegeEntryHandler.cs
```diff
@@ -44,6 +48,95 @@ public ClientSiegeEntryHandler(IMessageBroker messageBroker, INetwork network, I
...
```
garrettluskey: If the approval reply never arrives (server throws before the RunSafe enqueue, teardown race, whatever), this early return silently eats every break-in click for the rest of the session — the exact softlock class this PR is fixing. Nothing clears the pending state except a reply or Dispose.

Give it an expiry: allow a new attempt after a few seconds and clear/end the stale request id. At minimum log when a click gets swallowed here so it's diagnosable.

source/GameInterface/Services/MapEvents/Handlers/ConversationRequestHandler.cs
```diff
@@ -44,28 +44,51 @@ internal class ConversationRequestHandler : IHandler
...
     private readonly IPlayerManager playerManager;
+    private readonly object pvpInteractionSync = new object();
 
+    // Client request state is game-thread-only; patch publishers run there and receive handlers use GameThread.RunSafe.
     private DateTime lastRequestSentUtc = DateTime.MinValue;
+    private string pendingConversationRequestId;
+    private string activeConversationRequestId;
+    private bool hasActiveConversationRequest;
+    private long conversationActivationVersion;
+    private bool isApplyingApprovedRestart;
```
garrettluskey: Make this a depth counter, not a bool. If an Apply-path approval ever gets pumped inside another restart, the nested finally resets this to false while the outer restart is still running — then a ConversationEnded published later in the outer restart slips past the guard and releases the hold that was just activated.

Consume makes that near-unreachable today (a nested same-party approval comes back Duplicate, which doesn't touch the flag), but you already solved this exact problem with approvedBreakInContinuationDepth in SiegeEntryFlowPatches. Do the same here.

source/GameInterface/Services/MapEvents/Handlers/ConversationRequestHandler.cs
```diff
@@ -444,14 +492,26 @@ private static bool IsInSiege(MobileParty party)
...
+                return;
+
+            if (requireRequestIdMatch && interaction.RequestId != requestId)
+                return;
+
+            if (!pvpDefenderByAttacker.TryRemove(attackerPeer, out interaction))
+                return;
+
+            conversationPartyTracker.EndPvpConversation(interaction.DefenderPartyId);
+            network.SendAll(new NetworkPlayerInteractionEnded(interaction.DefenderPartyId));
```
garrettluskey: This holds pvpInteractionSync across network.SendAll (same in NotifyPvpInteractionStarted), so the game thread and the poller thread are now coupled through network I/O. LiteNetLib sends just enqueue so it works today, but capture what you need inside the lock and send after releasing it. Keep the begin-before-broadcast ordering — that part is right and the comment explains it well.

source/Coop.Core/Server/Services/SiegeEvents/Handlers/ServerSiegeEntryHandler.cs
```diff
@@ -54,6 +61,109 @@ public ServerSiegeEntryHandler(
...
+        bool alreadyEntered)
+    {
+        if (!party.IsActive) return false;
+        if (party.CurrentSettlement != null && !alreadyEntered) return false;
+        if (party.BesiegerCamp != null) return false;
+
+        var mapEventSide = party.Party?.MapEventSide;
+        var validEnteredMapEvent = alreadyEntered &&
+            ReferenceEquals(party.MapEvent, settlement.Party?.MapEvent) &&
+            party.Party.Side == BattleSideEnum.Defender;
```
garrettluskey: party.Party is read with ?. a few lines up and then dereferenced bare here. It can't actually be null on a MobileParty, so pick a stance — drop the ?. above or guard both. As written it reads like the null case was half-considered.

garrettluskey (dismissed):
The core design is right — authoritative entry replicated before the approval, request-scoped holds, and the test coverage is genuinely thorough. Requesting changes for the pending-request softlock below; the rest are smaller but worth fixing while you're in here.

source/GameInterface/Services/MapEvents/Handlers/ConversationRequestHandler.cs
ShoT-UPfps: queued `Started` and `Ended` under `pvpInteractionSync` and drain them through one sender after releasing it. this keeps begin-before-broadcast and preserves transition order when an end races or reenters a start. added a forced-interleaving regression.

source/GameInterface/Services/MapEvents/Handlers/ConversationRequestHandler.cs
ShoT-UPfps: changed this to a depth counter and added a nested approval regression so the outer restart keeps the guard active.

source/Coop.Core/Client/Services/SiegeEvents/Handlers/ClientSiegeEntryHandler.cs
ShoT-UPfps: added an expiry using `ObjectCreationTimeout`. a retry restores the staged location state and gets a fresh request id, and stale replies are ignored. matching rejection also exits the debrief and holds the party so it cant reopen the encounter.

source/Coop.Core/Server/Services/SiegeEvents/Handlers/ServerSiegeEntryHandler.cs
ShoT-UPfps: cached `party.Party` once and use it consistently without the nullable access.

source/Coop.Core/Client/Services/SiegeEvents/Handlers/ClientSiegeEntryHandler.cs
```diff
@@ -25,13 +28,22 @@ internal class ClientSiegeEntryHandler : IHandler
     private readonly INetwork network;
     private readonly IObjectManager objectManager;
     private readonly ISiegeEventInterface siegeEventInterface;
+    private PendingBreakInContinuation pendingBreakInContinuation;
```
garrettluskey, line 31: Nit: this field has the same game-thread-only invariant as the conversation request state (the attempt publisher is a patch prefix on the game thread, the approval handler hops through `RunSafe`, and `Dispose` touches it during teardown) — worth the same one-line comment `ConversationRequestHandler` got, so the next reader doesn't have to re-derive it.

garrettluskey (approved):
Round three — I re-verified each claimed fix against the head rather than taking the thread replies on faith, and they all hold:

- The fixture teardown unmark works from any thread (`UnmarkGameThread` just zeroes the marked id), so async continuations can no longer leave a recycled thread id promoted past the fixture's lifetime.
- The break-in pending state got the full treatment: 5s expiry (`ObjectCreationTimeout`), retry restores the staged `LocationEncounter` and mints a fresh request id, stale replies are id-filtered, and a matching rejection exits the debrief via `FinishLocalPlayerSiegeLeave` — with the hold/`TargetParty` cleanup pinned by the timeout test.
- `approvedRestartDepth` is an int with `++`/`--` in `finally`, matching the `SiegeEntryFlowPatches` pattern.
- The PvP Started/Ended notifications queue under `pvpInteractionSync` and drain through a single sender outside the lock; begin-before-broadcast survived the rework and the forced-interleaving test pins the order. Lock ordering is one-directional into the tracker, so no inversion risk.
- `partyBase` is cached once in `CanApplyBreakInContinuation`, no more mixed null-conditional access.
- `Dispose` restores the staged `LocationEncounter`, with the disposal-before-approval regression covering it.

I also re-swept the new logic for fresh holes and came up empty. The activation-version guard converges under every approval interleaving I could construct (FIFO delivery, pumped-during-restart, evicted context falling back to Duplicate). The server-detected null-request-id refresh can't strand a hold, because the client's Duplicate activation moves its active id to null in lockstep, so the eventual `ConversationEnded(null)` still matches. And `PartyEnterSettlement`'s besieger early-return can't silently no-op the break-in: `BesiegerCamp != null` is rejected before it, and the return value re-checks `CurrentSettlement` so a silent skip becomes a rejection the client recovers from.

CodeFactor is still red, but the remaining notices are in `MapEventDebugCommands.cs`, which this PR doesn't touch — pre-existing, not yours to fix here. One optional nit inline; take it or leave it.

PR #2502, opened 2026-07-30
[2429] Validate siege entry authoritatively

### LINKED ISSUE #2429: [SIEGES] Rejoining the siege of Amitatys can place player at Castle Garontor
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

Initially I was at Castle Garontor attacking a NPC lord with no players. Upon attempting to enter the battle, a player disconnected and we received a time out connection, however I was somehow able to keep playing alone while Swadian relaunched the server.

## Steps to reproduce

It slid my character across the map at 4.8 move speed north east, my last known movement input. On the way of being slid across the map, I clicked the siege of Amitatys and tried to join the siege and interact with it in hopes to "redirect my character and stop myself from sliding"; upon realizing this did nothing, I left the game and rejoined.

## Additional information

Upon rejoining I was no longer in battle or sliding across the map, but instead in siege with Amitatys while being located at Castle Garontor at the border of the Aserai. Swadian tried to join me and soft locked his game. Swadian typed in the command console: coop.debug.besiegercamp.set_leader_party; he was able to force end the siege event, and it deleted my character from existence but kept the siege tents outside the city. Furthermore it locked his client up completely, and he had to ALT-F4 to get out.

## Discord source

https://discord.com/channels/709516043332354119/1531317328690806966/1531317328690806966

## Images and videos

[](<https://cdn.discordapp.com/attachments/1531317328690806966/1531317329193861281/image.png?ex=6aa4c225&is=6aa370a5&hm=070eac9a01273f095df6cf89e5b7dd39599dd4955a99b16c8e4814a84f7f6a2c&>) ([Discord message](https://discord.com/channels/709516043332354119/1531317328690806966/1531317328690806966))

[](<https://cdn.discordapp.com/attachments/1531317328690806966/1531317329919606875/image.png?ex=6aa4c225&is=6aa370a5&hm=2fa09b4f44202a8a62bb3d7342fc39756b965280de4c478d2c81d5f41020308b&>) ([Discord message](https://discord.com/channels/709516043332354119/1531317328690806966/1531317328690806966))

[](<https://cdn.discordapp.com/attachments/1531317328690806966/1531317330351493280/image.png?ex=6aa4c226&is=6aa370a6&hm=fd59f886f4bbf27af47b8838b16142592528a79fc059fdfb57a02cf1b84bb59b&>) ([Discord message](https://discord.com/channels/709516043332354119/1531317328690806966/1531317328690806966))

[](<https://cdn.discordapp.com/attachments/1531317328690806966/1531317330758602904/image.png?ex=6aa4c226&is=6aa370a6&hm=e50aa5780ab378ab3906379ebb99d43a96df8bf17976561a90f77370352f5d78&>) ([Discord message](https://discord.com/channels/709516043332354119/1531317328690806966/1531317328690806966))

[](<https://cdn.discordapp.com/attachments/1531317328690806966/1531317331442139136/image.png?ex=6aa4c226&is=6aa370a6&hm=129c1b299c03d62391017a3ede076a4e66bb0fd6f3106f570645e0b9a300f90a&>) ([Discord message](https://discord.com/channels/709516043332354119/1531317328690806966/1531317328690806966))

[](<https://cdn.discordapp.com/attachments/1531317328690806966/1531322282423877883/image.png?ex=6aa4c6c2&is=6aa37542&hm=88b4d1ee235917afcd837ecb9602b41715b4a3fd000e7dffa66cbc5b1c9a84a1&>) ([Discord message](https://discord.com/channels/709516043332354119/1531317328690806966/1531322282927198431))

[](<https://cdn.discordapp.com/attachments/1531317328690806966/1531704890932330546/image.png?ex=6aa4d997&is=6aa38817&hm=d08762f43ca88831ab6e0634987a144184fd93b9a593d6c46186e9799ffc7b29&>) ([Discord message](https://discord.com/channels/709516043332354119/1531317328690806966/1531704890915684485))

## Logs

Client log: [Download message.txt](<https://cdn.discordapp.com/attachments/1531317328690806966/1531318678774091806/message.txt?ex=6aa4c367&is=6aa371e7&hm=6842abcda4694941e2bc84e8cf90248cf36d92906466032be5d58b641953984b&>) · [0.0.3+c2cbc01308b04a574d08c42a79c8a2b8f1dd7646 - Download message.txt](<https://cdn.discordapp.com/attachments/1531317328690806966/1531705336845832242/message.txt?ex=6aa4da02&is=6aa38882&hm=7ecaf2e3e70a3bf18c4ba22cc1c1a430f2a111822d13797aec6fa6ab6b26dc83&>)

Server log: Not provided

## Additional report details

- On the latest patch (`Coop 07-27-2026.7z`), a second occurrence happened while the player controlled a settlement under enemy siege: attempting to join the siege event soft-locked them into besieging their own settlement, and trying to set themselves as siege leader did not advance the camp and kept them stuck in their own settlement instead of joining the defense. ([Discord message](https://discord.com/channels/709516043332354119/1531317328690806966/1531698267392577546))

- Attempting to break into a siege can soft-lock the player after their troops are lost, showing only a `continue` option; selecting it does not move the player to defend the town and no further prompt appears. ([Discord message](https://discord.com/channels/709516043332354119/1531317328690806966/1531705082964349009))

- While in the siege-event menu (options to assault camp, break in to help defenders, or leave), if the besieging party completes the siege, the client crashes and freezes. ([Discord message](https://discord.com/channels/709516043332354119/1531317328690806966/1531706260611666060))

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

A stale settlement encounter can be reused after the player's party has moved, changed siege camp, or become the settlement defender.

## What is the new behavior?

Resolves #2429

Siege interaction requests now use the player's current party, settlement, siege camp, battle, faction, and ownership state. Invalid or stale requests return the player to the correct map or encounter state.

## Bot Changelog Entry

- Fixed stale siege interactions putting players into the wrong siege or into an attack against their own settlement.


source/Coop.Core/Server/Services/MobileParties/Handlers/ServerSettlementExitEnterHandler.cs
```diff
@@ -65,12 +67,12 @@ private void Handle(MessagePayload<NetworkRequestStartSettlementEncounter> obj)
         {
             if (!objectManager.TryGetObjectWithLogging(payload.PartyId, out MobileParty mobileParty))
             {
-                network.Send(peer, new NetworkSettlementEncounterRejected(payload));
+                RejectSettlementEncounter(peer, payload, "your party is no longer available");
```
garrettluskey, line 83: The siege side of this PR now rejects a request whose peer doesn't own the party, but this handler still trusts `payload.PartyId` as-is: a stale or forged request can start an encounter for another player's party (or an AI lord's), and `PartyEnterSettlement` applies it authoritatively and broadcasts it. That's the same stale-id class #2429 is about. The client-side publisher (EncounterManagerPatches) only ever sends the local player's own party, so the guard can't break a legit flow. Take `IPlayerManager` here too and mirror the siege check before these lookups:

```csharp
if (!playerManager.TryGetPlayer(peer, out var player) || player.MobilePartyId != payload.PartyId)
{
    RejectSettlementEncounter(peer, payload, "your party is not controlled by you");
    return;
}
```

The end-encounter request below has the same hole, worth closing while you're in here.

source/Coop.Core/Server/Services/SiegeEvents/Handlers/ServerSiegeEntryHandler.cs
```diff
@@ -169,45 +172,203 @@ private void HandleBesiege(MessagePayload<NetworkRequestBesiegeSettlement> paylo
...
+
+        if (!party.IsActive || party.Party == null)
+            rejectionReason = "your party is inactive";
+        else if (settlement.Party == null || !settlement.IsFortification)
+            rejectionReason = "the target is not a fortification";
+        else if (party.MapEvent != null)
+            rejectionReason = "your party is already in a map event";
+        else if (party.CurrentSettlement != null && party.CurrentSettlement != settlement)
+            rejectionReason = "your party is inside another settlement";
+        else if (party.BesiegerCamp != null)
```
garrettluskey, line 416: A repeat Besiege of the settlement the party is already besieging lands here and gets "your party is already in another siege camp" — "another" is wrong in that case, and the outcome is asymmetric with Join, which short-circuits to an approve when the party is already in the target camp. Extend the short-circuit above to Besiege (the client continuation just re-seats the preparation menu, which is what a retry wants), or at least special-case the reason text when `party.BesiegerCamp` is the target settlement's own camp.

source/Coop.Core/Server/Services/SiegeEvents/Handlers/ServerSiegeEntryHandler.cs
```diff
@@ -169,45 +172,203 @@ private void HandleBesiege(MessagePayload<NetworkRequestBesiegeSettlement> paylo
...
+            rejectionReason = "your party cannot begin this siege";
+        else if (action == SiegeEntryAction.Join &&
+            (settlement.SiegeEvent == null ||
+            !settlement.SiegeEvent.CanPartyJoinSide(party.Party, BattleSideEnum.Attacker)))
+            rejectionReason = "your party cannot join the attacking side";
+
+        return rejectionReason == null;
+    }
+
+    private static bool IsWithinInteractionDistance(MobileParty party, Settlement settlement)
```
garrettluskey: This is a line-for-line twin of the block added in ServerSettlementExitEnterHandler, and the two copies already disagree on the fallback: there a null `EncounterModel` skips the check, here it rejects — and the player would see "too far from the settlement" for a reason that has nothing to do with distance. Extract one shared helper for the encounter-distance rule so the port/blockade logic and the slack can't drift apart, and give the `0.5f` a name that says what it's for (absorbing position-sync drift).

garrettluskey (dismissed):
This is the right shape: entry requests are finally validated against the server's own party/settlement state, the rules match vanilla's encounter distances (ports and blockades included), and rejections now always answer the requester — the old TryGet failures just went silent and left the client waiting at the menu. The E2E coverage of both the reject and approve paths is thorough.

Requesting changes mainly for one gap: the settlement-encounter request still trusts the client-supplied party id, while the siege side now enforces peer ownership. Details inline, plus two smaller consistency items.

One thing on the description: "reconnecting clears invalid siege links instead of rebuilding the wrong siege" isn't in this diff — nothing here touches the reload/reconnect path — and the bot changelog entry repeats it. If that part is a follow-up PR, trim it from this description so the changelog doesn't announce something that didn't ship.

source/Coop.Core/Server/Services/MobileParties/Handlers/ServerSettlementExitEnterHandler.cs
ShoT-UPfps: fixed in `8f28a4fe`. start and end settlement encounter requests now verify the peer controls the party before any tracker or world mutation, and rejected leave returns a terminal `Suppressed` result.

source/Coop.Core/Server/Services/SiegeEvents/Handlers/ServerSiegeEntryHandler.cs
ShoT-UPfps: fixed in `8f28a4fe`. repeating `Besiege` against the current target camp now returns approved and keeps the existing siege.

source/Coop.Core/Server/Services/SiegeEvents/Handlers/ServerSiegeEntryHandler.cs
ShoT-UPfps: fixed in `8f28a4fe`. both handlers now use the injected transient `ISettlementEncounterDistanceValidator`, the position-sync slack is named, and a missing encounter model gets an accurate rejection reason.

garrettluskey (approved):
All three items from the last round check out.

The ownership guard now covers both the start and end settlement-encounter requests, runs before any tracker consume or world mutation, and the rejected leave's terminal `Suppressed` result unwinds cleanly on the client (a foreign requester's pending state never matches, so it's a no-op there). The kingdom-creation suppression test correctly moved to the new always-answer contract. The repeat-Besiege path now short-circuits to an approve, and the new E2E test pinning the original SiegeEvent instance is exactly the regression coverage that case needed. The distance rule lives in one validator with the slack named and the missing-model fallback rejecting with an honest reason; swapping in the allow-all stub for the headless integration environment while E2E exercises the real one is the right split.

Also confirmed the eager `(NetPeer)obj.Who` cast in the end-encounter handler is safe — the only production sender is the client's network send, so the sender is always a peer.

One heads-up: the CodeFactor check is red but its report page lists no issues for this PR, so there's nothing actionable I can point at — looks like a stale or flaky status rather than a real finding.

Thanks for the thorough turnaround, including the description trim.

PR #2503, opened 2026-07-30
Recover army wait during siege transitions

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Joining an AI army as it begins a siege can leave the player on an incomplete army-wait screen with no usable menu options or map-event screen.

## What is the new behavior?

The army-wait flow now waits for a complete replicated siege state, enters siege preparation once that state is ready, and returns the player to the correct encounter menu when the army moves into another activity.

## Bot Changelog Entry

- Fixed players becoming stuck without menu options when an AI army begins a siege.


source/GameInterface/Services/Armies/Patches/PlayerArmyWaitBehaviorPatches.cs
```diff
@@ -1,28 +1,172 @@
...
+
+        var liveCamp = siegeEvent.BesiegerCamp;
+        if (liveCamp == null ||
+            mainPartyCamp != liveCamp ||
+            leaderCamp != liveCamp ||
+            liveCamp.SiegeEvent != siegeEvent ||
+            siegeEvent.BesiegedSettlement != targetSettlement ||
+            leaderParty.BesiegedSettlement != targetSettlement)
+        {
+            return AttachedArmySiegeState.Incomplete;
```
garrettluskey, line 162: This branch classifies every siege at the target settlement as our own siege mid-replication, but both camps being null while `liveCamp` exists is a persistent legitimate state, not a sync window: a Defender army relieving a besieged friendly settlement, or a Besieger army whose target another enemy faction reached first (vanilla models that exact case in `Army.IsAnotherEnemyBesiegingTarget`). In those states this returns `Incomplete` for as long as the foreign siege lasts — days of game time — and `ArmyWaitMenuTickPrefix` bails before the generic-state transition, so the member's menus freeze: the switch to `army_wait_at_settlement` when the leader waits at a town never fires, and neither do the siege-defense/encounter transitions once the army reaches the defended settlement. Nothing else on the client recovers these menus; army_wait handling lives only in this file.

When neither `mainPartyCamp` nor `leaderCamp` is set, none of our siege membership is in flight — that's someone else's siege and should read as `None`. The transient windows this branch is protecting all have at least one camp set:

```csharp
if (mainPartyCamp == null && leaderCamp == null)
    return AttachedArmySiegeState.None;
```

before the `liveCamp` comparison block.

source/GameInterface/Services/Armies/Patches/PlayerArmyWaitBehaviorPatches.cs
```diff
@@ -1,28 +1,172 @@
...
+        if (GetAttachedArmySiegeState(MobileParty.MainParty, out var settlement) != AttachedArmySiegeState.Ready)
+            return false;
+
+        StartAttachedArmySiege(settlement);
+        return true;
+    }
+
+    private static void StartAttachedArmySiege(Settlement settlement)
+    {
+        PlayerSiege.StartPlayerSiege(BattleSideEnum.Attacker, isSimulation: false, settlement);
```
garrettluskey: Both existing client-side call sites of this pair run it under `AllowedThread` — `ClientSiegeEntryHandler.HandleJoinApproved` (via `StartLocalPlayerJoinedSiege`) and `SiegeEventInterface.RestoreReloadedPlayerBesieging`. Here it runs bare from the tick prefixes, so any coop patch gate that checks `AllowedThread` inside the `OnPlayerSiegeStarted` handlers or the `menu_siege_strategies` activation silently no-ops on this path only. Wrap the body in `using (new AllowedThread())` to match the other entry points.

source/GameInterface/Services/Armies/Patches/PlayerArmyWaitBehaviorPatches.cs
```diff
@@ -1,28 +1,172 @@
...
 using TaleWorlds.CampaignSystem.Encounters;
 using TaleWorlds.CampaignSystem.GameMenus;
 using TaleWorlds.CampaignSystem.Party;
 using TaleWorlds.CampaignSystem.Settlements;
+using TaleWorlds.CampaignSystem.Siege;
+using TaleWorlds.Core;
 
 namespace GameInterface.Services.Armies.Patches;
 
+internal enum AttachedArmySiegeState
```
garrettluskey, line 21: Nit: new type with no doc comment, and the class summary below still only describes the leave/abandon relay — worth a line each now that this file also owns the siege-transition recovery.

garrettluskey (dismissed):
This lands the right shape. Vanilla's `PlayerArmyWaitBehavior.OnTick` is nothing but the attached-army siege kick, and it dereferences `AttachedTo.Army.AiBehaviorObject` unguarded — running it against a half-replicated army graph is exactly what broke the wait screen. Suppressing it and re-driving the same transition behind a full-graph check is the correct move, and gating the vanilla menu tick on stable army wiring also covers the unguarded `Army.LeaderParty` dereference at the end of vanilla's `ArmyWaitMenuTick`. The skip-constructed fixtures hitting the state machine directly are a nice touch, and this doesn't overlap the existing `ClientSiegeEntryHandler` flows, which only cover player-initiated entry.

Requesting changes for one real gap: the state machine treats any siege on the army's target settlement as our own siege mid-replication. A siege that belongs to someone else — a defender army relieving a besieged friendly settlement, or a besieger army whose target another enemy faction reached first — parks the client in `Incomplete` for as long as that siege lasts, which kills every army_wait menu transition. Details inline.

The red CodeFactor check is a single Complex Method notice on `GetAttachedArmySiegeState`; splitting the camp-coherence checks into a helper would clear it, but I'm not blocking on that.

source/GameInterface/Services/Armies/Patches/PlayerArmyWaitBehaviorPatches.cs
ShoT-UPfps: updated the enum and patch class summaries.

source/GameInterface/Services/Armies/Patches/PlayerArmyWaitBehaviorPatches.cs
ShoT-UPfps: wrapped both local siege calls in AllowedThread to match the other client entry paths.

source/GameInterface/Services/Armies/Patches/PlayerArmyWaitBehaviorPatches.cs
ShoT-UPfps: fixed, foreign sieges now return None when neither army party has a besieger camp. added coverage for that state.

garrettluskey (approved):
All three items from the last round are in and correct:

- Foreign sieges no longer read as ours mid-replication: the both-camps-null guard returns `None` before the live-camp comparison, so a defender army relieving a besieged settlement (or a besieger whose target another enemy reached first) falls through to the generic-state transition instead of freezing `army_wait`. The new `ForeignSiegeAtTarget` test pins exactly that shape — live siege at the target, both our camps null.
- Both siege-start paths now run under `AllowedThread` via `StartAttachedArmySiege`, matching the join-approved and rejoin-restore entry points.
- Enum and class summaries updated.

I also rechecked the one window the new guard changes: our own siege where `settlement.SiegeEvent` replicates before either camp assignment now reads `None` for a tick instead of `Incomplete`. That just lets the vanilla wait tick run against stable army wiring until the camps arrive, then the state machine takes over — no stuck state either way.

Build, unit, and all E2E shards are green. CodeFactor's remaining notice is the same Complex Method count on `GetAttachedArmySiegeState`; still not worth blocking over.

PR #2520, opened 2026-07-31
Host-Coop button spawns the bundled dedicated server

### LINKED ISSUE #2517: Host-Coop Button Spawn Dedicated Server
# For New Contributors

## Description

Clicking the host-coop button should start the dedicated server and fall back to the game instance if that cannot be found.

The dedicated server will sit in the Coop module under the DedicatedServer folder on steam, so we should use that path to try to resolve the Dedicated Server.

## Location

## Definition of Done

garrettluskey:
## PR Checklist

## PR Type

The Host button always spawns a second Bannerlord.exe game instance as the co-op server.

## What is the new behavior?

Resolves #2517

The Host click first resolves `BannerlordCoopServer.exe` from the Coop module's `DedicatedServer` folder (where the Steam bundle nests it) and spawns it with the coop session tokens (`/coopsave`, `/coopowner`, `/coopvisibility`, `/cooppassword`). It falls back to the game-instance server when the dedicated server is not installed, or when extra community modules are active (the dedicated server's pinned module list could not host them).

The dedicated-server side (accepting the tokens, hosting the clicked save from the game's own save folder) lives in the DedicatedServer repo.

## Bot Changelog Entry

- Hosting a co-op game now starts the bundled dedicated server when it is installed, instead of a second game instance.

## Other information

### DIFF
--- a/source/Coop.Core/Common/Session/ServerProcessManager.cs
+++ b/source/Coop.Core/Common/Session/ServerProcessManager.cs
@@ -54,12 +54,25 @@ public class ServerProcessManager : IDisposable
             CleanupLocked();
 
             var currentProcess = Process.GetCurrentProcess();
-            var exePath = ManagedServerLauncher.GetEngineExecutablePath();
-            var arguments = ServerLaunchArguments.BuildManagedServerArguments(
-                ManagedServerLauncher.GetActiveModuleIds(), saveName, currentProcess.Id, password, visibility);
 
             // The arguments may contain the hosted-server password, so never write them to a log.
-            Logger.Information("Spawning co-op server for save '{SaveName}': {Exe}", saveName, exePath);
+            string arguments;
+            var exePath = ResolveDedicatedServerExecutable();
+            if (exePath != null)
+            {
+                arguments = ServerLaunchArguments.BuildDedicatedServerArguments(
+                    saveName, currentProcess.Id, password, visibility);
+                Logger.Information("Spawning bundled dedicated co-op server for save '{SaveName}': {Exe}",
+                    saveName, exePath);
+            }
+            else
+            {
+                exePath = ManagedServerLauncher.GetEngineExecutablePath();
+                arguments = ServerLaunchArguments.BuildManagedServerArguments(
+                    ManagedServerLauncher.GetActiveModuleIds(), saveName, currentProcess.Id, password, visibility);
+                Logger.Information("Spawning co-op game-instance server for save '{SaveName}': {Exe}",
+                    saveName, exePath);
+            }
 
             var process = new Process
             {
@@ -97,6 +110,25 @@ public class ServerProcessManager : IDisposable
         {
             CleanupLocked();
         }
+    }
+
+    /// <summary>
+    /// The dedicated server bundled with the Coop module, when it is installed and can host
+    /// the active module set; null falls the spawn back to a game-instance server.
+    /// </summary>
+    private static string ResolveDedicatedServerExecutable()
+    {
+        var exePath = ManagedServerLauncher.GetDedicatedServerExecutablePath();
+        if (exePath == null) return null;
+
+        if (!ManagedServerLauncher.CanDedicatedServerHostActiveModules())
+        {
+            Logger.Information(
+                "Dedicated server skipped: the active community modules need a game-instance server");
+            return null;
+        }
+
+        return exePath;
     }
 
     private void OnServerExited(object sender, EventArgs e)
--- a/source/GameInterface/Services/GameState/ManagedServerLauncher.cs
+++ b/source/GameInterface/Services/GameState/ManagedServerLauncher.cs
@@ -1,17 +1,31 @@
-﻿using System.Diagnostics;
+using GameInterface.Services.Modules;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using TaleWorlds.ModuleManager;
+using ModuleInfo = GameInterface.Services.Modules.ModuleInfo;
 
 namespace GameInterface.Services.GameState;
 
 /// <summary>
-/// Resolves how to spawn the managed dedicated-server process: the game engine executable
-/// and the active module list, both read from the running game so the spawned server loads
-/// the same engine and mods as the hosting client.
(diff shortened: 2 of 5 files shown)


source/Coop.Core/Common/Session/ServerProcessManager.cs
```diff
@@ -54,12 +54,25 @@ public void Start(string saveName, string password, ServerVisibility visibility)
...
-            var arguments = ServerLaunchArguments.BuildManagedServerArguments(
-                ManagedServerLauncher.GetActiveModuleIds(), saveName, currentProcess.Id, password, visibility);
 
             // The arguments may contain the hosted-server password, so never write them to a log.
-            Logger.Information("Spawning co-op server for save '{SaveName}': {Exe}", saveName, exePath);
+            string arguments;
+            var exePath = ResolveDedicatedServerExecutable();
+            if (exePath != null)
+            {
+                arguments = ServerLaunchArguments.BuildDedicatedServerArguments(
```
ShoT-UPfps, line 63: we need the bundled launcher to accept these `/coop*` arguments before this merges, otherwise every Host click with the bundled server exits with code 2

source/GameInterface/Services/GameState/ManagedServerLauncher.cs
```diff
@@ -23,6 +37,67 @@ public static string GetEngineExecutablePath()
...
+    /// True when every module is loadable by the dedicated server's pinned module list:
+    /// official game modules, this mod itself, and the DedicatedServer.* host modules
+    /// (the same set <see cref="Modules.Validators.ModuleValidator"/> exempts from matching).
+    /// </summary>
+    public static bool CanDedicatedServerHostModules(IEnumerable<ModuleInfo> modules)
+    {
+        if (modules == null) return false;
+
+        return modules.All(module =>
+            module.IsOfficial ||
```
ShoT-UPfps: shouldnt this be `module.IsOfficial && !module.IsDlc`? `IsOfficial` includes DLC, but the pinned dedicated-server module list doesnt, so this can select the bundled server for a module set it cant host

ShoT-UPfps (changes requested):

source/Coop.Core/Common/Session/ServerProcessManager.cs
garrettluskey: The launcher side is now up: Bannerlord-Coop-Team/BannerlordCoop.DedicatedServer#14 — the launcher parses the /coop* tokens, forwards them to the engine child, and CoopServerHost re-parses them there to override server-config.json for the run; the hosted save resolves from the game's own save folder. Agreed on ordering: that PR should merge first, and the workshop bundle must be rebuilt from a launcher build that contains it before this ships. Since bundle-workshop.ps1 nests the dedicated server inside the same workshop item as the module, the two update atomically for users — an old-launcher/new-mod mismatch can only come from bundling in the wrong order.

source/GameInterface/Services/GameState/ManagedServerLauncher.cs
garrettluskey: Good catch — fixed in 49822ef70: the check is now `module.IsOfficial && !module.IsDlc` (matching ModuleValidator.ShouldValidateModule, which the doc comment already claimed to mirror), with a regression test that a DLC-enabled set falls back to the game-instance server.

source/Coop.Core/Common/Session/ServerProcessManager.cs
garrettluskey: BannerlordCoop.DedicatedServer#14 merged, so the bundled launcher now accepts and forwards these arguments.

source/GameInterface/Services/GameState/ManagedServerLauncher.cs
```diff
@@ -1,17 +1,31 @@
...
 public static class ManagedServerLauncher
 {
+    /// <summary>The dedicated-server deployment folder nested inside the Coop module.</summary>
+    public const string DedicatedServerFolderName = "DedicatedServer";
+
+    /// <summary>The dedicated-server launcher executable at that deployment's root.</summary>
+    public const string DedicatedServerExecutableName = "BannerlordCoopServer.exe";
+
+    /// <summary>The id of this mod's module, whose folder carries the dedicated server.</summary>
+    public const string CoopModuleId = "Coop";
```
ShoT-UPfps, line 27: support the nightly module id here. current nightlies rewrite the active id to `CoopNightly`; `ModuleHelper.GetModuleFullPath` indexes loaded modules by that id, so this `Coop` lookup throws and is caught as "no dedicated server", while `CanDedicatedServerHostModules` rejects `CoopNightly` too. a nightly bundle therefore falls back to `Bannerlord.exe` even when the bundled launcher is present. resolve the active stable/nightly Coop module and add a nightly-id test.

ShoT-UPfps (commented):
one P2 correctness issue on `49822ef7`: the hard-coded `Coop` module id makes `CoopNightly` builds skip the bundled launcher and fall back to `Bannerlord.exe`. both prior correctness threads are resolved.

source/GameInterface/Services/GameState/ManagedServerLauncher.cs
garrettluskey: fixed in `44abc20ff`: the bundled path now resolves the active stable or nightly Coop module id, and `CoopNightly` is accepted by the pinned-module check. added regression tests for both cases.

gamingmenacers-svg (commented):
reviewed the exact head and also checked the current `BannerlordCoop.DedicatedServer` launcher contract. the `/coopsave`, `/coopowner`, `/cooppassword`, and `/coopvisibility` arguments match the merged launcher implementation; the stable/nightly module lookup, DLC/community fallback, quoting, and process-exit ordering are covered coherently. i did not find a new code defect in these five files.

this is still a draft 1,560 commits behind `development`, and every hosted build/test job is skipped. before merging, update it onto current `development`, rebuild the workshop bundle from a dedicated-server launcher containing `BannerlordCoop.DedicatedServer#14`, and run the host-click integration on stable + nightly with a stock set, DLC/community fallback, protected save, and all visibility modes. the current code is not merge-ready without that dependency and validation evidence.

PR #2527, opened 2026-07-31
Compile protobuf contracts individually on Wine-Mono

nolanryan489:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

On Wine-Mono the join never completes. Both peers sit on Join Coop Campaign until the tunnel times
out.

protobuf-net compiles a serializer for each contract via Reflection.Emit on first use and Wine-Mono
rejects some of that IL with InvalidProgramException, "Invalid IL code in (wrapper dynamic-method)".
A failed attempt leaves the MetaType resolvable so anything sent repeatedly recovers on its next
send, which is why CampaignTimePacket throws once and then streams normally and the problem is easy
to miss. One-shot handshake messages get no second chance, and NetworkTransferNewHero is published
exactly once from CharacterCreationState.Handle_CharacterCreationFinished.

It stops the whole session rather than one client because the throw aborts
JoinCampaignBaselineSender.Send partway, leaving queued packets with nothing driving them, and
OverloadedPeerManager then pauses campaign time for every peer while it waits for one that can never
catch up.

#2430 handles Wine-Mono by keeping compiled serializers and replacing the uninitialized-object
factory for value types. The factory is not the only thing being rejected. The emitted TypeSerializer
is rejected too, so UseConstructor does not help these contracts.

## What is the new behavior?

Opened at Seras's suggestion from discord said that ShoT_UP has been working on this in internal builds, so this may just be a second data point rather than the fix you want.

Auto-compile stays off on Mono so nothing is emitted implicitly, then each registered contract gets
one CompileInPlace attempt. Contracts the runtime accepts keep compiled serializers, contracts it
rejects stay on the reflection path, and Windows is untouched. This keeps the intent of "Keep
protobuf compiled on Wine-Mono" from #2430 rather than reverting to interpreted everywhere: a runtime
that accepts the IL loses nothing, and one that rejects all of it degrades to what auto-compile off
would have done anyway.

The attempt is deferred until after registration rather than run from AfterApplyDefaultBehaviour.
Compiling a contract also builds serializers for its nested types, which outruns their own
configuration and makes protobuf-net throw "The type cannot be changed once a serializer has been
generated". ConfigureMonoValueType now tolerates that for types whose serializer already exists.

Verified against Mono 6.4.0, the runtime Bannerlord bundles, with contracts mirroring the ones that
failed in game (PartyBehaviorUpdateData, MobilePartyJoinState, RomanceStateData[] and the messages
wrapping them). Stock settings throw Unable to wrap on 6 of 8 contracts on the first attempt. With
this change all pass on the first attempt, 32 contracts compiled and none fell back on that runtime.
Wire output is identical between the compiled and reflection paths in both directions, so this is not
a format change and a patched client still talks to an unpatched one.

In game two Linux players, one CachyOS and one SteamOS, went from a permanent hang to a completed
join with hero transfer, NetworkJoinSync and the usual tournament, mercenary, romance and siege
traffic. No Unable to wrap in either log and no catch-up pauses.

Common.Tests passes, 102/102.

One thing worth flagging separately: the published Workshop build never calls ConfigureRuntimeModel.
Its only call site is CoopMod.NoHarmonyInit from 0285b453 and the shipped Coop.dll predates it, so
Linux players will not see any of this until a build containing that call site goes out.

## Bot Changelog Entry

- Fixed multiplayer joins hanging forever on Linux, Steam Deck and SteamOS.

## Other information

Contracts that do fall back serialize more slowly than compiled ones. That is limited to Mono and
only to contracts the runtime actually rejects, so Windows is unaffected. live two-player traffic
peaked around 440 messages a second, roughly 78 KB/s, with queues draining and no catch-up pauses. I
have not tested 8 players or large battles.

### DIFF
--- a/source/Common.Tests/Serialization/ProtoBufSerializerTests.cs
+++ b/source/Common.Tests/Serialization/ProtoBufSerializerTests.cs
@@ -35,13 +35,15 @@ public class ProtoBufSerializerTests
     }
 
     [Fact]
-    public void ConfigureRuntimeModel_KeepsCompilationAndUsesDefaultInitializationForStructs()
+    public void ConfigureRuntimeModel_DisablesAutoCompileOnMonoAndUsesDefaultInitializationForStructs()
     {
         var model = RuntimeTypeModel.Create();
         ProtoBufSerializer.ConfigureRuntimeModel(model, isMonoRuntime: true);
         var metaType = model.Add(typeof(SkipConstructorStruct), applyDefaultBehaviour: true);
 
-        Assert.True(model.AutoCompile);
+        // Auto-compile is off on Mono so no contract is emitted implicitly; compilation is then
+        // attempted per type, and a runtime that rejects the IL simply keeps the reflection path.
+        Assert.False(model.AutoCompile);
         Assert.True(metaType.UseConstructor);
 
         var expected = new SkipConstructorStruct(42, "linux");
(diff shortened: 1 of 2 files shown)


ShoT-UPfps:
I don't think this is needed, the serializer IL is already updated so that winemono should be able to run it, it's just not released yet.

If you still have this issue on a nightly build or the next release build we can revisit it. But for now I would close the PR.

Appreciate the initiative!

PR #2531, opened 2026-07-31
Fix food consumption refresh

jordanbrymora:
## PR Checklist

## PR Type

Starving player parties buying food don't update their starving status until the next daily tick. This doesn't match vanilla and was originally implemented this way to avoid tying the campaign behavior's version number to the player parties.

## What is the new behavior?

Created a dictionary to store last checked roster versions to keep the check in line with vanilla. Now when a starving player party gets food morale penalties etc. will update much sooner.

### DIFF
--- a/source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
+++ b/source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
@@ -5,7 +5,6 @@ using GameInterface.Services.MobileParties.Extensions;
 using GameInterface.Services.MobileParties.Interfaces;
 using GameInterface.Services.Players;
 using HarmonyLib;
-using System;
 using System.Collections.Generic;
 using TaleWorlds.CampaignSystem;
 using TaleWorlds.CampaignSystem.CampaignBehaviors;
@@ -93,6 +92,7 @@ internal class FoodConsumptionBehaviorPatches
     {
         if (!ContainerProvider.TryResolve<IFoodConsumptionBehaviorInterface>(out var foodConsumptionBehaviorInterface)) return false;
 
+        // Custom implementation to handle IsMainParty -> IsPlayerParty replacements and client notifications
         foodConsumptionBehaviorInterface.PartyConsumeFood(__instance, mobileParty, starvingCheck);
 
         return false;
@@ -104,6 +104,7 @@ internal class FoodConsumptionBehaviorPatches
     {
         if (!ContainerProvider.TryResolve<IFoodConsumptionBehaviorInterface>(out var foodConsumptionBehaviorInterface)) return false;
 
+        // Custom implementation to handle IsMainParty -> IsPlayerParty replacement and client notification
         foodConsumptionBehaviorInterface.CheckAnimalBreeding(__instance, party);
 
         return false;
--- a/source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
+++ b/source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
@@ -66,7 +66,7 @@ internal class FoodConsumptionBehaviorPatches
             var playerPartyId = player.MobilePartyId;
             if (!objectManager.TryGetObjectWithLogging<MobileParty>(playerPartyId, out var playerParty)) continue;
 
-            if (!ShouldTickFoodChange(playerManager, playerParty)) return false;
+            if (!ShouldTickFoodChange(playerManager, playerParty)) continue;
 
             int versionNo = playerParty.Party.ItemRoster.VersionNo;
 
--- a/source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
+++ b/source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
@@ -3,6 +3,7 @@ using GameInterface.Extentions;
 using GameInterface.Services.MapEvents.Patches;
 using GameInterface.Services.MobileParties.Extensions;
 using GameInterface.Services.MobileParties.Interfaces;
+using GameInterface.Services.ObjectManager;
 using GameInterface.Services.Players;
 using HarmonyLib;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
```diff
@@ -50,18 +52,37 @@ public static bool DailyTickPartyPrefix(FoodConsumptionBehavior __instance, Mobi
...
+                playerPartyLastItemVersions[playerParty] = versionNo;
+            }
+
+            if (playerParty.Party.IsStarving)
+            {
+                if (playerPartyLastItemVersions[playerParty] != versionNo)
+                {
+                    playerPartyLastItemVersions[playerParty] = versionNo;
+
+                    __instance.PartyConsumeFood(playerParty, true);
```
ShoT-UPfps, line 87: this needs the same map event guard as `DailyTickPartyPrefix`, otherwise an allowed consume that changes `VersionNo` while the party is still starving will call `PartyConsumeFood` again on the next tick.

source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
```diff
@@ -50,18 +52,37 @@ public static bool DailyTickPartyPrefix(FoodConsumptionBehavior __instance, Mobi
...
 
+    private static Dictionary<MobileParty, int> playerPartyLastItemVersions = new();
+
     [HarmonyPatch(nameof(FoodConsumptionBehavior.OnTick))]
     [HarmonyPrefix]
     public static bool OnTickPrefix(FoodConsumptionBehavior __instance, float dt)
     {
-        // Moved to be part of the daily tick.
-        // Avoids tying the server hero's item roster version number to the behavior's version number 
+        foreach (var playerParty in Campaign.Current.CampaignObjectManager.GetPlayerMobileParties())
```
ShoT-UPfps: shouldnt we skip disconnected parties here too? otherwise any roster change can call `PartyConsumeFood` even though `DailyTickPartyPrefix` blocks food consumption for this party.

source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
```diff
@@ -50,18 +52,37 @@ public static bool DailyTickPartyPrefix(FoodConsumptionBehavior __instance, Mobi
...
+        {
+            int versionNo = playerParty.Party.ItemRoster.VersionNo;
+
+            if (!playerPartyLastItemVersions.ContainsKey(playerParty))
+            {
+                playerPartyLastItemVersions[playerParty] = versionNo;
```
ShoT-UPfps: shouldnt a missing entry compare as `-1` first? setting it to `versionNo` here means a newly tracked starving party can never refresh on its first tick.

source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
ShoT-UPfps: `GetPlayerMobileParties()` scans `_mobileParties` and allocates a list, so this now does a full party scan on every campaign tick. can we iterate the registered player parties instead?

source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
```diff
@@ -50,18 +52,37 @@ public static bool DailyTickPartyPrefix(FoodConsumptionBehavior __instance, Mobi
 
         if (!ContainerProvider.TryResolve<IFoodConsumptionBehaviorInterface>(out var foodConsumptionBehaviorInterface)) return false;
 
-        // Custom implementation to move check for starving player parties into daily tick from OnTick
         foodConsumptionBehaviorInterface.DailyTickParty(__instance, party);
 
         return false;
     }
 
+    private static Dictionary<MobileParty, int> playerPartyLastItemVersions = new();
```
ShoT-UPfps: can this state be scoped to the campaign? this static dictionary never removes keys, so it keeps parties from old campaigns alive for the rest of the process.

ShoT-UPfps (changes requested):
cmt

source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
jordanbrymora: Fixed

source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
jordanbrymora: Fixed

source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
jordanbrymora: Fixed

source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
jordanbrymora: Fixed

source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
jordanbrymora: I'm not sure what the best approach for this is. Should I make an extension for `Campaign.Current` somehow?

source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
```diff
@@ -40,28 +43,49 @@ public static bool DailyTickPartyPrefix(FoodConsumptionBehavior __instance, Mobi
...
-        // Avoids tying the server hero's item roster version number to the behavior's version number 
+        if (!ContainerProvider.TryResolve<IPlayerManager>(out var playerManager)) return false;
+        if (!ContainerProvider.TryResolve<IObjectManager>(out var objectManager)) return false;
+
+        foreach (var player in playerManager.Players)
+        {
+            var playerPartyId = player.MobilePartyId;
+            if (!objectManager.TryGetObjectWithLogging<MobileParty>(playerPartyId, out var playerParty)) continue;
+
+            if (!ShouldTickFoodChange(playerManager, playerParty)) return false;
```
ShoT-UPfps: shouldnt this be `continue`? otherwise one disconnected player or player outside the AI join window stops the loop and every player after them misses the food refresh.

jordanbrymora: Fixed

source/GameInterface/Services/MobileParties/Patches/Disable/DisableFoodConsumptionBehavior.cs
ShoT-UPfps: no campaign extension needed, use a `ConditionalWeakTable<FoodConsumptionBehavior, Dictionary<MobileParty, int>>` keyed by `__instance` here. then the dictionary is collected with the campaign behavior instead of living for the rest of the process.

jordanbrymora: Added

ShoT-UPfps (approved):
lgtm

PR #2534, opened 2026-07-31
Fix player army instantly disbanding

NatteTosti69:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Player created armies could instantly disband when not having enough food or not being at war.

Player created armies only check the cohesion of the army to decide if it gets disbanded.

### DIFF
--- a/source/GameInterface/Services/Armies/Patches/ArmyPatches.cs
+++ b/source/GameInterface/Services/Armies/Patches/ArmyPatches.cs
@@ -18,6 +18,7 @@ using TaleWorlds.CampaignSystem.Party;
 using TaleWorlds.CampaignSystem.Settlements;
 using TaleWorlds.Core;
 using TaleWorlds.Library;
+using TaleWorlds.LinQuick;
 using TaleWorlds.Localization;
 namespace GameInterface.Services.Armies.Patches;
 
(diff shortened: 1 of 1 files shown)


source/GameInterface/Services/Armies/Patches/ArmyPatches.cs
```diff
@@ -183,6 +184,62 @@ private static void Soasdoasd(MobileParty __instance, MobileParty value)
...
 
+    /// <summary>
+    /// Replaces MobileParty.MainParty to IsPlayerParty
+    /// else player created armies can get instantly disbanded
+    /// </summary>
+    [HarmonyPatch(typeof(Army), nameof(Army.CheckArmyDispersion))]
+    [HarmonyPrefix]
+    private static bool CheckArmyDispersionPrefix(Army __instance)
+    {
+        if (__instance.LeaderParty.IsPlayerParty())
```
ShoT-UPfps: `Army.HourlyTick` checks `LeaderParty != MobileParty.MainParty` and can disband the army below 30 cohesion before it gets here. we need to replace that comparison with `IsPlayerParty` too, otherwise this `<= 0.1f` check never runs for a client-led army in that case.

source/GameInterface/Services/Armies/Patches/ArmyPatches.cs
```diff
@@ -183,6 +184,62 @@ private static void Soasdoasd(MobileParty __instance, MobileParty value)
         value?.UpdateVersionNo();
     }
 
+    /// <summary>
+    /// Replaces MobileParty.MainParty to IsPlayerParty
+    /// else player created armies can get instantly disbanded
+    /// </summary>
+    [HarmonyPatch(typeof(Army), nameof(Army.CheckArmyDispersion))]
+    [HarmonyPrefix]
+    private static bool CheckArmyDispersionPrefix(Army __instance)
```
ShoT-UPfps: we already have `MainPartyComparisonTranspiler` for this exact replacement. can we use that here instead of copying all of `CheckArmyDispersion`? otherwise any vanilla changes to the other dispersion checks wont run.

ShoT-UPfps (changes requested):
cmt

source/GameInterface/Services/Armies/Patches/ArmyPatches.cs
NatteTosti69: was not aware of this lol, is quite handy tho

ShoT-UPfps (approved):
lgtm

PR #2539, opened 2026-07-31
Reduce large field battle client CPU usage

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Field battles with more than 400 troops can suffer severe client FPS loss because Coop checks every controlled player and AI combatant twice around each mission update.

## What is the new behavior?

Large field battles now limit the early input check to the local player while still processing every authoritative player and AI combatant after the mission update. Player input timing, discrete combat action replication, packet cadence, and network traffic remain unchanged. AI actions first observed during the display tick are sent at the post-native boundary instead of the pre-native boundary; transient sub-frame poses that begin and end inside that window are not sent.

| Aggregate metric | Development | Feature | Change |
| --- | ---: | ---: | ---: |
| Pre-native visits | 3,498,804 | 14,224 | −99.59% |
| Combined action-scan CPU | 4,986.6 ms | 3,338.4 ms | −33.05% |

## Bot Changelog Entry

- Improve client performance in field battles with hundreds of troops without changing combat animations or network traffic.


garrettluskey (commented):
Went through the whole diff. The core change is sound: the pre-native scan now only touches the main player, the post-native pass still diffs every owned agent, and the extracted `PollAgentAction` matches the old loop body line for line, with the per-agent `ClearForLocalAgent` side effect preserved through `ClearLocalAgentStates()`. Mount gait fix and the new E2E coverage look good, CI is green.

Left a handful of inline comments. The two in `MissionActionDiagnostics` (orphaned mount-rewind counters, and the timeline filter turning several `AddEvent` calls into dead code) are the ones I'd like handled before merge since they'd mislead the next perf investigation. The rest are take-it-or-leave-it.

source/Missions/Diagnostics/MissionActionDiagnostics.cs
```diff
@@ -0,0 +1,888 @@
...
+            channel,
+            targetActionIndex,
+            agent.GetCurrentActionType(channel).ToString(),
+            startProgress,
+            sequence,
+            restart,
+            sameAction);
+    }
+
+    public static void RecordMountProgressReplay(
```
garrettluskey: This no longer has any callers — the producer was the `SetCurrentActionProgress` replay in `ApplyMount` that this PR removed. So `mountProgressReplays`, `mountProgressWraps` and `mountProgressRewinds` will always read 0 in the snapshot, which looks like "no rewinds are happening" when it really means "not measured anymore". I'd drop the method and the three counters, or at least pull them out of the snapshot output.

source/Missions/Diagnostics/MissionActionDiagnostics.cs
```diff
@@ -0,0 +1,888 @@
...
+        int actionIndex,
+        string actionType,
+        float progress,
+        long sequence,
+        bool restart,
+        bool sameAction,
+        double durationMilliseconds = 0d,
+        float currentProgress = 0f)
+    {
+        if (!kind.StartsWith("animation-", StringComparison.Ordinal)) return;
```
garrettluskey: This filter quietly drops every kind that doesn't start with `animation-`, which makes the `outbound-update` event in `RecordOutboundAction` and the `source + "-command"` event at the end of `RecordActionCommand` dead code — they never reach the timeline, but their args (`GetCurrentActionType().ToString()` etc.) still get built on every command while tracing. I get that this is the payload bounding, but I'd rather see the dead `AddEvent` calls deleted than filtered out here — the next person who adds an event kind is going to burn time figuring out why it never shows up.

source/Missions/Agents/Handlers/AgentActionHandler.cs
```diff
@@ -132,243 +135,50 @@ public void PollActionsAfterNativeTick()
...
-                agent.GetCurrentActionType(1),
-                ref state.HasAction1DefendingAction,
-                ref state.Action1DefendingAction);
-            if (defendFlags == Agent.MovementControlFlag.None
-                && !AgentActionData.IsGuardMode(guardMode))
+        }
+        else
+        {
+            remoteActionProcessor.ClearLocalAgentStates();
+            Agent mainAgent = mission.MainAgent;
```
garrettluskey, line 165: Worth a sentence in the PR description: with this split, an AI action that game logic realizes between the display tick and the next mission tick goes out one boundary later than before (post-native of the same frame instead of pre-native), and a transient that starts and gets replaced inside that window is never sent at all. I think that's the right trade — it's a sub-frame pose peers shouldn't replay, and the discrete-action latch still catches discrete-to-locomotion transitions — but "attacks, guards, wield changes remain unchanged" is almost-but-not-exactly true. `PollActions_PreNative_EmitsOnlyMainPlayerInput` pins the new contract nicely.

source/Missions/Agents/Handlers/RemoteAgentActionProcessor.cs
```diff
@@ -711,6 +732,12 @@ private RemoteActionApplyResult TryApplyRemoteAction(
                 action.Data,
                 guardActionChannel,
                 previousGuard);
+#if DEBUG
+        using (MissionActionDiagnostics.MeasureRemoteApply(
+                   agentId,
+                   action.Sequence))
+#endif
+        {
```
garrettluskey, line 750: Nit: the body inside these braces kept its old indentation, and in Release the braces compile to a bare block, so it reads like an accident. Indenting `action.Data.Apply(...)` one level under the brace would make the shape obvious in both configs.

source/GameInterface/Services/Party/Commands/LargeBattleRosterFixtureCommands.cs
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        return FormatState(
+            fixture == null ? "none" : "active",
+            firstParty,
+            secondParty);
+    }
+
+    [CommandLineArgumentFunction("large_battle_roster_restore", "coop.debug.mobileparty")]
+    public static string Restore(List<string> args)
```
garrettluskey, line 136: Two edge cases here, since `fixture` is a static that survives session changes: begin a fixture, quit to menu, load a different campaign, and restore will mutate the dead campaign's party objects and report success. Re-resolving the party through the object manager before restoring (and bailing if it's gone) would close that. Also, restore zeroes hero elements too — if a snapshot hero died during the fixture battle you'll re-add a dead hero and the fingerprint still matches. Fine for the intended begin -> battle -> restore flow, just worth a comment so nobody trips on it later.

source/GameInterface/Services/Party/Commands/LargeBattleRosterFixtureCommands.cs
```diff
@@ -0,0 +1,308 @@
...
+            fixtureTroop,
+            addedPerParty);
+
+        return
+            $"LARGE_BATTLE_ROSTER_FIXTURE_STARTED troop={FixtureTroopId} addedPerParty={addedPerParty}\n" +
+            FormatState("active", firstParty, secondParty);
+    }
+
+    [CommandLineArgumentFunction("large_battle_roster_status", "coop.debug.mobileparty")]
+    public static string Status(List<string> args)
```
garrettluskey, line 101: `fixture` is a server-side static, so running this on a client always prints `state=none` even while a fixture is active on the server. Either gate it to the server like begin/restore, or report something like `state=unknown-on-client` so it can't be read as "no fixture".

garrettluskey (dismissed):
Flipping this to request changes for the two MissionActionDiagnostics items from my inline comments above — the orphaned mount-rewind counters and the AddEvent filter that turns several calls into dead code. Both would actively mislead the next perf investigation that reaches for these diagnostics. The other inline comments are optional.

source/Missions/Agents/Handlers/AgentActionHandler.cs
```diff
@@ -381,6 +191,255 @@ private void PollActions(bool afterNativeTick)
...
+#endif
+        if (!actionSyncedAgent) return;
+
+        int action0 = agent.GetCurrentAction(0).Index;
+        int action1 = agent.GetCurrentAction(1).Index;
+        _localAgentStates.TryGetValue(info.AgentId, out var state);
+        bool hadState = state.HasObservation;
+        bool isPlayerControlled =
+            agent.Controller == AgentControllerType.Player;
+        bool retainInputBoundary =
```
garrettluskey, line 223: This can go stale now. Pre-native only refreshes the input boundary for the current `MainAgent`, but this retain path fires for *any* player-controlled agent whose `HasInputBoundaryObservation` is set — and that flag is only cleared when the agent stops being player-controlled. So an agent that stays alive and Player-controlled but stops being `Mission.MainAgent` replays its last captured defend flags/guard mode on every post-native poll, forever. Peers see a shield frozen up, and `CatchUpJoiner` feeds joiners the same stale state.

The old code couldn't hit this because the pre-native loop refreshed these values for every player-controlled agent each frame.

Cheapest fix: also require `agent == Mission.Current.MainAgent` here. Or clear `HasInputBoundaryObservation` when it's consumed post-native — pre-native re-arms it every frame, so either way retention scopes back down to a single frame.

source/Missions/Agents/Handlers/AgentActionHandler.cs
```diff
@@ -132,243 +135,50 @@ public void PollActionsAfterNativeTick()
...
-                action1,
-                agent.GetCurrentActionType(1),
-                ref state.HasAction1DefendingAction,
-                ref state.Action1DefendingAction);
-            if (defendFlags == Agent.MovementControlFlag.None
-                && !AgentActionData.IsGuardMode(guardMode))
+        }
+        else
+        {
+            remoteActionProcessor.ClearLocalAgentStates();
```
garrettluskey, line 164: Nothing in the suite pins this call. If someone drops it, adopted agents keep their retained remote guard state after an authority migration and no test goes red — the new tests only cover what pre-native *doesn't* send. One test that seeds remote state for a local agent, calls `PollActions()`, and asserts it got cleared would close that.

source/Missions/CoopMissionController.cs
```diff
@@ -92,12 +93,16 @@ public override void OnPreDisplayMissionTick(float dt)
         coopMissionComponent.AgentMovementHandler.Interpolator
             .ReplayLookDirections();
 
-        // The prior native Agent tick can realize an action after OnMissionTick sampled its input.
-        // Diff again here so peers receive that exact action instead of retaining the earlier pose.
+        // Native Agent processing realizes authoritative AI and player actions after the input boundary.
+        // Diff them here so peers receive the displayed action instead of retaining an earlier pose.
         coopMissionComponent.AgentActionHandler.PollActionsAfterNativeTick();
```
garrettluskey, line 99: AI action sync now rides entirely on `OnPreDisplayMissionTick`. That holds today because every battle participant renders, but if a headless participant ever runs a live mission, AI action broadcast silently stops. Put that assumption in a comment here or on `PollActionsAfterNativeTick` so nobody moves battle hosting somewhere headless without tripping over it.

source/Missions/Agents/Handlers/RemoteAgentActionProcessor.cs
```diff
@@ -233,6 +237,15 @@ public int GetOutgoingBattleHostEpoch()
         return assignment.Epoch;
     }
 
+    public void ClearLocalAgentStates()
```
garrettluskey, line 240: Heads up: this is still O(all local agents) every pre-native tick, and these iterations don't show up in the pre-native visit counter — so the −99.59% in the PR description measures the expensive sampling only, not the loop itself. Fine by me, but if you want the loop actually gone, iterate `_agentStates` instead (usually tiny — only actually-remote agents plus recent adoptions) and clear the entries that are locally controlled, rather than probing the dictionary once per local agent. Happy for that to be a follow-up.

source/GameInterface/Services/Party/Commands/LargeBattleRosterFixtureCommands.cs
```diff
@@ -0,0 +1,358 @@
...
+                -element.Number,
+                -element.WoundedNumber,
+                0,
+                false);
+        }
+        roster.RemoveZeroCounts();
+
+        foreach (TroopRosterElement element in snapshot.MemberRoster)
+        {
+            // This also restores heroes that died during the fixture battle.
```
garrettluskey: `AddToCounts` doesn't resurrect anyone — if a hero died during the fixture battle this puts a dead hero back into a live roster, which party UI doesn't love. Debug-only so not blocking, but print a warning in the restore output when a captured element's hero is dead so whoever runs this knows the state is off.

garrettluskey (dismissed):
Requesting changes for one thing: the stale input-boundary retention flagged below. It's a narrow window but it's a real desync the old code couldn't hit, and the fix is a one-line guard. Everything else inline is minor or optional.

For what it's worth I diffed the extracted `PollAgentAction` against the old loop body line by line — it's a faithful move. The main-agent gating also matches what the old loop would have selected: `ControllerAgentMap` and `CurrentAuthority` move together under the registry lock, so host migration can't make them diverge. The mount progress-replay removal is a good catch and the new tests pin it down properly.

source/Missions/Diagnostics/MissionActionDiagnostics.cs
ShoT-UPfps: removed `RecordMountProgressReplay` and all three counters from the trace snapshot in `cf6fa7cd1`.

source/Missions/Diagnostics/MissionActionDiagnostics.cs
ShoT-UPfps: deleted the filtered-out `AddEvent` calls and removed the prefix filter in `cf6fa7cd1`.

source/Missions/Agents/Handlers/AgentActionHandler.cs
ShoT-UPfps: updated the PR description with the post-native boundary delay and omitted sub-frame pose behavior.

source/Missions/Agents/Handlers/RemoteAgentActionProcessor.cs
ShoT-UPfps: fixed the indentation in `cf6fa7cd1`.

source/GameInterface/Services/Party/Commands/LargeBattleRosterFixtureCommands.cs
ShoT-UPfps: restore now rejects previous campaigns, re-resolves both parties through the current object manager, and warns when the snapshot contains a dead hero.

source/GameInterface/Services/Party/Commands/LargeBattleRosterFixtureCommands.cs
ShoT-UPfps: gated status to the server in `cf6fa7cd1`.

source/Missions/Agents/Handlers/AgentActionHandler.cs
ShoT-UPfps: fixed in `68a980d9e`: input-boundary retention now requires the current `MainAgent`, with a former-main regression test.

source/Missions/Agents/Handlers/AgentActionHandler.cs
ShoT-UPfps: this is covered by `LocalAuthorityTransfer_ClearsRemoteGuardBeforePolling`; it seeds retained state, transfers local authority, calls `PollActions`, and asserts the guard is cleared.

source/Missions/CoopMissionController.cs
ShoT-UPfps: added the headless participant assumption next to `PollActionsAfterNativeTick` in `68a980d9e`.

source/Missions/Agents/Handlers/RemoteAgentActionProcessor.cs
ShoT-UPfps: changed `ClearLocalAgentStates` to scan `_agentStates` and clear only locally controlled entries in `68a980d9e`.

source/GameInterface/Services/Party/Commands/LargeBattleRosterFixtureCommands.cs
ShoT-UPfps: restore output now warns when the captured roster contains a dead hero in `68a980d9e`.

source/Missions/Agents/Handlers/AgentActionHandler.cs
```diff
@@ -381,6 +191,256 @@ private void PollActions(bool afterNativeTick)
...
+        int action0 = agent.GetCurrentAction(0).Index;
+        int action1 = agent.GetCurrentAction(1).Index;
+        _localAgentStates.TryGetValue(info.AgentId, out var state);
+        bool hadState = state.HasObservation;
+        bool isPlayerControlled =
+            agent.Controller == AgentControllerType.Player;
+        bool retainInputBoundary =
+            afterNativeTick
+            && isPlayerControlled
+            && agent == Mission.Current.MainAgent
```
garrettluskey, line 226: This closes the poll path, but the latch itself survives: a former-main agent that stays Player-controlled hits neither the set branch below (pre-native main only) nor the clear branch (`!isPlayerControlled` only), so `HasInputBoundaryObservation` stays true with frozen values indefinitely. `CatchUpJoiner` still reads it without this guard — a joiner connecting later gets the frozen defend flags/guard mode (shield up), and regular polling never corrects it because the live state produces no further diff.

I'd clear the latch instead of guarding a second call site: make the clear branch `else if (!isPlayerControlled || agent != Mission.Current.MainAgent)`. Then the latch only ever exists on the current main agent and `CatchUpJoiner`'s existing condition becomes safe by construction — nothing to keep in sync. A `CatchUpJoiner` variant of your former-main test would pin it.

garrettluskey (dismissed):
Round two: four of the five are properly closed and I verified each — the `MainAgent` guard plus the former-main regression test, the `_agentStates` scan in `ClearLocalAgentStates` (no mutation-during-enumeration, `ClearForLocalAgent(id, null)` degrades fine, same selection criterion as the old loop), the headless comment, and the dead-hero warning. You're also right about `LocalAuthorityTransfer_ClearsRemoteGuardBeforePolling` — I missed it, that's exactly the coverage I asked for.

Still requesting changes for one leftover from the retention fix, inline below. Same class of bug, same size of fix.

source/Missions/Agents/Handlers/AgentActionHandler.cs
ShoT-UPfps: cleared the latch for every non-main player in `33cbbfc3f` and extended the former-main regression test through `CatchUpJoiner`.

garrettluskey (approved):
Checked `33cbbfc3f` — the widened clear branch is exactly right: the latch can no longer outlive being the current main, which makes `CatchUpJoiner` safe by construction rather than by a second guard. I also traced the extended regression test both ways: with the latch alive the frozen guard flows to the joiner and `Assert.Empty(DirectPacketSends)` fails, with the fix the former main reads as None and gets skipped. Current-main flow is untouched (set branch wins pre-native, neither clear condition fires for the live main post-native).

That closes everything from both rounds. The poll restructure held up well under scrutiny — nice work.

PR #2546, opened 2026-07-31
Fix campaign time freezing permanently on the dedicated server

### LINKED ISSUE #2540: [TIME CONTROL] Game time freezes and requires a server restart
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

Game time can freeze while players are performing different activities. Pressing pause and play, or rejoining the server, does not resume time; a server restart is required.

## Steps to reproduce

Not provided

## Additional information

Game version: 1.4.7 (latest build).

## Discord source

https://discord.com/channels/709516043332354119/1532834941573005353/1532834941573005353

## Logs

Client log: [0.0.3+d238190b9ce7b38153320525145f651376c6dadd - Download GregLog.log](<https://cdn.discordapp.com/attachments/1532834941573005353/1532836122894270564/GregLog.log?ex=6aa459e2&is=6aa30862&hm=d3567f1fd57c5d53b2492cf64abbb2734cc4cfc370ad535f6dfe0c0d6d49bbd4&>) · [0.0.3+d238190b9ce7b38153320525145f651376c6dadd - Download JayLog.log](<https://cdn.discordapp.com/attachments/1532834941573005353/1532836123200458812/JayLog.log?ex=6aa459e2&is=6aa30862&hm=1b501d4fe23a4a9a2c0ade24e8c3cb28271a1c9d46a7dfec9ad60592109d6713&>) · [0.0.3+d238190b9ce7b38153320525145f651376c6dadd - Download KorrLog.log](<https://cdn.discordapp.com/attachments/1532834941573005353/1532836123666022430/KorrLog.log?ex=6aa459e2&is=6aa30862&hm=dacd0e559fc5c81a60644653b8929ea74a17cd2986def197393ab70ac298433a&>)

Server log: [Download coop-server-20260731-191512.log](<https://cdn.discordapp.com/attachments/1532834941573005353/1532836122567245835/coop-server-20260731-191512.log?ex=6aa459e2&is=6aa30862&hm=3d3a255454c885ee0c479f4fc02ca2454b8180984897fb9b928e500af403e448&>)

## Additional report details

- The time freeze occurred at about 20:19. ([Discord message](https://discord.com/channels/709516043332354119/1532834941573005353/1532836123812827207))

garrettluskey:
Fixes #2540 — campaign time freezes on the server and only a restart brings it back.

## Root cause

`MapState.OnTick` (native) short-circuits the entire campaign tick while a save is pending:

```csharp
protected override void OnTick(float dt)
{
    base.OnTick(dt);
    if (Campaign.Current.SaveHandler.IsSaving)
    {
        Campaign.Current.SaveHandler.SaveTick();
        return;                     // ← OnMapModeTick, and so Campaign.RealTick, never run
    }
```

`SaveHandler.IsSaving => !SaveArgsQueue.IsEmpty()`, and the queue is dequeued in exactly one place — `SaveHandler.OnSaveCompleted`, the callback handed to `MBSaveLoad.AutoSaveCurrentGame`. `Game.SaveAux` parks that callback in **`Game._currentActiveSaveData`, a single-slot field**, drained by `Game.OnTick` once the async write finishes:

```csharp
SaveOutput saveOutput = SaveManager.Save(this, metaData, saveName, driver);
if (!saveOutput.IsContinuing) OnSaveCompleted(saveOutput, onSaveCompleted);
else _currentActiveSaveData = new Tuple<...>(saveOutput, onSaveCompleted);   // single slot
```

Start a second `Game.Current.Save` while the first async write is still in flight and the slot is overwritten — the first save's callback is lost forever. `SaveArgsQueue` never drains, `IsSaving` latches `true`, and the campaign never ticks again. Only a process restart clears it, because the queue lives on `Campaign.Current`.

The dedicated server runs **two independent, uncoordinated save paths**:

| | interval | path |
|---|---|---|
| Native `SaveHandler` autosave | ~30 min | `SetSaveArgs` → `MBSaveLoad.AutoSaveCurrentGame` → callback → dequeue |
| DS `CoopServerHost.SaveNow()` | 5 min | direct `Game.Current.Save(…, AsyncFileSaveDriver, _ => {})` |

Native autosave was still enabled server-side — `SaveHandlerClientBlockPatch` blocked `SetSaveArgs` only when `ModInformation.IsClient`.

## Evidence from the reported server log

```
20:15:56.521  AutoSaveCurrentGame: saveauto3          ← native save enqueued, blocks the game thread
20:15:58.015  Save Process completed in 1.486s        ← parks its callback in _currentActiveSaveData
20:15:58.031  ------Saving with new context. Save name: SR Test 1------   ← DS 5-min save, 16 ms later
20:15:58.226  ---------------SAVE STATISTICS---------  ← native async write finishes, too late
20:15:59.830  ------Successfully saved------          ← ONE line, for TWO saves
```

`"------Successfully saved------"` is printed only by `SaveOutput.PrintStatus()`, called only from `Game.OnSaveCompleted`. Every other save in the hour-long log produces exactly one; this pair produces one for two saves — the direct fingerprint of the overwritten slot. Campaign time, `parties=` and `mapEvents=` are frozen from that moment to the end of the log.

The earlier pair at 19:45 had 1.36 s of clearance between the native save completing and the DS save starting, logged two `Successfully saved` lines, and did not freeze.

**Why it was intermittent with no repro steps:** the two schedules drift into collision. The native trigger creeps ~1.5 s later each 30-minute cycle (`_lastAutoSaveTime` is stamped at *completion*); the DS's creeps ~0.02 s. 19:45 → 2.5 s apart; 20:15 → 0.97 s apart, landing inside the native save's blocking serialize phase, so `MaybeAutosave` fired on the very next free frame. Nothing to do with what players were doing.

Every reported symptom follows: `pulse: timeMode=StoppablePlay` still looks healthy because that field reads the mod's cached `TimePatches.CurrentMode`; pause/play still logs `Applying server time control` and changes nothing; networking stays alive because the coop `GameThread` queue is pumped by the DS's `CoopDriver.PumpGameLoop`, not by `MapState.OnTick`; rejoining cannot help; the pulse interval snaps back to the exact frame-cap spacing because `OnTick` returns immediately.

## The fix

Block the native scheduler on headless hosts too, leaving one save mechanism per process:

```csharp
internal static bool Prefix() => ModInformation.IsServer && !ModInformation.IsHeadlessHost;
```

**The discriminator is deliberately not `IsServer`.** A blanket server block would regress the non-Steam in-process server — `CoopartiveMultiplayerExperience.Handle(AttemptHost)` makes the clicking player's own process the server when `!SessionDiscovery.SteamAvailable`, and there the native autosave is the only thing that writes a `.sav`.

New `ModInformation.IsHeadlessHost` keys on `BANNERLORD_USER_DIR`, the marker headless and container hosts already set and which three existing consumers already treat as exactly this distinction (`CoopSaveManager`, `ModConfig`, `CampaignDifficultyHandler`). It was verified live in the #2540 log itself, which shows `mod-config.json loaded (…\server-data\…)` and `difficulty: playerReceivedDamage VeryEasy -> Realistic` — both only happen when that variable is set. `CampaignDifficultyHandler`'s identical private copy of the predicate is deduped onto the shared property.

Nothing is lost on the dedicated server: its own save goes through `Game.Save` too, so `SavePatches` still publishes `GameSaved` and the paired coop session json is still written next to the `.sav`.

## Changes

- `SaveHandlerClientBlockPatch` → `SaveHandlerAutoSaveBlockPatch` (renamed — the old name is now wrong), predicate extended to headless hosts, with the full mechanism documented on the class
- `ModInformation.IsHeadlessHost` added
- `CampaignDifficultyHandler` deduped onto it
- `coop.debug.save.force_autosave` now reports the headless case honestly instead of claiming the save should succeed
- 4 new tests in `SaveHandlerAutoSaveBlockPatchTests` — this patch had no coverage at all before

## Testing

- `GameInterface.Tests`: 720 tests, 709 passed, 11 skipped, 0 failed
- `Coop.Tests` + `Common.Tests`: 623 tests, 622 passed, 1 skipped, 0 failed
- Clean build, no new compiler warnings

## Follow-ups not in this PR

1. **DS-side belt-and-braces** (other repo): `CoopServerHost.SaveNow` should defer a frame while `Campaign.Current.SaveHandler.IsSaving` or a `Game.Current` save is in flight. The coop-side change alone closes the observed collision; this would be the second layer.
2. **A defensive prefix on `Game.Save`** refusing or queueing when `_currentActiveSaveData != null` would close the hole for any future save path, not just this pair.
3. **`isSaving=` in the DS pulse plus a watchdog** — the log carried no direct signal for this state; it had to be inferred from a missing log line.
4. **`TournamentSessionHandler`'s server-side `SaveHandler.IsSaving` guard** is now always false on the dedicated server. Harmless today only because the DS's save blocks the game thread outright, so there is no interleaving window — but it is a latent dead guard.


garrettluskey:
Superseded by Bannerlord-Coop-Team/BannerlordCoop.DedicatedServer#13, which fixes the same root cause (full writeup carried over there) with a dedicated-server-owned prefix on SaveHandler.SetSaveArgs instead of a headless-host heuristic in the mod. The suppression belongs with the component that owns the replacement scheduler: 'the native autosave is redundant' is only true where CoopServerHost.SaveNow exists, and the BANNERLORD_USER_DIR inference here could wrongly disable persistence on a server process that inherits the marker without running SaveNow. The incidental cleanups in this branch (ModInformation.IsHeadlessHost + the CampaignDifficultyHandler dedup) can be salvaged separately if wanted.

PR #2547, opened 2026-07-31
Add coop.delete_player command

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

There is no way to delete a player. A registered player's hero and party live forever, and rejoining with the same controller id always resumes the existing character.

## What is the new behavior?

Adds a client-only `coop.delete_player` console command. The client sends the request to the server; the server resolves the player from the requesting connection (client-sent ids are advisory only), removes the registration everywhere (`NetworkPlayerRemoved` goes out first so other clients lift the player-party destroy protection), disconnects the requester, then kills the hero and destroys the party with patches live so the existing sync flows (hero field sync, party destroy broadcast, clan cascades) replicate the results. Requests that cannot be applied (unknown connection, party in a battle or siege) are denied with a reason and nothing changes. Rejoining after deletion goes through character creation again.

Also adds `IPlayerManager.RemovePlayer` (the inverse of `AddPlayer`) and a note in `HeroSync` documenting that `_heroState` must not be registered there (the `HeroFieldPatches` transpiler already owns that store).

## Bot Changelog Entry

- Added a `coop.delete_player` console command: deletes your character on the server and disconnects you, so you can rejoin and create a new one.

## Other information

The E2E flow test does not assert the other client's hero death: the harness runs `Hero.ChangeState` during environment startup before `PatchAll`, so that method's transpiled intercept never fires in-process (production patches before any campaign code runs and is unaffected; the relay itself is covered by `HeroFieldTests`).

### DIFF
--- a/source/GameInterface/Services/Players/Handlers/PlayerDeletionHandler.cs
+++ b/source/GameInterface/Services/Players/Handlers/PlayerDeletionHandler.cs
@@ -1,0 +1,239 @@
+using Common;
+using Common.Logging;
+using Common.Messaging;
+using Common.Network;
+using Common.Util;
+using GameInterface.Services.ObjectManager;
+using GameInterface.Services.Players.Messages;
+using LiteNetLib;
+using Serilog;
+using System;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.Actions;
+using TaleWorlds.CampaignSystem.Party;
+using TaleWorlds.Library;
+
+namespace GameInterface.Services.Players.Handlers;
+
+/// <summary>
+/// Handles the coop.delete_player flow. The client forwards <see cref="PlayerDeleteRequested"/> to
+/// the server as <see cref="NetworkRequestDeletePlayer"/>; the server resolves the player from the
+/// requesting connection (never from client-sent ids), deletes its registration everywhere
+/// (<see cref="NetworkPlayerRemoved"/> lifts the player-party destroy protection on the other
+/// clients first), kicks the requester, then kills the hero and destroys the party with patches
+/// live so the existing death/destroy sync replicates both to the remaining clients. A request
+/// that cannot be applied is answered with <see cref="NetworkDeletePlayerDenied"/> and the
+/// requester stays connected.
+/// </summary>
+internal class PlayerDeletionHandler : IHandler
+{
+    private static readonly ILogger Logger = LogManager.GetLogger<PlayerDeletionHandler>();
+
+    private readonly IMessageBroker messageBroker;
+    private readonly INetwork network;
+    private readonly IObjectManager objectManager;
+    private readonly IPlayerManager playerManager;
+
+    public PlayerDeletionHandler(
+        IMessageBroker messageBroker,
+        INetwork network,
+        IObjectManager objectManager,
+        IPlayerManager playerManager)
+    {
+        this.messageBroker = messageBroker;
+        this.network = network;
+        this.objectManager = objectManager;
+        this.playerManager = playerManager;
+
+        messageBroker.Subscribe<PlayerDeleteRequested>(Handle_PlayerDeleteRequested);
+        messageBroker.Subscribe<NetworkRequestDeletePlayer>(Handle_NetworkRequestDeletePlayer);
+        messageBroker.Subscribe<NetworkPlayerRemoved>(Handle_NetworkPlayerRemoved);
+        messageBroker.Subscribe<NetworkDeletePlayerDenied>(Handle_NetworkDeletePlayerDenied);
+    }
+
+    public void Dispose()
+    {
+        messageBroker.Unsubscribe<PlayerDeleteRequested>(Handle_PlayerDeleteRequested);
+        messageBroker.Unsubscribe<NetworkRequestDeletePlayer>(Handle_NetworkRequestDeletePlayer);
+        messageBroker.Unsubscribe<NetworkPlayerRemoved>(Handle_NetworkPlayerRemoved);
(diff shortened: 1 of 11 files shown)


source/GameInterface/Services/Players/Handlers/PlayerDeletionHandler.cs
```diff
@@ -0,0 +1,228 @@
...
+    /// normal sync flows (the hero state via the hero field sync, the party via the destroy
+    /// broadcast); this message only carries the session-level registration, which no game-state
+    /// sync can.
+    /// </summary>
+    private void Handle_NetworkPlayerRemoved(MessagePayload<NetworkPlayerRemoved> payload)
+    {
+        if (ModInformation.IsServer) return;
+
+        var data = payload.What;
+        GameThread.RunSafe(() =>
```
ShoT-UPfps, line 179: this removal needs to finish before `RemotePlayerHeroHandler` checks for a duplicate. a loading client can replay the old player, removal, and new player back to back, and the poll-thread check at `RemotePlayerHeroHandler.cs:55` still sees the old registration and drops the new hero.

source/GameInterface/Services/Players/Handlers/PlayerDeletionHandler.cs
```diff
@@ -0,0 +1,228 @@
...
+        // the native death flow, then destroy whatever party is left, mirroring vanilla's defeat
+        // teardown order.
+        if (hero != null && hero.IsAlive)
+        {
+            // (showNotification: false, isForced: true) — forced skips the can-die model checks,
+            // same as the companion removal flow.
+            TryStep("hero kill", () => KillCharacterAction.ApplyByRemove(hero, false, true));
+        }
+
+        if (party != null && party.IsActive)
```
ShoT-UPfps: we still need to destroy inactive player parties here. captivity parks the registered party with `IsActive = false`, so deleting while captured skips this block and leaves the old party registered after the player creates a new one.

source/GameInterface/Services/Players/Handlers/PlayerDeletionHandler.cs
```diff
@@ -0,0 +1,228 @@
...
+        var data = payload.What;
+        GameThread.RunSafe(() =>
+        {
+            if (!playerManager.TryGetPlayer(data.ControllerId, out var player))
+            {
+                Logger.Debug("Deleted player {ControllerId} was not registered on this client", data.ControllerId);
+                return;
+            }
+
+            playerManager.RemovePlayer(player);
```
ShoT-UPfps: we need to call `hero.ChangeState(Hero.CharacterStates.Dead)` here under `AllowedThread`, because `NetworkHeroStateChanged` only assigns `_heroState`. otherwise the clan alive-lord cache and `CampaignObjectManager.AliveHeroes` still contain the deleted hero.

source/GameInterface/Services/Players/PlayerManager.cs
```diff
@@ -193,6 +203,47 @@ public void ClearPeer(NetPeer peer)
         }
     }
 
+    /// <inheritdoc cref="IPlayerManager.RemovePlayer(Player)"/>
+    public bool RemovePlayer(Player player)
+    {
+        if (player == null || !_players.Remove(player)) return false;
```
ShoT-UPfps: we need to synchronize `_players` before removing here. delete runs on the game thread while join handlers enumerate this `HashSet` on the poll thread, so a concurrent join can throw and drop its validation or existing-player transfer.

ShoT-UPfps (changes requested):
comments

source/GameInterface/Services/Players/Handlers/PlayerDeletionHandler.cs
garrettluskey: Fixed in 1b6406021: the removal now applies with GameThread.RunSafe(blocking: true), so it is visible before the poll thread handles the next replayed message (matching RemotePlayerHeroHandler's own blocking apply).

source/GameInterface/Services/Players/Handlers/PlayerDeletionHandler.cs
garrettluskey: Fixed in 1b6406021: the IsActive gate is gone - the party is destroyed whenever it still resolves, and it is re-resolved after the kill in case the death cascade already destroyed it. Clients drop their copies through the registry-level destroy replication regardless of local IsActive. Added ServerDeleteRequest_WithDeactivatedParty_StillDestroysIt to cover the captive case.

source/GameInterface/Services/Players/Handlers/PlayerDeletionHandler.cs
garrettluskey: Fixed in 1b6406021: the handler runs hero.ChangeState(Dead) under AllowedThread again (guarded on IsAlive), so the clan alive-lord caches and CampaignObjectManager lists drop the hero; the later NetworkHeroStateChanged confirm is an idempotent field overwrite. The flow test asserts the remaining client's hero death again.

source/GameInterface/Services/Players/PlayerManager.cs
garrettluskey: Fixed in 1b6406021: _players and controllerToPeer are now guarded by a single registrySync lock (formerly peerSync) across AddPlayer/TryGetPlayer/RemovePlayer/SetPeer/ClearPeer, and Players returns a snapshot so join-side enumeration never observes the mutation.

ShoT-UPfps (approved):
lgtm

PR #2550, opened 2026-08-01
[2541] Keep troop upgrade options open during time progression

### LINKED ISSUE #2541: [TROOPS] Troop upgrade bar closes while time is progressing
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

While time is progressing, opening the troop menu causes the troop upgrade bar to close on its own, preventing upgrades.

## Steps to reproduce

Not provided

## Additional information

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1532855726953857115/1532855726953857115

## Images and videos

- [Video: Mount__Blade_II_Bannerlord_2026.07.31_-_16.56.10.02.mp4](<https://cdn.discordapp.com/attachments/1532855726953857115/1532855731601277220/Mount__Blade_II_Bannerlord_2026.07.31_-_16.56.10.02.mp4?ex=6aa46c25&is=6aa31aa5&hm=c5d0a84b6cbe78cf0f7e94efea68c996cba85e9579308d4cfe4b7565a0edf193&>) ([Discord message](https://discord.com/channels/709516043332354119/1532855726953857115/1532855726953857115))

## Logs

Client log: [0.0.3+d238190b9ce7b38153320525145f651376c6dadd - Download Coop_client.log](<https://cdn.discordapp.com/attachments/1532855726953857115/1532856107582754816/Coop_client.log?ex=6aa46c7f&is=6aa31aff&hm=8b40e1d8ea69c7a03a67d4ddd262325356c28741e4210e06f5b2a2e238ade0a0&>)

Server log: [Download coop-server-20260731-163403.log](<https://cdn.discordapp.com/attachments/1532855726953857115/1532859962005520434/coop-server-20260731-163403.log?ex=6aa47016&is=6aa31e96&hm=2e7023458b202a6829ee43b9b0d50712ddbf396c0973e3967c42746644fe46ed&>)

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Opening a troop's upgrade choices while campaign time is progressing causes the choices to close when an authoritative roster update refreshes the Party screen, preventing the player from completing an upgrade.

## What is the new behavior?

Resolves #2541

The selected troop row and its upgrade choices now remain open through authoritative Party-screen roster refreshes, while the normal fallback remains when that troop is no longer present.

## Bot Changelog Entry

- Fixed troop upgrade choices closing while campaign time progresses.

### DIFF
--- a/source/E2E.Tests/Util/ObjectBuilders/MobilePartyBuilder.cs
+++ b/source/E2E.Tests/Util/ObjectBuilders/MobilePartyBuilder.cs
@@ -12,6 +12,7 @@ internal class MobilePartyBuilder : IObjectBuilder
         var spawnSettlement = GameObjectCreator.CreateInitializedObject<Settlement>();
         var leaderHero = GameObjectCreator.CreateInitializedObject<Hero>();
         var clan = GameObjectCreator.CreateInitializedObject<Clan>();
+        // Real clans always have a culture; surrender paths read it while building encounter state.
         clan.Culture = GameObjectCreator.CreateInitializedObject<CultureObject>();
         
         leaderHero.Clan = clan;
--- a/source/E2E.Tests/Util/ObjectBuilders/MobilePartyBuilder.cs
+++ b/source/E2E.Tests/Util/ObjectBuilders/MobilePartyBuilder.cs
@@ -12,6 +12,7 @@ internal class MobilePartyBuilder : IObjectBuilder
         var spawnSettlement = GameObjectCreator.CreateInitializedObject<Settlement>();
         var leaderHero = GameObjectCreator.CreateInitializedObject<Hero>();
         var clan = GameObjectCreator.CreateInitializedObject<Clan>();
+        clan.Culture = GameObjectCreator.CreateInitializedObject<CultureObject>();
         
         leaderHero.Clan = clan;
         clan.SetLeader(leaderHero);
--- a/source/GameInterface/Services/Party/Commands/PartyCommands.cs
+++ b/source/GameInterface/Services/Party/Commands/PartyCommands.cs
@@ -139,7 +139,7 @@ internal class PartyCommands
     [CommandLineArgumentFunction("restore_position", "coop.debug.mobileparty")]
     public static string RestorePositionCommand(List<string> strings)
     {
-        if (!ModInformation.IsServer) return "Command can only be run on the server.";
+        if (ModInformation.IsClient) return "Command can only be run on the server.";
         if (strings.Count != 4 ||
             !float.TryParse(strings[1], NumberStyles.Float, CultureInfo.InvariantCulture, out var positionX) ||
             !float.TryParse(strings[2], NumberStyles.Float, CultureInfo.InvariantCulture, out var positionY) ||
@@ -288,6 +288,86 @@ internal class PartyCommands
     }
 
     /// <summary>
+    /// Sets one member-roster troop's exact state and reports the state it replaced.
+    /// </summary>
+    [CommandLineArgumentFunction("set_troop_state", "coop.debug.mobileparty")]
+    public static string SetTroopStateCommand(List<string> strings)
+    {
+        if (ModInformation.IsClient) return "Command can only be run on the server.";
+        if (strings.Count != 6)
+            return "Usage: coop.debug.mobileparty.set_troop_state <party id> <character id> <exists> <number> <wounded count> <xp>";
+
+        if (!bool.TryParse(strings[2], out var shouldExist) ||
+            !int.TryParse(strings[3], out var number) ||
+            !int.TryParse(strings[4], out var woundedCount) ||
+            !int.TryParse(strings[5], out var xp))
+            return "Exists must be true or false, and number, wounded count, and xp must be integers.";
+        if (number < 0 || woundedCount < 0 || woundedCount > number || xp < 0 || (number == 0 && xp != 0) ||
+            (!shouldExist && (number != 0 || woundedCount != 0 || xp != 0)))
+            return "State requires number >= 0, wounded between 0 and number, xp >= 0, zero xp when number is zero, and all zero values when exists is false.";
+
+        if (!TryGetObjectManager(out var objectManager)) return "Unable to resolve ObjectManager.";
+        if (!objectManager.TryGetObject(strings[0], out MobileParty party))
+            return $"Party with id {strings[0]} not found.";
+        if (!objectManager.TryGetObject(strings[1], out CharacterObject character))
+            return $"Character with id {strings[1]} not found.";
+        if (character.IsHero) return "Hero roster elements are not supported by this command.";
+
+        var roster = party.MemberRoster;
+        int index = roster.FindIndexOfTroop(character);
+        bool oldExists = index >= 0;
+        var oldState = oldExists ? roster.GetElementCopyAtIndex(index) : default;
+
+        if (index < 0 && shouldExist)
+        {
+            roster.AddToCounts(character, System.Math.Max(number, 1), removeDepleted: false);
+            index = roster.FindIndexOfTroop(character);
+        }
+
+        if (index >= 0)
+        {
+            int currentWounded = roster.GetElementWoundedNumber(index);
+            if (currentWounded > number) roster.SetElementWoundedNumber(index, number);
+            roster.SetElementNumber(index, number);
+            roster.SetElementWoundedNumber(index, woundedCount);
+            roster.SetElementXp(index, xp);
+            if (!shouldExist) roster.RemoveZeroCounts();
+            roster.InitializeCachedData();
+        }
+
+        return $"TROOP_STATE_SET party={strings[0]} character={strings[1]} " +
+               $"oldExists={oldExists} oldNumber={oldState.Number} oldWounded={oldState.WoundedNumber} oldXp={oldState.Xp} " +
+               $"newExists={shouldExist} newNumber={number} newWounded={woundedCount} newXp={xp}";
+    }
+
+    /// <summary>
+    /// Selects a real right-member row so live tests can observe its inline upgrade choices.
+    /// </summary>
+    [CommandLineArgumentFunction("select_party_screen_troop", "coop.debug.mobileparty")]
+    public static string SelectPartyScreenTroopCommand(List<string> strings)
+    {
+        if (ModInformation.IsServer) return "Command can only be run on a client.";
+        if (strings.Count != 1)
+            return "Usage: coop.debug.mobileparty.select_party_screen_troop <character id>";
+        if (!TryGetObjectManager(out var objectManager)) return "Unable to resolve ObjectManager.";
+        if (!objectManager.TryGetObject(strings[0], out CharacterObject character))
+            return $"Character with id {strings[0]} not found.";
+        if (!(Game.Current?.GameStateManager?.ActiveState is PartyState))
+            return "No active party screen.";
+
+        var partyVm = (ScreenManager.TopScreen as GauntletPartyScreen)?._dataSource;
+        if (partyVm == null) return "No active Party screen view model.";
+
+        var row = partyVm.MainPartyTroops.FirstOrDefault(vm => vm.Character == character);
+        if (row == null) return $"{strings[0]} is not in the right member roster.";
+
+        if (!row.IsSelected) partyVm.ExecuteSelectCharacterTuple(row);
+        return $"PARTY_SCREEN_TROOP_SELECTED character={strings[0]} selected={row.IsSelected} " +
+               $"upgradeTargets={row.Upgrades.Count} ready={row.NumOfReadyToUpgradeTroops} " +
+               $"upgradeable={row.NumOfUpgradeableTroops}";
+    }
+
+    /// <summary>
     /// Creates a real pending Party-screen transfer for live synchronization tests.
     /// </summary>
     [CommandLineArgumentFunction("stage_party_screen_transfer", "coop.debug.mobileparty")]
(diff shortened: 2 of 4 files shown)


source/GameInterface/Services/Party/PartyScreenRosterRefresher.cs
```diff
@@ -237,7 +240,31 @@ private static void RefreshRecruitablePrisoners(
...
+            .Concat(partyVm.MainPartyPrisoners)
+            .Concat(partyVm.OtherPartyTroops)
+            .Concat(partyVm.OtherPartyPrisoners)
+            .FirstOrDefault(row =>
+                ReferenceEquals(row.Character, selectedCharacter) &&
+                row.Side == selectedSide &&
+                row.Type == selectedType);
+        if (replacement == null) return;
+
+        partyVm.SetSelectedCharacter(replacement);
```
garrettluskey: Heads up: `SetSelectedCharacter` isn't just an assignment. The `CurrentCharacter` setter runs `RefreshCurrentCharacterInformation()` (that's the part that actually re-opens the upgrades) and then `ExecuteRemoveZeroCounts()`, which runs `PartyScreenLogic.RemoveZeroCounts()` over all four current rosters.

So every refresh that restores a selection now purges zero-count elements mid-apply — something this class otherwise only does explicitly via `TryRemoveZeroCounts`. `TryRebase`/`Write` can intentionally leave an `exists=true, number=0` element in the visible roster between a set-to-zero and its paired `ZeroCountsRemoved` message; this purges it a message early, and the next `TryApply` for that character reads the missing element as a local player edit (`localExistenceChanged` flips).

I traced it and it does converge — the later `TryRemoveZeroCounts` just no-ops, and a native row click fires this exact chain anyway — so keeping it is defensible. But make it deliberate: either drop a comment here saying the purge is intended (native-click-equivalent), or restore without the setter side effects:

```csharp
partyVm._currentCharacter = replacement;
partyVm.RefreshCurrentCharacterInformation();
partyVm.OnPropertyChangedWithValue(replacement, nameof(partyVm.CurrentCharacter));
```

source/GameInterface/Services/Party/PartyScreenRosterRefresher.cs
```diff
@@ -237,7 +240,31 @@ private static void RefreshRecruitablePrisoners(
...
+
         logic.OnReset(false);
+
+        if (!restoreSelection) return;
+
+        var replacement = partyVm.MainPartyTroops
+            .Concat(partyVm.MainPartyPrisoners)
+            .Concat(partyVm.OtherPartyTroops)
+            .Concat(partyVm.OtherPartyPrisoners)
+            .FirstOrDefault(row =>
```
garrettluskey: None of this is tested. Every test in `PartyScreenRosterRefresherTests` hits the `partyVm == null` early-out, so the matching rules — upgraded-away troop falls back, same character on both sides disambiguated by `Side`, member vs prisoner by `Type` — run for the first time on a live client.

Pull the capture + matching into an internal helper that takes the four row lists plus the captured `(character, side, type)` and returns the replacement row. Then the existing test class covers it without needing a real `PartyVM`, and the VM-touching shell stays thin.

If that's genuinely not doable, at minimum put the two-client repro steps (`set_troop_state` / `select_party_screen_troop` / `party_screen_troop_state`) in the PR description so someone can actually verify this.

source/GameInterface/Services/Party/Commands/PartyCommands.cs
```diff
@@ -287,6 +287,89 @@ public static string SetTroopWoundedCommand(List<string> strings)
         return $"Set {strings[1]} oldWounded={element.WoundedNumber} newWounded={woundedCount}.";
     }
 
+    /// <summary>
+    /// Sets one member-roster troop's exact state and reports the state it replaced.
+    /// </summary>
+    [CommandLineArgumentFunction("set_troop_state", "coop.debug.mobileparty")]
+    public static string SetTroopStateCommand(List<string> strings)
```
garrettluskey, line 294: CodeFactor is red on this method ("Complex Method") and it'll block the PR. The parse/validate block is most of the bulk — pull it out into something like `ValidateTroopStateArgs(strings, out shouldExist, out number, out woundedCount, out xp)` that returns the error string (null when valid). Clears the check and the mutation sequence gets easier to read.

source/GameInterface/Services/Party/Commands/PartyCommands.cs
```diff
@@ -335,25 +418,27 @@ public static string PartyScreenTroopStateCommand(List<string> strings)
...
                $"pending={logic.IsThereAnyChanges()}";
     }
 
-    private static (int number, int wounded) GetRosterElement(
+    private static (int number, int wounded, int xp) GetRosterElement(
         TroopRoster roster,
         CharacterObject character)
     {
         var index = roster.FindIndexOfTroop(character);
         if (index < 0) return default;
```
garrettluskey: `default` prints an absent troop as `0/0/0`, but the VM branch above prints `-1/-1/-1` for a missing row. Now that `set_troop_state` can create `exists=true, number=0` on purpose, the `visible*`/`baseline*` fields can't tell "absent" from "present at zero" — exactly the states these live tests need to distinguish. Return `(-1, -1, -1)` here too. Nothing parses this format yet (I checked), so it's a free change.

source/GameInterface/Services/Party/Commands/PartyCommands.cs
```diff
@@ -287,6 +287,89 @@ public static string SetTroopWoundedCommand(List<string> strings)
...
+               $"newExists={shouldExist} newNumber={number} newWounded={woundedCount} newXp={xp}";
+    }
+
+    /// <summary>
+    /// Selects a real right-member row so live tests can observe its inline upgrade choices.
+    /// </summary>
+    [CommandLineArgumentFunction("select_party_screen_troop", "coop.debug.mobileparty")]
+    public static string SelectPartyScreenTroopCommand(List<string> strings)
+    {
+        if (ModInformation.IsServer) return "Command can only be run on a client.";
```
garrettluskey, line 374: Nit: this file now spells the same gate three ways — `ModInformation.IsServer` here, `!ModInformation.IsClient` two commands down, and `ModInformation.IsClient` on the server-only ones (which your `restore_position` change aligns with). They're all the same check; `IsClient` is literally `!IsServer`. Pick one spelling per direction so a grep finds every gate. Not blocking.

source/E2E.Tests/Util/ObjectBuilders/MobilePartyBuilder.cs
```diff
@@ -12,6 +12,7 @@ public object Build()
         var spawnSettlement = GameObjectCreator.CreateInitializedObject<Settlement>();
         var leaderHero = GameObjectCreator.CreateInitializedObject<Hero>();
         var clan = GameObjectCreator.CreateInitializedObject<Clan>();
+        clan.Culture = GameObjectCreator.CreateInitializedObject<CultureObject>();
```
garrettluskey, line 16: What needed this? Nothing in the PR uses the clan culture and no E2E test was added — I ran the suites on this branch and nothing depends on it. If this was groundwork for a selection-restore E2E test that got cut, I'd rather have the test in the PR. If some path NREs without it, say which one in a comment. Not blocking either way — it matches what `CaravanPartyComponentTests` already does.

garrettluskey (dismissed):
Requesting changes. The fix itself is right — I checked against the decompiled `PartyVM`: `OnReset` rebuilds every row unselected and never reassigns `CurrentCharacter`, so the selection (and the upgrade panel that hangs off it) always died on refresh. Re-selecting through `SetSelectedCharacter` re-runs `InitializeUpgrades`, which is what brings the choices back, and since all rebuilt rows start unselected the single-selection invariant holds. `set_troop_state` also only goes through patched mutators, so it replicates.

Three things before merge:

1. The restore drags `ExecuteRemoveZeroCounts` into the authoritative refresh path — needs to be a deliberate choice, see inline.
2. The new matching logic has zero coverage; every existing refresher test bails at `partyVm == null`. See inline.
3. CodeFactor is red on `SetTroopStateCommand`.

CI hasn't run on this PR, so I ran it locally on the branch: `GameInterface.Tests` all 716 green (including the 8 refresher/delta-handler tests), plus `ArmyDestructionTests` E2E for the builder change. The `PARTY_SCREEN_TROOP_STATE` format change breaks nothing — no automated consumers. The `restore_position` gate flip is a no-op (`IsClient` ≡ `!IsServer`), just idiom alignment.

source/GameInterface/Services/Party/PartyScreenRosterRefresher.cs
ShoT-UPfps: fixed, selection restoration now updates `_currentCharacter` and refreshes its details without running `ExecuteRemoveZeroCounts`.

source/GameInterface/Services/Party/PartyScreenRosterRefresher.cs
ShoT-UPfps: added coverage for exact `(character, side, type)` matching and the removed-selection fallback.

source/GameInterface/Services/Party/Commands/PartyCommands.cs
ShoT-UPfps: pulled the parse and validation block into `ValidateTroopStateArgs`.

source/GameInterface/Services/Party/Commands/PartyCommands.cs
ShoT-UPfps: missing roster rows now return `(-1, -1, -1)`.

source/GameInterface/Services/Party/Commands/PartyCommands.cs
ShoT-UPfps: standardized the client-only command gates on `ModInformation.IsServer`.

source/E2E.Tests/Util/ObjectBuilders/MobilePartyBuilder.cs
ShoT-UPfps: added a comment explaining that the surrender encounter path reads clan culture.

source/GameInterface/Services/Party/PartyScreenRosterRefresher.cs
```diff
@@ -237,7 +268,176 @@ private static void RefreshRecruitablePrisoners(
...
+    private static void RefreshSort(
+        PartyScreenLogic logic,
+        TaleWorlds.CampaignSystem.ViewModelCollection.Party.PartyVM partyVm,
+        PartyScreenSelectionIdentity identity)
+    {
+        var sortType = logic.GetActiveSortTypeForSide(identity.Side);
+        if (sortType != PartyScreenLogic.TroopSortType.Count) return;
+
+        var comparer = new TaleWorlds.CampaignSystem.ViewModelCollection.Party.PartyVM.TroopVMComparer(
+            logic.GetComparer(sortType));
```
garrettluskey: This can sort the wrong way. The comparers in `_defaultComparers` are shared between the two sides, and `UpdateComparersAscendingOrder(command.IsSortAscending)` stamps the direction onto all of them globally on every sort command — the per-side direction only lives in `GetIsAscendingSortForSide`. Native gets away with using `GetComparer(...)` as-is because the VM's `SortTroops` handler only ever runs right after the logic synced the direction for that exact side's command. This runs outside any sort command, so: sort the right side by count ascending, sort the left side by anything descending, then let an authoritative right-side count change land — the right list re-sorts descending while its header still says ascending.

One line before wrapping:

```csharp
var troopComparer = logic.GetComparer(sortType);
troopComparer.SetIsAscending(logic.GetIsAscendingSortForSide(identity.Side));
var comparer = new TaleWorlds.CampaignSystem.ViewModelCollection.Party.PartyVM.TroopVMComparer(troopComparer);
```

(`SetIsAscending` and `GetIsAscendingSortForSide` are both public.) Mutating the shared comparer is fine — that's exactly what every native sort command does.

garrettluskey (dismissed):
Re-reviewed `9b1e8f2`. All six earlier points are addressed — replied in the threads, everything's resolved (CodeFactor is still red on `SetTroopStateCommand`, but it's advisory here, see thread).

Also went through the new in-place refresh against the decompiled VM, and it holds up:

- `RefreshTroop` mirrors what native does per row — the `Troop` setter refreshes tooltip/transfer state, `ThrowOnPropertyChanged` fires the count/upgrade bindings, and the add/remove cases correctly fall back to a full reset.
- Skipping `OnReset` when no VM is attached is safe: `OnReset` only invokes `AfterReset`, and `PartyVM` is its only subscriber.
- XP-only changes skipping the top-info refresh while still reinitializing upgrades is exactly right.
- CI is green now — build, unit tests, all 8 E2E shards.

One real bug in the new code though, hence requesting changes again: `RefreshSort` can sort with a stale direction. See inline — it's a one-line fix.

source/GameInterface/Services/Party/PartyScreenRosterRefresher.cs
garrettluskey: Verified — matches what the setter did minus the purge, and `UpdateRecruitable` covers the rest of `SetSelectedCharacter`. The in-place `RefreshTroop` path making full resets rare is even better. Resolved.

source/GameInterface/Services/Party/PartyScreenRosterRefresher.cs
garrettluskey: Seam plus both tests look good. Resolved.

source/GameInterface/Services/Party/Commands/PartyCommands.cs
garrettluskey: Still red after the split — same annotation, the threshold is tighter than I figured. Pulling the mutation block into an `ApplyTroopState` helper should clear it. Not blocking on it though: CodeFactor is not enforced here (#2300 merged red), so your call.

source/GameInterface/Services/Party/Commands/PartyCommands.cs
garrettluskey: Resolved.

source/GameInterface/Services/Party/Commands/PartyCommands.cs
garrettluskey: Resolved.

source/E2E.Tests/Util/ObjectBuilders/MobilePartyBuilder.cs
garrettluskey: Comment works. Resolved.

source/GameInterface/Services/Party/PartyScreenRosterRefresher.cs
ShoT-UPfps: fixed, `RefreshSort` now restores the saved direction for the affected side before wrapping the shared comparer.

garrettluskey: Verified. Resolved.

garrettluskey (approved):
Fix confirmed — the comparer direction gets re-synced from the side's saved state before the sort, which is exactly what native does on every sort command. CI is green across the board (build, unit, all 8 E2E shards). CodeFactor is still flagging `set_troop_state` but that's advisory and already discussed in-thread. Good to merge.

PR #2573, opened 2026-08-01
Enforce exact co-op build compatibility

ShoT-UPfps:
## PR Type

- [x] Bug fix
- [x] Build related changes

local and playtest builds can lose the Git revision and advertise only the shared semantic version, so servers built from different commits can appear compatible. direct connections also only compare module manifest versions.

## What is the new behavior?

builds require the exact Git revision in the co-op build version. the client sends that build version during connection validation and the server rejects a mismatch before module validation.

## Bot Changelog Entry

- Servers running a different co-op build now show as incompatible and reject connections.


ShoT-UPfps:
i think this is fixed now? so closing

PR #2584, opened 2026-08-02
[2572] Prevent duplicate captive lord release conversations

### LINKED ISSUE #2572: [CAPTIVITY] Captive lord release triggers duplicate follow-up conversation and relation rewards
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

Releasing captive lords causes the post-release conversation to happen twice and applies relation gains twice for each lord.

## Steps to reproduce

1. Fight/cheat against an enemy army with captive lords.
2. Free one or more Khuzait lords.
3. Observe the follow-up conversation and relationship bonus firing twice.

## Additional information

The duplicate behavior is reported as not occurring for lords that are defeated and captured.

## Discord source

https://discord.com/channels/709516043332354119/1533170717221322986/1533170717221322986

## Logs

Client log: [0.1.1+87a4520ab6fd2a9e4e6aad964ccd395296a507fc - Download Coop_client.log](<https://cdn.discordapp.com/attachments/1533170717221322986/1533170717611266128/Coop_client.log?ex=6aa4e8c0&is=6aa39740&hm=3e2b9ba821493e5ab800d22df9dc93aae86b5a42de2f4527c7414e2ef4cbc932&>)

Server log: [Download coop-server-20260801-095646.log](<https://cdn.discordapp.com/attachments/1533170717221322986/1533170717942747278/coop-server-20260801-095646.log?ex=6aa4e8c0&is=6aa39740&hm=d4718d81163b301d6ee45856a1319da530ca649e01cc67219489e3cbb6045695&>)

## Crash report link

- [Google Drive file](<https://drive.google.com/file/d/1ff4Q2WOXA6vkbSwNukS0eBVK-Pt9JBDO/view?usp=sharing>) ([Discord message](https://discord.com/channels/709516043332354119/1533170717221322986/1533170717221322986))

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Releasing captive lords after defeating their captor can reopen each lord's follow-up conversation and grant the relation reward twice.

## What is the new behavior?

Resolves #2572

Each freed captive lord now appears once in the post-battle conversation sequence and grants the +4 relation reward only once.

## Bot Changelog Entry

- Fixed an issue where if you freed a lord that was being held captive by another lord, you'd talk to them twice.


source/GameInterface/Services/Heroes/Patches/LordConversationsCampaignBehaviorPatches.cs
```diff
@@ -78,6 +80,7 @@ public static bool ConversationTalkLordFreedToLordCaptureOnConsequencePrefix()
 
         var message = new TakeLordPrisoner(Campaign.Current.MainParty.Party, Hero.OneToOneConversationHero);
         MessageBroker.Instance.Publish(null, message);
+        MarkFreedLordConversationHandled(Hero.OneToOneConversationHero);
```
garrettluskey, line 91: `conversation_player_fails_to_release_prisoner_on_consequence` needs this too. vanilla runs `TakePrisonerAction.Apply(PartyBase.MainParty, ...)` synchronously there so `DoFreeOrCapturePrisonerHeroes` stops offering the conversation, but ours defers it to `Handle_NetworkTakeLordPrisoner`, so until the round trip lands the hero is still a prisoner of the enemy party and passes the `IsPrisoner && PartyBelongedToAsPrisoner != PartyBase.MainParty` filter, and the 'you're my prisoner now' conversation can reopen and republish `TakeLordPrisoner`.

garrettluskey (dismissed):
the mechanism is right. vanilla advances the post-battle conversation queue by running the captivity action synchronously in the consequence handler, and we defer that over the network, so advancing `_capturedAlreadyPrisonerHeroes` manually is the fix, and the `IsPrisoner` early returns stop the second message from reapplying the reward. the `CultureObject.HasFeat` test switch is right too, disabling `PartyBaseHelper.HasFeat` never fired since our own prefix returns false first.

one hole: the keep-as-prisoner choice has the same race and no mark, see inline.

source/GameInterface/Services/Heroes/Patches/LordConversationsCampaignBehaviorPatches.cs
ShoT-UPfps: fixed, the keep-as-prisoner consequence now advances `_capturedAlreadyPrisonerHeroes` before the network round trip too

garrettluskey (approved):
mechanism checks out against the decompiled vanilla. `DoFreeOrCapturePrisonerHeroes` never removes entries from `_capturedAlreadyPrisonerHeroes`, it just reoffers the last hero still matching `IsPrisoner && PartyBelongedToAsPrisoner != PartyBase.MainParty`, and vanilla only advances that because every consequence handler runs its captivity action synchronously. we defer the action over the network, so removing the entry when we publish is the right replacement, and the `IsPrisoner` early returns stop a repeated message from reapplying the relation reward. both keep-as-prisoner paths are marked now, which covers the hole from the last review, and the defeat/capture conversations are fine left alone because `_capturedHeroes` gets removed by the `ConversationEndOneShot` handler.

the test swap is right too. disabling `PartyBaseHelper.HasFeat` never fired since our own prefix returns false before the `Priority.Last` disable runs, and `CultureObject.HasFeat` is where the null `_cultureFeats` would actually throw. unit test matches the `SiegeAftermathPendingTests` setup and everything is green.

PR #2592, opened 2026-08-02
Fix joins breaking permanently when a player's hero goes missing

garrettluskey:
## PR Checklist

## PR Type

If a player's registration names a hero that no longer exists in the campaign, that player can never join again — and the failure is invisible to them.

`ResolveCharacterState.Handle_ClientValidate` finds the registration, fails to resolve its hero, and routes the player to character creation. The character created there calls `AddPlayer` with the **same controller id**. `Player` has no value equality and the registry was a `HashSet<Player>`, so the second registration was accepted. From that point `TryGetPlayer(controllerId, ...)` ran `SingleOrDefault` over two matches and threw:

```
System.InvalidOperationException: Sequence contains more than one matching element
  at PlayerManager.TryGetPlayer(String controllerId, Player& player)
  at PlayerManager.SetPeer(String controllerId, NetPeer peer)
  at CreateCharacterState.Handle_NetworkTransferNewHero(MessagePayload`1 obj)
```

That throw is inside a `MessageBroker` handler, so it dies in the poller: the rest of the handler never runs, no reply is sent, and the client sits on the validation screen until `ValidateModuleState`'s 30-second deadline expires. Every later reconnect from that controller hits the same throw, now from `Handle_ClientValidate`.

Seen live on a dedicated server: one poisoned controller, then 13 failed rejoins over the next hour while other players joined normally. The corresponding client log shows only `Timed out after 30s waiting for the server to validate the connection`, repeated — the module check itself passed every time.

## What is the new behavior?

**One registration per controller.** The registry is keyed by controller id, so a lookup can never be ambiguous and `TryGetPlayer` cannot throw. `AddPlayer` refuses a second registration for a controller that already has one and logs which heroes collided. `RemovePlayer` matches on the instance, so a caller holding a superseded `Player` cannot deregister its successor.

**The dead registration is dropped, not duplicated.** `ResolveCharacterState` deregisters a registration whose hero is missing before routing its owner to character creation, so the replacement takes the controller id cleanly. This is the root fix — the duplicate is never created.

**Joins fail loudly.** `Handle_ClientValidate` no longer lets an unexpected throw escape into the poller; it drops the connection so the client surfaces a disconnect immediately instead of hanging, matching the existing treatment of the module-validation step. It disconnects rather than answering "no hero" because that answer would push the player into creating a second character.

**Existing poisoned saves heal.** A session file that already carries two entries for one controller loads the first and logs the skip instead of silently registering both; the owner heals the leftover by joining, which drops the entry naming the missing hero.

## Bot Changelog Entry

- Fixed a player being permanently unable to rejoin a server — stuck on the "Validating modules" screen — after their character went missing from the save.

## Other information

Tests: `PlayerManagerTests` covers the refused duplicate, the deregister-then-replace path, and the superseded-instance guard. `ResolveCharacterStateTests` covers the stale registration being dropped on the way to character creation, and a throw during resolution not answering or advancing the connection.

Full unit suites green locally: 1484 passed, 14 skipped, 0 failed.

### DIFF
--- a/source/Coop.Core/Server/Services/Save/Handlers/SaveGameHandler.cs
+++ b/source/Coop.Core/Server/Services/Save/Handlers/SaveGameHandler.cs
@@ -1,4 +1,5 @@
-﻿using Common.Messaging;
+﻿using Common.Logging;
+using Common.Messaging;
 using Common.Network;
 using Coop.Core.Server.Services.Save.Messages;
 using GameInterface.CoopSessionData;
@@ -7,6 +8,7 @@ using GameInterface.Registry.Messages;
 using GameInterface.Services.Heroes.Messages;
 using GameInterface.Services.Players;
 using GameInterface.Services.Save.Messages;
+using Serilog;
 using System.Collections.Generic;
 using System.Linq;
 
@@ -17,6 +19,8 @@ namespace Coop.Core.Server.Services.Save.Handlers;
 /// </summary>
 internal class SaveGameHandler : IHandler
 {
+    private static readonly ILogger Logger = LogManager.GetLogger<SaveGameHandler>();
+
     private readonly IMessageBroker messageBroker;
     private readonly ICoopSaveManager saveManager;
     private readonly ICoopSessionProvider coopSessionProvider;
@@ -121,7 +125,15 @@ internal class SaveGameHandler : IHandler
 
         foreach (var player in savedSession.Players)
         {
-            playerRegistry.AddPlayer(player);
+            // A save written before controller ids were made unique can carry two entries for one
+            // controller; AddPlayer keeps the first. The owner heals the leftover by joining — a
+            // registration naming a missing hero is dropped in ResolveCharacterState — but say so,
+            // because a silently skipped registration is exactly what made this hard to diagnose.
+            if (!playerRegistry.AddPlayer(player))
+                Logger.Warning(
+                    "Skipped saved registration for controller {ControllerId} (hero {HeroId}): " +
+                    "that controller is already registered",
+                    player?.ControllerId, player?.HeroId);
         }
     }
 }
--- a/source/Coop.Core/Server/Services/Save/Handlers/SaveGameHandler.cs
+++ b/source/Coop.Core/Server/Services/Save/Handlers/SaveGameHandler.cs
@@ -6,11 +6,14 @@ using GameInterface.CoopSessionData;
 using GameInterface.CoopSessionData.Save.Data;
 using GameInterface.Registry.Messages;
 using GameInterface.Services.Heroes.Messages;
+using GameInterface.Services.ObjectManager;
 using GameInterface.Services.Players;
+using GameInterface.Services.Players.Data;
 using GameInterface.Services.Save.Messages;
 using Serilog;
 using System.Collections.Generic;
 using System.Linq;
+using TaleWorlds.CampaignSystem;
 
 namespace Coop.Core.Server.Services.Save.Handlers;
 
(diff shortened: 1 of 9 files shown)


source/GameInterface/Services/Players/PlayerManager.cs
```diff
@@ -121,9 +130,32 @@ public PlayerManager(ILogger logger, IObjectManager objectManager, IControllerId
...
+                // owns that decision, and the join flow already deregisters a dead registration
+                // before creating a replacement. Reaching this means an unexpected double-add, or
+                // a save written before this fix that carries two entries for one controller.
+                if (!ReferenceEquals(registered, player))
+                    logger.Error(
+                        "Controller {ControllerId} is already registered to hero {RegisteredHeroId}; " +
+                        "refusing to also register hero {HeroId}",
+                        player.ControllerId, registered.HeroId, player.HeroId);
+
+                return false;
```
ShoT-UPfps, line 155: could we make CreateCharacterState stop when this returns false? two overlapping joins for one controller can both reach character creation before either upload completes. otherwise the second peer is associated with the first player while it and the other clients are sent the rejected second player.

source/Coop.Core/Server/Services/Save/Handlers/SaveGameHandler.cs
```diff
@@ -121,7 +125,15 @@ private void Handle_AllGameObjectsRegistered(MessagePayload<AllGameObjectsRegist
 
         foreach (var player in savedSession.Players)
         {
-            playerRegistry.AddPlayer(player);
+            // A save written before controller ids were made unique can carry two entries for one
+            // controller; AddPlayer keeps the first. The owner heals the leftover by joining — a
+            // registration naming a missing hero is dropped in ResolveCharacterState — but say so,
+            // because a silently skipped registration is exactly what made this hard to diagnose.
+            if (!playerRegistry.AddPlayer(player))
```
ShoT-UPfps, line 134: could we prefer the duplicate whose hero still resolves when loading? a poisoned save can contain the missing old registration first and the valid replacement second, so keeping the first skips the only usable registration. the next join then removes that entry and creates another character while the valid hero is left unregistered.

source/GameInterface/Services/Players/PlayerManager.cs
garrettluskey: Good catch — fixed in 9930fe77.

You are right that it is reachable: everything after `AddPlayer` assumes the peer owns the player it just created, so a refusal bound the peer to whichever registration already held the controller (`SetPeer` resolves by controller id, not by the instance), broadcast the refused player to every other client via `NetworkNewPlayerHeroCreated`, and handed the joiner the refused player’s ids in `NetworkHeroRecieved`.

`CreateCharacterState` now disconnects the joining peer and returns, the same as the `TryCreatePlayer` failure directly above it. Covered by `NetworkTransferNewHero_ControllerAlreadyRegistered_DisconnectsWithoutAnnouncing`, which asserts only that peer is ejected, `SetPeer` is never called, and nothing is sent.

source/Coop.Core/Server/Services/Save/Handlers/SaveGameHandler.cs
garrettluskey: Agreed, and that is exactly the order the save from this bug had — the dead registration first, the replacement appended after it. Fixed in 9930fe77.

Loading now picks the registration whose hero still resolves rather than the one that comes first. `AllGameObjectsRegistered` means the campaign objects are registered, so that lookup is authoritative about which registration still has something behind it. `SaveGameHandler` takes an `IObjectManager` for it.

Two details worth flagging:
- When neither hero resolves there is nothing to prefer, so it keeps the first — the controller still ends up with exactly one registration, which its owner heals by joining.
- A missing hero only breaks ties. A lone registration is always restored even if its hero is gone, otherwise that player would be sent to character creation for no reason.

`SaveGameHandlerTests` covers all three, with the duplicate case run in both orders.

ShoT-UPfps (approved):
lgtm

PR #2597, opened 2026-08-02
Harden mission membership and relay authorization

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Rapid mission transitions, reconnects, or stale leave messages can leave players registered in the wrong battle, publish incorrect departure or host-migration events, keep empty battles occupied, or forward relay traffic between different missions and players who already left.

## What is the new behavior?

The server now uses each player's current connection identity, keeps one authoritative mission membership per player, ignores failed leaves, removes every stale membership on disconnect, and forwards relays only when the sender and recipient are current members of the same named mission. Battle membership, departures, host migration, cleanup, and relay routing now remain consistent through rapid transitions and reconnects.


source/Coop.Core/Server/Services/Instances/Handlers/ServerMissionMembershipHandler.cs
```diff
@@ -55,33 +60,40 @@ public void Dispose()
 
     private void Handle_MissionEntered(MessagePayload<NetworkMissionEntered> payload)
     {
-        var peer = (NetPeer)payload.Who;
+        if (payload.Who is not NetPeer peer)
+            return;
+
         var message = payload.What;
+        if (!TryGetCurrentController(peer, out var controllerId))
```
garrettluskey, line 71: this drops the message without logging anything, and `Handle_MissionLeft` does the same at line 127. since this gate is now the main reason an entry or leave never lands, can we log when it rejects? otherwise a membership that never shows up in `TryGetControllers` is hard to trace back to this line. Debug level is probably right because a replaced peer tripping it is expected.

source/Coop.Core/Server/Services/Instances/Handlers/ServerMissionMembershipHandler.cs
```diff
@@ -55,33 +60,40 @@ public void Dispose()
...
             {
                 return;
             }
 
-            messageBroker.Publish(this,
-                new MissionMemberEntered(message.ControllerId, message.InstanceId, isFirstMember));
+            foreach (var departure in result.PreviousDepartures)
+                PublishDeparture(departure, wasRetreat: true);
+
+            if (result.Status == MissionEntryStatus.Entered)
```
garrettluskey: reconnected entries skip `MissionMemberEntered`, so `completionTracker.ResetMember` never runs for them. before this change a re-entry always published the event and wiped the member's completion state. in the crash-reconnect case the old peer's `HandleDisconnect` finds no membership after the route is replaced, so no departure publishes either. so a result report the player submitted before the crash now survives the reconnect and can conclude the battle later. is that intended, or should a reconnect reset their report the way the old re-entry did?

garrettluskey (changes requested):
the membership rework reads well: one authoritative membership per controller/peer pair under the `gate`, identity checked against `PlayerManager` before acting, and the relay fence that pairs every `RevokeRelay` with the queued leave/disconnect. the new unit and integration tests cover the duplicate-entry, reconnect, and stale-leave races, and the `ConcurrentDictionary` to `HashSet` swap is safe since every read now goes through a locked `MissionManager` method. two things below: a dropped-message path that logs nothing, and a question about what a reconnect does to battle completion state.

source/Coop.Core/Server/Services/Instances/Handlers/ServerMissionMembershipHandler.cs
ShoT-UPfps: added Debug logs for rejected entry and leave messages in `4ec9858`.

source/Coop.Core/Server/Services/Instances/Handlers/ServerMissionMembershipHandler.cs
ShoT-UPfps: fixed in `4ec9858`. reconnects now publish `MissionMemberEntered` so the completion report resets, while exact duplicate entries still skip it. added tests for both paths.

garrettluskey (approved):
the two things from last round are in: rejected entries and leaves log at debug now, and reconnects publish `MissionMemberEntered` again so the completion state resets while exact duplicate entries still skip it. i walked the relay fence pairing across the leave/disconnect/reconnect orderings and the new identity gate against `PlayerManager` (the peer link is set on the poll thread during the join handshake, before any mission message, so the gate can't race a reconnect), and it holds up. the `TryBindComponent` repair at commit matches the barrier flow too. build is still running on this push.

PR #2599, opened 2026-08-02
[2583] Keep party state current after companion conversations

### LINKED ISSUE #2583: [COMPANIONS] Party does not refresh after upgrading units and talking to a companion without saving
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

Upgrading units and then talking to a companion before saving leaves the client with an out-of-date view of the party/companions.

## Steps to reproduce

1. Upgrade one or more units.\n2. Without saving, talk to a companion.\n3. Notice the client-side party state becomes incorrect until reconnecting.

## Additional information

To correct the companion/group state, the player must quit the server and rejoin.

## Discord source

https://discord.com/channels/709516043332354119/1533277119327506453/1533277119327506453

## Logs

Client log: [0.1.1+87a4520ab6fd2a9e4e6aad964ccd395296a507fc - Download Coop_client.log](<https://cdn.discordapp.com/attachments/1533277119327506453/1533277120044859583/Coop_client.log?ex=6aa4a318&is=6aa35198&hm=0c561a2d7e0b30d8903cf8e16da15b5509e3b8f79bca4a1cb234580b397abac0&>)

Server log: [Download coop-server-20260801-184419.log](<https://cdn.discordapp.com/attachments/1533277119327506453/1533277119734612148/coop-server-20260801-184419.log?ex=6aa4a318&is=6aa35198&hm=911041de2ab57c708b03adbd52d31e931dc71f583a5de545e918768b99069cd5&>)

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Upgrading troops and then assigning a companion role through conversation can leave the client's party and companion information out of date until the player reconnects.

## What is the new behavior?

Resolves #2583

Party and companion information now stays current after troop upgrades and companion role conversations, without requiring a reconnect.

## Bot Changelog Entry

- Fixed an issue where your party could stop refreshing after you talk to a companion


source/GameInterface/Services/Companions/Commands/CompanionsCommands.cs
```diff
@@ -68,6 +73,241 @@ public static string ListWanderersCommand(List<string> strings)
...
+        if (args.Count != 0) return usage;
+        var companion = FindRoleFixtureCompanion();
+        if (companion == null) return "The companion-role fixture hero was not found.";
+        if (Hero.MainHero == null) return "The local main hero is unavailable.";
+
+        bool initializedExSpouses = companion.ExSpouses == null;
+        if (initializedExSpouses)
+        {
+            AccessTools.Field(typeof(Hero), nameof(Hero._exSpouses))
+                .SetValue(companion, new MBList<Hero>());
```
garrettluskey: can we skip the reflection here? `TaleWorlds.CampaignSystem` is on the publicize list and we already touch this field directly elsewhere (`HeroBinaryPackage.cs:48` reads `Object._exSpouses`), so this can just be `companion._exSpouses = new MBList<Hero>();`. the `HarmonyLib` using goes away too.

source/GameInterface/Services/Clans/Handlers/ClanManagementRefreshHandler.cs
```diff
@@ -1,92 +1,85 @@
...
-    }
-
     private void Handle_RefreshPartiesList(MessagePayload<RefreshPartiesList> obj)
     {
         GameThread.RunSafe(() =>
         {
-            currentClanManagementVM?.ClanParties?.RefreshPartiesList();
-            currentClanManagementVM?.ClanMembers?.RefreshMembersList(); // Needed to refresh clan members who can be party leaders
-        });
+            if (!(ScreenManager.TopScreen is GauntletClanScreen clanScreen)) return;
```
garrettluskey, line 39: is there any flow where the clan screen is alive but not the top screen, e.g. the party screen pushed over it? if so a refresh arriving then gets dropped while the screen is still around, and the old cached VM code did refresh in that case. if the screens are exclusive ignore this.

garrettluskey (changes requested):
binding the clan refreshes to the live screen instead of a cached VM is the right move, it drops the VM reference the handler held forever and the constructor patch, and it matches what `ClanNameHandler` and `HeroDataHandler` already do. the debug fixture coverage for the live conversation flow is thorough.

one change: the `_exSpouses` init reaches for reflection on a field thats already publicized. plus a question on the `TopScreen` gating below.

source/GameInterface/Services/Companions/Commands/CompanionsCommands.cs
ShoT-UPfps: _exSpouses is publicized but it's still readonly, so the direct assignment fails with CS0191. i switched this to the existing HeroBinaryPackage.Hero_ExSpouses setter and removed the Harmony dependency.

source/GameInterface/Services/Clans/Handlers/ClanManagementRefreshHandler.cs
ShoT-UPfps: the clan screen can stay underneath party, but when it becomes active again GauntletClanScreen.OnActivate calls RefreshCategoryValues(), which rebuilds fiefs, members, parties, and income. RefreshPartiesList() also recreates the role VMs from current party state, so dropping this while clan isn't top won't leave stale data.

garrettluskey (approved):
both points from last round are handled. the ex-spouses init reuses the existing `HeroBinaryPackage.Hero_ExSpouses` setter now (direct assignment is out because the field is readonly), and the HarmonyLib using is gone. on the TopScreen question: i decompiled `GauntletClanScreen`, and `OnActivate` builds a fresh `_dataSource` every time the screen activates, so a refresh dropped while clan sits under another screen gets rebuilt when it comes back to the top. the one codefactor hit is just the advisory complex-method flag on the new debug command.

PR #2601, opened 2026-08-02
[2107] Hide stale and duplicate Steam lobby entries

### LINKED ISSUE #2107: [DEDICATED SERVER] Dedicated server stays in Steam lobbies after the game executable is killed
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

After the game executable is killed while the client is running, a dedicated server already visible in Steam lobbies remains listed even after refreshing or returning to the main menu. The server is only removed when a connection attempt is made. No screens or logs were provided.

## Steps to reproduce

1. Have a dedicated server visible in Steam lobbies.
2. While the client is running, kill the game executable process.
3. Return to the main menu or refresh the Steam lobbies list.
4. The server remains listed.
5. Attempt to connect to it; the server then disappears.

## Additional information

The reporter reported no screens and no logs for this issue.

## Discord source

https://discord.com/channels/709516043332354119/1527722004613038300/1527722004613038300

## Logs

Client log: Not provided

Server log: Not provided

## Consolidated duplicate evidence

- #2305 reports the same stale Steam lobby entry after the dedicated server closes; it contains no additional reproduction details or attachments.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Killing a dedicated server can leave its Steam lobby visible, and restarting the server can make the same server appear multiple times in the Steam lobby browser.

## What is the new behavior?

Resolves #2107

Expired dedicated-server lobbies disappear from the Steam lobby browser, duplicate entries from the same live server are shown once, and distinct live servers remain visible.

## Bot Changelog Entry

- Fixed an issue where quitted servers can stil lshow up in the Steam lobby browser.


source/Coop.Steam/SteamLobbyLeaseRenewer.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Coop.Steam;
+
+public interface ISteamLobbyLeaseRenewer : IDisposable
+{
+    void Start(Action renew);
+    void Stop();
+}
+
+/// <summary>Schedules required Steam lobby lease writes on the game thread.</summary>
+public class SteamLobbyLeaseRenewer : ISteamLobbyLeaseRenewer
```
garrettluskey: new file came in with LF endings. everything else in the repo is CRLF and there's no .gitattributes to normalize it on checkout, so please convert this file to CRLF (the BOM is already there).

source/Coop.Steam/SteamLobbyLeaseRenewer.cs
```diff
@@ -0,0 +1,55 @@
...
+    private readonly TimeSpan interval;
+    private Timer timer;
+    private bool disposed;
+
+    public SteamLobbyLeaseRenewer()
+        : this(RenewalInterval)
+    {
+    }
+
+    internal SteamLobbyLeaseRenewer(TimeSpan interval)
```
garrettluskey: this internal ctor is never called, DI binds the default one and no test passes an interval. drop it, or give it a test if you want the custom interval covered.

garrettluskey (dismissed):
the lease approach here reads well: the server renews a 60 second lease every 15 seconds through `GameThread.RunSafe`, the browser drops expired entries and keeps only the newest lease per `ServerSteamId`, and every ambiguous case fails open so a live server never gets hidden. coverage looks good too, renewal, steam time going away, repeated renewal failures, and the dedupe paths all have tests, and the live-test fixtures from the earlier commits are fully backed out of the net diff.

two small fixes, then this is good to go: the new file needs CRLF endings, and the unused internal ctor should either go away or get used. codefactor's `Complex method` flag on `BuildSummaries` is advisory and i wouldnt change the code for it.

source/Coop.Steam/SteamLobbyLeaseRenewer.cs
ShoT-UPfps: fixed in 7c545e8, removed the unused ctor and use RenewalInterval directly.

source/Coop.Steam/SteamLobbyLeaseRenewer.cs
ShoT-UPfps: fixed in 7c545e8, the committed file is now BOM + CRLF.

garrettluskey (approved):
both earlier notes are fixed: `SteamLobbyLeaseRenewer.cs` is BOM + CRLF now, and the unused ctor is gone with `Start` using `RenewalInterval` directly. re-read the whole PR at the new head and it holds together: renewals marshal onto the game thread with `GameThread.RunSafe`, three failed renewals in a row withdraw and recreate the advertisement, and the browser fails open everywhere it cant be sure (missing or malformed lease, tied expiries, steam time unavailable). a `uint.MaxValue` lease written when steam time is down gets replaced by a real expiry on the next successful renewal, and a metadata update during a time outage keeps both the published lease and the renewal failure count, with tests pinning that down. codefactor's Complex Method on `BuildSummaries` is still advisory, i wouldnt change the code for it.

PR #2602, opened 2026-08-02
[2514] Prevent field-battle replication starvation

### LINKED ISSUE #2514: [FIELD BATTLES] Players split into separate instances during a two-player field battle
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

In a 2-player field battle, each player can enter a separate instance of the same battle. The other player’s units are not visible, and neither side sees the other’s army.

## Steps to reproduce

1. Start a 2-player field battle.
2. Join the same field-battle instance together.
3. Observe that each player only sees their own units and cannot see the other player’s army.

## Additional information

Game version is 1.4.7. Logs show the issue around 00:15. Screenshots were provided and show late-battle blank army stats for the friend.

## Discord source

https://discord.com/channels/709516043332354119/1532565057982959676/1532565057982959676

## Images and videos

[](<https://cdn.discordapp.com/attachments/1532565057982959676/1532565058201194546/20D5951.JPG?ex=6aa4aeef&is=6aa35d6f&hm=6687540c9686619bf0f4b9bd9d888e0ea4be4225f4e2019fe5b751a9079e1c4f&>) ([Discord message](https://discord.com/channels/709516043332354119/1532565057982959676/1532565057982959676))

## Logs

Client log: [0.0.3+fc991dfbeae26e0cfc3958f7be30b01d5b6ac535 - Download Coop_client_4.log](<https://cdn.discordapp.com/attachments/1532565057982959676/1532565336224829500/Coop_client_4.log?ex=6aa4af31&is=6aa35db1&hm=5eb15efa24d30ab0701d4cfc0ad04ff904f6cca79f2dd733b514e447f45389ce&>) · [0.0.3+fc991dfbeae26e0cfc3958f7be30b01d5b6ac535 - Download Coop_client.log](<https://cdn.discordapp.com/attachments/1532565057982959676/1532565336585670886/Coop_client.log?ex=6aa4af31&is=6aa35db1&hm=6ee9f2d79069b40f2c7efc3d07e1897db325273fdf2ce00daca94c56021a4395&>)

Server log: [Download coop-server-20260730-232038.log](<https://cdn.discordapp.com/attachments/1532565057982959676/1532565337038389438/coop-server-20260730-232038.log?ex=6aa4af32&is=6aa35db2&hm=86d32af5164afe31e515bc1c269f6a0d98cc2bb9a2d75c952eb6a1ae030df21f&>)

## Consolidated related reports

- #2169: Player and troop visibility can desynchronize in only one direction. The invisible side can still attack, and multiple client-log reports reproduce the symptom.
- #2450: Siege and field battles above 1,000 battle size can diverge in party participation, troop counts, movement, gate state, and damage. That report includes two client logs, a server log, and a save.
- #1943: Village-raid puppets can appear in the wrong location while their attacks still damage the client.

The original issue bodies, discussion, screenshots, logs, and save remain available in the linked duplicate issues.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Large two-player field battles can leave each player seeing only their own army while the other side's troops arrive tens of seconds late or never become usable, making both players appear to be fighting in separate battle instances.

## What is the new behavior?

Resolves #2514

Field-battle troops now reach both players before deployment and activation complete, while movement traffic stays responsive without delaying battle state. Late joiners also receive the current troops and deployment state before their battle is activated.

## Bot Changelog Entry

- Fixed large field battles appearing split between players because other armies arrived late or not at all. Both players must update before playing together.


source/Missions/Battles/CoopBattleController.cs
```diff
@@ -376,23 +378,29 @@ protected override void SendJoinInfo(string controllerId)
...
+                    Session.HostEpoch,
+                    battleState));
+            }
         }
+
+        // An active battle needs a barrier so later reliable sends cannot pass its replay. During mission
+        // loading the game-thread queue does not drain yet, so the ordinary pre-activation handshake defers.
+        GameThread.RunSafe(
+            ReplayJoinState,
+            blocking: Deployment.IsActivated,
```
garrettluskey: `GameThread.Run` has a 30s `BlockingTimeout`, and `RunSafe` only guards the action — the `TimeoutException` / `OperationCanceledException` from the wait itself escapes this handler onto the Poller thread. We've been burned by Poller-thread throws before (silent network death). The action stays queued after a timeout, so catching here degrades to the non-blocking path instead of risking the poll loop:

```csharp
try
{
    GameThread.RunSafe(ReplayJoinState, blocking: Deployment.IsActivated, context: nameof(SendJoinInfo));
}
catch (Exception e) when (e is TimeoutException || e is OperationCanceledException)
{
    Logger.Warning(e, "[BattleSync] Join replay barrier for {Controller} was abandoned; the replay stays queued", controllerId);
}
```

source/Missions/Battles/CoopBattleController.cs
```diff
@@ -376,23 +378,29 @@ protected override void SendJoinInfo(string controllerId)
...
+            {
+                network.Send(controllerId, new NetworkBattleResultSnapshot(
+                    Session.InstanceId,
+                    Session.OwnControllerId,
+                    Session.HostEpoch,
+                    battleState));
+            }
         }
+
+        // An active battle needs a barrier so later reliable sends cannot pass its replay. During mission
```
garrettluskey: Worth spelling out here that the strict snapshots-before-activation guarantee only holds on this blocking path. A pre-activation joiner can still receive `NetworkBattleActivated` from our inline commit broadcast before the queued replay goes out — benign today (puppets spawn unpaused, the activation receipt is idempotent), but without the caveat written down someone could later "fix" it into a load-time deadlock.

source/Missions/Battles/PuppetSpawner.cs
```diff
@@ -98,27 +102,63 @@ public void Dispose()
...
+            context: nameof(Handle_NetworkSpawnBattleAgents));
     }
 
-    private void SpawnPuppet(BattleAgentSpawnData data)
+    private void SpawnPuppetBatch(
+        NetworkSpawnBattleAgents message,
+        BattleAgentSpawnData[] agents)
     {
-        if (data.AgentId == Guid.Empty) return;
+        if (Mission.Current == null) return;
```
garrettluskey: This drops the whole decoded batch with no trace — the early return also skips the "Applied" log below. It should only happen during teardown, since the queue doesn't drain during load, but a one-line log here will save whoever ends up debugging vanished troops.

source/Missions/Battles/PuppetSpawner.cs
```diff
@@ -375,8 +415,9 @@ public void DrainPendingPuppets()
         lock (pendingPuppetLock)
         {
             if (pendingPuppets.Count == 0) return;
-            pending = pendingPuppets.ToArray();
-            pendingPuppets.Clear();
+            int count = Math.Min(MaxBufferedSpawnsPerTick, pendingPuppets.Count);
+            pending = pendingPuppets.GetRange(0, count).ToArray();
```
garrettluskey: Nit: `GetRange` already allocates a copy, then `ToArray` copies it again. Allocate the array and `pendingPuppets.CopyTo(0, pending, 0, count)` does it in one.

source/Missions/Agents/Handlers/MovementTrafficBudget.cs
```diff
@@ -0,0 +1,96 @@
...
+    void ReportFrame(int deferredSnapshots, float maximumDeferredAgeSeconds);
+    void Clear();
+}
+
+/// <summary>Per-route token bucket that leaves most of Steam's minimum send rate available to reliable traffic.</summary>
+public sealed class MovementTrafficBudget : IMovementTrafficBudget
+{
+    private static readonly ILogger Logger = LogManager.GetLogger<MovementTrafficBudget>();
+
+    internal const int DefaultBytesPerSecond = 1024 * 1024;
```
garrettluskey, line 22: `TrySpend` counts payload bytes once per `SendAll`, but actual egress is that times peer count. Fine for the 2-player battles this was tuned on; on a bigger mesh the effective cap drifts. Worth a comment on this constant so the next person tuning it knows the assumption.

source/Missions/Agents/Handlers/MovementBatchSender.cs
```diff
@@ -10,25 +10,52 @@ namespace Missions.Agents.Handlers;
...
+    void EndFrame(int deferredSnapshots, float maximumDeferredAgeSeconds);
 
     void Clear();
 }
 
+public readonly struct MovementSendResult
+{
+    public int SentCount { get; }
+    public int DeferredCount { get; }
+    public int SentBytes { get; }
```
garrettluskey: `SentBytes` is computed and summed but nothing reads it — either feed it into the `EndFrame` telemetry or drop it.

source/E2E.Tests/Services/Missions/MountedPuppetMovementTests.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using System.Reflection;
+using Common;
```
garrettluskey: Leftover using? Nothing else changed in this file.

source/E2E.Tests/Services/Missions/BattleAgentSpawnBatchCodecTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            encoded.BatchIndex,
+            encoded.BatchCount,
+            encoded.Purpose,
+            encoded.PayloadSha256,
+            agents: null!);
+
+        Assert.False(codec.TryDecode(wire, out _));
+    }
+
+    private static BattleAgentSpawnData[] DecodeWireMessage(
```
garrettluskey, line 118: These tests round-trip the inner payload through real protobuf, but the outer `NetworkSpawnBattleAgents` never touches the serializer — every wire message here is hand-built. Given our surrogate history (readonly fields, the `Guid` TransferId, the enum), one `DeepClone` round-trip assert on an encoded batch is cheap insurance.

garrettluskey (dismissed):
The ordering fix is right — troops before the finished/activated markers on one ReliableOrdered stream, and marshaling the marker handlers onto the game-thread queue so they can't pass earlier spawn applies. Recording last-sent movement state only on actual send is also the correct call; without it the new deferral would silently eat deltas. Test coverage is strong across the board.

Requesting changes for one thing: the blocking join barrier can throw on the network thread (see inline). The rest is notes and nits.

One thing for the changelog: this changes the wire format of `NetworkSpawnBattleAgents` (field 1 went from a record array to bytes), so old and new builds can't do battles together. Module validation should catch the mismatch at join, but call out that both players need to update.

garrettluskey (approved):
all addressed. the DeepClone round trip is the right shape, `Agents` is ProtoIgnore so the clone hits the real payload decode path instead of the in-process bypass. lgtm

PR #2604, opened 2026-08-02
Command and support for resetting skills

jordanbrymora:
## PR Checklist

## PR Type

## Bot Changelog Entry

Add a command to reset character skills that can be used until arenas are implemented.

### DIFF
--- a/source/GameInterface/Services/HeroDevelopers/Commands/HeroDeveloperCommands.cs
+++ b/source/GameInterface/Services/HeroDevelopers/Commands/HeroDeveloperCommands.cs
@@ -107,6 +107,33 @@ internal class HeroDeveloperCommands
         else return $"Unable to find hero with name or id of {strings[0]}";
     }
 
+    /// <summary>
+    /// Reset all skills of a hero and give focus/attribute points back based on level.
+    /// Example:
+    /// coop.debug.herodeveloper.resetskills
+    /// </summary>
+    [CommandLineArgumentFunction("resetskills", "coop.debug.herodeveloper")]
+    public static string ResetSkillsCommand(List<string> strings)
+    {
+        if (ModInformation.IsClient) return "Command can only be run on the server.";
+
+        if (strings.Count != 1) return "Usage: resetskills heroName";
+
+        StringBuilder stringBuilder = new StringBuilder();
+        foreach (var hero in Hero.AllAliveHeroes)
+        {
+            if (hero.Name.ToString() == strings[0] || hero.StringId == strings[0])
+            {
+                hero.HeroDeveloper.ResetCharacterStats();
+
+                stringBuilder.AppendLine($"{strings[0]}'s skills were reset.");
+            }
+        }
+
+        if (stringBuilder.Length > 0) return stringBuilder.ToString();
+        else return $"Unable to find hero with name or id of {strings[0]}";
+    }
+
     private static SkillObject GetSkillByName(string skillName)
     {
         var property = typeof(DefaultSkills).GetProperty(skillName, BindingFlags.Public | BindingFlags.Static | BindingFlags.IgnoreCase);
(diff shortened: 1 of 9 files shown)


source/GameInterface/Services/HeroDevelopers/Patches/ResetSkillsPatches.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    [HarmonyPatch(nameof(PerkHelper.ClearPerksForSkill))]
+    [HarmonyPostfix]
+    public static void ClearPerksForSkill(Hero hero)
+    {
+        if (ModInformation.IsClient) return;
+
+        hero.PartyBelongedTo?.MemberRoster?.UpdateVersion();
+
+        var message = new UpdateRosterVersionAfterPerkChange(hero.PartyBelongedTo.MemberRoster);
```
ShoT-UPfps, line 52: `hero.PartyBelongedTo` can still be null here even though line 34 handles it. dont publish the roster update when there isnt a party, otherwise resetting a partyless hero throws.

source/GameInterface/Services/HeroDevelopers/Handlers/ClearFocusesHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    private void Handle_NetworkClearFocuses(MessagePayload<NetworkClearFocuses> obj)
+    {
+        var data = obj.What;
+
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<HeroDeveloper>(data.HeroDeveloperId, out var heroDeveloper)) return;
+
+            heroDeveloper.ClearFocuses();
```
ShoT-UPfps: we already autosync `HeroDeveloper._newFocuses`, including `ClearAllProperty`, so this sends and applies the clear twice. i think this handler and the manual messages can be removed, the second call is outside `AllowedThread` and logs `Client updated unmanaged SkillObject`.

source/GameInterface/Services/HeroDevelopers/Commands/HeroDeveloperCommands.cs
```diff
@@ -107,6 +107,33 @@ public static string AddFocusPointsCommand(List<string> strings)
...
+        if (ModInformation.IsClient) return "Command can only be run on the server.";
+
+        if (strings.Count != 1) return "Usage: resetskills heroName";
+
+        StringBuilder stringBuilder = new StringBuilder();
+        foreach (var hero in Hero.AllAliveHeroes)
+        {
+            if (hero.Name.ToString() == strings[0] || hero.StringId == strings[0])
+            {
+                hero.HeroDeveloper.ResetCharacterStats();
```
ShoT-UPfps, line 127: this cheat probably shouldnt be run on the server, but our users have other ideas. `ResetCharacterStats()` calls vanilla `PerkHelper.ClearPerksForSkill`, which uses `PartyBase.MainParty.MemberRoster`; the server has no main party so this throws partway through the reset.

source/GameInterface/Services/HeroDevelopers/Handlers/ClearFocusesHandler.cs
jordanbrymora: I've already tested without it. ClearAllProperty doesn't sync with clients which is why I added this manually.

Moved the client handle into allowed thread to suppress the error.

source/GameInterface/Services/HeroDevelopers/Patches/ResetSkillsPatches.cs
jordanbrymora: Fixed

source/GameInterface/Services/HeroDevelopers/Commands/HeroDeveloperCommands.cs
jordanbrymora: Changed to use prefix replacing the entire function to avoid `MainParty` call on server.

ShoT-UPfps (approved):
lgtm

PR #2638, opened 2026-08-03
Publish nightly client installer artifacts

ShoT-UPfps:
## PR Type

- [x] CI related changes

The nightly client is only published to the private server-build handoff and the short-lived GitHub artifact.

## What is the new behavior?

The workflow also uploads the client to the public nightly R2 bucket and publishes `nightly/client.json` after the archive is verified. The public manifest lets the guided installer find the latest completed client nightly without exposing R2 credentials.

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -239,6 +239,7 @@ jobs:
           HEAD_SHA: ${{ steps.release.outputs.head_sha }}
         run: |
           set -eu
+          apk add --no-cache rclone >/dev/null
           test -n "$R2_ACCOUNT_ID"
           test -n "$R2_PUBLIC_BUCKET"
           export RCLONE_CONFIG_R2_ENDPOINT="https://$R2_ACCOUNT_ID.r2.cloudflarestorage.com"
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -21,6 +21,8 @@ concurrency:
 
 env:
   R2_PRIVATE_BUCKET: bannerlordcoop-build-inputs
+  R2_PUBLIC_BUCKET: bannerlordcoop-nightly-releases
+  R2_PUBLIC_BASE_URL: https://pub-bf6bfe4b880e4d1b83f4b09b10419f78.r2.dev
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-release.yml
```diff
@@ -225,6 +227,37 @@ jobs:
...
+          RCLONE_CONFIG_R2_TYPE: s3
+          RCLONE_CONFIG_R2_PROVIDER: Cloudflare
+          RCLONE_CONFIG_R2_ACCESS_KEY_ID: ${{ secrets.R2_PUBLIC_ACCESS_KEY_ID }}
+          RCLONE_CONFIG_R2_SECRET_ACCESS_KEY: ${{ secrets.R2_PUBLIC_SECRET_ACCESS_KEY }}
+          R2_ACCOUNT_ID: ${{ secrets.R2_ACCOUNT_ID }}
+          RELEASE_DATE: ${{ steps.release.outputs.release_date }}
+          FILE_NAME: ${{ steps.release.outputs.file_name }}
+          HEAD_SHA: ${{ steps.release.outputs.head_sha }}
+        run: |
+          set -eu
```
garrettluskey, line 241: rclone isn't installed in this step, it's only on the path because the previous step ran `apk add --no-cache rclone` in the same job. the other two rclone steps in the repo (nightly-release.yml:212 and manual-build.yml:176) each install it themselves. please add `apk add --no-cache rclone >/dev/null` after `set -eu` otherwise this step breaks if the private publish step is ever removed or reordered.

garrettluskey (dismissed):
shape of this looks right: archive goes up first, remote size is verified with `rclone lsl`, and `client.json` is published last so installers never see incomplete bytes, same handoff pattern as the private bucket step. one small change requested: the new step uses rclone without installing it, it only works because the previous step happens to install it in the same job.

.github/workflows/nightly-release.yml
ShoT-UPfps: updated

garrettluskey (approved):
The rclone issue is fixed: the public publish step installs it itself now instead of leaning on the previous step, and it got test -n guards on the account id and bucket while it was in there. Everything else held up from the last pass: archive uploads before the manifest, remote size is verified against the local archive before client.json goes up, and the public manifest carries no secrets.

PR #2640, opened 2026-08-03
List the game binaries and their versions in crash reports

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

A coop crash report folder holds `report.txt`, the copied logs, `dump.dmp` and `shareable.zip`. It records the mod build, but nothing about the install it was running against, so a stale, partially updated or hand-patched `bin/Win64_Shipping_Client` is indistinguishable from a real bug when reading a dump.

## What is the new behavior?

Crash reports now include `game-binaries.json`, listing every file in the game's binaries directory with its size, last write time and, for DLLs and EXEs, its file and product version. It is written into the report folder and into `shareable.zip`, and `report.txt` gains a `Game binaries directory:` / `Game binaries:` summary line.

- The module resolves the directory through Bannerlord's own helpers — `BasePath.Name` + `Common.ConfigName` — and passes it to the reporter in `COOP_CRASH_GAME_BINARIES`. The reporter is a plain process with no TaleWorlds assemblies, so the resolution has to happen module-side. If the resolved path does not exist (headless hosts), it falls back to the executable's own directory; if nothing resolves, the report says so instead of guessing.
- Traversal stops two directories deep. A full walk pulls in the bundled `mono` tree — 5473 files, 1.1 MB of JSON and 99 s of cold disk right after the game died. Bounded, a real install is 736 files, 172 KB and 0.4 s, and still covers `Microsoft.NETCore.App/<version>` and `Watchdog`. Directories that were not descended into are counted in the manifest (`skippedDirectories`) and in the `report.txt` summary rather than silently dropped.
- Unreadable subdirectories and files are skipped individually, so one locked folder cannot cost the whole manifest, and a failure to resolve or write it never blocks the rest of the report.

Verified against the real install: 736 entries, valid JSON, TaleWorlds DLLs listed with size and timestamp. `Coop.CrashReporter.Tests` covers the listing, nested paths, the depth limit, path escaping, and both "not captured" paths (9 tests green); `Coop` and `Coop.CrashReporter` build clean.

## Bot Changelog Entry

- Crash reports now include a list of the game's installed files and their versions, making it easier to spot a broken or out-of-date game install.

## Other information

Sample entry:

```json
{ "path": "TaleWorlds.CampaignSystem.dll", "bytes": 5462464, "modifiedUtc": "2026-07-25T19:47:30.5117697Z", "fileVersion": "1.0.0.0", "productVersion": "1.0.0" }
```

### DIFF
--- a/source/Coop.CrashReporter/CrashReportCollector.cs
+++ b/source/Coop.CrashReporter/CrashReportCollector.cs
@@ -58,6 +58,15 @@ namespace Coop.CrashReporter
         {
             process.WaitForExit();
             int exitCode = process.ExitCode;
+
+            // Walked before anything waits on a dump: the zero exit-code path spends up to
+            // 30 seconds looking for one, and a launcher update inside that window would rewrite
+            // the very files being listed. A clean exit pays for a walk whose result is then
+            // thrown away, but the game's own binaries are still in the file cache it just used
+            // them from, so that is a fraction of a second in a process about to exit anyway.
+            GameBinariesManifest binaries = GameBinariesManifest.Capture(
+                options.GameBinariesDirectory);
+
             string dumpPath = exitCode == 0
                 ? WaitForMatchingDump(CrashDumpDiscoveryTimeout)
                 : null;
@@ -69,14 +78,9 @@ namespace Coop.CrashReporter
             Directory.CreateDirectory(logsDirectory);
             List<string> copiedLogs = CopyLogs(logsDirectory);
 
-            // Written before the dump wait so the listed versions are the ones that were on
-            // disk at the crash, not whatever a launcher update replaced them with meanwhile.
             string manifestPath = Path.Combine(reportDirectory, GameBinariesManifest.FileName);
             bool manifestWritten;
-            string binariesSummary = GameBinariesManifest.Write(
-                manifestPath,
-                options.GameBinariesDirectory,
-                out manifestWritten);
+            string binariesSummary = binaries.Write(manifestPath, out manifestWritten);
 
             if (dumpPath == null)
                 dumpPath = WaitForMatchingDump(CrashDumpDiscoveryTimeout);
--- a/source/Coop.CrashReporter/CrashReportCollector.cs
+++ b/source/Coop.CrashReporter/CrashReportCollector.cs
@@ -69,6 +69,15 @@ namespace Coop.CrashReporter
             Directory.CreateDirectory(logsDirectory);
             List<string> copiedLogs = CopyLogs(logsDirectory);
 
+            // Written before the dump wait so the listed versions are the ones that were on
+            // disk at the crash, not whatever a launcher update replaced them with meanwhile.
+            string manifestPath = Path.Combine(reportDirectory, GameBinariesManifest.FileName);
+            bool manifestWritten;
+            string binariesSummary = GameBinariesManifest.Write(
+                manifestPath,
+                options.GameBinariesDirectory,
+                out manifestWritten);
+
             if (dumpPath == null)
                 dumpPath = WaitForMatchingDump(CrashDumpDiscoveryTimeout);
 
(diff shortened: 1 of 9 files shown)


source/Coop.CrashReporter/CrashReportCollector.cs
```diff
@@ -69,6 +69,15 @@ public int Run(Process process)
             Directory.CreateDirectory(logsDirectory);
             List<string> copiedLogs = CopyLogs(logsDirectory);
 
+            // Written before the dump wait so the listed versions are the ones that were on
+            // disk at the crash, not whatever a launcher update replaced them with meanwhile.
+            string manifestPath = Path.Combine(reportDirectory, GameBinariesManifest.FileName);
+            bool manifestWritten;
+            string binariesSummary = GameBinariesManifest.Write(
```
ShoT-UPfps: the zero exit-code crash path already waited up to 30 seconds in `WaitForMatchingDump` before this runs. can we capture the manifest before that wait, otherwise a launcher update can replace the files before we list them?

source/Coop.CrashReporter/GameBinariesManifest.cs
```diff
@@ -0,0 +1,330 @@
...
+            {
+                return new string[0];
+            }
+        }
+
+        private static string[] TryGetFiles(string directory)
+        {
+            try
+            {
+                return Directory.GetFiles(directory);
```
ShoT-UPfps: `MaximumFileCount` doesnt bound this because `Directory.GetFiles` allocates every path before the loop can stop at 20,000. can we enumerate lazily or bounded here, otherwise a large unpacked directory can still kill the crash reporter before it writes the report?

source/Coop.CrashReporter/GameBinariesManifest.cs
```diff
@@ -0,0 +1,330 @@
...
+        // An unreadable subdirectory must cost its own contents, not the whole manifest.
+        private static string[] TryGetDirectories(string directory)
+        {
+            try
+            {
+                return Directory.GetDirectories(directory);
+            }
+            catch (IOException)
+            {
+                return new string[0];
```
ShoT-UPfps: use `Array.Empty<string>()` instead of `new string[0]` here and in `TryGetFiles`

ShoT-UPfps (changes requested):
comments

source/Coop.CrashReporter/CrashReportCollector.cs
garrettluskey: Good catch — done in deac486e9. Capturing and writing are split now: `GameBinariesManifest.Capture` runs immediately after `process.WaitForExit()`, before either `WaitForMatchingDump` call, and `Write` just persists that snapshot once the report directory exists. `capturedUtc` is now the walk time rather than the write time too.

A clean exit does pay for a walk that gets thrown away, but the game's binaries are still in the file cache it just used them from, so it's ~0.4s in a process that's about to exit. Pinned by `Capture_SnapshotsTheDirectoryBeforeTheManifestIsWritten`, which replaces a file between capture and write and asserts the manifest still reports the original.

source/Coop.CrashReporter/GameBinariesManifest.cs
garrettluskey: Right, the cap was doing nothing there. Both walks enumerate lazily now (`EnumerateSafely` wraps `Directory.EnumerateFiles`/`EnumerateDirectories` and swallows per-directory IO/access failures around `MoveNext`), so `MaximumFileCount` actually stops the walk mid-directory.

While in there I capped the queue the same way — `MaximumDirectoryCount` (5000) — since a directory with a million children could otherwise grow the pending stack even though the file cap held. Anything dropped by either cap is still counted in `skippedDirectories`/`truncated` and surfaces in the report.txt summary line.

source/Coop.CrashReporter/GameBinariesManifest.cs
garrettluskey: Both `new string[0]` sites are gone entirely — the lazy rewrite replaced `TryGetDirectories`/`TryGetFiles` with an iterator that just `yield break`s on failure, so there's no empty array to allocate anymore.

ShoT-UPfps (approved):
lgtm

PR #2648, opened 2026-08-03
[2320] Show a connecting screen with a cancel for every join attempt

### LINKED ISSUE #2320: Add Connecting Popup
# For New Contributors

## Description

Add a connecting to server popup with a cancel button.

When the cancel button is clicked, cancel the trying to connect loop

## Location

## Definition of Done

douglasholm:
## PR Checklist

## PR Type

A join attempt gives the player nothing to look at and no way out. The Join screen stays up while the client dials, and on failure it silently redials every 3 seconds. Cancelling only closes the screen — the retry loop keeps running behind the menu, so if the target server comes up minutes later the player is pulled into a session they thought they had walked away from.

## What is the new behavior?

Resolves #2320

Every join attempt — direct IP, Steam, and hosting's own loopback join — raises Bannerlord's loading screen naming what it is contacting, with a Cancel button. Cancel ends the session, which disposes the network and stops the retry loop, and returns to the Join screen with the address intact; the confirmation is a line in the corner message log rather than a pop-up. Unsolicited endings (server exited, join timeout, refused password) keep their pop-ups.

The screen lives on the engine's loading window through the existing `ILoadingInterface` — an earlier revision drew it on a screen of its own, on a claim about the loading window that re-testing disproved; @ShoT-UPfps was right to push back. The one custom piece is a small layer above the window carrying the Cancel button, which the window's text-only view model cannot.

It is owned by `MainMenuState`, the client's initial state, which already hides that window on the way in and re-shows it on `NetworkConnected` — so ownership follows the state's lifetime rather than a flag in the session orchestrator. A `JoinIntent` picks the wording, so a Steam join never names an endpoint the player did not type, and cancelling a host's own loopback wait says the standalone server stays open.

## Validation

- Units 1721 passed / 0 failed / 14 skipped, Release, at 68029ee9. E2E runs in CI on this branch; the earlier hand-run figure was measured against a superseded commit and is not repeated here.
- Direct join, in game on 1.4.7: the screen holds `Contacting localhost:4200...` across 14 redials; Cancel logs `Player cancelled a "PlayerDirect" join attempt`, the log then stops growing (the redial loop is dead, not just hidden), the screen comes down and `Connection attempt cancelled` appears.
- Hosting's loopback join, on a Release build: the spawned standalone server kept logging for 95 seconds after Cancel and exited only when its own window was closed.
- Hosting's loopback join, Release build 68029ee9, in game on 1.4.7: the heading stays `Hosting Coop Server` across the handshake while the description advances from `Waiting for the server to load the campaign save...` to `Applying patches...`, and the `Stop Waiting` button is withdrawn on connect. Frames 4s apart, below.
- The Steam-initiated join is unit-covered but was not playtested — it needs a lobby and a second machine.

CodeFactor reports `Handle(AttemptHost)` as a Complex Method. That is its advisory annotation; the added branch is the refused-start guard that stops callers acting on a container belonging to another attempt.

1-connecting-screen

2-after-cancel

3-hosted-heading-holds

## Bot Changelog Entry

- Joining a co-op server now shows the standard Bannerlord loading screen naming the server being contacted, with a Cancel button.
- Cancelling a join attempt now actually stops it instead of leaving it retrying in the background, and returns you to the join screen with the address you typed still filled in.

### DIFF
--- a/source/Coop.Core/CoopartiveMultiplayerExperience.cs
+++ b/source/Coop.Core/CoopartiveMultiplayerExperience.cs
@@ -52,7 +52,7 @@ namespace Coop.Core
         // A spawned server has to load the whole campaign save before it binds its port.
         public static readonly TimeSpan HostedServerStartTimeout = TimeSpan.FromMinutes(5);
 
-        private const string StartRefusedNotice =
+        public const string StartRefusedNotice =
             "Another co-op session is still starting; try joining again shortly";
         private const string HostClientStartFailedNotice =
             "Could not start the co-op client; the standalone server remains open until you close it";
@@ -342,10 +342,9 @@ namespace Coop.Core
 
             try
             {
+                // A refusal leaves the process-wide tunnel and lobby to the start still in flight.
                 if (!StartAsClient(configuration, intent: JoinIntent.PlayerSteam))
                 {
-                    joinEndpointPreparer.TearDownActiveTunnel();
-                    messageBroker.Publish(this, new SessionJoinAbandoned());
                     InformationManager.DisplayMessage(new InformationMessage(StartRefusedNotice));
                     return;
                 }
--- a/source/Coop.Core/CoopartiveMultiplayerExperience.cs
+++ b/source/Coop.Core/CoopartiveMultiplayerExperience.cs
@@ -52,6 +52,11 @@ namespace Coop.Core
         // A spawned server has to load the whole campaign save before it binds its port.
         public static readonly TimeSpan HostedServerStartTimeout = TimeSpan.FromMinutes(5);
 
+        private const string StartRefusedNotice =
+            "Another co-op session is still starting; try joining again shortly";
+        private const string HostClientStartFailedNotice =
+            "Could not start the co-op client; the standalone server remains open until you close it";
+
         public CoopartiveMultiplayerExperience(
             bool standaloneServerProcess = false,
             Action<string> setCrashPhase = null)
@@ -103,7 +108,10 @@ namespace Coop.Core
                 EnableSteamInvites = connectMessage.EnableSteamInvites,
             };
 
-            StartAsClient(configuration, advertisementConfig);
+            if (!StartAsClient(configuration, advertisementConfig, JoinIntent.PlayerDirect))
+            {
+                InformationManager.DisplayMessage(new InformationMessage(StartRefusedNotice));
+            }
         }
 
         private void Handle(MessagePayload<AttemptHost> obj)
@@ -198,7 +206,12 @@ namespace Coop.Core
 
             try
             {
-                StartAsClient(configuration, advertisementConfig);
+                if (!StartAsClient(configuration, advertisementConfig, JoinIntent.HostLoopback))
+                {
+                    hostedSession = false;
+                    InformationManager.DisplayMessage(new InformationMessage(HostClientStartFailedNotice));
+                    return;
+                }
 
                 container.Resolve<SteamJoinWatchdog>().Arm(configuration.Address, configuration.Port,
                     timeout: HostedServerStartTimeout,
@@ -212,8 +225,7 @@ namespace Coop.Core
                 Logger.Error(ex, "Hosted co-op session failed to start");
                 hostedSession = false;
                 DestroyContainer();
-                InformationManager.DisplayMessage(new InformationMessage(
-                    "Could not start the co-op client; the standalone server remains open until you close it"));
+                InformationManager.DisplayMessage(new InformationMessage(HostClientStartFailedNotice));
             }
         }
 
@@ -330,7 +342,13 @@ namespace Coop.Core
 
             try
             {
-                StartAsClient(configuration);
+                if (!StartAsClient(configuration, intent: JoinIntent.PlayerSteam))
+                {
+                    joinEndpointPreparer.TearDownActiveTunnel();
+                    messageBroker.Publish(this, new SessionJoinAbandoned());
+                    InformationManager.DisplayMessage(new InformationMessage(StartRefusedNotice));
+                    return;
+                }
 
                 container.Resolve<SteamJoinWatchdog>().Arm(prepared.Address, prepared.Port, prepared.Tunneled);
             }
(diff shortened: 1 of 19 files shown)


ShoT-UPfps:
I think this has to be during the loading screen with the bannerlord supplied backgrounds and it has to be supported for both direct IP and Steam.

ShoT-UPfps (changes requested):
cmts

douglasholm:
> I think this has to be during the loading screen with the bannerlord supplied backgrounds and it has to be supported for both direct IP and Steam.

Reworking this now, hopefully I'll have the new version ready tomorrow! Thank you for clearing that up for me.

ShoT-UPfps:
> > I think this has to be during the loading screen with the bannerlord supplied backgrounds and it has to be supported for both direct IP and Steam.
> 
> Reworking this now, hopefully I'll have the new version ready tomorrow! Thank you for clearing that up for me.

np, thanks for the help!

ShoT-UPfps:
@douglasholm, pushing back on the "Why the connecting state is a screen of our own rather than the engine's loading window" a bit. We already have the custom loading status on the bottom left of the loading screen without rolling our own custom screen, why can't we have the same for this button?

ShoT-UPfps (changes requested):
cmt

douglasholm:
You're right, and thanks for pushing on it twice. I re-tested in game and the engine's loading window paints perfectly well over the co-op join screen: full background art, title and description, live text updates through `SetLoadingMessage`, and a clean restore when it hides. My earlier claim that it "paints nothing at a quiescent main menu" was a bad generalisation from one observation, and it should not have gone into the description as a fact.

The branch now puts the connecting status on the engine's loading screen through the same `ILoadingInterface` the mod already uses for the loading status in the bottom-left corner, so there is no custom screen for it any more. The one custom piece left is a small layer above the loading window holding the Cancel button: the loading window's view model only takes text (`TitleText`/`DescriptionText`), so a clickable widget needs its own layer for input restrictions and focus. If there is an existing way to get a button onto that window that I have missed, I will happily drop that layer too.

Both direct IP and Steam joins go through it; body and screenshots are updated to match.

douglasholm:
@ShoT-UPfps after testing around a bit, I hope that I've managed to put together what the issue asked for and what you have in mind as the solution. The UX makes sense, and the UI additions are minimal and non-intrusive.

source/Coop.Core/CoopartiveMultiplayerExperience.cs
```diff
@@ -103,7 +116,88 @@ private void Handle(MessagePayload<AttemptJoin> obj)
...
+        {
+            // Claimed rather than read: the teardown below is deferred, so the button is still
+            // clickable when a second press arrives. Dropping the repeat here — where whether a
+            // teardown is already under way is actually known — is what lets the button publish
+            // every press instead of latching on one that may never take effect.
+            JoinIntent? intent;
+            lock (joinUiGate)
+            {
+                intent = activeJoinIntent;
+                activeJoinIntent = null;
```
ShoT-UPfps: we need a cancellation-in-progress state here instead of clearing the intent permanently. if any of the deferred UI/finalizer work throws, `RunSafe` swallows it and every later press returns at line 145, so the retry keeps running behind a dead cancel button.

source/GameInterface/Services/UI/JoinCancel/JoinCancelVM.cs
```diff
@@ -0,0 +1,46 @@
...
+    private string cancelButtonText;
+
+    public JoinCancelVM(string cancelButtonText)
+        : this(cancelButtonText, MessageBroker.Instance)
+    {
+    }
+
+    public JoinCancelVM(string cancelButtonText, IMessageBroker messageBroker)
+    {
+        this.messageBroker = messageBroker ?? throw new ArgumentNullException(nameof(messageBroker));
```
ShoT-UPfps: please use the statement form `if (messageBroker == null) throw ...`, we dont use `x ?? throw` for plumbing.

source/Coop.Core/CoopartiveMultiplayerExperience.cs
```diff
@@ -626,6 +784,108 @@ private void StartAsClientCore(
...
+        /// </remarks>
+        private JoinAttemptPresentation ClaimJoinAttemptUI(JoinIntent intent, INetworkConfig config)
+        {
+            var presentation = JoinAttemptPresentation.For(intent, config.Address, config.Port);
+
+            var loadingInterface = container.Resolve<ILoadingInterface>();
+
+            // A headless server process has no loading window to draw any of this on. Claiming nothing
+            // is also what makes the raise a no-op there.
+            if (!loadingInterface.IsLoadingScreenAvailable) return presentation;
```
ShoT-UPfps: `IsLoadingScreenAvailable` is also false while the loading manager is being recreated during a state transition. a Steam join can reach this in that window, then we never claim the loading screen or cancel button.

source/Coop.Core/CoopartiveMultiplayerExperience.cs
```diff
@@ -103,7 +116,88 @@ private void Handle(MessagePayload<AttemptJoin> obj)
...
+                return;
+            }
+
+            Logger.Information("Player cancelled a {Intent} join attempt", intent.Value);
+
+            // Deferred to the next update rather than run inline. This arrives from inside the cancel
+            // button's own click dispatch, and removing our layers touches state the engine is
+            // iterating at that moment.
+            //
+            // Latent limit: the teardown waits on the container gate, which a DEBUG build's pre-connect
```
ShoT-UPfps: please move this into the commit message, we keep comments to the non-obvious why and dont add 6+ line rationale blocks.

source/Coop.Core/CoopartiveMultiplayerExperience.cs
```diff
@@ -198,7 +292,7 @@ private void Handle(MessagePayload<AttemptHost> obj)
 
             try
             {
-                StartAsClient(configuration, advertisementConfig);
+                StartAsClient(configuration, advertisementConfig, JoinIntent.HostLoopback);
```
ShoT-UPfps: `StartAsClient` can return false here, but we still resolve and arm the watchdog from the other attempt's `container`. check the return like the direct path does, same for `StartResolvedJoin`.

ShoT-UPfps (changes requested):
cmts

douglasholm:
@ShoT-UPfps the cancellation-state point pushed the whole screen onto `MainMenuState`, since that flag really belongs to whatever owns the screen and its lifetime is the attempt. It's `FinishCancel` now, and `cancelling` resets if the deferred work throws, so a failed teardown can't leave a dead button behind. That moved the code your comments were on, so they all show as outdated now, but each one is addressed.

The `?? throw` is gone, nothing gates on `IsLoadingScreenAvailable` any more since the state claims the screen in `Connect()`, and the rationale blocks are down to class docs.

Two of them I had only half fixed. `CoopMod.cs:459` was still ignoring the return value after I converted the other call sites, so that one join path gave no feedback at all when a start was refused. And the hosted join's title wasn't following the intent, so the heading flipped from "Hosting Coop Server" to "Connecting to Coop Server" the moment the handshake landed. Both are fixed, and I've playtested the hosted flow on 1.4.7 with screenshots in the description.

If any of these aren't what you meant, I'd rather redo them than guess.

Thank you again for the patience here!

ShoT-UPfps (approved):
lgtm

PR #2624, opened 2026-08-02
Barter fixes

EgardA:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] All new classes have class-level documentation comments, if there are any at all
- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update
- [ ] Other... Please describe:

## What is the current behavior?

Marriage and other barters were failing and blocking players from finishing barters

## What is the new behavior?

Resolves #

Barters should be fully functional while not showing the popups on the server.

## Bot Changelog Entry

Fixed barter functionality.
Fixed lord recruitment via defection.

PR #2652, opened 2026-08-03
Kingdom membership replication, lord recruitment, and barter fixes

alikhan-tbl:
## PR Checklist

## PR Type

Lord recruitment is missing entirely, kingdom membership changes only half-replicate, and several barter paths fail or misreport their result.

## What is the new behavior?

**Restore lord recruitment via defection**
`DisableLordDefectionCampaignBehavior` prefixed `RegisterEvents` with `false`, which also removed `OnSessionLaunched -> AddDialogs` and so deleted the whole recruit dialogue tree, orphaning the `JoinKingdomAsClan` barter sync that was already written.
Removed rather than narrowed to `OnDailyTick`: the client is the only machine that fills the persuasion-attempt list, so suppressing that tick there makes a lord who once refused you refuse forever.

**Replicate kingdom membership changes to clients**
`Clan._kingdom` is AutoSynced but the kingdom's own `_clans`/`_fiefsCache` are not, and AutoSync cannot express a null reference at all, so a clan that left, was expelled or rebelled never lost its kingdom on clients.
Postfixing `ChangeKingdomAction.ApplyInternal` covers every membership path at once and broadcasts the authoritative kingdom including "no kingdom"; settlement banners are redrawn since their colour follows the owner's realm.

**Support barters started from a settlement menu**
Talking to a lord from a castle or town menu creates no agent interaction and no location mission, so no engagement was ever held and every barter from it was refused after the player had already played the persuasion.
Added a `Settlement` conversation context where the server validates co-location instead of a lock, which a player who is not there cannot fake.

**Award Charm XP for client persuasion**
A client who won the recruitment persuasion gained nothing, because client-side XP writes are blocked and the server never runs the dialogue.
The client now ships only the outcome enums and the server derives the XP from the same vanilla model, so no number crosses the wire.

**Do not report an applied barter as rejected**
The marriage and peace handlers could apply their mutation and then still fall into the catch and send a rejection, so the client rolled back a marriage or peace that had really happened.
Once the mutation is committed the result is reported as success.

**Hold a conversation target for one player at a time**
Two players both at war with the same lord could each run the persuasion and each pay, with the lord defecting twice.
A party is now held by exactly one player; simultaneous attackers still converge on the MapEvent through the existing branch.

**Let clients grant settlements to their vassals**
The kingdom screen's Give Settlement was a silent no-op on a client, which is the natural action right after recruiting a lord.
The client forwards the gift and the server re-derives authority from the requesting peer.

**Keep barter presentation off the server**
`OnBarterAccepted` and campaign cutscenes ran on the headless host, showing popups nobody can dismiss and blocking it during a wedding scene.
Both are suppressed server-side only; clients still run their own.

**Make vassal and mercenary offers co-op aware**
The behaviour was switched off wholesale, but the real problem is that its handlers read `Hero.MainHero`/`Clan.PlayerClan` and so act for the host rather than the player the event concerns.
The two that keep offer bookkeeping consistent are reimplemented against the co-op player checks; offers to players stay disabled pending routing.

**Also fixed**
Dead participants are rejected between authorize and request; `LordBarterHandler.Dispose` no longer clears its peer dictionaries off the game thread; `coop.debug.clan.join_kingdom` and `change_clan_kingdom` replicate membership instead of lying about the result; value rejections now log the computed offer value with the kingdom roster counts and tell the player the shortfall.

## Other information

Overlaps #2624, which independently fixes the settlement-menu context and the defection dialogue. Happy to rebase on it.

Full suite green: 2629 passed, 0 failed. Verified live with a server and client pair - a lord defected, his clan and castle transferred, and both sides agreed.


EgardA:
Converting this to a draft as I have talked with author in DMs, this PR does about the same thing that two other PRs does but will later complement atleast my PR.

alikhan-tbl:
Closing - reopening from the correct account.

PR #2656, opened 2026-08-03
[2649] Partition agent action polling in large battles

### LINKED ISSUE #2649: Partition post-native agent action polling in large battles
## Description

The post-native mission action poll scans every locally controlled agent on every call. The one-client 900-vs-900 profile from #2641 recorded 1,155,154 registry visits over 1,154 calls in 10.03 seconds, but only 487 action updates. The post-native path took 774.19 ms in the DEBUG diagnostic window, while the selective pre-native player path made 1,154 visits and took 13.31 ms.

This is a follow-up to #2641 and uses the profiling support from #2643.

## Requirements

- Keep the main player and explicitly dirty action/guard agents on the every-frame path.
- Partition locally controlled AI agents across native-tick polls instead of scanning all of them every call.
- Keep a bounded rotating fallback scan because native AI actions can change without a managed setter.
- Preserve discrete actions, guard transitions, mounted guards, host migration, catch-up, action sequence ordering, and reliable packet behavior.
- Add focused automated coverage for scan partitioning, dirty-agent priority, fallback latency, agent removal, and controller changes.
- Instrument calls, registry visits, active visits, updates, packets, bytes, and worst/average detection latency.

## Live verification

- Use exactly one client (`testclient`); never launch a second client.
- Use the exact uncapped 900-vs-900 field-battle fixture from #2641.
- Capture baseline and changed deployment/active-combat profiles on the same client process.
- Report FPS, normalized CPU, post-native time per call, visits per call, update traffic, and observed action/guard latency.
- Reject the change if combat presentation, guard timing, or action ordering regresses even when CPU improves.

## Definition of done

- [ ] The post-native scan visits fewer agents per call with bounded full-coverage latency.
- [ ] Action and guard synchronization tests remain green.
- [ ] Independent one-client 900-vs-900 evidence shows the CPU and frame-time impact.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Large battles repeatedly scan every locally controlled agent after each native tick, spending CPU on hundreds of agents even when only a few actions or guard states change.

## What is the new behavior?

Resolves #2649

Large battles spread background AI action checks across four bounded frame groups while immediately checking the player and agents with recent action or guard activity. Action ordering and synchronization remain unchanged, with less polling work per frame.

Matched one-client 900-vs-900 profiling measured the targeted action-check path as follows:

| 900-vs-900 measurement | Current | New |
| --- | ---: | ---: |
| Action-poll CPU | 77.58 ms/s | 39.73 ms/s (-48.8%) |
| Per-agent action-check CPU | 31.77 ms/s | 4.31 ms/s (-86.4%) |
| Agents visited per polling call | 988.26 | 260.00 (-73.7%) |
| Action-check time per call | 933.02 us | 528.79 us (-43.3%) |

## Bot Changelog Entry

- Reduced CPU time spent checking AI actions during large battles.

### DIFF
--- a/source/E2E.Tests/Services/Missions/AgentActionPollSchedulerTests.cs
+++ b/source/E2E.Tests/Services/Missions/AgentActionPollSchedulerTests.cs
@@ -1,0 +1,204 @@
+﻿using Missions.Agents.Handlers;
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace E2E.Tests.Services.Missions;
+
+public class AgentActionPollSchedulerTests
+{
+    [Fact]
+    public void FallbackPartitions_VisitEveryAgentWithinOneSweep()
+    {
+        var scheduler = new AgentActionPollScheduler();
+        var agentIds = new List<Guid>();
+        for (int partition = 0;
+             partition < AgentActionPollScheduler.FallbackPartitionCount;
+             partition++)
+        {
+            agentIds.Add(CreateAgentId(partition, 0));
+            agentIds.Add(CreateAgentId(partition, 1));
+        }
+
+        var visits = new Dictionary<Guid, int>();
+        for (int poll = 0;
+             poll < AgentActionPollScheduler.FallbackPartitionCount;
+             poll++)
+        {
+            scheduler.BeginPoll(
+                registryVersion: 1,
+                refreshedAgentIds: poll == 0 ? agentIds : null);
+            foreach (Guid agentId in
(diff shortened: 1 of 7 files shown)


ShoT-UPfps:
closing this because the four-frame partition makes ordinary AI action polling depend on client FPS. at 40 FPS that drops it to 10 Hz, or 100 ms between polls, and it gets worse as FPS drops, so short action transitions could be missed.

PR #2657, opened 2026-08-03
Hideout sync

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

No hideouts

## What is the new behavior?

Resolves #

Hideouts

## Bot Changelog Entry

Hideouts added.


ShoT-UPfps:
so this doesn't allow multiplayer hideouts, right?

EgardA:
> so this doesn't allow multiplayer hideouts, right?

Correct, thinking about it we should add a block so two people can't interact with the same hideout at the moment.

source/GameInterface/Services/Settlements/SettlementSync.cs
```diff
@@ -42,7 +42,7 @@ public SettlementSync(AutoSyncRegistry AutoSyncRegistry)
         AutoSyncRegistry.AddProperty(AccessTools.Property(typeof(Settlement), nameof(Settlement.BribePaid)));
         AutoSyncRegistry.AddProperty(AccessTools.Property(typeof(Settlement), nameof(Settlement.SiegeEvent)));
         AutoSyncRegistry.AddProperty(AccessTools.Property(typeof(Settlement), nameof(Settlement.IsActive)));
-        //AutoSyncRegistry.AddProperty(AccessTools.Property(typeof(Settlement), nameof(Settlement.IsVisible)));
+        AutoSyncRegistry.AddProperty(AccessTools.Property(typeof(Settlement), nameof(Settlement.IsVisible)));
```
ShoT-UPfps, line 44: `_isVisible` is already synced at line 21. registering `IsVisible` too publishes once from the property prefix and again when its setter writes `_isVisible`.

source/GameInterface/Services/Hideouts/Patches/Disable/HideoutCampaignBehavior.cs
```diff
@@ -0,0 +1,54 @@
...
+using TaleWorlds.CampaignSystem.Party;
+using TaleWorlds.CampaignSystem.Settlements;
+using TaleWorlds.Core;
+
+namespace GameInterface.Services.Hideouts.Patches.Disable;
+
+[HarmonyPatch(typeof(HideoutCampaignBehavior))]
+internal class HideoutCampaignBehaviorPatch
+{
+    [HarmonyPatch(nameof(HideoutCampaignBehavior.HourlyTickSettlement))]
```
ShoT-UPfps, line 61: i think we need a client request that runs the hideout menu consequences on the server. vanilla calls `ArrangeHideoutTroopCountsForMission`, `SetNextPossibleAttackTime`, and `SetCleanHideoutRelations` on the client, so the server never gets the roster changes, cooldown, or rewards.

source/GameInterface/Services/Hideouts/Patches/Disable/HideoutCampaignBehavior.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            float hideoutSpottingDistance = Campaign.Current.Models.MapVisibilityModel.GetHideoutSpottingDistance();
+
+            if (ContainerProvider.TryResolve<IPlayerManager>(out var playerManager) == false)
+                return false;
+
+            if (ContainerProvider.TryResolve<IObjectManager>(out var objectManager) == false)
+                return false;
+
+            foreach (var item in playerManager.Players)
```
ShoT-UPfps, line 80: skip inactive player parties here. disconnected players stay registered with `IsActive = false`, so their parked party can keep spotting hideouts while someone else advances time.

source/GameInterface/Services/MapEvents/Patches/MapEventPatches.cs
```diff
@@ -397,26 +397,36 @@ private static void Postfix_CanPartyJoinBattle(
...
     {
-        if (ModInformation.IsClient)
+        if (ModInformation.IsClient || !StartsWithPlayerParty(attackerParty, defenderParty))
             return;
 
-        var attackerIsPlayer = attackerParty.MobileParty?.IsPlayerParty() == true;
-        var defenderIsPlayer = defenderParty.MobileParty?.IsPlayerParty() == true;
+        // Initialize adds settlement occupants to the battle before its postfixes run. Open the window now so
+        // the AI-join gate does not reject and eject native hideout defenders during that initialization.
+        playerBattleWindows.GetValue(
```
ShoT-UPfps: we need to separate the initialization window from the reinforcement broadcast check. opening this before `MapEvent.Initialize` makes every AI party added during initialization resend every party and full roster accumulated so far.

source/GameInterface/Services/Settlements/SettlementSync.cs
Gerseras: We should then remove the field sync otherwise we would have to add our own patch as target method otherwise spotted hideouts won't be visible on client side

source/GameInterface/Services/Hideouts/Patches/Disable/HideoutCampaignBehavior.cs
ShoT-UPfps: `OnTroopRosterManageDone` still runs the original after sending this request, so the client adjusts the roster and then applies the relative roster deltas from the server on top of it. we need to stop the local roster mutation otherwise the client can remove the excess troops twice.

ShoT-UPfps (changes requested):
cmt

ShoT-UPfps (approved):
lgtm

PR #2607, opened 2026-08-02
fixed some bugs related to kingdom

NatteTosti69:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ x ] All new classes have class-level documentation comments, if there are any at all
- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ x ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update
- [ ] Other... Please describe:

## What is the current behavior?

Clans were not properly added to the kingdom clan tab, kingdom colors of allies and enemies were not properly updated. policies would sometimes not work at all. garrison were also unstable

## What is the new behavior?

Resolves #2559 
Resolves #2524 
Resolves #2631
Resolves #2724 

Clans are now properly added to the kingdom clan tab, kingdom colors of allies and enemies are properly updated and policies work correctly. garrison can now also be properly accessed and donated to

## Bot Changelog Entry
 Clans are now properly added to the kingdom clan tab, kingdom colors of allies and enemies are properly updated and policies work correctly

PR #2668, opened 2026-08-04
Tracking: co-op barter, siege and battle fixes (split into #2748-#2758)

alikhan866:
> **Not for review.** This is the index for the set below. Please review the individual PRs — each fixes a single problem, and nine of the eleven can be reviewed and merged on their own, in any order.

Split out of the original branch on review feedback that a PR should fix a single problem, then flattened further: rather than one deep stack, almost all of these now target `development` directly.

## Independent — target `development`, mergeable in any order

| PR | Fixes |
|----|-------|
| #2748 | Client persuasion earned no Charm XP, and a lord's refusal never expired on a client |
| #2750 | Clients could not grant fiefs, and gifting paid no relation |
| #2751 | Retreat ran client-side and never reached the other players |
| #2753 | Garrisons never sortied; relief forces joined the besiegers' side |
| #2754 | Debug commands for driving siege and sortie scenarios |
| #2755 | A captor froze forever; army joins were invisible to everyone else |
| #2756 | Nearby AI never joined a player's battle; Surrender went dead |
| #2757 | Battle deployment wedged and sides spawned nothing |
| #2758 | Silently dropped party state, and a 4,200 line/sec log flood |

## The only two with a real dependency

- **#2749** needs **#2748** — its `Dispose` fix changes marshalling that #2748 introduces.
- **#2752** needs **#2751** — `BreakInCasualtiesPatch.cs` is created there and only modified here.

## Closed

- **#2747** (kingdom membership replication) — duplicate of #2607, which fixes the same problem at a better choke point.
- **#2759** (analyzer complexity) — refactored methods written by six of the others, so it has no valid parent now that they merge in parallel. Worth redoing against `development` in one pass once these land.

## Verification

- Every branch builds `Coop.sln` in Release **on its own**, against current `development`.
- Full suite green; two clients plus a dedicated server were used to verify the kingdom, barter and battle paths live.
- No dedicated-server or headless code in any branch — checked per branch, files and diffs.

## Dropped as superseded by `development`

Found while rebasing, and removed rather than duplicated: the break-in entry flow (`development` has its own server-approved continuation), barter presentation suppression on the server, new-hero replication and TextObject int attributes (rebased to an empty diff), and the vassal/mercenary offer behaviour. The recruit-dialogue half of #2613 is also already fixed there, which is why #2748 no longer claims it.


source/GameInterface/Services/MapEvents/Interfaces/BattleRetreatInterface.cs
```diff
@@ -0,0 +1,225 @@
...
+
+        var sacrificed = GetTroopsSacrificed(battle);
+        var commanded = CollectCommandedParties(requester);
+
+        ApplyTroopSacrifice(requester, commanded, sacrificed);
+        RemoveGoods(requester);
+
+        campClearedPartyIds = ClearBesiegerCamps(requester, commanded);
+
+        requester.TeleportPartyToOutSideOfEncounterRadius();
```
ShoT-UPfps, line 72: could we remove this party from the authoritative map event before reporting success? otherwise the server still has it on the defender side, so the player remains in the battle and a repeated click can charge the retreat again.

source/GameInterface/Services/MapEvents/Patches/BreakInCasualtiesPatch.cs
```diff
@@ -0,0 +1,45 @@
...
+    private static bool ApplyBreakInPrefix(ref TroopRoster casualties, ref int armyCasualtiesCount)
+    {
+        if (CallOriginalPolicy.IsOriginalAllowed()) return true;
+        if (ModInformation.IsServer) return true;
+
+        casualties = TroopRoster.CreateDummyTroopRoster();
+        armyCasualtiesCount = 0;
+
+        var mainParty = MobileParty.MainParty;
+        var settlement = mainParty?.CurrentSettlement ?? mainParty?.BesiegedSettlement;
```
ShoT-UPfps: could this use the encounter settlement instead of the party's current/besieged settlement? during a normal break-in the party is still outside and is not a besieger, so both values are null and this suppresses vanilla without ever requesting casualties.

source/GameInterface/Services/MapEvents/Handlers/NearbyPartyReinforcementHandler.cs
```diff
@@ -0,0 +1,203 @@
...
+        if (mapEvent.IsRaid) return "raid";
+        if (mapEvent.IsSiegeAssault) return "siege assault";
+        if (mapEvent.IsForcingSupplies) return "forcing supplies";
+        if (mapEvent.IsForcingVolunteers) return "forcing volunteers";
+
+        if (mapEvent.MapEventSettlement?.IsHideout == true) return "hideout";
+
+        // Only player battles reinforce, and only while the window the broadcast path checks is still open -
+        // otherwise the join would apply on the server and never reach the clients.
+        if (!InteractionPatches.IsWithinAiJoinWindow(mapEvent))
```
ShoT-UPfps, line 134: could we open the AI join window before publishing `PlayerJoinedBattle`? that publish is synchronous and currently runs before `playerBattleWindows.GetValue`, so this guard always rejects the immediate reinforcement scan.

source/GameInterface/Services/Kingdoms/Handlers/KingdomHandler.cs
```diff
@@ -57,6 +61,78 @@ public KingdomHandler(
...
+        var data = payload.What;
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Clan>(data.ClanId, out var clan)) return;
+
+            Kingdom kingdom = null;
+            if (data.KingdomId != null &&
+                !objectManager.TryGetObjectWithLogging<Kingdom>(data.KingdomId, out kingdom)) return;
+
+            if (ReferenceEquals(clan.Kingdom, kingdom)) return;
```
ShoT-UPfps: could we still refresh the settlement visuals when the kingdom value already matches? the existing `_kingdom` AutoSync packet arrives before this explicit message, so this return skips `RefreshSettlementVisuals` and the clan's fiefs keep their old colours.

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
```diff
@@ -20,21 +22,145 @@ public class SettlementOwnershipHandler : IHandler
...
+            if (!CanGift(requestingHero, settlement, newOwner, out var reason))
+            {
+                Logger.Warning("Settlement gift of {Settlement} by {Hero} rejected: {Reason}",
+                    settlement.StringId, requestingHero.StringId, reason);
+                return;
+            }
+
+            // ApplyByGift re-enters ChangeOwnerOfSettlementPatch on the server, which publishes
+            // SettlementOwnershipChanged and replicates to every client.
+            GameThread.RunSafe(
```
ShoT-UPfps, line 95: could we move `CanGift` into this game-thread action and recheck it immediately before applying the gift? otherwise ownership or kingdom membership can change after the earlier check and a stale request can transfer a fief the requester can no longer grant.

source/GameInterface/Services/MapEvents/TroopSupply/CoopTroopSupplier.cs
```diff
@@ -245,6 +250,57 @@ public int TotalTroops
...
+            // Rounded, so the owners' slices can differ from the allocation by at most one troop per side -
+            // invisible against a battle size in the hundreds, and the next wave re-derives from scratch.
+            // ponytail: exact apportionment would need each owner to know the others' shares; not worth a
+            // wire field and a migration path for a possible off-by-one.
+            var share = (int)Math.Round(sideAllocation * (double)owned / total, MidpointRounding.AwayFromZero);
+
+            // Never round an owner with troops down to nothing. A share of zero on the side holding the local
+            // player's party is what CoopBattleMissionSpawnHandler treats as "origin missing" - it aborts the
+            // battle - so a small allocation must still field someone.
+            if (share <= 0) share = 1;
```
ShoT-UPfps: could we apportion the side allocation so the owners' shares sum to the requested count? forcing every positive owner to one lets a one-troop allocation spawn one troop per owner, and larger waves can also exceed the allocation.

source/GameInterface/Services/MobileParties/Patches/LordDefectionRetryPatches.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        AccessTools.Method(typeof(LordDefectionCampaignBehavior), nameof(LordDefectionCampaignBehavior.OnDailyTick))
+    };
+
+    [HarmonyPrefix]
+    static bool Prefix(LordDefectionCampaignBehavior __instance)
+    {
+        switch (ModConfigProvider.ModOptions.LordDefectionRetries)
+        {
+            case LordDefectionRetryMode.NeverExpire:
```
ShoT-UPfps: could these modes change the failed-attempt gate rather than only the daily prune? vanilla stops blocking an attempt after one week even if its record remains, so `NeverExpire` still expires after a week, while `AlwaysRetry` waits until the next daily tick instead of clearing immediately.

ShoT-UPfps (changes requested):
comments

source/GameInterface/Services/MapEvents/Interfaces/BattleRetreatInterface.cs
alikhan866: Done in 05ec3da89. The retreating party (and anything it commands) now clears its MapEventSide before success is reported, which calls RemovePartyInternal on the side it was on. It runs after the loss model and camp clearing, since both need the battle intact.

source/GameInterface/Services/MapEvents/Patches/BreakInCasualtiesPatch.cs
alikhan866: Done in 75cf0308c. PlayerEncounter.EncounterSettlement is tried first, with MapEvent.MapEventSettlement, CurrentSettlement and BesiegedSettlement as fallbacks for breaking out and for a besieger. If none resolves it is now logged rather than passing silently.

source/GameInterface/Services/MapEvents/Handlers/NearbyPartyReinforcementHandler.cs
alikhan866: Done in c4d69f538. The window is created before the Publish, so the synchronous handler sees the state the announcement describes.

source/GameInterface/Services/Kingdoms/Handlers/KingdomHandler.cs
alikhan866: Done in 37958b033. Only the clan move is conditional now; RefreshSettlementVisuals always runs.

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
alikhan866: Done in 3c36e11d9. CanGift is re-derived inside the game-thread action immediately before the transfer; the original call stays as a cheap pre-filter so an invalid request never reaches the game thread.

source/GameInterface/Services/MapEvents/TroopSupply/CoopTroopSupplier.cs
alikhan866: Done in 9a89dd877. Each party now carries its offset within the side, so an owner takes floor(alloc*(offset+count)/total) - floor(alloc*offset/total). Those offsets partition the side exactly once, so the shares sum to the allocation across all owners with no coordination. The zero floor now applies only to the owner holding the receiver's own party, where a zero share reads as "origin missing" and aborts the battle - every other owner takes its apportioned share, including none. SideOffset is additive with a default of 0. Covered by new tests including a deliberately indivisible 3-owner/7-troop split.

source/GameInterface/Services/MobileParties/Patches/LordDefectionRetryPatches.cs
alikhan866: Done in 466f8bad8. Both modes now patch CanAttemptToPersuade rather than the daily prune: NeverExpire blocks while any unsuccessful attempt survives (and still suppresses the prune so one does), AlwaysRetry never blocks and lets vanilla tidy the list. You were right that patching only the prune changed nothing - the gate stops at one week and the prune only drops year-old records.

source/GameInterface/Services/MapEvents/Interfaces/BattleRetreatInterface.cs
ShoT-UPfps: could the retreat broadcast the same NetworkPartyLeftBattle removal used by the ordinary leave path? RemovePartyInternal does not auto-replicate, so another player who stays in the battle still keeps the retreating force in their local map event.

source/GameInterface/Services/MapEvents/TroopSupply/CoopTroopSupplier.cs
ShoT-UPfps: this is exact until the receiver top-up at line 315. could we cover multiple player owners here? with a one-troop allocation, every owner whose interval rounds to zero raises itself to one, while the owner holding the apportioned troop also supplies one, so the side still exceeds the allocation.

garrettluskey:
A PR should fix a single problem. Please split this up

source/GameInterface/Services/MapEvents/Interfaces/BattleRetreatInterface.cs
alikhan866: Good catch, and yes. `LeaveBattle` now collects the parties it detaches and broadcasts `NetworkPartyLeftBattle` for each, instead of relying on the `MapEventSide` setter alone — `RemovePartyInternal` uses `RemoveAt`, which bypasses the collection sync, so the retreating force was detached on the server only and every other player still fighting kept it in their local map event, counted in their battle and still on their side's roster.

Reusing the ordinary leave path's message rather than inventing a second removal protocol, as you suggested. Both flags are false: `FinishLocalMenus`, because the retreating client's own teardown comes from the `NetworkBattleRetreatResolved` that follows and finishing menus twice would close one it has already moved past; and `LeaveSiege`, because the retreat clears its own besieger camps and reports them through `campClearedPartyIds`, so re-running the siege leave would clear them a second time. The map-event removal is the only part wanted from this message.

source/GameInterface/Services/MapEvents/TroopSupply/CoopTroopSupplier.cs
alikhan866: You're right on both counts — the interval apportionment is exact, and the top-up is what breaks it. With N human-owned parties on a side, every owner whose interval floors to zero raises itself to one while the owner whose interval actually covers the apportioned troop also supplies one, so the side can field up to N-1 more than the allocation.

I've bounded and documented it rather than removed it, because a zero share is a hard failure rather than a cosmetic one. `CheckDeployment` skips a side whose reservation falls short of `InitialSpawnNumber`, and that `continue` skips the side's plan-making as well as its spawn — so an owner that rounds to zero doesn't just field nobody itself, it leaves `IsPlanMade` false for its team and *nothing on that side spawns at all*. We hit exactly that in testing: the player had no agent, `Mission.InitialPlayerAgent` stayed null, and `DeploymentMissionController.SetupTeams` threw on it every tick with every player in the battle stuck on the deployment screen.

Two changes here. The top-up now also requires that party to still *have* troops, so an exhausted player party stops conjuring one troop per wave for the rest of the battle — that was the case where the overshoot compounded. And the bound is written down: it only applies while the allocation is smaller than the number of human owners on the side, since against any ordinary wave every player's interval already covers a troop and nothing tops up.

Making it exact needs each owner to know the others' shares — a wire field on the reserve plus a migration path — to save at most a handful of troops in the first wave of a battle that has more players on one side than troops to field. That seemed the wrong trade, but I'll do it if you'd prefer the invariant to hold unconditionally.

ShoT-UPfps: the exhausted-party guard limits how often this happens, but could the allocation still remain exact across active player owners? the allocation < N bound does not hold for uneven parties: if one player owns 1 of 1000 troops, its share of an allocation of 100 is still zero and tops up to one while the other owners already supply all 100.

source/GameInterface/Services/MapEvents/Patches/CoopDeploymentPlayerAgentGuardPatch.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        Invoke(__instance, "BeforeDeploymentFinished");
+        Invoke(__instance, "UnhideAgentsOfSide", __instance.IsPlayerAttacker
+            ? BattleSideEnum.Defender
+            : BattleSideEnum.Attacker);
+
+        try
+        {
+            mission?.OnDeploymentFinished();
+            mission?.SetFallAvoidSystemActive(true);
```
ShoT-UPfps: could this recovery restore the same mission flags as vanilla FinishDeployment? SetupTeams sets DisableDying to true, but this path never sets it back to false and also leaves fall avoidance enabled, so a leaderless battle can continue with agents unable to die.

source/GameInterface/Services/MapEvents/Interfaces/BattleRetreatInterface.cs
```diff
@@ -0,0 +1,297 @@
...
+    {
+        if (left.Count == 0) return;
+        if (!ContainerProvider.TryResolve<INetwork>(out var network)) return;
+        if (!ContainerProvider.TryResolve<IObjectManager>(out var objectManager)) return;
+
+        foreach (var party in left)
+        {
+            if (!objectManager.TryGetId(party, out var partyId)) continue;
+
+            network.SendAll(new NetworkPartyLeftBattle(partyId, leaveSiege: false, finishLocalMenus: false));
```
ShoT-UPfps: could attached player parties get their own leave and menu teardown when the army leader retreats? setting the leader MapEventSide to null cascades to attached parties, but this message uses finishLocalMenus: false and the later retreat result only closes the requester encounter, so an attached player can lose its party from the map event while remaining in the battle UI.

source/GameInterface/Services/MapEvents/Patches/CoopDeploymentPlayerAgentGuardPatch.cs
```diff
@@ -0,0 +1,112 @@
...
+
+    [HarmonyFinalizer]
+    private static Exception Finalizer(DeploymentMissionController __instance, Exception __exception)
+    {
+        if (__exception == null) return null;
+        if (!BattleSpawnGate.IsCoopBattleActive) return __exception;
+
+        // Only the missing-player-agent case is swallowed. Anything else is a real failure and must keep
+        // propagating rather than being hidden behind a half-finished deployment.
+        if (Mission.Current?.InitialPlayerAgent != null) return __exception;
```
ShoT-UPfps: could this verify that SetupTeams reached the expected player-agent dereference before swallowing the exception? InitialPlayerAgent is also null during earlier side and AI setup, so an unrelated failure there is suppressed and the deployment tail runs against partially initialized teams.

source/GameInterface/Services/MapEvents/Patches/CoopDeploymentPlayerAgentGuardPatch.cs
alikhan866: You're right, and it was worse than the two flags you named. I diffed the recovery against native `FinishDeployment` instruction by instruction and it was missing four things:

- `Mission.DisableDying` stayed `true` — agents could not die, exactly as you said.
- Fall avoidance was set the **wrong way**. `SetupTeams` turns it on for the duration of deployment and `FinishDeployment` turns it off; this turned it on again.
- `Mission.AllowAiTicking` was never set.
- The per-unit pass was missing entirely. That is what calls `SetIsAIPaused(false)` on every agent, so the troops that had just spawned stayed frozen — in a leaderless battle, nothing on the field moved at all.

The tail now mirrors native in order and omits only the two lines that need the agent that does not exist (`SetDetachableFromFormation(true)` and `Controller = Player`). `UnhideAgentsOfSide` also follows native now, which unhides the defender side and only when the player is the attacker — the previous version called it unconditionally with a side picked from `IsPlayerAttacker`.

source/GameInterface/Services/MapEvents/Patches/CoopDeploymentPlayerAgentGuardPatch.cs
alikhan866: Agreed — `InitialPlayerAgent` is null right through steps 1-4, so "it is null now" was not evidence of anything and would have swallowed a failure in enemy-side or AI setup, then run the deployment tail over half-built teams. That trades a visible stall for silent corruption, which is the worse of the two.

Three things now have to agree before the exception is swallowed:

- it is a `NullReferenceException`;
- it was raised in `SetupTeams` **itself**, not in anything it called — the dereference is inline, so a throw inside `OnSetupTeamsOfSide` or `SetupAIOfEnemySide` leaves that frame on top and no longer matches (prefix match on the name, since Harmony renames the original);
- agents exist, so the spawning steps demonstrably ran.

Anything else propagates untouched.

source/GameInterface/Services/MapEvents/Interfaces/BattleRetreatInterface.cs
alikhan866: Good catch. `FinishLocalMenus` is now per-party rather than a single flag for the whole broadcast: `false` for the requester, whose teardown comes from the `NetworkBattleRetreatResolved` that follows, and `true` for every other party that left.

That covers the case you describe — a commanded or attached party is dropped by the cascade, but `NetworkBattleRetreatResolved` only closes the requester's encounter, so another player pulled out by their army leader's retreat kept the battle UI over a party that was no longer in the map event. `ApplyNetworkLeave` gates the teardown on `isMainParty`, so `true` reaches exactly the one client that owns each party and is a no-op everywhere else.

source/GameInterface/Services/MapEvents/TroopSupply/CoopTroopSupplier.cs
alikhan866: Your counterexample is correct and my stated bound was wrong. I've fixed the comment rather than leave a claim in the code that does not hold: an owner with 1 troop of a 1000-strong side scales to zero for an allocation of 100, tops itself up to one, and the other owners still supply all 100 — so `allocation < N` was never the condition. The real bound is one extra per owner whose interval floors to zero, up to `N-1` for `N` human-owned parties, at any allocation size.

On making it exact across active player owners: I don't think it can be done locally. Exactness requires some owner to give up the troop the topped-up owner just claimed, and that owner is a different client which has no idea this one rounded to zero — each computes only its own interval. The only place with the whole picture is the server, so closing it properly means the server apportioning centrally and sending each owner its number, which is a field on the reserve plus a migration path.

I've left the behaviour as-is because the failure it prevents is not proportional to its cost: an owner that fields nothing does not merely lose its own troops, it leaves `IsPlanMade` false for its team and `CheckDeployment` then skips the entire side, so nobody on that side spawns. Happy to do the server-side apportionment if you'd rather have the invariant hold unconditionally — it's a contained change and I'd rather ask than guess at how much wire churn you want here.

ShoT-UPfps: yes, please keep the allocation exact and move the apportionment to the server. otherwise every active player owner whose interval rounds to zero still adds an extra troop beyond the engine requested.

source/GameInterface/Services/MapEvents/Interfaces/BattleRetreatInterface.cs
ShoT-UPfps: this still does not include the attached player in `left`. `CollectCommandedParties` skips player parties, and detaching the requester only adds the requester even though `MapEventSide` clears its attachments, so no `finishLocalMenus: true` message is sent for that player.

alikhan866: You're right — I made the flag per-party but never fixed the list it iterates, so the one case the broadcast exists for was still the one it missed.

`CollectAttachedPlayerParties` now runs **before** anything detaches and adds those parties to `left`. Reading it early matters: the assignment that cascades is the same one that clears the attachment list, so after the detach there is nothing to enumerate.

`CollectCommandedParties` still skips player parties, which is deliberate and separate — one player's retreat must not spend another player's troops or clear their camp. The cascade removes them from the map event either way; what was missing was telling them.

source/GameInterface/Services/MapEvents/TroopSupply/CoopTroopSupplier.cs
alikhan866: Understood — I'll make it exact. Writing down the design first, because the obvious reading of "move it to the server" runs into a snag worth naming.

The server cannot apportion directly today: the allocation is `InitialSpawnNumber`, which the *engine* derives on each client from the battle-size split after `Init`, and the server never sees it. So "server computes each owner's count and sends it" would mean the server also reproducing the battle-size calculation — more coupling than this is worth, and fragile against any engine-side change.

What gets the same guarantee with the data already on the wire is to make the *decision* server-authoritative rather than the arithmetic. `PartyReserve` already carries `SideOffset`, so the cumulative-flooring shares are exact by construction; the only thing that breaks the sum is the local top-up, which each owner applies blind to the others. If the server also sends the `(offset, count)` intervals of the **player-owned parties on the side**, every client can compute the identical picture:

- base shares by cumulative flooring — exact, sums to the allocation;
- the set of player-owned parties whose base is zero — identical on every client, since they all see the same intervals;
- one troop granted to each of those, paid for by decrementing the largest base share, ties broken by lowest offset — deterministic, so every client reaches the same answer.

Sum stays exactly the allocation, every player still fields an agent, and no client needs to know what any other client *owns* — only which intervals are player-held. It is additive on the wire with a default of empty, so an older reserve just reproduces today's behaviour.

I'd rather land this as its own change than fold it into this PR: it alters the reserve format, so a mid-session client and server on different builds would disagree about the split, and I want it to carry its own tests for the tie-breaking and the zero-allocation edges. Flag it if you'd prefer it here and I'll move it in.

alikhan866: Done — the allocation is now exact, including your counterexample.

The fix turned out simpler than the compensation scheme I sketched above: reserve the troop **before** the split rather than adding it after. The server sends how many parties on the side belong to a player (`PlayerOwnedPartyCount`) and each one's rank among them (`PlayerOwnedRank`); every owner then takes one troop off the top per player-owned party, apportions the remainder across the whole side by the same cumulative flooring, and adds its own reserved troop back.

That sums to exactly the allocation by construction — the flooring covers the remainder exactly once, and there are exactly as many reserved troops as player-owned parties — so no owner ever has to give one up, and no owner needs to know what the others hold. It also means the server never has to reproduce the engine's battle-size calculation, which was the snag with apportioning centrally.

A wave smaller than the number of player-owned parties is the one case somebody misses out; the ranks settle it identically everywhere because the server assigns them.

Tests cover your case directly (1 troop of 1000 against an allocation of 100 — exact, and the small owner still fields its player), three owners including an AI-held one across a range of allocations, and the too-small-wave tie-break. Both fields are additive with 0/-1 defaults, so a reserve from a peer that does not send them falls back to the previous behaviour rather than fielding nobody.

garrettluskey:
Thank you for responding to the scope feedback by splitting and flattening this work. The independent `development` bases, the two documented real dependencies, and closing the duplicate/artificial stack members were the right changes. #2748 and #2749 also show that the smaller review cycle can work well.

The remaining issue is review-queue strategy rather than implementation. Thirteen PRs were opened essentially at once, and there are currently nine functional PRs from this batch marked ready. That presents more review work than the team can reasonably process at one time.

For this batch, please:

1. Keep #2750 as the current review item and choose at most one additional PR as the next item; move the rest back to draft.
2. Keep dependent PRs such as #2752 draft until their parent is accepted.
3. Mark a PR ready only after its scope is stable, the standard PR template is completed, and red checks are resolved or clearly explained.
4. Recheck overlap with `development` and existing PRs before requesting review so we do not repeat the #2747 duplicate review cycle.
5. Revisit PRs whose own titles/bodies describe multiple independent outcomes—especially #2755, #2756, and #2758. Split them again unless the changes are inseparable, in which case explain the single root problem clearly in the body.
6. Keep inline review responses concise: what changed, the commit, and the validation performed.

Please leave this PR as a draft index/not-for-review. For future batches, a tracking issue is preferable to retaining a large duplicate-diff tracking PR.

PR #2671, opened 2026-08-04
[1779] Hide disconnected player parties on server start

### LINKED ISSUE #1779: [ISSUE] Disconnected Player Parties Visible on Server Start
# For New Contributors

## Description

When the server starts player parties are visible. They only go invisible on disconnect

## Location

## Definition of Done

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

After restarting a server from a save, parties belonging to players who have not reconnected remain visible and interactable on the campaign map.

## What is the new behavior?

Resolves #1779

Saved parties for offline players are hidden and inactive as soon as the server finishes loading. Their party returns when they reconnect, while parties still involved in battles remain available until the battle ends.

## Bot Changelog Entry

- Hide saved parties belonging to offline players until they reconnect.


source/Coop.Core/Server/Services/Players/Handlers/PlayerPartyVisibilityHandler.cs
```diff
@@ -60,16 +61,33 @@ public PlayerPartyVisibilityHandler(
...
+    {
+        if (ModInformation.IsClient) return;
+
+        foreach (var player in playerManager.Players)
+        {
+            if (playerManager.IsConnected(player)) continue;
+            if (!objectManager.TryGetObjectWithLogging<MobileParty>(player.MobilePartyId, out var party))
+                continue;
+
+            ParkParty(player, party, "its saved player is offline");
```
garrettluskey, line 91: does this run before the map builds any party visuals at load? if so `RemoveVisual` is a no-op here. is there anything that keeps the visual manager from building figures for parked parties afterwards, which `PartyVisualLifetimePatches` would then replicate to clients?

source/Coop.Core/Server/Services/Players/Handlers/PlayerPartyVisibilityHandler.cs
```diff
@@ -89,31 +107,37 @@ private void Handle_PlayerDisconnected(MessagePayload<PlayerDisconnected> payloa
...
+                party.StringId,
+                mapEvent.StringId,
+                reason);
+            return;
+        }
 
-            party.IsActive = false;
+        var wasActive = party.IsActive;
+        party.IsActive = false;
+        party.IsVisible = false;
```
garrettluskey, line 137: Nit: anything on the server read `_isVisible` directly? `PartyIsSpottedServerPatch` already forces `IsSpotted` to `IsActive`, and `IsVisible` isn't in `MobilePartySync` so this never reaches clients.

garrettluskey (commented):
solid fix. reusing the existing park flow keeps the map event case working, and `SavedPlayerRegistrationsRestored` publishes synchronously out of `MapScreen.OnInitialize`, so `ParkParty` already runs on the game thread without extra wrapping. unit tests cover the new handler and the publish order. two questions inline, neither blocking.

source/Coop.Core/Server/Services/Players/Handlers/PlayerPartyVisibilityHandler.cs
ShoT-UPfps: `MapScreen` calls `InitializeVisuals` from its constructor, so `MobilePartyVisualManager` builds the existing party visuals before the `MapScreen.OnInitialize` postfix publishes `CampaignReady`. `RemoveVisual` removes that existing figure here, and after initialization the manager only builds another one for `MobilePartyCreated`, which this saved party does not fire.

source/Coop.Core/Server/Services/Players/Handlers/PlayerPartyVisibilityHandler.cs
ShoT-UPfps: yes, vanilla `MobilePartyVisual` reads `MobileParty.IsVisible` while updating its native entity, which reads this backing field. it isnt networked, `NetworkDestroyPartyVisual` is still what removes the client figures.

garrettluskey (dismissed):
Re-review at the merged head: only change since the last review is merging development, and both of my earlier questions are settled. The park at load does hit real figures because MapScreen builds them in its constructor before CampaignReady publishes, and after init the visual manager only adds figures on MobilePartyCreated, which a saved party never fires. And the IsVisible write isn't redundant, vanilla MobilePartyVisual reads it while updating the native entity, with NetworkDestroyPartyVisual still handling the client figures. The SavedPlayerRegistrationsRestored publish is in place in SaveGameHandler and the park path reuses the disconnect flow, so the MapEvent deferral and un-park on PlayerCampaignEntered still apply.

garrettluskey (approved):
Re-check at the new head: the only movement since the last review is another development merge, the functional diff against development is unchanged from what I approved, and the SavedPlayerRegistrationsRestored wiring is intact through the merge.

PR #2675, opened 2026-08-04
Multiplayer quests

Thaipho:
Re-enables quests behind an allowlist, with real multiplayer sync for each (ownership so only the actual quest owner can complete it, proper handling of values that could differ between peers, party spawns gated to the host). 35 of the 40 vanilla quest types so far

**Known limitation:** three upcoming quest types (Gang Leader Needs Weapons, Merchant Army of Poachers, LandLord Company of Trouble - not yet in this PR) trigger their fights through a scripted, dialogue-driven battle launch rather than the normal encounter path. The mechanism that lets other players join the same battle instance is only wired up for normal encounters, so for these three, each player would currently end up in their own separate fight instead of a shared one. A fix is already planned just not implemented yet. The fix is to extend the existing hook to also catch those two, routing them into the same already-working shared-battle system instead of building something new from scratch.


garrettluskey:
Can we start with trade quests? This is a bit too much to add all at once

Thaipho:
@garrettluskey want me to separate the trade requests and open a a new PR

garrettluskey:
PR too broad, closing

PR #2681, opened 2026-08-04
[2651] Remove per-party pending map event resolution

### LINKED ISSUE #2651: Remove per-party pending map event resolution from movement ticks
## Description

Every moving-party tick enters `PendingMapEventPartyMovementPatch` and resolves `IMapEventInitializationBarrier` from Autofac before scanning barrier states for the party. In the #2641 one-client 900-vs-900 active-combat trace, `ContainerProvider.TryResolve` accounted for 1,685 ms beneath a 1,849 ms pending-check stack and a 4,227 ms `ParallelTickMovingParties` subtree over 31.02 seconds. Deployment showed the same background overhead.

This is a follow-up to #2641 and uses the profiling support from #2643.

## Requirements

- Bind the lifetime-scoped initialization barrier once instead of resolving it per moving party.
- Replace `states.Values.Any(...)` with an O(1) pending-party view.
- Publish worker-readable state safely; do not expose a mutable collection to concurrent campaign workers.
- Clear the binding and pending state with the lifetime scope so a later session cannot observe stale parties or services.
- Preserve pending map-event movement blocking, commit, abort, party attach/detach, and disposal behavior.
- Do not suppress client campaign simulation as part of this PR.
- Instrument resolve attempts, pending lookups, pending hits, state-view updates, and moving-party CPU.

## Live verification

- Use exactly one client (`testclient`); never launch a second client.
- Use the exact uncapped 900-vs-900 field-battle fixture from #2641.
- Capture baseline and changed deployment/active-combat profiles on the same client process.
- Report FPS, normalized CPU, pending-check time, service-resolution count, lookup count/hit rate, and `ParallelTickMovingParties` time.
- Exercise a pending map-event transition so the optimized predicate is proven behaviorally, not only by a zero-hit benchmark.

## Definition of done

- [ ] Moving-party ticks do not resolve the barrier from Autofac per party.
- [ ] Pending-party reads are O(1) and safe across campaign worker threads.
- [ ] Map-event initialization behavior remains correct through commit, abort, and disposal.
- [ ] Independent one-client 900-vs-900 evidence shows the CPU and frame-time impact.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Large battles repeatedly perform the same pending map-event lookup and scan for every moving party, adding client CPU work even when almost no parties are waiting for initialization.

## What is the new behavior?

Resolves #2651

The client reuses one session lookup and checks pending parties directly, preserving movement blocking while map events initialize and releasing parties when initialization completes, is cancelled, or aborts.

| 900-vs-900 client metric | Current | New |
| --- | --- | --- |
| Pending transition | 2.026 µs/lookup, 3.158 s total | 0.180 µs/lookup, 0.327 s total (91% less per lookup) |
| Deployment (20 s) | 2.169 µs/lookup, 3.257 s total | 0.091 µs/lookup, 0.157 s total (96% less per lookup) |
| Active combat (25 s) | 2.274 µs/lookup, 4.149 s total | 0.091 µs/lookup, 0.173 s total (96% less per lookup) |
| Pending-path resolutions | 1.50–1.82 million per window | 0 per window, with 1 scope binding |

## Bot Changelog Entry

- Reduced CPU usage during battles

### DIFF
--- a/source/GameInterface/Services/MapEvents/Initialization/MapEventInitializationBarrierBinding.cs
+++ b/source/GameInterface/Services/MapEvents/Initialization/MapEventInitializationBarrierBinding.cs
@@ -1,0 +1,42 @@
+﻿using Autofac;
+using Autofac.Core;
+using System;
+using System.Collections.Concurrent;
+
+namespace GameInterface.Services.MapEvents.Initialization;
+
+internal sealed class MapEventInitializationBarrierBinding : IDisposable
+{
+    private static readonly ConcurrentDictionary<IComponentRegistry, IMapEventInitializationBarrier> Bindings =
+        new ConcurrentDictionary<IComponentRegistry, IMapEventInitializationBarrier>();
+
+    private readonly IComponentRegistry componentRegistry;
+    private readonly IMapEventInitializationBarrier barrier;
+    private bool disposed;
+
+    public MapEventInitializationBarrierBinding(
+        ILifetimeScope lifetimeScope,
+        IMapEventInitializationBarrier barrier)
+    {
+        componentRegistry = lifetimeScope.ComponentRegistry;
+        this.barrier = barrier;
+
+        if (!Bindings.TryAdd(componentRegistry, barrier))
+            throw new InvalidOperationException("A map-event initialization barrier is already bound to this scope");
+    }
+
+    public void Dispose()
+    {
+        if (disposed) return;
+        disposed = true;
+        Bindings.TryRemove(componentRegistry, out _);
+    }
+
+    internal static bool TryGet(
+        ILifetimeScope lifetimeScope,
+        out IMapEventInitializationBarrier barrier)
+    {
+        barrier = null;
+        return lifetimeScope != null && Bindings.TryGetValue(lifetimeScope.ComponentRegistry, out barrier);
+    }
+}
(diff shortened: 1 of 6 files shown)


source/GameInterface/Services/MapEvents/Initialization/MapEventInitializationBarrierBinding.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Autofac;
+using Autofac.Core;
+using System;
+using System.Collections.Concurrent;
+
+namespace GameInterface.Services.MapEvents.Initialization;
+
+internal sealed class MapEventInitializationBarrierBinding : IDisposable
```
garrettluskey, line 8: Nit: does this one need an interface? we usually declare one at the top of the file for new classes, but since nothing injects it and the tests hit the static `TryGet` directly, maybe not.

garrettluskey (approved):
the snapshot approach checks out. every spot that touches `state.Parties` or `state.Announced` re-publishes (`Register`, `SetServerPartyPending` including the rollback path, `CommitServer`, `AbortServer`/`DestroyGraph`, `HandlePendingParty`, `FinishClient`, `AttachClient`, `Dispose`), so the volatile read in `IsPartyPending` sees the same set the old scan did, just without the per-party resolve and walk. binding lookup keyed by component registry keeps the server and client containers isolated in the e2e environment, and the new tests cover isolation, cancellation, abort, and concurrent readers. CI is green. one small question inline, nothing blocking.

PR #2682, opened 2026-08-04
Sync trade agreements

### LINKED ISSUE #2655: [ECONOMY] Income Information Shows 101 Denars Instead of 975 per Day
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

The Income Information display reports 101 denars per day, although the daily income should be 975 denars.

## Steps to reproduce

Not provided

## Additional information

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1533963523493728266/1533963523493728266

## Images and videos

[](<https://cdn.discordapp.com/attachments/1533963523493728266/1533963524588437644/2026081.JPG?ex=6aa47f5b&is=6aa32ddb&hm=0fed1cdfa538daab5161890d53142c1fe67720a3ad28a89f347a58328dc80f14&>) ([Discord message](https://discord.com/channels/709516043332354119/1533963523493728266/1533963523493728266))

## Logs

Client log: Not provided

Server log: Not provided

## Additional report details

- The save game is located on the dedicated server in `DedicatedServer\server-data\Game Saves`. ([Discord message](https://discord.com/channels/709516043332354119/1533963523493728266/1534017459919650986))

- All clients on the server had this problem. ([Discord message](https://discord.com/channels/709516043332354119/1533963523493728266/1534017905803526295))

## Save file

- [Download 4thTimesTheCharm.sav](<https://cdn.discordapp.com/attachments/1533963523493728266/1534017994647146577/4thTimesTheCharm.sav?ex=6aa4b216&is=6aa36096&hm=d4d3144378916bf45210a800c49759b1466eff27bfd7bd48160af51ba004aad0&>) ([Discord message](https://discord.com/channels/709516043332354119/1533963523493728266/1534017995502911528))

### LINKED ISSUE #2725: [ECONOMY] Displayed wage amounts do not match amounts received on a dedicated server
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

In-game wage amounts do not match the amounts actually received across multiple wages on a dedicated server. For example, the game showed 6,700 income while the reporter lost about 200.

## Steps to reproduce

Not provided

## Additional information

Relaunching the server and logging out and back in were tried.

## Discord source

https://discord.com/channels/709516043332354119/1534431106902921287/1534431106902921287

## Logs

Client log: [0.1.1+324d59795711727d40bc1aa3eb56a652ae656306 - Download Coop_client.log](<https://cdn.discordapp.com/attachments/1534431106902921287/1534431108446289970/Coop_client.log?ex=6aa4e154&is=6aa38fd4&hm=c07ec52ec7b50b4cfd25bf67b895f550fd09e135dbf0d482f5addb45776bcdc2&>)

Server log: [Download coop-server-20260804-193822.log](<https://cdn.discordapp.com/attachments/1534431106902921287/1534431107704029245/coop-server-20260804-193822.log?ex=6aa4e154&is=6aa38fd4&hm=7b25b7e92dc4d5659b34c89951600580474f12289e6e448093e83c384892b361&>)

## Images and videos

[](<https://cdn.discordapp.com/attachments/1534431106902921287/1534434060880056410/image.png?ex=6aa4e414&is=6aa39294&hm=027d012ac29f2130733fbf2533a2add3650a52135163fd2dfe6a30548ec919b4&>) ([Discord message](https://discord.com/channels/709516043332354119/1534431106902921287/1534434061190303814))

[](<https://cdn.discordapp.com/attachments/1534431106902921287/1534687004200271924/image.png?ex=6aa47e26&is=6aa32ca6&hm=75b18e76d7012f8b31c5293039bb1c1b2d7bf39d39cb2aead60af79e29a5e724&>) ([Discord message](https://discord.com/channels/709516043332354119/1534431106902921287/1534687004598468790))

jordanbrymora:
## PR Checklist

## PR Type

The `TradeAgreementsCampaignBehavior` is both not disabled and not synced. This means that all trade agreements are different between the server and clients. Because the server's gold calculation is what actually causes the gold change, players can see a different amount than what their own client calculated when part of a kingdom with a trade agreement.

## What is the new behavior?

Resolves #2655
Resolves #2725

Synced `TradeAgreementsCampaignBehavior` to keep trade agreements synced between server and clients.

## Bot Changelog Entry

Fixed mismatching gold between predicted and actual change when in a kingdom with trade agreements.


source/GameInterface/Services/UI/Notifications/Messages/NetworkNotifyTributePaymentEnded.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Common.Messaging;
+using ProtoBuf;
+
+namespace GameInterface.Services.UI.Notifications.Messages;
+
+[ProtoContract(SkipConstructor = true)]
+internal readonly struct NetworkNotifyTributePaymentEnded : ICommand
+{
+    public readonly string ClanId;
+    public readonly string PayerFactionId;
```
garrettluskey, line 13: `ClanId` and `PayerFactionId` need `[ProtoMember]` tags. protobuf ignores untagged fields on a `[ProtoContract]`, so both ids deserialize as null and every tribute-ended notification stops at the first lookup.

source/GameInterface/Services/Kingdoms/Patches/TradeAgreementsCampaignBehaviorPatches.cs
```diff
@@ -0,0 +1,150 @@
...
+        // Vanilla uses a check for player hostility to apply penalty for player.
+        // Player hero that caused the war to be declared isn't available here.
+        var targetHero = kingdom.Leader;
+        if (detail == DeclareWarAction.DeclareWarDetail.CausedByPlayerHostility)
+        {
+            // TODO: Apply the penalty to the hero of the client that declared the war once it is passed down here
+            // Traits are not synced yet either.
+            //targetHero = playerHero;
+            //TraitLevelingHelper.OnTradeAgreementBroken();
+            return false;
```
garrettluskey: this drops the whole -50 relation penalty for every `ApplyByPlayerHostility` war. player-party hostile encounters use that detail, so carry the initiating player hero into this path and apply the penalty instead of returning.

garrettluskey (changes requested):
the trade agreement mutations are applied on the game thread and the id lookups stay inside the queued actions. two paths still lose behavior: the tribute notification payload has no protobuf members, and player-hostility wars skip the agreement penalty.

source/GameInterface/Services/Kingdoms/Patches/TradeAgreementsCampaignBehaviorPatches.cs
jordanbrymora: Re-worked to use ResolvedMainHeroContext from the location that a war was declared from and kept the kingdom leader as a fallback if no player hero is available.

source/GameInterface/Services/UI/Notifications/Messages/NetworkNotifyTributePaymentEnded.cs
jordanbrymora: Fixed

source/GameInterface/Services/Kingdoms/Handlers/TradeAgreementsHandler.cs
```diff
@@ -0,0 +1,309 @@
...
+
+            if (tradeAgreementsBehavior.TryGetTradeAgreement(newTradeAgreement.Kingdom1, newTradeAgreement.Kingdom2, out var _))
+            {
+                Logger.Debug("Skipping forming new trade agreement between {kingdom1} and {kingdom2} because an agreement is already present.",
+                    newTradeAgreement.Kingdom1.Name,
+                    newTradeAgreement.Kingdom1.Name);
+                return;
+            }
+
+            tradeAgreementsBehavior._tradeAgreements.Add(newTradeAgreement);
```
garrettluskey, line 201: `MakeTradeAgreement` also raises `OnTradeAgreementSigned`, which adds the trade-agreement log entry. raise that event after this add, otherwise connected clients miss the entry while the server and late-join saves have it.

garrettluskey (changes requested):
the tribute payload and player-hostility relation penalty now keep the data and initiating hero from the previous review. connected clients still skip the vanilla signed-agreement event when applying a new agreement.

source/GameInterface/Services/Kingdoms/Handlers/TradeAgreementsHandler.cs
jordanbrymora: It'd be better to handle logs more like the default notifications where the server fires the events and then sends notifications to clients when appropriate. Otherwise we have to spread out where different logs come from and fire the events manually on clients when needed. Firing the events on clients is also more risky as they can lead to other behaviour then just the logging (just not for this specific case).

garrettluskey (changes requested):
the tribute payload and player-hostility penalty are fixed, and the agreement mutations are game-thread ordered. the existing `TradeAgreementsHandler.cs:201` signed-agreement event comment still applies, connected clients miss the resulting trade-agreement log entry.

jordanbrymora:
> the tribute payload and player-hostility penalty are fixed, and the agreement mutations are game-thread ordered. the existing `TradeAgreementsHandler.cs:201` signed-agreement event comment still applies, connected clients miss the resulting trade-agreement log entry.

Repeated from previous comment:

It'd be better to handle logs more like the default notifications where the server fires the events and then sends notifications to clients when appropriate. Otherwise we have to spread out where different logs come from and fire the events manually on clients when needed. Firing the events on clients is also more risky as they can lead to other behaviour then just the logging (just not for this specific case).

Logging the trade should therefore be out of scope for this PR and should be added when `DefaultLogsCampaignBehavior` updates clients for all logs for issue https://github.com/Bannerlord-Coop-Team/BannerlordCoop/issues/2994.

source/GameInterface/Services/Kingdoms/Handlers/TradeAgreementsHandler.cs
garrettluskey, line 201: `MakeTradeAgreement` also raises `OnTradeAgreementSigned`, which creates the trade-agreement log. adding directly to `_tradeAgreements` means connected clients miss that result while late joiners load the agreement state. replicate the resulting log/notification on this path.

garrettluskey (changes requested):
the agreement mutations and id lookups stay game-thread ordered, and the tribute and player-hostility paths are fixed. connected clients still miss the signed-agreement outcome.

source/GameInterface/Services/Kingdoms/Handlers/TradeAgreementsHandler.cs
jordanbrymora: Fixed

garrettluskey (approved):
the new-agreement apply now adds the state and raises `OnTradeAgreementSigned` on the same game-thread action, so connected clients receive the same log/notification outcome as the server and late joiners. the previous blocker is fixed and all checks pass.

PR #2705, opened 2026-08-04
Settlement NPC host ownership and migration

garrettluskey:
## What this does

Settlement-mission NPCs (ambient crowd, roster heroes, animals — every native non-player agent in town center / tavern / lord's hall / castle courtyard / village interiors) are now **spawned and owned by one elected host player** and replicated to the other players as movement-driven puppets, with **adopt-in-place host migration** when that player leaves or disconnects. This replaces the old model where every client rolled its own seeded-RNG crowd that diverged the moment the mission started.

It mirrors the field-battle host architecture end to end: server-authoritative election with a successor line and epoch-ordered assignments, host-side capture patches feeding batched LZ4 spawn records over the mission mesh, and a migrator that transfers authority and revives settlement AI on promotion.

## How it works

- **Election** (`LocationHostHandler` + `ILocationHostRegistry`): first mission-ready client wins; later entrants append in arrival order; the epoch watermark survives instance teardown so re-entered settlements cannot resurrect a stale assignment. Deliberately a separate registry from battles — each handler ignores departures for instance ids its own registry does not hold.
- **Suppression** (`LocationNativeSpawnSuppressionPatches`): every native population entry point (`SpawnLocationCharacters` and the per-agent paths, the six `MissionHelper` animal spawns, the 30s `AgentPassageUsageTick` roster churn) is skipped while `LocationNpcGate.ShouldSuppressNativeSpawns` holds — all clients until the assignment lands, non-hosts forever. Every seam is decompile-verified (doc/SettlementRequirements.md, Appendix A).
- **Capture + replication** (`LocationAgentSpawnedPatch`/`LocationMonsterSpawnedPatch` → `LocationOwnedAgentReplicator`): the host runs the real native pass once confirmed (`LocationPopulationDirector`) and every non-player spawn is registered under a minted compact movement scope and broadcast in bounded batches. Animals get their own `SpawnMonster` capture — they never pass `SpawnAgent(AgentBuildData, bool)`.
- **Roster-bound puppets** (`LocationPuppetRosterBinder` + `LocationPuppetSpawner`): spawn records carry the NPC roster identity; peers bind an existing local entry (server-synced heroes) or reconstruct one, and build the puppet **from that entry origin**. Native bookkeeping (`IsAlreadySpawned`, `GetLocationCharacter(agent.Origin)`, passage guards) then recognizes puppets on every client — which is what makes migration exact and structurally prevents duplicate crowds on a promoted host.
- **Migration** (`LocationAuthorityMigrator`): departures despawn only the leaver player agent; NPC puppets await adoption. The promoted client transfers authority, forgets stale interpolation targets, and re-creates each NPC settlement AI via the verified native tail (`CampaignAgentComponent.CreateAgentNavigator` + `entry.AddBehaviors`), with a stationary-AI fallback for unresolvable entries and an absent-controller sweep for NPCs inherited through earlier migrations.
- **Mid-mission dynamics**: host-side removals (passage exits, churn) and deaths replicate as reasoned despawns on the same reliable-ordered stream, with tombstones so buffered records cannot resurrect a despawned NPC. Late joiners get a full catch-up replay.
- **Conversations**: the existing server conversation lock now broadcasts hold/release, and the host pauses the reserved NPC so a remote player conversation anchors to a stationary agent.

## Commits (one per phase, each independently green)

- 11b20616a — groundwork: decompile verification (V1–V9) into doc/SettlementRequirements.md, dead-code cleanup (instance-id unification, dead mesh echo handler, inert ambient machinery)
- 7b0287ebd — election: registry, session, gate, server-authoritative handler + 11 E2E election tests incl. battle/location registry isolation
- 77941fb33 — spawning: suppression, capture, roster-bound puppet replication, population director
- 8527d781e — migration: departure fork + adopt-in-place
- 1cff08154 — despawn + death replication (v1: peers fade dead NPCs; the wire carries the reason for a later blow-replay upgrade)
- d09eb2bf8 — conversation hold + requirements finalization

## Verification

- Coop.Tests **554 passed** (incl. new registry/session/gate/binding-map units; `TestComponentBase` mocks `ILocationHostRegistry` like the battle registry)
- `HarmonyPatchesAll` **green** — all new patches apply against the real game DLLs (incl. the byref-pinned `SpawnMonster` overload and the nested `MissionHelper` statics)
- E2E location + battle-host + conversation suites **79 passed**; battle suites stayed green throughout (the battle stack is touched in exactly one behavior-gated line in `RemoveControllerParty`)

## Outstanding before this is player-proven

- The **2–3 client live checklist** at the top of doc/SettlementRequirements.md (identical crowds, migration with no duplicate crowd, churn staying mirrored, late-joiner catch-up, conversation hold)
- Mock-engine mirror tests for the spawn/adoption verticals (needs `SpawnMonster`/action-set fixture shims — one follow-up harness task)
- Noted follow-ups: death blow-replay fidelity on peers, non-host NPC damage routing, ambient seed-patch retirement


source/Missions/Locations/LocationAuthorityMigrator.cs
```diff
@@ -0,0 +1,222 @@
...
+                adopted.Count, controllerId, reason, revived, stationary);
+        }, context: nameof(AdoptAgentsFrom));
+    }
+
+    // [Game thread] Turn an inert puppet into a locally simulated settlement NPC. Humans get the
+    // native AI stack re-created from their roster entry; an unresolvable entry (or an animal) falls
+    // back to plain engine AI — stationary for humans (SR-031), native idle wandering for animals.
+    private bool ReviveSettlementAi(Agent agent)
+    {
+        agent.Controller = AgentControllerType.AI;
```
garrettluskey, line 261: ReviveSettlementAi flips every adopted non-human straight to Controller = AI, and that includes scene horses - they are IsMount agents from SpawnMonster. BattleAuthorityMigrator deliberately keeps adopted mounts authority-only ("an adopted MOUNT only changes authority - it is not a combatant: no AI controller, no formation, no wake"), and handing a riderless horse an engine AI controller is exactly the class of edge we kept relearning in battle migration. Sheep and chickens should keep getting AI (they are native-AI animals); mounts should not. Authority transfer and interpolator.Forget above already cover them:

```csharp
if (agent.IsMount) return true; // authority-only, mirroring battle mount adoption
agent.Controller = AgentControllerType.AI;
agent.SetIsAIPaused(false);
```

source/GameInterface/Services/Locations/Conversations/Handlers/LocationNpcHoldHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        GameThread.RunSafe(() =>
+        {
+            var mission = Mission.Current;
+            if (mission == null) return;
+
+            if (!objectManager.TryGetObject<Location>(locationId, out var location) || location == null) return;
+            if (!objectManager.TryGetObject<CharacterObject>(characterId, out var character) || character == null) return;
+
+            foreach (var entry in location.GetCharacterList())
```
garrettluskey: location.GetCharacterList() can return null, and every other consumer in this service guards it (SettlementPopulationTracker, the roster patches, the binder, even the debug command). Here the NRE gets swallowed by RunSafe, so the hold silently never lands and the NPC keeps strolling mid-conversation with nothing in the log to explain it. Guard it like the rest:

```csharp
foreach (var entry in location.GetCharacterList() ?? Enumerable.Empty<LocationCharacter>())
```

source/GameInterface/Services/Locations/Patches/LocationAgentSpawnedPatch.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        string itemId = null;
+        if (equipmentElement.Item != null)
+            objectManager.TryGetId(equipmentElement.Item, out itemId);
+
+        string harnessItemId = null;
+        if (harnessRosterElement.Item != null)
+            objectManager.TryGetId(harnessRosterElement.Item, out harnessItemId);
+
+        if (itemId == null)
```
garrettluskey, line 77: When the animal item id does not resolve, the record is dropped with no log at all. That is precisely the case where one peer's town is missing its sheep and there is nothing to grep - we lean on mutation logging for exactly this kind of live diagnosis. One warning naming the item StringId before the return is enough.

source/Missions/Locations/LocationPuppetSpawner.cs
```diff
@@ -0,0 +1,430 @@
...
+        // passage guards) and is what a promoted host re-binds AI through.
+        LocationCharacter entry = null;
+        if (data.RosterEntry != null)
+            rosterBinder.TryBindOrReconstruct(data.RosterEntry, out entry);
+
+        AgentBuildData buildData;
+        string actionSetCode;
+        if (entry != null)
+        {
+            buildData = entry.GetAgentBuildData();
```
garrettluskey: Nit: chain .NoHorses(true) on the bound-entry build - every native settlement spawn passes it (SpawnWanderingAgentWithInitialFrame always does). The SpawnEquipment override makes it moot in practice since the host's agent spawned horseless, but a roster entry whose AgentData carries battle equipment would mint a surprise mount on peers if the equipment override is ever skipped.

garrettluskey (commented):
The architecture lands the way it should: the separate host registry with the registry-miss discriminator, puppets built from local roster-entry origins so native bookkeeping recognizes them on every client, and the state-derived promotion check in the population director are all the right shape, and the election/isolation E2E coverage is real coverage, not ceremony. This would be a request-changes review, but GitHub does not allow that on one's own PR, so it rides as a comment: one correctness issue in the migration path and two robustness gaps that would silently eat functionality in a live run, plus one parity nit.

On the red CI: shards 1 and 7 fail the same tests with the same generic asserts on the development baseline (runs 30708066130 / 30707777657 - the known order-flake set: MapEventCollection, BattleSurrender, EncounterSurrenderOption, VillageHostileAction), so they carry no signal for this PR. The full local E2E suite on this branch is 1154 passed / 1 failed, and the one failure is TournamentSpectatorRulesTests.OrangeItem_UsesNativeMeshAndIsRegisteredForSupportedGameTypes reading source/deploy/ModuleData/tournament_spectator_throwables.xml - a full-deploy artifact that does not exist in a git worktree; the test predates this branch, this PR touches no tournament or deploy files, and the same test is green in this PR's CI. Net: nothing in the red is introduced here, which also means the live checklist in doc/SettlementRequirements.md is doing the real acceptance work.

source/Missions/Locations/LocationAuthorityMigrator.cs
garrettluskey: Fixed in ed12adafa.

source/GameInterface/Services/Locations/Conversations/Handlers/LocationNpcHoldHandler.cs
garrettluskey: Fixed in ed12adafa.

source/GameInterface/Services/Locations/Patches/LocationAgentSpawnedPatch.cs
garrettluskey: Fixed in ed12adafa.

source/Missions/Locations/LocationPuppetSpawner.cs
garrettluskey: Fixed in ed12adafa.

source/Missions/Locations/LocationPuppetSpawner.cs
```diff
@@ -0,0 +1,432 @@
...
+        if (instanceId != null && instanceId != session.InstanceId) return;
+
+        lock (withdrawnControllerLock)
+        {
+            withdrawnControllers.Add(controllerId);
+        }
+
+        lock (pendingPuppetLock)
+        {
+            pendingPuppets.RemoveAll(data => data.OwnerControllerId == controllerId);
```
ShoT-UPfps: we drop every buffered old-host record here before the promotion applies. if this client only spawned part of the catch-up it adopts that subset and loses the rest, and if it spawned none `LocationPopulationDirector` rolls a new crowd with new ids because `bindingMap.Count` is 0. can we preserve or re-key these records and use explicit migration state instead?

source/Missions/Locations/LocationPuppetSpawner.cs
```diff
@@ -0,0 +1,432 @@
...
+        GameThread.RunSafe(
+            () => SpawnPuppetBatch(message, agents),
+            context: nameof(Handle_NetworkSpawnLocationAgents));
+    }
+
+    private void SpawnPuppetBatch(
+        NetworkSpawnLocationAgents message,
+        LocationAgentSpawnData[] agents)
+    {
+        if (Mission.Current == null)
```
ShoT-UPfps, line 212: can we gate these spawns on the same mission-ready state as `HandleJoinInfo`? `Mission.Current` exists while the mission is still loading, so the first host can spawn this crowd before teams and the local player agent are initialized.

source/Missions/Locations/LocationPuppetSpawner.cs
```diff
@@ -0,0 +1,432 @@
...
+                var agent = info.Agent;
+                if (agent != null && agent.IsActive() && agent.Health > 0)
+                {
+                    bool hideMount = agent.HasMount && agent.MountAgent != null && agent.MountAgent.IsActive();
+                    agent.FadeOut(false, hideMount);
+                }
+
+                coopMissionComponent.AgentMovementHandler.Interpolator.Forget(agent);
+                registry.RemoveAgent(id);
+                bindingMap.Forget(id);
```
ShoT-UPfps, line 175: we need to replicate the roster move here, not only fade the agent. vanilla `LocationComplex.ChangeLocation` removes the entry from the source location and adds it to the destination, otherwise a promoted client has stale passage rosters and the npc cannot be selected to return.

source/GameInterface/Services/Locations/Conversations/Handlers/LocationNpcHoldHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+
+    private void Handle_NetworkLocationNpcReleased(MessagePayload<NetworkLocationNpcReleased> payload)
+    {
+        SetNpcPaused(payload.What.LocationId, payload.What.CharacterId, paused: false);
+    }
+
+    private void SetNpcPaused(string locationId, string characterId, bool paused)
+    {
+        if (ModInformation.IsServer) return;
+        if (!LocationNpcGate.IsCoopLocationMissionActive || !LocationNpcGate.IsLocalHostConfirmed) return;
```
ShoT-UPfps, line 66: we drop hold state on every non-host here. if the host leaves during someone else's conversation, the successor has no hold to restore and `ReviveSettlementAi` unpauses the npc while the conversation is still active.

source/GameInterface/Services/Locations/Conversations/Handlers/LocationNpcHoldHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        GameThread.RunSafe(() =>
+        {
+            var mission = Mission.Current;
+            if (mission == null) return;
+
+            if (!objectManager.TryGetObject<Location>(locationId, out var location) || location == null) return;
```
ShoT-UPfps: please use `TryGetObjectWithLogging` for these lookups, otherwise a missed hold silently leaves the npc moving.

ShoT-UPfps (changes requested):
Comments

source/GameInterface/Services/Locations/Conversations/Handlers/LocationNpcHoldHandler.cs
garrettluskey: Fixed in 60471271e. Holds are tracked on every client in ILocationNpcHoldRegistry regardless of host status, and adoption re-applies the pause to a held NPC (both the bulk migration sweep and late-adopted stragglers). One residual documented in SR-040: a client that joined AFTER the hold broadcast and gets promoted within that same conversation has no hold state - closing that needs a server-side hold replay on assignment change, which I would rather do as a follow-up than couple the conversation handler to host assignments in this PR.

source/GameInterface/Services/Locations/Conversations/Handlers/LocationNpcHoldHandler.cs
garrettluskey: Fixed in 60471271e - both lookups use TryGetObjectWithLogging now.

source/Missions/Locations/LocationPuppetSpawner.cs
garrettluskey: Fixed in 60471271e. Despawn records now carry the destination location id (resolved host-side from where vanilla ChangeLocation left the entry), and receivers move the bound entry between the location lists via the roster binder before fading the agent - promoted hosts keep truthful passage rosters.

source/Missions/Locations/LocationPuppetSpawner.cs
garrettluskey: Fixed in 60471271e. A departed HOST's records are retained now (only a plain player's are pruned): the promoted client spawns each retained record and hands it straight to the migrator for late adoption, and migration is explicit on LocationHostAuthorityAcquired.WasMigration instead of inferred from bindingMap.Count. The promoted host only runs the population pass when the departed host provably left nothing behind (no bindings AND no pending records) - that case is a first crowd, not a duplicate one.

source/Missions/Locations/LocationPuppetSpawner.cs
garrettluskey: Fixed in 60471271e. The spawner now has the same readiness gate as the join-info path: records buffer until the controller registers the local player agent (NotifyMissionReady), and the drain holds off too.

source/Missions/Locations/LocationPuppetSpawner.cs
ShoT-UPfps: the destination array contains null for every death or plain removal, and protobuf-net throws when a repeated string contains a null element. can we send a non-null value here, otherwise those despawns and any mixed passage batch never serialize?

ShoT-UPfps (changes requested):
cmt

source/Missions/Locations/LocationPuppetSpawner.cs
garrettluskey: updated

ShoT-UPfps (approved):
lgtm

PR #2718, opened 2026-08-05
Fix residual E2E flake: serialize patch churn, restore the game-thread mark, reset leaked battle statics

garrettluskey:
## Problem

The campaign-time heartbeat fix (#2569, `1bba6186f`) reduced but did not end the repo-wide E2E flake: post-fix runs — including on `development` itself — still drop 8–15 `MapEvents`/`Villages`/`SiegeEvents` tests per shard, different subsets each run, all with one signature: **server-created `MapEvent`/`MapEventSide` state silently never materializes on the client replica** (`Assert.NotNull` on `MapEventSide` in the `SetMainPartyInBattle` helpers, `TryGetObject<MapEvent>` misses). Recent examples: runs 30935011666 (a credits-XML-only branch), 30934817042, 30953611637, 30952103612, and the 12 non-hideout failures in 30890911211.

Diagnosis was done by mining those runs' logs and auditing the harness. Three residual mechanisms cooperate, all harness-level:

### 1. Harmony patch churn was unguarded against concurrent packet simulation
`EnvironmentInstance.Call` serialized on its own static `_lock`, while `SimulateMessage`/`SimulatePacket` only took `GameInstance.@lock` (via `StaticScope`) — and `PatchScope` rewrote detours **outside** that lock. Any background thread delivering a packet could execute patched game code mid-patch/unpatch, silently dropping a detour. `Call` now takes `GameInstance.@lock` itself (the lock every `Simulate*` already honors). Monitor reentrancy is load-bearing and preserved: routed sends synchronously nest `Simulate*` inside a `Call` on the same thread.

### 2. The game-thread mark was stolen by worker threads and never restored
`StaticScope` called `MarkGameThread()` with no restore. A `Call` on a pool thread (`Task.Run(() => Server.Call(...))` in `PlayerPartyInteractionFlowTests`) permanently stole the mark, after which every `GameThread.RunSafe` client-apply — the map-event commit broadcast path (`MapEventInitializationBarrier` → `FinishClient`) — silently **enqueued onto a queue nobody pumps** instead of running inline. That is the desync. Stale queued actions then executed inside a *later* test's pump against a torn-down container — visible in the CI logs as the swallowed `VillagePatches.cs:44` NRE (`Failed to run action on the game thread: "(none)"`) and an Autofac `ObjectDisposedException`.

`StaticScope` now saves/restores the mark (new `GameThread.GameThreadId` / `RestoreGameThread`), and `E2ETestEnvironment` discards queued actions at construction and teardown (new `GameThread.DiscardQueuedActions`, which releases blocked waiters), with teardown re-marking its own thread so handler `Dispose` bodies run inline.

### 3. Battle-mode gating statics leaked across environments under reused ids
`ServerBattleModeArbiter.modes`, `BattleModeRegistry`, `BattleConclusionGate`, `BattleSpawnGate`, and `BattleSimulationReplay` are process-wide and keyed by object-manager ids — which restart at `MapEvent_Created_1` in every fresh environment. A claim leaked by a test that never finalized its battle silently gated joins, raid updates, and surrender for whichever unrelated class ran next — producing the order-dependent "different subsets each run" pattern (xUnit class order shifts between builds). The environment ctor now resets them next to the existing `PlayerPartyInteractionDialogState.Clear()` precedent.

### Supporting hardening
- `MeshNetworkRouter` deliveries now suspend the sender's `AllowedThread` allowance exactly like `TestNetworkRouter.Deliver` — receivers were inheriting the sender's allowance and silently skipping every receive-path patch.
- `MessageCollection`/`PacketCollection` guard their lists — background-thread sends were racing test-thread assertions with no exception on loss.
- `OutputSinkManager` snapshots its callback list under a lock and survives a dead xUnit `ITestOutputHelper` being logged to from a background thread after its test ended.

## Verification (local, Release)

- `CoopUnitTests.slnf`: **1,535 passed, 0 failed** (Common, CrashReporter, Coop, IntegrationTests, GameInterface)
- Full `E2E.Tests` suite in a single process: **1,171 passed, 0 failed, 4 skipped** (7m59s)
- The previously-failing suites also green when run as standalone chunks (MapEvents 187/187, Villages+Siege+Missions 646/646)

The flake does not reproduce locally (it is load-dependent, CI-only), so the real proof is CI stability over the next batch of PR runs — but each fixed mechanism is individually confirmed against the CI log breadcrumbs it produces.

Related but **not** part of this flake (deterministic branch-side failures found during log mining, owners notified separately): hideout-sync's `_isSpotted` intercept test mismatch, and more-barters' `AmbiguousMatchException` on the new `EndEngagement` overload.

### DIFF
--- a/.github/scripts/run-e2e-shard.sh
+++ b/.github/scripts/run-e2e-shard.sh
@@ -105,12 +105,14 @@ esac
 assigned_case_count=$(printf '%s\n' "$assignment" | cut -f2)
 shard_filter=$(printf '%s\n' "$assignment" | cut -f3-)
 echo "running E2E shard $shard_index/$shard_count with $assigned_case_count discovered test cases"
+# Per-test result lines (not ErrorsOnly): flake triage needs to reconstruct which tests ran,
+# in what order, before a failure — an errors-only log makes that unrecoverable afterwards.
 "$dotnet_cmd" test "$test_project" \
     -c Release \
     --no-build \
     --no-restore \
     --filter "$shard_filter" \
-    --consoleLoggerParameters:ErrorsOnly
+    --logger "console;verbosity=normal"
 
 elapsed_seconds=$(( $(date +%s) - start_seconds ))
 echo "E2E shard $shard_index completed in ${elapsed_seconds}s"
--- a/source/Common.Tests/Utils/MessageCollection.cs
+++ b/source/Common.Tests/Utils/MessageCollection.cs
@@ -10,7 +10,20 @@ public class MessageCollection : IEnumerable<IMessage>
 {
     public readonly List<IMessage> Messages = new List<IMessage>();
 
-    public int Count => Messages.Count;
+    // Test-spawned background threads can send (append) while the test thread enumerates for an
+    // assertion; an unguarded List race loses entries or tears the enumeration with no exception.
+    private readonly object gate = new object();
+
+    public int Count
+    {
+        get
+        {
+            lock (gate)
+            {
+                return Messages.Count;
+            }
+        }
+    }
 
     /// <summary>
     /// Gets an iterator for all messages with type <typeparamref name="TMessage"/>
@@ -19,7 +32,7 @@ public class MessageCollection : IEnumerable<IMessage>
     /// <returns>Iterator for all messages with type <typeparamref name="TMessage"/></returns>
     public IEnumerable<TMessage> GetMessages<TMessage>() where TMessage : IMessage
     {
-        return Messages
+        return Snapshot()
             .Where(msg => typeof(TMessage).IsAssignableFrom(msg.GetType()))
             .Select(msg => (TMessage)msg);
     }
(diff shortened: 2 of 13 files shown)


garrettluskey:
**Update — the first commit did not eliminate the CI flake** (shards 3/5/6 dropped the familiar cluster on this PR's own run), so I mined those three shard logs plus six historical runs in depth. Key findings:

1. **The failing set is a deterministic function of shard membership and is independent of execution order.** Three pre-fix shards with identical composition produced byte-identical 8-method failure sets in three visibly different class orders (one had `CoopBattleFinalizeTests` failing ~16s into the process, another at ~52s). Whenever the group is co-scheduled it fails; shard indices shift between branches only because the greedy method-level balancer re-packs the discovered test list. This kills pure "class X leaked statics into class Y" theories — including part of commit 1.
2. **Ruled out affirmatively**: leaked `Mission.Current`/`MissionState.Current` (wrong polarity — a leak *suppresses* the barrier's cleanup; and that path nulls a side, it never removes objects from the client ObjectManager, which is what the logs show); `MapEventCreationCoordinator.Instance` (nine of eleven failures never traverse it); a leaked `AllowedThread` (the server provably sent in the failing blocks); and this PR's own diff (identical clusters pre/post).
3. **The remaining suspects are three silent exits**, one of which is almost certainly the mechanism: (a) `MapEventInitializationBarrier`'s abort path — it destroys the entire battle graph on the server *and every client* with only a generic ObjectManager id-miss as a trace, and the `GauntletMapEventVisual` id-miss line appears exactly once in 10 of 11 failing blocks on shard 5; (b) `GameThread.Run` enqueueing a client-side apply onto a queue nothing pumps (or dropping it on a cancelled ambient token) with no log; (c) a garbage-collected weak broker subscription silently unsubscribing itself — which would neatly explain load-dependence, order-independence, and local irreproducibility at once.

Commit 2 therefore instruments every one of those exits with named log lines and switches the E2E shard step to per-test result logging (the errors-only console made the pass timeline — the one thing every ordering theory needs — unrecoverable). No behavioral changes. The next red run will name the mechanism outright; commit 1 stands on its own as real hazard removal (patch-churn lock hole, stolen game-thread mark, cross-environment queue/static leaks).

ShoT-UPfps (approved):

PR #2740, opened 2026-08-05
[2446] Synchronize battle entry with authoritative participants

### LINKED ISSUE #2446: [FIELD BATTLES] Battle entry desync can leave participants, reserves, or mission state missing
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

Joining a battle appears to load normally, then fails with an error and the mission ends as invalid because the local player origin is missing from battle reserves.

## Steps to reproduce

1. Open a battle encounter and click attack.
2. Join the spawned battle mission.

## Additional information

Observed on LATEST STABLE 1.47. Logs show field-battle initialization and reserve sync before the failure: the battle uses FieldBattleMissionInitializer, opens a coop field battle, and then ends the mission after the local player origin is null.

## Discord source

https://discord.com/channels/709516043332354119/1531715379590729978/1531715379590729978

## Images and videos

[](<https://cdn.discordapp.com/attachments/1531715379590729978/1531715379750240426/image.png?ex=6aa4e35c&is=6aa391dc&hm=77cbd7ec545a2d51b150644c75f0a4f06211063e16a0996742ab53133006480c&>) ([Discord message](https://discord.com/channels/709516043332354119/1531715379590729978/1531715379590729978))

[](<https://cdn.discordapp.com/attachments/1531715379590729978/1531715380198772767/image.png?ex=6aa4e35c&is=6aa391dc&hm=13f1e8f90edbb79dfc4a047441fcd540ca7cd67a329c44b6aba11835a644a284&>) ([Discord message](https://discord.com/channels/709516043332354119/1531715379590729978/1531715379590729978))

[](<https://cdn.discordapp.com/attachments/1531715379590729978/1531715380517671143/image.png?ex=6aa4e35c&is=6aa391dc&hm=24d1c1a112430dfaa45f7dfc9240b10066b17e24ab44fd31f1e565be35696b25&>) ([Discord message](https://discord.com/channels/709516043332354119/1531715379590729978/1531715379590729978))

## Logs

Client log: Not provided

Server log: Not provided

## Additional report details

- The attack flow logs show `MapEvent_Created_36560` resolved to `FieldBattleMissionInitializer`, then opened a coop field battle as player side `Attacker`, with the battle host requesting own reserves at mission entry before requesting the P2P battle instance. ([Discord message](https://discord.com/channels/709516043332354119/1531715379590729978/1531715465305522290))

- For `MapEvent_Created_36560`, coop battle logs show the spawn handler started before reserves were fully synchronized (defender not yet populated), then reserve synchronization updated `Attacker` twice (`1/53` then `2/62`) and `Defender` once (`1/4`) with `receiver party null`; after both sides reported populated, the spawn handler still logged `Local player origin missing from battle reserves` and marked the mission invalid. ([Discord message](https://discord.com/channels/709516043332354119/1531715379590729978/1531715482196115467))

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Joining a field battle can leave a player's party missing from the participant list or reserves. The mission then fails because the local player is missing, or players enter separate or incomplete battles where parties, troops, or the battle result are absent.

## What is the new behavior?

Resolves #2446

A player now enters a battle only after their party has joined it on the server. Only battle participants enter the mission and receive reserves, so players arrive in the same battle with their parties and troops present. Leaving a tournament also clears its mission state before the next battle.

## Bot Changelog Entry

- Fixed battle entry desyncs that could leave players, parties, troops, or reserves missing from battles.


source/GameInterface/Services/MapEvents/Handlers/BattleJoinLeaveHandler.cs
```diff
@@ -131,10 +137,39 @@ private void Handle_PlayerJoinBattleAttempted(MessagePayload<PlayerJoinBattleAtt
         if (!objectManager.TryGetIdWithLogging(data.MapEvent, out var mapEventId)) return;
         if (!objectManager.TryGetIdWithLogging(data.JoiningParty, out var partyId)) return;
 
+        var requestId = Guid.NewGuid().ToString();
```
garrettluskey, line 142: should this have a `ModInformation.IsClient` guard like the other handlers in this file? right now only the client-side prefix publishes `PlayerJoinBattleAttempted`, but if that ever fires on the server we would `SendAll` a join request at ourselves.

source/GameInterface/Services/MapEvents/Handlers/BattleJoinLeaveHandler.cs
```diff
@@ -131,10 +137,39 @@ private void Handle_PlayerJoinBattleAttempted(MessagePayload<PlayerJoinBattleAtt
...
+            if (!pendingJoinRequests.TryGetValue(reply.PartyId, out var pendingRequestId) ||
+                !string.Equals(pendingRequestId, reply.RequestId, StringComparison.Ordinal))
+            {
+                return;
+            }
+
+            pendingJoinRequests.TryRemove(reply.PartyId, out _);
+            if (!reply.Accepted)
+            {
+                Logger.Warning("Server rejected battle join for party {PartyId} and map event {MapEventId}",
```
garrettluskey: after a reject the party is locally out of the battle since the prefix clears `_mapEventSide`, and this just logs. is there anything that recovers the client here, or are we relying on the replicated state catching up (the membership add for `already in a map event`, finalization for a concluded event)?

garrettluskey (dismissed):
The authoritative membership gate is the right fix for #2446, and the pending-request coalescing plus the E2E coverage for outsiders and late joiners look solid. Two smaller things below before this merges.

garrettluskey (approved):
re-reviewed at the new head. both earlier comments are addressed: `Handle_PlayerJoinBattleAttempted` has the `ModInformation.IsServer` guard now, covered by `ServerJoinAttemptEvent_DoesNotSendClientRequest`, and the reject path actually recovers the client, `Handle_NetworkJoinBattleReply` calls `PlayerEncounter.LeaveBattle()` and switches back to `join_encounter`, covered by `RejectedJoinRestoresEncounterAndAllowsRetry` including the retry with a fresh request id.

the rest of the flow holds up on a full pass. the server answers every join request in the `finally`, so the client's `pendingJoinRequests` always clears and a reject never strands the coalescing entry. `joined` now comes from a real `FindMapEventParty` instead of `TryGetRequestingPlayer`, and mission start only goes to parties with a `MapEventParty`, with the new `TryGetValidBattle` membership check on the client matching it, so the optimistic `_mapEventSide` back-reference can no longer open a mission. `EncounterAttackConsequencePatch` sends the main party's `MobileParty` id, which is what `TryGetRequestingParticipant` compares against, and the proto field renumber is fine because the handshake blocks mismatched versions. tournament `OnLeaving` clearing the mission context is wired through the existing `IMissionContext` registration.

PR #2742, opened 2026-08-05
Guard the engine tick against unhandled exceptions

Onlysudden:
## PR Checklist

## PR Type

Three paths let one unhandled exception take down more than the operation that caused it.

**Queued game-thread actions.** `GameThread.Update` drains them inside the engine's tick (`UpdateableList.UpdateAll` ← `OnApplicationTick`), and `RunQueuedTask` has a `try`/`finally` with no `catch`. An action queued through plain `Run` rather than `RunSafe` therefore kills the process and abandons everything queued behind it — on a headless server, the session for everyone. Live example: the save snapshot in `TransferSaveState` guards its flush and broadcasts individually, but `saveInterface.SaveCurrentGame()` itself is unguarded, so a campaign state where saving throws turns one join attempt into a server-wide crash.

**Server receive.** `CoopServer.OnNetworkReceive` is unguarded while `CoopClient`'s copy of the same eight lines is. The two forked, which is exactly why the fix never reached the server. It runs on the poll thread inside `netManager.PollEvents()`, which `Update` calls before flushing coalesced sends and aggregated messages, so an exception abandons the rest of the tick for every player.

**Updateable entries.** `UpdateableList.UpdateAll` calls each entry unguarded, so a faulting entry ends the process and skips the entries behind it.

Related: #2669 reports this class of traffic on a dedicated server and notes the sender cannot be identified from the log. #2713 reports a server that stopped progressing while one error repeated ~15 times a second.

## What is the new behavior?

Related to #2669, #2713

1. Added a `catch` to `GameThread.RunQueuedTask`, because the drain runs inside the engine's tick and an escaping exception costs both the process and the rest of the queue. The action is logged with the label the queue already carries, and the wait handle is still set so a blocking caller resumes instead of waiting out its timeout.
2. Routed the instrumented branch of `GameThread.Update` through `RunQueuedTask` rather than repeating its body. Its cancellation pre-check stays, now with a note that it exists for per-label attribution rather than correctness.
3. Moved the receive dispatch into `CoopNetworkBase.HandleReceivedPayload` and reduced both `OnNetworkReceive` overrides to one line, because the duplication is what let the server go unguarded. The two handler dependencies it needs move to the base constructor with it, which is the churn in `CoopClient`, `CoopServer` and `CoopNetworkBaseTests`. The client gains the peer id in its log lines and the tests now cover both endpoints.
4. Extracted `Poller`'s repeat-suppression into `Common/Util/FaultLogThrottle` and applied it to every guard here that can fire per frame or per packet — the queue drain, `RunSafe`'s wrapper, `UpdateableList` and the receive path. Unthrottled, a permanent fault logs a stack trace on every occurrence, and the log sink drops its middle when full, evicting the first occurrence. A fault is keyed by its source paired with the exception message, so a noisy source neither hides another's fault nor collects its count, and two sources failing in alternation stay throttled. The tracked set is bounded at 256 entries, because a fault message can carry an id. `RunSafe` and `EnqueueSafe` hand the wrapper the resolved caller label, so each call site keys separately and the error line names it. `Poller` now uses the extracted type instead of its own copy.
5. Made `TransferSaveState` disconnect the joiner when the snapshot throws, matching the adjacent `!saveResults.Success` branch. Without it the guard leaves the peer parked in that state with no chunks, no queue and no timeout.
6. Made `IGameStateInterface.LoadSaveData` return whether the campaign loaded, and `ReceivingSavedDataState` end the session through `ICoopFinalizer` when it did not, as the other pre-campaign states do, which is the one added argument in its `ClientLogic` factory. Entering `LoadingState` after a failed load waits on a `CampaignReady` that cannot arrive, and a return to the main menu on its own leaves the peer connected while the server's connection sits in its own `LoadingState`, queueing world updates for a client that will never enter the campaign. The joiner gets the reason in a pop-up, which the forced loading screen would otherwise hide.
7. Made `SteamGameServerBoot.RunCallbacks` respect the shutdown flag as well as `started`, removing the fault at its source: the pump stays registered after teardown and would otherwise dispatch into the unloaded native module on every remaining tick.
8. Corrected the docs the guard invalidated — `Run`'s remarks now scope the suppression to queued work and point at `RunSafe` for the inline path, `RunSafe`'s summary states what is actually still unique to it, `WaitWhilePumping`'s comment no longer claims to guard a failing queued action, and `ValidateModuleState`'s note on why it prefers `RunSafe` gives the reason that still holds, which is log attribution.
9. Let session cancellation through the receive guard instead of logging it, so teardown ends the poll tick through `UpdateWithinSession` — which already catches exactly this — rather than being reported as a packet failure once per queued event while `PollEvents` keeps dispatching into a session being disposed.
10. Routed the "neither `IPacket` nor `IMessage`" branch through the same throttle. An unregistered type id deserializes to `null` and reaches that branch without an exception, at packet rate, so it was the one path that logged in full on every packet.

11. Resolved `UpdateableList`'s logger per call rather than in a static field, as `BootPatches` does for the same reason. `CoopMod.Updateables` is a static property, so the runtime initialises it on entry to `NoHarmonyInit()` — before that method installs the binding redirects in `AssemblyHellscape` — and a static logger field builds the Serilog graph at that moment, killing the process during module load with a `TypeInitializationException`. Found by running the branch in game; neither suite can cover it, since both run without the game's module loader.

Tests: `GameInterface.Tests/Utils/GameThreadGuardTests.cs` alongside the existing `GameThread` tests, reusing their pump and collection (the drain keeps going after a queued action throws, a throwing blocking action releases its caller, the inline path still propagates while `RunSafe` swallows there, and a repeatedly failing action is reported once); `Coop.Tests/Network/CoopNetworkBaseReceiveTests.cs` (throwing handler, throwing deserializer and unregistered type id each leave the tick intact, dispatch is verified, a good payload after a run of failures still lands, repeated unhandled payloads are reported once, and session cancellation passes through while any other cancellation stays guarded); `Common.Tests/UpdateableListTests.cs` (including two entries raising the same fault, which must both be named, and a shape test pinning that the logger is not held in a static field); `Common.Tests/Utils/FaultLogThrottleTests.cs` (per-source counting, alternating sources staying throttled, faults with no exception, and the capacity bound); plus `ReceivingSavedDataStateTests` for the aborted join.

Worth knowing when reviewing: without the `GameThread` guard the test host itself crashes, because `TestGameLoopPump` drives `Update` on a background thread with no guard of its own — the same failure mode as the engine's tick, which is what makes these tests meaningful rather than tautological.

## Bot Changelog Entry

- Fixed the server shutting down when a single background action or an unreadable network packet failed; it now logs the failure and keeps the session running.
- Fixed joining players being left on the loading screen forever when the host's save could not be sent or loaded — the join now ends with an explanation instead of hanging.
- Server logs now name the player whose packet failed to process, and stop repeating the same error every frame.

## Other information

The guards are last-resort and all log at error level, so suppressed failures stay visible. No success path changes.

Verified in game on 1.4.7 against a build of the merge base: the branch hosts a sandbox save, a second instance joins and plays, and the session ends without a crash dump. That run is what surfaced item 11; details are in the thread.

Two deliberate scope notes. A blocking `Run` caller still cannot tell a completed action from a failed one in general; the two places where that mattered (the join snapshot and the client save load) now report their outcome explicitly, and a general mechanism for the remaining ~20 blocking call sites belongs in its own PR. And `LiteNetP2PClient.OnNetworkReceive` holds a third copy of the receive dispatch that is still unguarded — it does not derive from `CoopNetworkBase`, so folding it in is a separate change.


source/Coop.Core/Client/States/ReceivingSavedDataState.cs
```diff
@@ -69,7 +73,14 @@ internal void Handle_NetworkGameSaveDataReceived(MessagePayload<NetworkGameSaveD
             "Loading Host Campaign",
             "Loading host save data...");
 
-        gameStateInterface.LoadSaveData(saveData);
+        if (!gameStateInterface.LoadSaveData(saveData))
+        {
+            // Entering LoadingState now would wait on a CampaignReady that cannot arrive, leaving the
+            // joiner on the loading screen with no error. Fall back to the main menu instead.
+            Logger.Error("Loading the host save failed; aborting the join");
+            Disconnect();
```
ShoT-UPfps: we need to run the normal `ICoopFinalizer` teardown here. `Disconnect()` only switches to `MainMenuState`, so the peer stays connected while the server remains in `LoadingState` and keeps queueing world updates for it

source/Coop.Core/Common/Network/CoopNetworkBase.cs
```diff
@@ -342,6 +354,55 @@ public void SendAllBut(NetPeer excludedPeer, IMessage message)
...
+            {
+                messagePacketHandler.PublishEvent(peer, message);
+            }
+            else
+            {
+                Logger.Error("Received payload deserialized to neither IPacket nor IMessage: {Type} from peer {PeerId}",
+                    received?.GetType(), peer?.Id);
+            }
+        }
+        catch (Exception ex)
```
ShoT-UPfps, line 395: rethrow `OperationCanceledException` here when `sessionCancellation.IsCancellationRequested`. otherwise cancellation wakes a handler blocked in `GameThread.Run`, this catch swallows it, and `PollEvents()` keeps dispatching while `Dispose()` is waiting for the poller

source/Coop.Core/Common/Network/CoopNetworkBase.cs
```diff
@@ -342,6 +354,55 @@ public void SendAllBut(NetPeer excludedPeer, IMessage message)
...
+            {
+                packetManager.HandleReceive(peer, packet);
+            }
+            else if (received is IMessage message)
+            {
+                messagePacketHandler.PublishEvent(peer, message);
+            }
+            else
+            {
+                Logger.Error("Received payload deserialized to neither IPacket nor IMessage: {Type} from peer {PeerId}",
```
ShoT-UPfps: this branch also needs to go through the throttle. an unregistered type id deserializes to `null`, so repeated unknown packets still log once per packet and never reach `receiveFaultThrottle`

source/Common/GameThread.cs
```diff
@@ -462,6 +460,13 @@ private static void RunQueuedTask(
                 task.Act?.Invoke();
             }
         }
+        catch (Exception e)
+        {
+            // Update runs inside the engine's tick, so an escaping exception kills the process and
+            // abandons the rest of the drain. RunSafe's guard, applied to plain Run callers too.
+            Logger.Error(e, "A queued game-thread action threw and was suppressed: {Label}",
```
ShoT-UPfps: this also needs throttling by label and error. a periodic action that keeps failing now writes a full stack every frame or packet, and `UpdateableList` cant throttle it because the exception is consumed here

source/Common/Util/FaultLogThrottle.cs
```diff
@@ -0,0 +1,65 @@
...
+        this.repeatInterval = repeatInterval;
+    }
+
+    /// <summary>
+    /// Records <paramref name="fault"/> and says what to log for it. <paramref name="repeats"/> carries
+    /// how many times it has repeated, for the <see cref="FaultLogAction.Summary"/> case.
+    /// </summary>
+    public FaultLogAction Classify(Exception fault, out long repeats)
+    {
+        string message = fault?.Message;
```
ShoT-UPfps: the throttle needs to include the source in its key. `UpdateableList` shares one instance across updateables and `CoopNetworkBase` shares one across peers, so identical errors hide the second source and the summary can blame the wrong one

ShoT-UPfps (changes requested):

source/Coop.Core/Client/States/ReceivingSavedDataState.cs
Onlysudden: Done in eb3b19e82, the failure path runs `coopFinalizer.Finalize` like the other pre-campaign teardowns. Left `Disconnect()` itself alone — in every client state it reacts to a connection that already dropped rather than ending the session, so the call site was the wrong part. The test pins the `EndCoopMode` publish, the pop-up, and that we no longer enter `LoadingState`.

source/Coop.Core/Common/Network/CoopNetworkBase.cs
Onlysudden: Done in 4789e05be with a filtered catch, so `UpdateWithinSession` ends the tick as it was written to. My guard was also reporting every clean disconnect as a packet failure. Tests cover the pass-through and a cancellation outside teardown still being guarded.

source/Coop.Core/Common/Network/CoopNetworkBase.cs
Onlysudden: Done in f5245d738, keyed by peer plus payload type so two bogus types stay distinct. Worth knowing that routing it through the old throttle as it stood would have dropped the first occurrence entirely, since the key was the exception message and a null exception matched the initial state — the keying in 352965f88 removes that. The test asserts one line for five identical unhandled payloads.

source/Common/GameThread.cs
Onlysudden: Done in 9e3dfcefa, and applied to `WrapSafe` too, since it consumes the fault before the drain sees it and has 647 call sites against 119 for plain `Run`. `RunSafe` and `EnqueueSafe` now hand it the resolved label, so callers that pass no context stop sharing one key and the error line names the caller.

source/Common/Util/FaultLogThrottle.cs
Onlysudden: Done in 352965f88, and a bit further than asked. Adding the source to the single-slot throttle alone would have made it worse: two sources alternating each look new, every call returns Full and the throttling disappears exactly where it is needed. It is now a dictionary keyed by source and message, bounded at 256 entries with a clear on overflow because a message can carry an id. `Poller` keeps its previous behaviour through the single-source overload.

ShoT-UPfps:
How have you tested this?

Onlysudden:
> How have you tested this?

Automated tests only so far, no live session. Here is what that covers and what it leaves open.

Everything below was run on the PR head `8968b8ad4` inside the CI image `garrettluskey/bannerlordcoop:latest`, with the workflow's own commands. `dotnet build CoopTests.slnf -c Release` builds clean. `dotnet test CoopUnitTests.slnf -c Release --no-build --no-restore` gives 1627 passed, 14 skipped, 0 failed. `E2E.Tests` run as one pass instead of the eight shards gives 1213 passed, 4 skipped, 0 failed.

The PR adds 24 tests, each tied to one guarded path. `GameThreadGuardTests` (4) for the queue drain, `CoopNetworkBaseReceiveTests` (8) for the shared receive path, `UpdateableListTests` (4), `FaultLogThrottleTests` (7), and the aborted join in `ReceivingSavedDataStateTests`.

A guard test can pass without pinning anything, so I reverted each fix on its own, rebuilt and reran. What each revert breaks:

- `catch` in `RunQueuedTask` back to a rethrow — the test host process crashes and the run aborts with `Unhandled exception. System.InvalidOperationException: repeated queued failure`, no assertion involved. `TestGameLoopPump` drives `Update` on a background thread with no guard of its own, which is the engine tick's failure mode.
- receive guard back to a rethrow — 4 of the 8 receive tests fail.
- `UpdateableList`'s guard back to a rethrow — 3 of its 4 tests fail.
- `source` dropped from the throttle key — `Classify_TheSameFaultFromTwoSources_IsReportedForEach` and `Classify_PastCapacity_ForgetsWhatItTrackedAndReportsInFullAgain` fail.
- throttle always returning `Full` — 6 tests fail across `Common.Tests`, `GameInterface.Tests` and `Coop.Tests`, the repeat-suppression assertion of every guard that uses it.
- teardown cancellation caught instead of passed through — `HandleReceivedPayload_WhenTeardownCancelsAHandler_LetsTheCancellationOut` fails.
- unhandled-payload branch logging unthrottled — `HandleReceivedPayload_WithRepeatedUnhandledPayloads_ReportsOnlyTheFirst` fails.
- failed host-save load going to the main menu instead of `ICoopFinalizer` — `NetworkGameSaveDataReceived_WhenLoadingFails_TearsDownTheSession` fails.

What this does not cover. There was no live session, so no dedicated server, no second client, and none of these faults watched happening in a real game. The E2E harness marks the test thread as the game thread and feeds payloads through `TestMessageBroker`, so it never reaches the poll-thread receive or the queued drain — those two are covered only by the unit tests above, which drive the real `GameThread.Update` through the existing pump and the real `HandleReceivedPayload` through a minimal `CoopNetworkBase` subclass, with the serializer and handlers mocked because a unit test has no live reader.

ShoT-UPfps:
The scope of this PR requires manual testing to ensure that it's working properly

Onlysudden:
> The scope of this PR requires manual testing to ensure that it's working properly

Ran it in game, and it paid off — the run found a startup crash this PR introduced.

`UpdateableList` held its logger in a static field. `CoopMod.Updateables` is a static property, so the runtime initialises it on entry to `NoHarmonyInit()`, before that method's first statement installs the binding redirects in `AssemblyHellscape`. Building the Serilog graph there throws `FileNotFoundException`, and the `TypeInitializationException` wrapping it kills the process about twelve seconds into module load, before any log file exists. Debug and Release alike. Fixed in c501d61f4 by resolving the logger per call, as `BootPatches` already does for the same reason.

Neither suite can see this. Both run without the game's module loader, so the window where the redirects are still missing does not exist there.

**Setup.** The module was built from this branch and from the merge base `36e5eac6` in the CI image, installed into a 1.4.7 install, hosting a vanilla sandbox save. The host instance was launched exactly as `ServerProcessManager` launches it and driven through the Debug live-test pipe from #2718. Only the six mod assemblies differ between the two packages, the other 44 files are byte-identical.

| Build | Outcome |
|---|---|
| `36e5eac6`, base | boots, Steam game-server logon, hosts the save, reaches `MapState` |
| branch, before the fix | dies after ~12 s, no log written, `0xE0434352` with `0x80131534`, WER points at `Common!LogManager` |
| branch, with the fix | boots, hosts, joiner connects, campaign runs, clean shutdown |

On the fixed build a two-instance session ran end to end — host plus a joiner on `/platformid`, character resolve, two in-game days advanced from the server, traffic both ways, clean exit on both sides with no dump. With `coop.debug.gamethread.instrument on` the instrumented drain branch produced 116 summaries with per-handler attribution; that branch had no coverage before. Unit suite on the fix: 1704 passed, 14 skipped, 0 failed.

**What did not reproduce.** The Steam-pump-after-shutdown fault fired on neither build, neither by quitting the process nor by returning to the main menu, so there is no live evidence for that guard — only its unit test and the mutation run. I would expect the dedicated server to be where it bites, and its release build checks the SHA-256 of the four assemblies this PR touches before opening its socket, so it will not load a branch build. The queue guard and the receive guard are unobservable live by construction: they act only once something has already thrown, and injecting a fault would mean testing a different build.

One note for the inline-path thread from the first round. Joining with a deliberately incomplete module set made `DebugCharacterCreationInterface.SkipCharacterCreationInternal` throw straight out of `GameThread.Run` into `MessageBroker`, which logged it and carried on — the inline path, behaving exactly as the remarks now describe.

garrettluskey:
closing this. there isnt a reproduced issue that requires changing `GameThread.Run`, `UpdateableList`, and the shared receive path to swallow exceptions. `RunSafe` already owns the guarded game-thread path, and changing plain `Run` makes a failed blocking action return as though it succeeded. catching every updateable failure can also leave campaign state partially applied and keep a broken server running.

#2669 only needed peer attribution and log handling and was reported fixed after a hotfix. #2713 doesnt establish these framework guards as the cause. the save-transfer/load failure handling and server receive diagnostics can come back as separate focused PRs tied to reproduced failures.

please dont use AI to expand an issue into broad framework changes without discussing the design first.

Onlysudden:
> closing this. there isnt a reproduced issue that requires changing `GameThread.Run`, `UpdateableList`, and the shared receive path to swallow exceptions. `RunSafe` already owns the guarded game-thread path, and changing plain `Run` makes a failed blocking action return as though it succeeded. catching every updateable failure can also leave campaign state partially applied and keep a broken server running.
> 
> #2669 only needed peer attribution and log handling and was reported fixed after a hotfix. #2713 doesnt establish these framework guards as the cause. the save-transfer/load failure handling and server receive diagnostics can come back as separate focused PRs tied to reproduced failures.
> 
> please dont use AI to expand an issue into broad framework changes without discussing the design first.

Fair on all three points. RunSafe is the guarded path, and plain Run should keep propagating — with the blocking rethrow dropped earlier in review, the queue guard did leave a failed blocking action returning as though it succeeded. The UpdateableList catch had no reproduced failure behind it either. Point taken on discussing the design first.

PR #2747, opened 2026-08-06
Replicate kingdom membership changes to clients

alikhan866:
### Issues

- **#2724** — "Canceling a kingdom leaves players marked as kingdom members". This is the null case above: the server clears `Clan._kingdom`, but AutoSync keys on an object id and cannot express a null reference, so clients kept the clan in the destroyed kingdom and refused to let it join another. The explicit `ClanKingdomMembershipChanged` message added here carries "no kingdom". Worth a retest by the reporter before closing — the report says a restart did not clear it, which this change would not explain on its own.
- **#2559** — "Joining a kingdom as a vassal does not change all colors", specifically *"settlements are not highlighted"*. That half is fixed here: settlement visuals are refreshed on every membership change, and the second commit covers the case where the refresh was skipped because the AutoSync packet had already made the values match. The banner and shield colours the reporter also mentions are **not** addressed by this PR.

### What was broken

`Clan._kingdom` is AutoSynced, but the kingdom's own `_clans` / `_fiefsCache` collections are not reliably intercepted, so clients kept listing a clan in its old realm until a reload. Worse, AutoSync cannot express a null reference at all — its templates key on the object id — so a clan that left, was expelled or rebelled never lost its kingdom on clients.

A clan's fiefs also kept flying the old kingdom's colours. The `_kingdom` AutoSync packet usually lands before the explicit membership message, so by the time that message arrived the value already matched, the equality check returned early, and `RefreshSettlementVisuals` was skipped.

### What it does now

- Postfixes `ChangeKingdomAction.ApplyInternal` on the server so every membership path — join, leave, expel, rebellion, clan destruction — republishes the collections and broadcasts the authoritative kingdom, including "no kingdom".
- Redraws settlement map banners on apply, since their colour comes from the owning clan's kingdom.
- Makes only the *move* conditional on the kingdom differing; the redraw always runs.
- `coop.debug.clan.join_kingdom` and `change_clan_kingdom` republish too. They previously called the vanilla action alone and so reported a result they had not actually replicated.

### Why this approach

Every other membership mutation in the repo republishes the collections explicitly (see `VassalServiceHandler.ApplyVassalage`); this follows the same pattern rather than adding a new sync mechanism for the null case.

---

### The stack

Split out of #2668 so each PR fixes a single problem. Review and merge bottom-up: every PR targets the branch below it, and the bottom targets `development`.

1. **#2747 — Replicate kingdom membership changes to clients** ← this PR
2. #2748 — Restore lord recruitment by persuasion
3. #2749 — Fix barters being refused, double-applied or misreported
4. #2750 — Let clients grant fiefs, and pay the relation for it
5. #2751 — Make retreating from a battle server-authoritative
6. #2752 — Fix breaking into a besieged settlement, and the capture aftermath menu
7. #2753 — Make besieged settlements sortie, and fix which side a relief force joins
8. #2754 — Add debug commands for driving siege and sortie scenarios
9. #2755 — Free a captor holding a captured player, and replicate army joins
10. #2756 — Pull nearby AI parties into a player's battle, and keep Surrender working
11. #2757 — Fix battle deployment wedging and sides spawning nothing
12. #2758 — Stop silently dropping party state, and a 4,200 line/sec log flood
13. #2759 — Split the methods the analyzer flagged as too complex

### DIFF
--- a/source/GameInterface/Services/Clans/Patches/ClanKingdomMembershipReplicationPatch.cs
+++ b/source/GameInterface/Services/Clans/Patches/ClanKingdomMembershipReplicationPatch.cs
@@ -1,0 +1,69 @@
+using Common;
+using Common.Logging;
+using Common.Messaging;
+using GameInterface.Services.Kingdoms;
+using GameInterface.Services.Kingdoms.Messages;
+using HarmonyLib;
+using Serilog;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.Actions;
+
+namespace GameInterface.Services.Clans.Patches;
+
+/// <summary>
+/// Republishes kingdom membership collections after any server-side kingdom change.
+/// </summary>
+/// <remarks>
+/// <see cref="Clan"/>._kingdom is AutoSynced, but the kingdom's own <c>_clans</c>, <c>_fiefsCache</c>,
+/// <c>_townsCache</c> and <c>_settlementsCache</c> are not reliably intercepted. Without this, clients
+/// agree the clan changed sides while the kingdom's roster still lists it in the old kingdom - the
+/// kingdom screen and every Kingdom.Clans consumer stay stale until a reload.
+///
+/// Individual call sites used to fix this one at a time, so each new membership path (expel, leave,
+/// rebellion, clan destruction, a host-initiated defection that never reaches LordBarterHandler) had
+/// to remember to do it, and most did not. Every one of them funnels through ApplyInternal, so the
+/// republish belongs here rather than in each caller.
+///
+/// MoveClanToKingdom and KingdomCollectionSync.AddClan are idempotent, so call sites that already
+/// republish explicitly stay correct.
+/// </remarks>
+[HarmonyPatch(typeof(ChangeKingdomAction), "ApplyInternal")]
+internal class ClanKingdomMembershipReplicationPatch
+{
+    private static readonly ILogger Logger = LogManager.GetLogger<ClanKingdomMembershipReplicationPatch>();
+
+    [HarmonyPrefix]
+    private static void Prefix(Clan clan, out Kingdom __state)
+    {
+        // Captured before the action mutates it; the postfix needs the kingdom being left.
+        __state = clan?.Kingdom;
+    }
+
+    [HarmonyPostfix]
+    private static void Postfix(Clan clan, Kingdom __state)
+    {
+        if (!ModInformation.IsServer || clan == null) return;
+
+        var newKingdom = clan.Kingdom;
+        if (ReferenceEquals(newKingdom, __state)) return;
+
+        if (!ContainerProvider.TryResolve<IKingdomMembershipState>(out var kingdomMembershipState))
+        {
+            Logger.Error("Kingdom membership changed for clan {Clan} before the membership state was available",
+                clan.StringId);
+            return;
+        }
+
+        kingdomMembershipState.MoveClanToKingdom(
+            __state,
+            newKingdom,
+            clan,
+            publishCollectionChanges: true,
+            republishExistingCollections: true);
+
+        // The kingdom's collections are handled above, but Clan._kingdom itself is an AutoSynced
+        // reference field and AutoSync cannot send a null one (it keys on the object id), so a clan
+        // that left, was expelled or rebelled stayed a member on every client. Send it explicitly.
+        MessageBroker.Instance.Publish(clan, new ClanKingdomMembershipChanged(clan, newKingdom));
+    }
+}
--- a/source/GameInterface/Services/Clans/Patches/ClanKingdomMembershipReplicationPatch.cs
+++ b/source/GameInterface/Services/Clans/Patches/ClanKingdomMembershipReplicationPatch.cs
@@ -21,11 +21,14 @@ namespace GameInterface.Services.Clans.Patches;
 ///
 /// Individual call sites used to fix this one at a time, so each new membership path (expel, leave,
 /// rebellion, clan destruction, a host-initiated defection that never reaches LordBarterHandler) had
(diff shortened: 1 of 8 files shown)


source/GameInterface/Services/Clans/Patches/ClanKingdomMembershipReplicationPatch.cs
```diff
@@ -0,0 +1,69 @@
...
+        if (ReferenceEquals(newKingdom, __state)) return;
+
+        if (!ContainerProvider.TryResolve<IKingdomMembershipState>(out var kingdomMembershipState))
+        {
+            Logger.Error("Kingdom membership changed for clan {Clan} before the membership state was available",
+                clan.StringId);
+            return;
+        }
+
+        kingdomMembershipState.MoveClanToKingdom(
```
ShoT-UPfps, line 60: now that this runs for every `ChangeKingdomAction`, can we remove the existing `MoveClanToKingdom(... republishExistingCollections: true)` calls in `LordBarterHandler`, `VassalServiceHandler`, and the debug commands? otherwise those paths send every old/new kingdom collection snapshot twice

ShoT-UPfps (changes requested):
comment

source/GameInterface/Services/Clans/Patches/ClanKingdomMembershipReplicationPatch.cs
alikhan866: Good catch — you're right, and it was worse than just the republish: those call sites were doing the whole `MoveClanToKingdom` a second time.

I checked the assumption in IL rather than trusting the call graph:

- Every public `ChangeKingdomAction` entry point (`ApplyByJoinToKingdom`, `ApplyByJoinToKingdomByDefection`, `ApplyByJoinFactionAsMercenary`, `ApplyByCreateKingdom`, `ApplyByLeaveKingdom`, `ApplyByLeaveKingdomAsMercenary`, `ApplyByLeaveWithRebellionAgainstKingdom`, `ApplyByLeaveByKingdomDestruction`, `ApplyByLeaveKingdomByClanDestruction`) calls `ApplyInternal`, so the patch covers all of them.
- `JoinKingdomAsClanBarterable.Apply` ends in one of those on **every** terminal path, so `LordBarterHandler`'s call was redundant too.

Removed, along with the `IKingdomMembershipState` dependency where nothing else in the type used it:

- `ClanDebugCommands.ChangeClanKingdom`, `.JoinKingdom`, `.LeaveKingdom`
- `LordBarterHandler` (the post-condition check that the clan landed in the kingdom stays — it verifies the patch did its job instead of repeating it)
- `VassalServiceHandler.ApplyVassalage`

Two kept deliberately, because the patch never fires for them — they drive `IKingdomMembershipState` as the *mutator* (`KingdomCollectionSync.AddClan` sets `Clan._kingdom` itself), with no `ChangeKingdomAction` involved:

- `KingdomDebugCommand`
- `PlayerPartyInteractionOutcomeHandler`

`KingdomHandler` is unaffected — it's the client applying a received message with `publishCollectionChanges: false`.

One path worth flagging, in case you know something I don't: promoting a mercenary that's **already in the target kingdom** (`VassalServiceHandler`, `clan.Kingdom == kingdom`). `EndMercenaryServiceAction.EndByBecomingVassal` only calls `Clan.EndMercenaryService`, which sets `IsUnderMercenaryService = false` — no `ChangeKingdomAction`, so the patch doesn't run there. I dropped the republish anyway because the kingdom's rosters don't change on that transition (the clan is already in `Kingdom.Clans`) and `MoveClanToKingdom` never published `IsUnderMercenaryService` in the first place. Happy to put it back if you'd rather keep the belt and braces.

Full suite green on the stack.

ShoT-UPfps (approved):
lgtm

alikhan866:
Closing this as a duplicate of #2607, which fixes the same problem and does it better.

#2607 replicates clan kingdom membership by prefixing `Clan.SetKingdomInternal` itself, which is a strictly more complete choke point than the `ChangeKingdomAction.ApplyInternal` postfix used here: it also catches direct `clan.Kingdom = ...` assignments, including the ones in `KingdomDebugCommand` and `PlayerPartyInteractionOutcomeHandler` that this PR's patch does not cover.

It is also better on the client side. Applying the change through `SetKingdomInternal` lets vanilla do the work — `LeaveKingdomInternal` / `EnterKingdomInternal` maintain both kingdoms' collections, and `UpdateBannerColorsAccordingToKingdom` recolours the banner — and re-raising `CampaignEventDispatcher.OnClanChangedKingdom` lets every view model refresh itself instead of being poked by hand.

That last point matters. While testing this PR with two clients we found that a clan's settlements kept the old realm's colours after it changed kingdom: the data replicated correctly but the map did not repaint. The cause is that a clan's `Banner` is one object mutated in place, so the nameplate cache's content check compares it against itself, and its only other guard is `Banner.GetVersionNo()` — the sum of each `BannerData._localVersion`, a counter bumped solely by the property setters. Replication writes the backing fields, so the counter drifts (measured: server 16, client 12) and the cached banner image is never rebuilt. #2607's approach avoids this entirely, because the client goes through the real code path rather than a replicated field write.

The two changes are also mutually incompatible rather than merely overlapping — #2607 removes `Clan._kingdom` from AutoSync and rewrites `KingdomMembershipState` to assign `clan.Kingdom`, which is the helper this PR builds on. They cannot both land.

Two tests here may be worth carrying over to #2607, which currently has its "tests have been added" box unchecked:

- `LeavingAKingdom_ReplicatesRosterRemovalToClients`
- `JoinKingdomCommand_ReplicatesKingdomRosterToClients`

They cover the roster-removal and "no kingdom" cases, which is where this started. Happy to open them as a small PR against that branch if it would help.

The rest of the stack that sat on top of this branch has been rebased onto `development` and is unaffected.

PR #2748, opened 2026-08-06
Restore lord recruitment by persuasion

alikhan866:
### Issues

**Related to #2613 ("Unable to recruit AI clans") — but that issue is already fixed, and this PR does not claim it.**

#2613 reported that the dialogue option to persuade an AI clan was completely missing, and correctly identified `LordDefectionCampaignBehavior` as the disabled behaviour behind it. That half was fixed independently in `development`: `DisableLordDefectionCampaignBehavior` no longer prefixes `RegisterEvents` with `false` (which had removed `OnSessionLaunched -> AddDialogs` and with it the whole recruit dialogue tree), and instead gates only `OnDailyTick` on `IsServer`. The issue was closed as completed on 5 Aug.

What remains after that fix, and is what this PR adds, is the client half — neither of which has an issue filed:

- **A client that wins the persuasion gains no Charm XP.** The XP writes are blocked client-side and the server never runs the dialogue, so the skill-up is simply lost for anyone who is not the host.
- **A refusal never expires on a client.** Vanilla's gate refuses while a matching unsuccessful attempt is under a week old, and only `OnDailyTick` prunes that list. `development`'s gate suppresses that tick on the client — and the client is the machine that *fills* the list — so nothing prunes it there and a lord who once refused refuses forever. `lordDefectionRetries` makes this explicit and restores singleplayer behaviour by default.

The rest of the PR is hardening on the authoritative path: rejecting participants who died between authorize and request, marshalling `Dispose` onto the game thread, and telling the player the exact shortfall instead of a flat refusal.

### What was broken

Recruiting a lord by persuasion was unusable in co-op for two separate reasons. The first — a missing dialogue option — has since been fixed in `development` and is **not** what this PR addresses: `DisableLordDefectionCampaignBehavior` used to prefix `RegisterEvents` with `false`, which removed `OnSessionLaunched -> AddDialogs` and deleted the whole recruit dialogue tree, and it now gates only `OnDailyTick`.

What this PR fixes is what is left once the option is reachable: a client that wins the persuasion gets nothing for it, and a lord who refuses one refuses forever.

A client who then won the persuasion gained nothing either. The XP writes are blocked client-side and the server never runs the dialogue.

### What it does now

- Removes the blanket patch rather than narrowing it. The daily tick it also suppressed only prunes the behaviour's private persuasion-attempt list and touches no shared state, and the client is the only machine that fills it.
- Awards Charm XP for a client's persuasion: the client ships only the per-attempt outcome enums and the server derives the XP from the same vanilla model the host uses, so no number crosses the wire. Skill and difficulty are constants in vanilla's consequence and are never taken from the client. The count is bounded and a non-defection barter carrying outcomes is rejected outright.
- Flushes the awarded XP with the barter result — both the skill and total XP writes are coalesced and would otherwise not reach the client until the next tick.
- Replicates kingdom membership after the barterables are applied, rejects participants who died between authorize and request, and marshals `Dispose` through the game thread.
- Logs the computed offer value with the kingdom roster counts on a value rejection and tells the player the shortfall. Both sides run the same acceptance test, so a disagreement is always an input disagreement; the client auto-balances the offer to the exact boundary, since a one-denar drift flips accept into reject.
- Guards the barter-less defection branch on the client, which would otherwise apply the kingdom change locally with nothing replicated. It is unreachable on 1.4.7 (its condition compiles to a hardcoded `false`) but a future game patch could enable it.

### Retry behaviour is now a setting

Vanilla keeps two independent rules and only one of them blocks the player: `CanAttemptToPersuade` refuses while a matching unsuccessful attempt is under a **week** old, while the daily prune only drops records over a **year** old. Suppressing the tick made every refusal permanent for the session — not a decision anyone took, just a side effect of disabling the behaviour wholesale.

`lordDefectionRetries` selects between `Vanilla` (default, matching singleplayer), `NeverExpire` (the old co-op behaviour) and `AlwaysRetry`. Both non-default modes patch the **gate**, not the prune: patching the prune changed nothing that mattered, because `NeverExpire` kept the record while the gate stopped blocking after a week anyway, and `AlwaysRetry` cleared the list only on the next daily tick, so "retry immediately" still meant waiting up to a day.

### Note on overlap with development

`development` has since narrowed the same behaviour's `RegisterEvents => false` to `OnDailyTick => IsServer`, which fixes the dialogue-tree half of this independently.

That gate is now **removed**, and this PR's mode-aware prefix is the only one left on `OnDailyTick`. The two do not compose: Harmony skips the original as soon as any prefix returns false, so `IsServer` silently won on clients and the prune never ran there. Because a record leaving the list is the only thing that ever releases the pre-gate, a refusal then lasted the whole session no matter which mode was configured. Suppressing it bought nothing either — `OnDailyTick`'s entire body is a call to `RemoveOldAttempts`, which prunes that machine's own local list and replicates nothing.

---

### Independent

---

### Verification

The behaviour these modes are meant to control is not decided by `CanAttemptToPersuade`. `conversation_lord_from_ruling_clan_on_condition` answers first, refusing on `Any(a => a.PersuadedHero == OneToOneConversationHero)` — a predicate that checks neither the age of the attempt nor whether it succeeded — and returns before the gate is consulted. That was confirmed by disassembling the shipped v1.4.7 assembly, not inferred.

| 1 | `AlwaysRetry` clears every attempt against that lord, including the successful ones | `AlwaysRetry_ClearsEveryAttemptAgainstThisLord_IncludingTheSuccessfulOnes` | Pass — fails against a "remove only failures" implementation |
| 2 | An unrelated lord's refusal is left intact | same test | Pass |
| 3 | `Vanilla` and `NeverExpire` leave the records untouched | `OtherModes_LeaveTheRecordsAlone` | Pass |
| 4 | No records, or no conversation partner, is not an error | `AlwaysRetry_WithNothingRecorded_DoesNotThrow` | Pass |
| 5 | Full suite | `Common`, `Coop`, `Coop.Tests`, `GameInterface.Tests`, `Coop.IntegrationTests`, `E2E.Tests` | 0 failed |

Case 1 pins the non-obvious part. Every persuasion *option* records its own attempt, so a failed persuasion leaves winning options behind; because the pre-gate's predicate ignores success, dropping only the unsuccessful records would let those keep refusing on their own.

Also exercised live on a server with two connected clients, with `lordDefectionRetries` set to `AlwaysRetry`, against a vassal clan leader staying in a town:

| # | Scenario | Result |
|---|----------|--------|
| 1 | Client fails a defection persuasion, then reopens the conversation immediately | Recruitment offered again; the "You have tried to persuade me before" refusal is gone |
| 2 | Same on a second connected client | Same |
| 3 | Defection barter refused on value | Server log names the requesting player's hero and clan, not the host's |

Worth recording for whoever tests this next: only a clan's **leader** can be recruited, and a lord from the kingdom's **ruling clan** is refused by a different vanilla branch — either one produces a refusal that looks like this bug but is not.


source/GameInterface/Services/MobileParties/Patches/LordDefectionRetryPatches.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            switch (ModConfigProvider.ModOptions.LordDefectionRetries)
+            {
+                case LordDefectionRetryMode.NeverExpire:
+                    // Age is deliberately ignored: any surviving refusal keeps blocking.
+                    __result = !HasUnsuccessfulAttempt(__instance, targetHero, reservationType);
+                    return false;
+
+                case LordDefectionRetryMode.AlwaysRetry:
+                    __result = true;
```
ShoT-UPfps, line 60: `AlwaysRetry` doesnt reach this gate. `conversation_lord_from_ruling_clan_on_condition` rejects any surviving failed attempt before `CanAttemptToPersuade` is called, so this still shows the refusal line instead of allowing another attempt.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.DefectionXp.cs
```diff
@@ -0,0 +1,116 @@
...
+            var breakdown = string.Join(", ", barter.GetOfferedBarterables()
+                .Select(b => $"{b.GetType().Name}x{b.CurrentAmount}={b.GetValueForFaction(targetHero.Clan)}"));
+
+            Logger.Warning(
+                "Lord barter refused on value. offerValue={OfferValue} threshold=-0.01 offered=[{Breakdown}] " +
+                "mainHero={MainHero} playerClan={PlayerClan} targetClan={TargetClan} " +
+                "targetKingdom={Kingdom} kingdomClans={ClanCount} kingdomFiefs={FiefCount}",
+                offerValue,
+                breakdown,
+                Hero.MainHero?.StringId,
```
ShoT-UPfps: use `playerHero` and `playerHero.Clan` here. this runs on the dedicated server, so `Hero.MainHero` and `Clan.PlayerClan` are null.

source/GameInterface/Services/MobileParties/Patches/LordDefectionRetryPatches.cs
```diff
@@ -0,0 +1,92 @@
...
+        nameof(LordDefectionCampaignBehavior.OnDailyTick))]
+    internal class OnDailyTickPatch
+    {
+        [HarmonyPrefix]
+        private static bool Prefix()
+        {
+            // Suppressed so nothing a refusal recorded is ever forgotten; the gate above reads it forever.
+            // AlwaysRetry lets the prune run normally - its gate ignores the list, and leaving vanilla to
+            // tidy up keeps the list from growing for the whole session.
+            return ModConfigProvider.ModOptions.LordDefectionRetries != LordDefectionRetryMode.NeverExpire;
```
ShoT-UPfps, line 146: this still doesnt restore the default mode on clients. `DisableLordDefectionCampaignBehavior.cs:11` also returns false from an `OnDailyTick` prefix, so `RemoveOldAttempts` never runs and a refusal still lasts for the session.

ShoT-UPfps (changes requested):
cmts

EgardA:
On phone so can't see too good but does this take into account that lord recruitments already are fixed and this is just improvements upon that?

source/GameInterface/Services/MobileParties/Patches/LordDefectionRetryPatches.cs
alikhan866: Fixed in ed64d6ee5. You were right, and the reason is worse than the gate simply being bypassed: I disassembled `conversation_lord_from_ruling_clan_on_condition` on the shipped v1.4.7 assembly, and its refusal predicate is `Any(a => a.PersuadedHero == OneToOneConversationHero)` — it checks neither the age of the attempt nor whether it succeeded. So a single prior attempt refuses forever, and it returns before `CanAttemptToPersuade` is reached. `AlwaysRetry` did nothing at all.

It now clears that lord's attempt records in a prefix on that condition, so the pre-gate finds nothing to refuse on. Clearing the records rather than forcing the result is deliberate: the method also rebuilds `_allReservations` and answers several unrelated branches, so it has to run.

One subtlety the fix has to respect — it drops ALL of that lord's attempts, not just the unsuccessful ones. Every persuasion OPTION records its own attempt, so a failed persuasion leaves winning options behind, and since the predicate ignores success those would keep refusing on their own. `AlwaysRetry_ClearsEveryAttemptAgainstThisLord_IncludingTheSuccessfulOnes` fails against the narrower "remove only failures" version.

Verified live with two connected clients: fail a defection persuasion, reopen the conversation immediately, and the recruitment option is offered again.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.DefectionXp.cs
alikhan866: Fixed in ed64d6ee5 — now logs `playerHero?.StringId` and `playerHero?.Clan?.StringId`.

Worth noting it was wrong on a listen host too, not only a dedicated one. There `Hero.MainHero` and `Clan.PlayerClan` are not null, they are the HOST's own hero and clan — so the log would have quietly named the wrong player rather than obviously logging nothing. Confirmed live that the refusal breakdown now names the requesting player.

source/GameInterface/Services/MobileParties/Patches/LordDefectionRetryPatches.cs
alikhan866: Fixed in ed64d6ee5 by removing `DisableLordDefectionCampaignBehavior` entirely, so the mode-aware prefix is the only one left on `OnDailyTick`.

The two were never composing. Harmony skips the original as soon as ANY prefix returns false, so `ModInformation.IsServer` silently won on clients and the prune never ran there. Combined with the pre-gate above — whose predicate ignores age, making a record leaving the list the only thing that ever releases it — that is exactly the session-long refusal you describe, and it applied whatever mode was configured.

Suppressing it bought nothing: `OnDailyTick`'s entire body is a single call to `RemoveOldAttempts` (checked in IL), which prunes that machine's own local list and replicates nothing. I have corrected the PR description too, which claimed the gate was kept and that the two composed.

Removing a pre-existing file does widen the diff beyond this PR's original file set — happy to neuter it in place instead if you would rather keep the Disable* family intact.

ShoT-UPfps (approved):
lgtm

PR #2749, opened 2026-08-06
Fix barters being refused, double-applied or misreported

alikhan866:
Four ways an authoritative barter could be refused, double-applied, or misreported.

### One player at a time on a conversation target

Two players both at war with the same lord could each open a conversation with him and each run the persuasion, then both pay for a barter and see him defect twice. The hostility carve-out that allowed this exists so simultaneous attackers converge on one `MapEvent`, which is right for combat and wrong for diplomacy: every server-side conversation outcome is authorised only as "this peer holds an engagement with this party", so a shared hold lets two players apply the same one-shot result.

A party is now held by exactly one player. Attackers still converge — once the holder starts the battle, the contender's retry joins that `MapEvent` through the existing attacker/defender branch.

### Barters started from a settlement menu

Talking to a lord from a castle or town menu opens a conversation with no agent interaction and no `CampaignMission.Current.Location`, so neither the map-party hold nor the location lock was ever acquired and every barter from such a conversation was refused with "the conversation is no longer active" — after the player had already played out the persuasion.

Adds a `Settlement` conversation context. There is nothing to lock, so authority comes from co-location instead: the server checks that both the requesting party and the target are inside the settlement named by the request, which a player who is not there cannot fake. The context is tried last, so map and location conversations keep their stronger hold. An unrecognised context is refused rather than validated as a settlement.

Applies to the lord defection and peace barters, which shared the gap.

### An applied barter reported as rejected

Both the marriage and peace handlers applied their mutation and could then still fall into the `catch` and send a rejection. The client rolled its UI back while the server had already married the pair or ended the war, leaving the two sides disagreeing about a change that really happened — and for marriage the gold had moved too.

They now track whether the mutation was committed and, once it has been, report success even if a later step throws. The authoritative state stands; claiming failure is the desync, not the fix. `LordBarterHandler` already did this and is unchanged.

### Dispose no longer blocks container teardown

`LordBarterHandler.Dispose` marshalled its cleanup to the game thread and waited. During teardown the loop may already have stopped pumping, so the wait ran its full 30s timeout and failed teardown, taking three conversation tests with it. The queued clear is discardable — the handler is being thrown away anyway — so it no longer waits.

---

### Depends on #2748

This PR targets `coop/lord-defection` rather than `development`, because its `Dispose` fix flips `blocking: true` to `false` on the `GameThread.RunSafe` marshalling that #2748 introduces — on plain `development` that call does not exist yet. Merge #2748 first.

The rest of the work was split out of #2668 into PRs that each fix a single problem; the others target `development` directly.

---

### Verification

Each of the three behavioural fixes was proven by removing the fix and confirming its test goes red, then restoring it and confirming green again — not merely asserted.

| 1 | Settlement context resolves when the lord leads an **active** party | `SettlementConversation_WithLordLeadingAnActiveParty_StillResolvesSettlementContext` | Pass — fails without the fix |
| 2 | Map-party context still resolves when neither side is in a settlement | `MapConversation_WithNeitherSideInASettlement_StillResolvesMapPartyContext` | Pass |
| 3 | Settlement barter accepted end to end | `GenericLordBarter_StationarySettlementConversation_ValidatesCurrentPresence` | Pass |
| 4 | A second player is refused the same lord, and the holder is unaffected | `SettlementConversation_WhenAnotherPlayerHoldsTheLord_RefusesTheSecondAuthorization` | Pass — fails without the fix |
| 5 | Cancelling releases the lord for the next player | `SettlementConversation_AfterTheHolderCancels_TheLordIsAvailableAgain` | Pass |
| 6 | Marriage reports success when the spouse links do not take after apply | `PersonalMarriageBarter_WhenSpouseLinksDoNotTakeAfterApply_StillReportsSuccess` | Pass — fails without the fix |
| 7 | Full suite | `Common`, `Coop`, `Coop.Tests`, `GameInterface.Tests`, `Coop.IntegrationTests`, `E2E.Tests` | 2742 passed, 0 failed, 18 skipped |

The existing settlement test asserted `Assert.Null(targetHero.PartyBelongedTo)`, so the map-party branch could never be reached and the misclassification stayed invisible. Case 1 gives the lord a real party, which is the in-game situation.

Also exercised live on a server with two connected clients, with a lord staying inside a town while leading an active 242-man party:

| # | Scenario | Result |
|---|----------|--------|
| 1 | Player barters with a lord from the town menu while that lord leads an active party | Accepted, gold transferred |
| 2 | Second player opens a barter with the same lord while the first player's barter screen is still open | Refused; the first player's barter completed unaffected, gold charged once |
| 3 | Both players barter with a lord on the campaign map, neither side in a settlement | Both accepted — the reordering leaves map conversations intact |

One thing this does not change: the reservation refuses the second player server-side without notifying them, so their barter screen still opens and is only refused on accept. Worth a follow-up, but it is a presentation issue rather than a correctness one.

### DIFF
--- a/source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
+++ b/source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
@@ -458,8 +458,23 @@ internal sealed partial class LordBarterHandler : IHandler
         playerParty = mobileParty.Party;
         targetParty = targetHero.PartyBelongedTo?.Party;
 
-        var context = (PeaceConversationContext)request.Context;
-        if (context == PeaceConversationContext.MapParty)
+        var conversationContext = (PeaceConversationContext)request.Context;
+        if (conversationContext == PeaceConversationContext.Settlement)
+        {
+            // A settlement-menu conversation acquires no engagement - there is no agent and no
+            // location mission to lock - so authority comes from co-location instead: both the
+            // requesting party and the target must actually be inside the settlement named by the
+            // request. That is as strong as the hold for this case, because a player who is not in
+            // the settlement cannot be talking to someone who is.
+            if (!objectManager.TryGetObject(request.ContextId, out Settlement conversationSettlement) ||
+                mobileParty.CurrentSettlement != conversationSettlement ||
+                targetHero.CurrentSettlement != conversationSettlement)
+            {
+                reason = "The lord settlement conversation is no longer active.";
+                return false;
+            }
+        }
+        else if (conversationContext == PeaceConversationContext.MapParty)
         {
             if (!objectManager.TryGetObject(request.ContextId, out PartyBase requestedParty) ||
                 requestedParty != targetParty ||
@@ -475,7 +490,7 @@ internal sealed partial class LordBarterHandler : IHandler
                 return false;
             }
         }
-        else if (context == PeaceConversationContext.Location)
+        else if (conversationContext == PeaceConversationContext.Location)
         {
             if (targetHero.CharacterObject == null ||
                 !objectManager.TryGetId(targetHero.CharacterObject, out var characterId) ||
@@ -488,13 +503,10 @@ internal sealed partial class LordBarterHandler : IHandler
         }
         else
         {
-            if (!objectManager.TryGetObject(request.ContextId, out Settlement settlement) ||
-                mobileParty.CurrentSettlement != settlement ||
-                targetHero.CurrentSettlement != settlement)
-            {
-                reason = "The lord settlement conversation is no longer active.";
-                return false;
-            }
+            // Refused rather than validated as a settlement conversation - accepting a context we do
+            // not understand is how an unvalidated barter gets through.
+            reason = "The lord conversation context is not supported.";
+            return false;
         }
 
         if (targetHero.IsPrisoner || targetHero.Clan == null)
--- a/source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
+++ b/source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
@@ -89,14 +89,16 @@ internal sealed partial class LordBarterHandler : IHandler
         messageBroker.Unsubscribe<PlayerDisconnected>(HandlePlayerDisconnected);
 
         // Every other access to these dictionaries happens on the game thread (handlers are drained
-        // there), so clearing them from the disposing thread is a data race. Marshal, as
-        // ConversationPartyTracker.Dispose does.
+        // there), so clearing them from the disposing thread would be a data race - marshal instead.
+        // NOT blocking: Dispose runs during container teardown, when the game loop may already have
+        // stopped pumping, and a blocking wait there just times out after 30s and fails the teardown.
+        // If the queued clear never runs, the handler is being discarded anyway.
         GameThread.RunSafe(() =>
         {
             authorizations.Clear();
             completedResults.Clear();
         },
-            blocking: true,
+            blocking: false,
             context: nameof(LordBarterHandler));
 
         LordBarterPatch.ClearPendingRequest();
--- a/source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
+++ b/source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
@@ -372,6 +372,19 @@ internal sealed partial class LordBarterHandler : IHandler
             return;
         }
 
+        // A map-party conversation is already exclusive: ConversationPartyTracker holds the target for
+        // exactly one peer. A settlement-menu conversation acquires no such hold - authority there is
+        // co-location - and co-location is NOT exclusive: every player standing in the settlement
+        // satisfies it, so without this two kingdom leaders could each authorize, each pay, and each
+        // move the same clan in turn. Reserve the target hero for one peer at a time.
+        if (IsTargetHeldByAnotherPeer(peer, authorization.TargetHeroId))
+        {
+            Logger.Warning(
+                "Rejected lord barter authorization for {TargetHeroId}: another player is already negotiating with that lord",
+                authorization.TargetHeroId);
+            return;
+        }
+
         authorizations[peer] = new LordBarterAuthorization(
             authorization.RequestId,
             authorization.TargetHeroId,
@@ -381,6 +394,26 @@ internal sealed partial class LordBarterHandler : IHandler
             authorization.TargetKingdomId,
             DateTime.UtcNow.Add(AuthorizationLifetime));
         completedResults.Remove(peer);
+    }
+
+    /// <summary>
+    /// Whether a DIFFERENT peer already holds a live authorization against this lord. Expired entries
+    /// do not reserve anything, so a player who walked away cannot block the lord for the rest of the
+    /// session - the authorization lifetime is what releases it.
+    /// </summary>
+    private bool IsTargetHeldByAnotherPeer(NetPeer peer, string targetHeroId)
+    {
+        if (string.IsNullOrEmpty(targetHeroId)) return false;
+
+        var now = DateTime.UtcNow;
+        foreach (var entry in authorizations)
+        {
+            if (entry.Key == peer) continue;
+            if (entry.Value.ExpiresAtUtc <= now) continue;
+            if (entry.Value.TargetHeroId == targetHeroId) return true;
+        }
+
+        return false;
     }
 
     private bool TryGetAuthorization(
@@ -739,7 +772,7 @@ internal sealed partial class LordBarterHandler : IHandler
     private sealed class LordBarterAuthorization
     {
         public string RequestId { get; }
-        private string TargetHeroId { get; }
+        public string TargetHeroId { get; }
         private int Context { get; }
         private string ContextId { get; }
         private int Kind { get; }
--- a/source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
+++ b/source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
@@ -205,25 +205,9 @@ internal sealed partial class LordBarterHandler : IHandler
             var previousTargetKingdom = kind == LordBarterKind.JoinKingdomAsClan
                 ? targetHero.Clan.Kingdom
                 : null;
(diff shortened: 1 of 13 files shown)


source/GameInterface/Services/Barters/Patches/LordBarterPatch.cs
```diff
@@ -336,19 +336,25 @@ internal static bool TryGetConversationContext(BarterData barterData, IObjectMan
             return true;
         }
 
-        var settlement = barterData.OffererParty?.MobileParty?.CurrentSettlement;
-        if (settlement != null && barterData.OtherHero?.CurrentSettlement == settlement &&
-            manager.TryGetId(settlement, out contextId))
+        if (barterData.OtherParty?.MobileParty?.IsActive == true && manager.TryGetId(barterData.OtherParty, out contextId))
```
ShoT-UPfps: a lord party stays active while it is inside a settlement, so this classifies settlement-menu barters as `MapParty` and the server rejects them because there is no conversation hold. keep the settlement check first.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
```diff
@@ -458,8 +460,23 @@ private bool TryResolveContext(NetPeer peer, NetworkRequestLordBarter request, o
         playerParty = mobileParty.Party;
         targetParty = targetHero.PartyBelongedTo?.Party;
 
-        var context = (PeaceConversationContext)request.Context;
-        if (context == PeaceConversationContext.MapParty)
+        var conversationContext = (PeaceConversationContext)request.Context;
+        if (conversationContext == PeaceConversationContext.Settlement)
```
ShoT-UPfps: co-location still lets every player in this settlement authorize the same lord at once. we need to reserve the target hero per peer here too, otherwise two kingdom leaders can both pay for and move the same clan in sequence.

source/GameInterface/Services/Barters/Handlers/MarriageBarterHandler.cs
```diff
@@ -211,6 +211,10 @@ private void ProcessRequest(object sender, NetworkRequestMarriageBarter request)
             var offeredBarterables = barterData.GetOfferedBarterables();
             foreach (var barterable in offeredBarterables)
                 barterable.Apply();
+            // Past this point the marriage and any gold have already moved on the authoritative
```
ShoT-UPfps: this still calls `Reject` at `MarriageBarterHandler.cs:223` after every barterable has been applied. if the spouse links are missing, the client rolls back even though gold may already have moved.

source/GameInterface/Services/MapEvents/ConversationPartyTracker.cs
```diff
@@ -158,8 +158,12 @@ public bool TryBeginEngagement(
...
             }
 
-            var existing = engagements.Values.FirstOrDefault(x => x.PartyId == partyId);
-            if (existing.PartyId != null) wasAiDisabled = existing.WasAiDisabled;
+            // One engager per party. Every server-side consequence of a map conversation is validated
+            // only as "this peer holds an engagement with this party" (LordBarterHandler,
+            // PeaceBarterHandler, MarriageBarterHandler, BanditBarterHandler), so a shared hold lets
+            // two players each apply the same one-shot outcome - e.g. two players persuading the same
+            // lord, both paying, and the lord defecting twice.
+            if (engagements.Values.Any(x => x.PartyId == partyId)) return false;
```
ShoT-UPfps, line 168: please update the shared-engagement docs with this change, `ConversationPartyTracker`, `ConversationPartyHold`, and `ConversationRequestHandler` still say hostile contenders share a target.

ShoT-UPfps (changes requested):
cmts

EgardA:
On phone so can't see too good but does this take into account that barters already are fixed and this is just improvements upon that?

source/GameInterface/Services/Barters/Patches/LordBarterPatch.cs
alikhan866: Fixed in 94a6a304c. The settlement branch now runs before the map-party branch in both `LordBarterPatch` and `PeaceBarterPatch`.

Confirmed the premise in IL first: `MobileParty.IsActive` is a plain backing field, cleared only by `RemoveParty`, `PlayerCaptivity` and `PreAfterLoad` — entering a settlement never touches it. So the map-party branch matched first for every settlement-menu barter, the server then demanded a hold that a settlement menu never acquires, and the barter was refused.

Reordering cannot steal an ordinary map conversation, because the settlement branch requires both sides to be inside the SAME settlement and a party talking on the map has no `CurrentSettlement`.

Worth noting why this got past the tests: `GenericLordBarter_StationarySettlementConversation_ValidatesCurrentPresence` asserts `Assert.Null(targetHero.PartyBelongedTo)`, so the map-party branch was unreachable there. `SettlementConversation_WithLordLeadingAnActiveParty_StillResolvesSettlementContext` gives the lord a real party and fails without the fix. Also verified live with two clients against a lord leading an active 242-man party inside a town — accepted, gold transferred.

source/GameInterface/Services/Barters/Handlers/LordBarterHandler.cs
alikhan866: Added in 94a6a304c. `ProcessAuthorization` now refuses to register an authorization when a different peer already holds a live one against the same lord.

The asymmetry you point at is the reason: a map-party conversation is already exclusive through `ConversationPartyTracker`, but a settlement conversation acquires no hold at all — its authority is co-location, and co-location is satisfied by every player standing in the settlement. Expiry still releases the reservation, so a player who walks away cannot pin a lord for the session.

Covered by `SettlementConversation_WhenAnotherPlayerHoldsTheLord_RefusesTheSecondAuthorization` (fails without the fix, and asserts the holder is unaffected and gold moves once) and `SettlementConversation_AfterTheHolderCancels_TheLordIsAvailableAgain`. Verified live: with one player holding the barter screen open, the second was refused and the first completed normally.

One limitation to flag: the refusal is silent, so the second player's barter screen still opens and is only refused on accept. Presentation rather than correctness, so I have left it for a follow-up unless you would rather it went in here.

source/GameInterface/Services/Barters/Handlers/MarriageBarterHandler.cs
alikhan866: Fixed in 94a6a304c. Once `mutationApplied` is set, the handler no longer calls `Reject` — a failure after that point is logged and success is still reported.

You were right that the earlier change did not cover this path: the spouse links are established by `MarriageBarterable.Apply` before the verification runs, so rejecting afterwards told the client to roll back a marriage the server had already performed, with the gold gone. The authoritative state stands; claiming failure is the desync.

`PersonalMarriageBarter_WhenSpouseLinksDoNotTakeAfterApply_StillReportsSuccess` suppresses the apply so the links do not take, and fails without the fix.

source/GameInterface/Services/MapEvents/ConversationPartyTracker.cs
alikhan866: Updated in 94a6a304c. The `ConversationPartyHold` comment claimed that every contender registered for a shared hostile encounter may interact with the target, which no longer describes the code — the hold is exclusive to a single holder. It now says so.

ShoT-UPfps (commented):
cmt

source/GameInterface/Services/MapEvents/ConversationPartyTracker.cs
ShoT-UPfps: this is still missing in `ConversationPartyTracker.cs:15`, `ConversationPartyTracker.cs:131`, `ConversationPartyHold.cs:53`, and `ConversationRequestHandler.cs:35`. they all still say hostile players can share a target or hold.

alikhan866: Fixed in 33556cbd0 — all four, plus a sweep for any I would otherwise have missed again.

I checked the behaviour rather than assuming, and you are right: `TryBeginEngagement` refuses outright when the target is already held.

```csharp
// One engager per party. ...
if (engagements.Values.Any(x => x.PartyId == partyId)) return false;
```

So a hold belongs to exactly one player, hostile or not, and every one of those comments described behaviour the guard makes unreachable. Last round I corrected only `ConversationPartyHold`'s inline comment and left the XML docs contradicting it, which is worse than not having touched it at all.

Corrected: the tracker's class remarks and its `TryBeginEngagement` summary, `ConversationPartyHold.TryEngage`, and the `ConversationRequestHandler` remarks. I also grepped the rest of `GameInterface` and `Coop.Core` for other share/hold claims instead of only the four listed — the remaining hits either describe the old behaviour deliberately, as the reason for the change, or are unrelated.

Comments only, no behaviour change.

One thing I left alone, in case you would rather it went too: `TryEndEngagement` still computes `shouldReleaseParty = !engagements.Values.Any(x => x.PartyId == endedPartyId)`, which is a leftover from when sharing existed — with the guard in place it is always true. Correct but dead, and outside what this thread asked for.

alikhan866: Following up on my own reply — 33556cbd0 was incomplete in two ways, both fixed in d2a5f1a33.

**A fifth stale doc I missed.** `ConversationPartyHold.IsInteractionBlocked` still read "not one of its registered **contenders**" — plural, the same shared-hold claim. My sweep grepped for "shar*", so a sentence phrased with "contenders" went straight through it. Now "not the single player holding it".

**I deleted a fact that is still true.** The old remarks tied MapEvent convergence to hostile players sharing the hold. Removing the sharing claim took the convergence with it — but attackers do still converge, by a different mechanism: once the holder starts the battle, the map-event branch in that same file approves the contender's next request so it joins that MapEvent. `ConversationPartyHold` explicitly points readers at `ConversationRequestHandler` for this, so deleting it left a dangling cross-reference. Restored with the correct mechanism rather than the old one.

I re-swept with a much wider net this time (contenders / every player / each player / both players / multiple players / engagers, across `GameInterface` and `Coop.Core`) instead of just "shar*". The only other candidate was `LocationConversationTracker`'s "Unlike a map-party hold there is nothing to freeze", which is about freezing the party rather than sharing it, so it is accurate and left alone.

Comments only, no behaviour change.

ShoT-UPfps (approved):
lgtm

PR #2750, opened 2026-08-06
Let clients grant fiefs, and pay the relation for it

### LINKED ISSUE #2790: [KINGDOMS] Give Settlement does nothing on a client, and gifting a fief awards no relation
## Severity of issue/bug

- [ ] High
- [x] Average
- [ ] Low

Two separate problems with granting a fief from the kingdom screen.

**1. On a client, "Give Settlement" does nothing.** The popup closes and ownership does not change. Nothing is logged and no error is shown, so it looks like the click was simply ignored. It is the natural thing to do right after recruiting a lord, so it is easy to hit.

The local call is blocked so the client cannot move a settlement on its own, but no request was forwarded to the server in its place, leaving the action a silent no-op.

**2. Gifting a fief awards no relation, on host or client.** In singleplayer a player gift goes through `KingdomManager.GiftSettlementOwnership`, which applies a relation change between giver and receiver before transferring ownership. Co-op calls `ChangeOwnerOfSettlementAction.ApplyByGift` directly, which transfers ownership but skips the relation. Gifting a fief is therefore worth nothing diplomatically, which is most of the point of doing it.

## Steps to reproduce

For (1), as a **client**:
1. Rule a kingdom that owns at least one settlement, with another clan as a vassal.
2. Open the kingdom screen -> Fiefs -> select a settlement -> **Give Settlement**.
3. Pick a recipient in your kingdom and confirm.
4. The popup closes and ownership is unchanged. Reopening the screen shows the original owner.

For (2), on **either** host or client:
1. Gift a settlement to a lord in your kingdom (on the host, where the transfer does work).
2. Check your relation with the recipient before and after - it is unchanged.

## Additional information

Both are addressed by #2750, which forwards the gift to the server, re-derives authority from the requesting peer rather than trusting the client, and applies the relation bonus on the authoritative path so it matches singleplayer.

**Not the same as #2612 / #2606**, which look similar but are a different route. Those go through the companion dialogue (`Clan.CreateCompanionToLordClan`), which has its own interception and is not touched by #2750. `ChangeOwnerOfSettlementAction.ApplyByGift` has exactly two callers in the game - `KingdomManager.GiftSettlementOwnership` (this issue) and `Clan.CreateCompanionToLordClan` (those issues) - so the two should be fixed and closed independently.

alikhan866:
Fixes #2790

### What was broken

The kingdom screen's **Give Settlement** was a silent no-op on a client: `ChangeOwnerOfSettlementPatch` blocked the local call and no request was ever sent, so the popup closed and nothing happened. It is the natural action right after recruiting a lord.

Gifting also awarded no relation. Vanilla routes a player gift through `KingdomManager.GiftSettlementOwnership`, which applies a relation change between giver and receiver before transferring ownership; co-op called `ChangeOwnerOfSettlementAction.ApplyByGift` directly, which does the transfer but not the relation. Gifting a fief in co-op was worth nothing diplomatically.

### What it does now

- The client forwards the gift and the server re-derives authority from the requesting peer. The request carries only two ids; the server checks that the requester either owns the settlement or rules its kingdom, and that the recipient is in that same kingdom.
- The relation bonus is applied on the authoritative path, matching singleplayer.
- The authority check is re-run on the game thread immediately before the transfer. `CanGift` ran on the network thread, but the transfer runs whenever the game thread reaches the queued action, and ownership or kingdom membership can change in between — the fief captured or sold, the clan out of the realm — and the request would still have transferred a settlement the requester could no longer give away. The original check stays as a cheap pre-filter so an invalid request never reaches the game thread; the one inside the action is what actually guards the transfer.

---

### Independent

---

### Review feedback

- **The lookups and `CanGift` now run on the game thread**, together with the apply. The separate pre-filter is gone rather than duplicated: resolving ids and testing authority reads campaign state (`OwnerClan`, `Kingdom`, `IsAlive`), and doing that on the network thread races the game loop and judges the request against a world a queued action may be about to change.

- **Authority is ownership, and only ownership.** Not `isOwner || isRuler`, and not `isOwner && isRuler`. `KingdomSettlementVM.ExecuteAnnex` opens the gift popup only when `settlement.OwnerClan.Leader == Hero.MainHero` and offers **annex** to everyone else, and `_onGrantFief` is invoked nowhere else — so a ruler cannot *gift* a vassal's fief (that is an annex), while a vassal *can* gift their own. `KingdomManager.GiftSettlementOwnership` carries no checks of its own. Requiring ownership also closes the replay hole: once the gift lands the giver is no longer the owner, so a duplicate request is refused instead of moving the fief twice.

- **A refused gift now reports back** to the requester via `NetworkSettlementGiftRejected`, surfaced client-side. The kingdom screen closes its popup on confirm, so a refusal that only reached the server log looked exactly like the silent no-op this PR set out to fix.

### Verification

Each behavioural fix was checked by reverting it and confirming its test goes red, then restoring it.

| 1 | The client's gift is forwarded rather than applied locally | `ClientGift_IsForwardedToTheServer_RatherThanAppliedLocally` | Pass |
| 2 | Ownership transfers and the relation bonus is paid | `ServerGift_FromTheOwner_TransfersOwnershipAndPaysTheRelationBonus` | Pass — fails without the relation fix |
| 3 | A requester who does not own the fief is refused, and told | `ServerGift_FromSomeoneWhoDoesNotOwnIt_IsRefusedAndTheRequesterIsTold` | Pass — fails under `isOwner \|\| isRuler` |
| 4 | A repeated request does not move the fief a second time | `ServerGift_RepeatedAfterTheFiefMoved_DoesNotTransferItASecondTime` | Pass — fails under `isOwner \|\| isRuler` |
| 5 | Full suite | `Common`, `Coop`, `Coop.Tests`, `GameInterface.Tests`, `Coop.IntegrationTests`, `E2E.Tests` | 2745 passed, 0 failed |

Also exercised live on a server with two connected clients, both in one kingdom, the ruler holding two towns:

| # | Scenario | Result |
|---|----------|--------|
| 1 | Client gifts a town it owns to the other player's clan from the kingdom screen | Ownership transfers, and the change is present on the server and both clients |
| 2 | Relation between giver and recipient after that gift | Rose from a reset baseline of 0, identically on the server and both clients — the bonus is applied server-side and replicates |
| 3 | The same client tries to gift that town again, now that it belongs to the other player | Not offered — the screen presents annex instead, matching vanilla |
| 4 | The client gifts its second town | Transfers normally |

The refusal message has no clean manual trigger, because the screen will not offer a gift the player is not entitled to make; it is reached by a stale or duplicate request, which case 4 of the automated table covers.


source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
```diff
@@ -20,21 +22,159 @@ public class SettlementOwnershipHandler : IHandler
...
+        /// request carries only two ids and is never trusted for who may give what away.
+        /// </summary>
+        private void Handle(MessagePayload<NetworkRequestSettlementOwnership> obj)
+        {
+            if (!ModInformation.IsServer) return;
+            if (!(obj.Who is NetPeer peer)) return;
+
+            var payload = obj.What;
+            if (!playerManager.TryGetPlayer(peer, out var player) ||
+                !objectManager.TryGetObject(player.HeroId, out Hero requestingHero))
```
ShoT-UPfps, line 89: these lookups and `CanGift` need to be inside `GameThread.RunSafe`, otherwise they read Campaign state on the network thread and can reject before an earlier game-thread apply runs.

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
```diff
@@ -20,21 +22,159 @@ public class SettlementOwnershipHandler : IHandler
...
+            {
+                reason = "the settlement has no owner clan";
+                return false;
+            }
+
+            // Either the owning clan's leader gives away their own fief, or the kingdom's ruler
+            // grants one held by their realm - the two cases vanilla's Give Settlement covers.
+            var kingdom = ownerClan.Kingdom;
+            bool isOwner = ownerClan.Leader == requestingHero;
+            bool isRuler = kingdom != null && kingdom.Leader == requestingHero;
```
ShoT-UPfps: shouldnt this be `isOwner && isRuler`? vanilla only opens the gift popup when the player owns the settlement and is the kingdom leader, otherwise a stale second request can move the fief again after the first gift made a vassal the owner.

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
```diff
@@ -20,21 +22,159 @@ public class SettlementOwnershipHandler : IHandler
...
+            if (!CanGift(requestingHero, settlement, newOwner, out var reason))
+            {
+                Logger.Warning("Settlement gift of {Settlement} by {Hero} rejected: {Reason}",
+                    settlement.StringId, requestingHero.StringId, reason);
+                return;
+            }
+
+            // ApplyByGift re-enters ChangeOwnerOfSettlementPatch on the server, which publishes
+            // SettlementOwnershipChanged and replicates to every client.
+            GameThread.RunSafe(
```
ShoT-UPfps, line 143: can we add an E2E test for the client gift path and relation bonus? this adds a new authoritative request with several rejection branches and none of it is covered.

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
```diff
@@ -20,21 +22,159 @@ public class SettlementOwnershipHandler : IHandler
...
+                () =>
+                {
+                    // Re-derived here as well, because the check above ran on the network thread and this
+                    // action only runs once the game thread gets to it. In between, ownership or kingdom
+                    // membership can change - the fief can be captured, sold, or the clan can leave the
+                    // realm - and the queued request would then transfer a settlement the requester is no
+                    // longer entitled to give away. This is the check that actually guards the transfer.
+                    if (!CanGift(requestingHero, settlement, newOwner, out var lateReason))
```
ShoT-UPfps: can we send a result back for these rejects? the popup is already closed, so a late validation failure still looks like the same silent no-op this is fixing.

ShoT-UPfps (changes requested):
Comments

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
alikhan866: Fixed in 63ac162d7. The whole request — the lookups, `CanGift` and the apply — now runs inside `GameThread.RunSafe`.

I removed the separate pre-filter rather than keeping a second copy of the check. Resolving the ids and testing authority reads campaign state (`OwnerClan`, `Kingdom`, `IsAlive`), so a "cheap" pre-filter on the network thread is exactly the thing you describe: it races the game loop and can refuse on state that a queued action is about to change. Saving one queued action is not worth judging the request against a world that has already moved on.

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
alikhan866: Good catch that the current rule is wrong — but I think `isOwner && isRuler` is wrong in the other direction, so 63ac162d7 makes it **ownership only**.

`KingdomSettlementVM.ExecuteAnnex` is what opens the popup:

```
IL_0031: settlement.OwnerClan.Leader
IL_003c: Hero.MainHero
IL_0041: bne.un.s IL_0050          // not the owner -> ANNEX (costs influence)
IL_0043: _onGrantFief.Invoke(...)  // owner -> GIFT popup
```

Vanilla gates the gift on ownership alone, offers **annex** to everyone else, and invokes `_onGrantFief` nowhere else; `KingdomManager.GiftSettlementOwnership` has no checks of its own. So requiring both would block a vassal gifting their own fief — which vanilla allows — while the current `||` lets a ruler gift a vassal's fief, which vanilla never offers as a gift.

Ownership also closes the hole you were pointing at: after the first gift the giver is no longer the owner, so the stale second request finds a different owner and is refused. Covered by `ServerGift_RepeatedAfterTheFiefMoved_DoesNotTransferItASecondTime`, which fails under the old rule.

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
alikhan866: Added in 63ac162d7 — `SettlementGiftTests`, four cases: the client forward, the transfer plus relation bonus, refusal for a non-owner, and a repeated request.

Each was checked by reverting its fix and confirming the test goes red, then restoring it. Worth noting because it nearly slipped past me: my first version of these tests passed against the OLD authority rule too, because the fixture had no kingdom so `isRuler` was never true and the assertions proved nothing. Giving the fixture a kingdom the giver actually rules made them discriminating.

Also exercised live with two connected clients — the gift transferred and replicated to the server and both clients, and the relation bonus landed identically on all three.

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
alikhan866: Added in 63ac162d7. Every refusal path now sends `NetworkSettlementGiftRejected` to the requesting peer, and the client surfaces it as a message; only the requester is told, since a refusal is not world state.

You are right that this was the important half — the kingdom screen closes its popup on confirm, so a late validation failure looked exactly like the silent no-op the PR set out to fix.

One limitation worth recording: this has no clean manual trigger, because the screen will not offer a gift the player is not entitled to make. It is reached by a stale or duplicate request, which is what the repeated-request test covers.

ShoT-UPfps: the rejection message still needs to be marshalled through `GameThread.RunSafe`; this handler runs on the poll thread and calls `InformationManager.DisplayMessage` directly.

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
ShoT-UPfps: this still needs both. `ExecuteAnnex` checks ownership, then `KingdomManagementVM.OnGrantFief` only opens `GiftFief` when `Kingdom.Leader == Hero.MainHero`; a non-ruler owner goes through `RelinquishSettlementOwnership` instead.

zzzzzzzzott:
Do you mind quickly uploading a before/after video of this bug?

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
alikhan866: Fixed in 415288ba3.

```csharp
GameThread.RunSafe(
    () => InformationManager.DisplayMessage(new InformationMessage(reason)),
    context: nameof(NetworkSettlementGiftRejected));
```

You are right, and it is the same mistake twice in one PR: I marshalled the request path onto the game thread after your first comment and left the reply path reading off the poller. `DisplayMessage` touches the UI message queue, so it belongs on the game thread for exactly the reason the lookups did.

source/GameInterface/Services/Settlements/Handlers/SettlementOwnershipHandler.cs
alikhan866: You are right and I was wrong. Fixed in 415288ba3 — it now requires both.

I argued ownership-only from `ExecuteAnnex` and stopped at the first gate. `KingdomManagementVM.OnGrantFief` is the second:

```
IL_0015: KingdomManagementVM.get_Kingdom()
IL_001a: Kingdom.get_Leader()
IL_001f: Hero.get_MainHero()
IL_0024: bne.un.s IL_0038                       // not the ruler -> IL_0038
IL_0032: KingdomGiftFiefPopupVM.OpenWith(settlement)
IL_0038: "Are you sure want to give this settlement back to your kingdom?"
```

So a non-ruler owner is offered `RelinquishSettlementOwnership`, which returns the fief to the realm — not a gift to a hero of their choosing. Accepting ownership alone let a client perform a transfer the game never offers them. Ownership is still required, so the replay hole stays closed.

Added `ServerGift_FromAnOwnerWhoDoesNotRuleTheKingdom_IsRefused`, which moves only the throne so it isolates the second gate. It is discriminating: red with the rulership check removed, green with it restored. None of the existing tests caught this — the fixture's giver rules the kingdom, so they pass under either rule, which is how the wrong rule survived my last round of tests as well.

garrettluskey (changes requested):
fix conflicts

garrettluskey:
Please address the comments otherwise we will have to close this PR @alikhan866

garrettluskey:

PR #2751, opened 2026-08-06
Make retreating from a battle server-authoritative

### LINKED ISSUE #1840: Sieges - Attacker retreat during defender deployment can soft-lock a PvP siege
## Description

In a player-attacker versus player-defender siege, if the attacker retreats while the defender is still in deployment, the defender does not receive a victory update.

The defender can open the scoreboard and retreat as well, after which the campaign map can remain soft-locked and report that two players are still in a map event.

## Steps to reproduce

1. Start a siege assault with a player attacker and a player defender.
2. Keep the defender in the deployment phase.
3. Have the attacker retreat.
4. Observe the defender's screen.
5. Have the defender open the scoreboard and retreat.
6. Return to the campaign map.

## Expected behavior

The defender receives the victory result when the attacker retreats, both missions close, and neither player remains registered in the map event.

## Actual behavior

The defender receives no victory update, and after both players retreat the campaign map can still report two players in the map event.

alikhan866:
Fixes #1840

### What was broken

Retreat ran entirely on the client, so the troop sacrifice and the besieger-camp teardown were applied locally and never reached anyone else. Beyond that, the retreating party was never actually taken out of the authoritative battle, and the removal that did happen did not replicate.

### What it does now

**Server-authoritative retreat.** The request is routed through the server, which reuses vanilla's own party-parameterised `SacrificeTroopsWithRatio` — no `MainParty` dereference, so a dedicated host can run it — and broadcasts the result. `TryToGetAwayPatches` covers both shapes the menu can be in: in a battle it publishes a retreat request; besieging with no battle it falls back to `BreakSiegeAttempted`. That replaces `LeaveSoldiersBehindAcceptPrefix`, which only handled the second shape and collided with the first. Commanded parties follow their leader out, but another player's party is never dragged along by someone else's retreat.

**The party actually leaves the battle.** `TryApplyRetreat` teleported the party away and applied its losses but left it on the defender side of the map event, so the server still had it in the encounter and a second click ran the whole retreat again — sacrificing troops and goods for a party that had already got away. Clearing `MapEventSide` is the whole removal: the setter calls `RemovePartyInternal` on the side the party was on and cascades to attached parties. It runs after the loss model and camp clearing, both of which need the battle intact.

**The removal replicates.** `PartyBase`'s `MapEventSide` setter removes through `RemovePartyInternal`, which uses `RemoveAt` and so bypasses the collection sync. The retreat therefore detached the force on the server only: every other player still fighting kept it in their local map event, counted in their battle and still on their side's roster. It now broadcasts the same `NetworkPartyLeftBattle` the ordinary leave path uses rather than inventing a second removal protocol.

**Everyone pulled out gets their own teardown.** The map-event removal cascades to commanded and attached parties, but only the requester got a `NetworkBattleRetreatResolved`, so another player dragged out by their army leader's retreat lost their party from the battle while still sitting in the battle UI. `finishLocalMenus` is now per-party — true for everyone except the requester, whose teardown already comes from the retreat result — and `ApplyNetworkLeave` gates it on `isMainParty`, so it reaches exactly the one client owning each party.

**Attached player parties are collected before the detach.** `CollectCommandedParties` deliberately skips player parties so one player's retreat never spends another's troops, but clearing the requester's `MapEventSide` cascades to attached parties regardless, player-led ones included. The exact case the removal broadcast was added for was the one it missed. They are now collected before the detach — the same assignment that cascades also clears the attachment list — and included in the broadcast.

---

### Also fixes #1840 — a PvP siege could strand the defender

**What was broken.** Leaving a battle *mission* without resolving it — the scoreboard's retreat — withdrew the leaver's agents on every client but never took its **party** out of the map event. The force was gone from the field yet still on its side, so the battle could not resolve and both players stayed counted in the map event.

The other player could not notice either. `BattleEndLogic.CheckIsEnemySideRetreatingOrOneSideDepleted` returns before evaluating a single end condition for as long as a `BattleDeploymentHandler` is on the mission, so a defender still deploying was structurally unable to see that its enemy had left, and sat there facing a side that no longer existed.

**What it does now.**

A mission left with no resolved result is reported to the server, which takes that party out of the battle. A *resolved* result is excluded: that battle concludes through the completion barrier instead, the path that owns the rosters.

When the leave empties a side, the battle goes through the coop finalize rather than `PartyBase`'s setter reaching vanilla's `RemovePartyInternal`, which finalizes the event itself and so bypasses the reserve cleanup, the siege that has to survive an unvictorious assault, and the encounter close every remaining player needs.

`BattleState` stays `None` throughout. A victory state is what forfeits the losing side's roster and captures its heroes; charging that to someone who walked out of an assault would take their whole army prisoner. Vanilla ends an abandoned assault with no victor and keeps the siege, which is what the finalize does.

The remaining player's client ends a mission that can no longer resolve. `CloseEncounter` deliberately leaves an active mission alone so a winner can read its scoreboard; a finalize with nothing decided is the opposite case, and left a mission nothing else would ever end.

---

### Automated tests

`BattleMissionRetreatTests` (E2E) — each of the two positive cases was confirmed to fail with the fix stubbed out, and the three guard cases to keep passing, so they discriminate rather than merely pass.

| Test | Covers |
| --- | --- |
| `LeavingABattleMissionUnresolved_IsForwardedToTheServer` | the client reports the unresolved leave rather than acting locally |
| `MissionRetreatEmptyingASide_FinalizesTheBattle_AndClosesEveryPlayersEncounter` | the reported case: the battle is finalized, nobody is left in the map event, the other player is told to close |
| `MissionRetreatWithAnAllyStillFighting_RemovesOnlyTheRetreatingParty` | one player leaving must not end the battle for the others; the removal replicates |
| `MissionRetreatOfAnAlreadyResolvedBattle_IsRefused` | a victory stays with the completion barrier, so the losing side is not pulled out of the battle it just lost |
| `MissionRetreatForAPartyThePeerDoesNotOwn_IsRefused` | ownership is re-derived from the peer; a client cannot end another player's battle |

`UnresolvableMissionEndRuleTests` (unit) — the client-side rule, extracted as a pure predicate because the headless harness has no live `Mission`.

| Case | Ends the mission? |
| --- | --- |
| No result, battle gone | yes — the stranded case |
| Winner (has a `MissionResult`) | no — the scoreboard survives |
| Battle still exists | no — the fight is not over |
| Close names no battle | no — not evidence of anything |
| No mission / already ending | no |

Full suite: **2746 passed, 0 failed, 18 skipped**.

### Manual two-client test

A player-led siege of Lochana Castle: client 1 besieging, client 2 inside commanding the defence, at war, siege engines built. Every result below was read back from the **server**, not from the clients' screens.

| # | Check | Result |
| --- | --- | --- |
| 1 | Defender enters the assault and stays in the deployment phase | pass |
| 2 | Attacker enters and starts the battle | pass |
| 3 | Attacker retreats from the mission | pass — server: *"Mission retreat of … emptied its side of …; finalizing the battle"* |
| 4 | **Defender leaves the deployment screen on its own** | pass — client: *"Battle … was finalized with nothing decided; ending the local mission that can no longer resolve"* |
| 5 | Neither player left registered in the map event | pass — the assault event is gone; the fast-forward lock counts 0 |
| 6 | The siege survives the abandoned assault | pass — camp leader and engines intact, defender back on the siege-defence menu |
| 7 | The retreat costs no troops | pass — the attacker keeps its full party; `BattleState` never leaves `None` |

Before the fix, step 4 left the defender frozen in deployment indefinitely.

---

### Independent


garrettluskey:

PR #2752, opened 2026-08-06
Fix breaking into a besieged settlement, and the capture aftermath menu

alikhan866:
### Breaking in to help the defenders was unreachable

The break-in menu option was force-disabled whenever a player led the besieging camp, which is the ordinary co-op case, so a besieged player had no way to reinforce their own settlement. It was disabled on the same reasoning as "attack the besiegers", but the two are not alike: attacking really does open a field battle, whereas the break-in's consequence only switches menu and the flow that follows creates no `MapEvent` at all — it sacrifices troops and walks the party inside. The entry itself is already routed through the server, so re-enabling the option does not hand a client authority over it.

`EnterSettlementAction`'s co-op prefix also vetoed any party attached to a map-event side, which is every party during a siege, so the authoritative entry the server applies under an `AllowedThread` was silently dropped. An explicitly allowed original now wins over that veto, with the idempotency check kept ahead of both so a party already inside cannot re-enter.

### The break-in cost nobody anything

During a normal break-in the party is still outside the walls and is not the besieger, so `CurrentSettlement` and `BesiegedSettlement` are both null. The casualties patch read only those, found nothing, and returned — having already suppressed vanilla.

The encounter settlement is the one that holds in that case, so it is tried first; the other two stay as fallbacks for breaking *out* from inside and for a besieger breaking through its own siege. If none resolves, that is now logged rather than passing silently.

### The capture aftermath menu could strand the player

Taking a settlement could leave the capturing player on a dead screen: the settlement encounter fired while the capture menu was still open and pushed a second menu over it, so neither could be dismissed.

The encounter now skips a settlement whose capture menu is still held. An earlier attempt keyed that check on `PlayerEncounter.Battle` being non-null, which is dead code — the battle is already null by the time a sortie has been torn down — so it keys on the hold itself instead. Castles escaped the hold entirely: `GetEncounterMenu` returns `castle_outside` on the same fall-through that gives a town `town_outside`, so that menu is held too.

---

### Depends on #2751

This PR targets `coop/battle-retreat` rather than `development`, because `BreakInCasualtiesPatch.cs` is created by #2751 and only modified here — standalone it is a modify/delete conflict. Merge #2751 first.

The rest of the work was split out of #2668 into PRs that each fix a single problem; the others target `development` directly.


garrettluskey:

PR #2753, opened 2026-08-06
Make besieged settlements sortie, and fix which side a relief force joins

### LINKED ISSUE #2622: [SIEGES] Player becomes an enemy after defeating a siege camp
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

After winning a siege camp encounter, the player is switched to the enemy side in the post-battle encounter. The encounter shows only two options, and selecting either option has no effect.

## Steps to reproduce

1. Join an army as a mercenary on a dedicated server. 2. Engage a siege camp while commanding the party (including re-issuing an attack after a retreat/formation change). 3. Win the battle. 4. Observe the post-battle encounter where you are treated as hostile.

## Additional information

Attempting to rejoin the server during this state causes a crash; reconnecting again returns the character to a free state. No logs were provided. Mod version noted as Nightly build - August 1, 2026 (Dedicated Server).

## Discord source

https://discord.com/channels/709516043332354119/1533584910344912896/1533584910344912896

## Images and videos

[](<https://cdn.discordapp.com/attachments/1533584910344912896/1533584910843904151/image.png?ex=6aa4703f&is=6aa31ebf&hm=2598f8a8cf3bd0d94c3a70a97aed02ae4b31c7d9056d1068f6a4c8a256910b76&>) ([Discord message](https://discord.com/channels/709516043332354119/1533584910344912896/1533584910344912896))

## Logs

Client log: Not provided

Server log: Not provided

alikhan866:
Fixes #2622 — the siege-camp assault put the player on the besiegers' side, which is why winning it left them treated as hostile. The factory redirect below builds the relief battle with the correct sides.

Five reasons a besieged settlement never fought back, and one that put the rescuer on the wrong side.

### Sorties were diverted into a conversation

A sallying garrison reached the player-interaction path and opened a conversation instead of starting the battle, so the sortie silently never happened. Live instrumentation showed the encounter exiting as `DivertedToConversation`; three separate static reviews of the same code had concluded the opposite, so the guard is anchored on that observation — a garrison whose own besieged town is the target starts a battle, never a conversation.

Connected besiegers now join a sortie on the defender side, mirroring the existing assault join, so a second player camped at the same settlement is pulled into the same battle rather than watching from outside.

### Militia were invisible to the sortie decision

Vanilla keeps militia on the walls during a sally-out, which also keeps them out of the strength comparison that decides whether to sortie at all. A town holding thousands of militia behind a small garrison therefore never sorties, which reads as the AI being broken even though it is faithful.

Measured at Marunath with 3000 militia and a 406-man garrison: defender strength 894.6 against a required 1538.3, so the gate failed every time. With militia counted it is 4005.5 and the sortie fires.

Both `Town.GetDefenderParties` and `GetNextDefenderParty` had to agree, or the roster and the party-by-index walk disagree about who is defending. The server has to route through them too — these patches previously deferred to vanilla on the server, and the sortie decision is made there, so a client-only change could never have affected it. Rebellious towns still hold their militia back, exactly as vanilla does. Configurable via `militiaJoinsSallyOut`.

### A retreat decided the siege

Vanilla scores a retreat as a decision, which during a siege is expensive. `KingdomManager.SiegeCompleted` hands the settlement over outright for a won `SallyOut`, so a garrison that turns and runs back through its own gate loses the town on the spot. `FinalizeEventAux`'s teardown lifts the camp for a `SiegeOutside` battle, so a relief force that breaks off ends the siege it never actually broke — and its troops are left standing in the open rather than back behind the walls they came from.

Co-op already handled this for assaults using vanilla's own `_keepSiegeEvent` lever; the same treatment is extended to sally-outs and siege-outside battles, and routed garrison and militia are sent back to the settlement they defend instead of being parked on `Hold` where the battle happened. If the *besieger* is the side that retreated, the siege is genuinely over and vanilla's lift is correct. Configurable via `resumeSiegeWhenEnemyRetreats`.

### Relieving a siege put the player on the besiegers' side

Taking "assault the siege camp" at a friendly besieged castle produced a battle whose attacker side held the player, the player's parties **and** all thirteen besieger parties, while the defender side was the castle itself. The player was assaulting their own castle alongside the enemy — and because both sides were then the same faction, the game split that faction across them, so players also fought their own troops.

The battle request reaches the server naming the settlement as defender, so `defender.IsSettlement` routes it to `StartSiegeMapEvent`. Redirecting the defender to the besieger camp lets the factory's existing `BesiegedSettlement` branch build a `SiegeOutside` event instead — a field battle outside the walls, relief force against besiegers, which is what the menu option has always claimed to do.

The redirect sits in the server's own factory rather than in the encounter. Two earlier attempts fixed the encounter pairing and were each correct at their layer, but something downstream substituted the settlement back in; the factory is the last point before the event exists, so nothing can override it, and being server-side it is authoritative for every client.

### The defenders you came to rescue never came out

A garrison only leaves the walls through a sally-out, and `SallyOutsCampaignBehavior` checks that on the settlement's hourly tick — which cannot run while a player sits in the relief battle, because campaign time is paused. The check that would send them out is one the situation can never reach.

The sortie now happens when the relief battle is created, the one moment it is possible. The roster comes from `GetInvolvedPartiesForEventType(SallyOut)`, so it is exactly who vanilla would send and it honours `militiaJoinsSallyOut` for free; the settlement's own party is skipped, since a fortification cannot march out to a field battle. Configurable via `garrisonJoinsSiegeRelief`.

---

### Independent


garrettluskey:

PR #2754, opened 2026-08-06
Add debug commands for driving siege and sortie scenarios

alikhan866:
Debug console commands for setting up siege and sortie scenarios by hand, so the behaviour can be driven directly instead of waiting for the campaign to produce it.

- **`set_troops`** replaces a party's rank and file with an exact count of one troop type. The existing `siege_buff` fills to 2000, which silently suppresses the very behaviour under test — the sally-out check only fires when the defenders out-ratio the besiegers 2:1.
- **`sally_state`** prints every guard `CheckSallyOut` evaluates for a settlement, plus each defending party's strength, so "the AI won't sortie" is diagnosable instead of guesswork.

---

### Independent


garrettluskey:

PR #2755, opened 2026-08-06
Free a captor holding a captured player, and replicate army joins

alikhan866:
### A lord who captured a player froze on the world map, permanently

Engaging a player applies three things, and `DisableAi` sets `_enableAgainAtHour` to `CampaignTime.Never`, so the party never self-recovers. The only release ran on the capturing client through `PlayerEncounter.Finish`, which captivity suppresses — so the hold was applied and never lifted, and the captor stood still instead of carrying the prisoner to a dungeon.

The server now ends the engagement held by a captured player's party, with a backstop on `PartyBelongedToAsPrisoner`: unlike `PrisonerTaken`, that fires on every capture path, including the ones that early-out.

It also disbands a captured player's army and republishes the captured party's position, which otherwise stayed where the client last saw it — the "ghost party still outside" and the two clients disagreeing about where a party stood.

The four silent early-returns in `ReleaseParty` now log. Each one was a way for the hold to be dropped with no trace of why.

### Joining a gathering army was invisible to everyone else

Accepting a lord's invitation to a gathering army ran only on the joining client, so no one else saw the party in the army. The join is now published from the dialogue consequence so it replicates like any other army membership change.

---

### Independent


garrettluskey:

PR #2756, opened 2026-08-06
Pull nearby AI parties into a player's battle, and keep Surrender working

alikhan866:
### Nearby AI parties never joined a player's battle

Vanilla does this from `PlayerEncounter.CheckNearbyPartiesToJoinPlayerMapEvent`, driven off `PlayerEncounter.Update`. Co-op suppresses that method outright, because on a client it would mutate the shared `MapEventSide` locally and desync — so nothing populated the AI join window, and a friendly army could sit beside your battle doing nothing.

The selection needs no porting: vanilla's method is a one-line delegate to `EncounterModel.FindNonAttachedNpcPartiesWhoWillJoinPlayerEncounter`, which takes only the two side lists, so a headless host can call it directly.

It is driven off `PlayerJoinedBattle`, published in the same block that opens the join window. `CampaignTick` alone is not enough — map time stops while a player sits in an encounter, so a tick-driven scan can go an entire battle without running, which is exactly when reinforcement is wanted. The tick scan stays for parties that arrive mid-battle. Skips are logged with a reason: a bare "nothing happened" is indistinguishable from "never ran".

### The window was opened after the announcement that reads it

`MessageBroker.Publish` is synchronous, so `NearbyPartyReinforcementHandler` ran *inside* the publish call — before the line below it created the window it asks about. `IsWithinAiJoinWindow` was therefore always false for the immediate scan, and nearby lords could only ever join on some later tick, if at all. The window is now created first, so the handler sees the state the announcement is telling it about.

### Surrender went dead for the rest of the encounter

The pre-battle encounter menu offers **Surrender** before any `MapEvent` exists, so the click published a request the server could not resolve — and the latch that guards against double-submission then made every later click a silent no-op.

It now latches only once a usable request can be formed, falling back to `MainParty`'s map event, and leaves the latch clear when there is genuinely none so the player can try again after the battle starts.

---

### Independent


garrettluskey:

PR #2757, opened 2026-08-06
Fix battle deployment wedging and sides spawning nothing

alikhan866:
The largest PR in the stack, and one problem: a battle shared between players did not size or spawn correctly. The commits are separated so each step is reviewable on its own.

### Deployment wedged with nobody spawned

Three distinct causes, all ending on the deployment screen:

- **A null roster during agent spawn.** Vanilla's `CalculateReadyToUpgradeSafe` walks `_mapEventParties.Find(p => p.Party == owner).Troops` and enumerates it with no null check on either step, while co-op fills `MapEventParty._roster` from a separate `NetworkUpdateMapEventParty` message — so a party can be tracked while its `Troops` is still null. The NRE lands under `BattleAgentLogic.OnAgentBuild`, inside `Mission.OnTick`, which kills the tick so `SetupTeams` never finishes. It caught whoever entered before their roster sync arrived, which is why joining in one order worked and the other did not. The existing guard only covered a party missing from the tracker entirely.
- **A side that could never reach its spawn target.** The engine spawns a side only once its reservation reaches `InitialSpawnNumber`, and `CheckDeployment` skips the **whole side — its plan-making included** — while the count falls short. The engine asked for the full deficit, `OwnedShareOf` handed back a fraction, and the gap closed geometrically without reaching zero. Live, on a client owning 382 of a 955-strong side against a 163-troop wave: 65, 39, 23, 14, 8, 5, 3, 2, 1, 1, 1 … stalling one short. Apportionment now happens once, at sizing, and supply hands over what it is asked for.
- **`SetupTeams` throwing every tick.** Native `SetupTeams` dereferences `Mission.InitialPlayerAgent` before it consults the co-op "no leader on field" handling, so whenever a player side reaches deployment without an agent the method throws every tick, with nothing logged to say why. A finalizer, not a prefix — both sides must still spawn, so the method has to run and be allowed to fail at the dereference. What is missing afterwards is the tail, so that is driven directly, in native order, minus the two player-agent lines. `FinishDeployment` cannot be reused: it dereferences the same null.

### Sides were sized from what one client owned

Two players sharing a side fielded more men than a larger enemy. Live case: attacker 955 across two clients, defender 1400 held by the host, and the battle came out 673 attackers against 300 defenders. The spawn logic splits a fixed battle size in proportion to the two totals it is handed, and each client passed only what its own supplier owned.

Sides are now measured whole — `BattleTroopReserveBuilder` accumulates each side's total across all parties before ownership is considered, so every client receives the same pair and computes the same split. Because the engine then asks every client for the whole side's allocation, `SupplyTroops` takes only this client's share of it; the rest arrive as replicated agents through `OwnedAgentReplicator`/`PuppetSpawner`. Serving the full allocation from each owner's pool would field the side once per owner — the same fault inverted.

### The owners' shares now add up exactly

Proportional rounding could put the owners' shares over the allocation, and the floor that stopped an owner rounding to zero was worse: it gave every owner with troops at least one, so a one-troop wave fielded one troop *per owner*.

The final form reserves the troop **before** the split rather than adding it after. The server sends how many parties on the side belong to a player and each one's rank among them; every owner takes one troop off the top per player-owned party, apportions the remainder across the whole side by cumulative flooring, and adds its own reserved troop back. The pieces sum to exactly the allocation — the flooring covers the remainder exactly once, and there are exactly as many reserved troops as player-owned parties — while no player is rounded down to nothing. A wave too small to give every player a troop is the one case somebody misses out, and the ranks settle it identically on every client because the server assigns them.

All new wire fields are additive with defaults that reproduce the previous behaviour, so a reserve from a peer that does not send them still deserialises.

### A client no longer simulates a battle the server owns

Leaving an encounter ends with vanilla simulating whatever is left of the battle. In co-op that decision belongs to the server, which has already resolved the event, so it is a desync by construction — and a crash: `SimulateBattleSetup` indexes the roster array by each side's mission side, and on a client whose sides were never assigned one, `MissionSide` is `None`, which is `-1`.

Observed live after a won sally-out: every click of "Leave…" threw `IndexOutOfRangeException` out of the menu consequence — twelve times in one session — so the encounter never finished and the player was pinned on the menu. The same block also stopped the campaign-side event resolving, so `MapEvent.OnBattleWon` never fired and a battle with 500 kills paid out no loot.

The first version of this guard was too wide: auto-resolve and "send troops" draw their battle screen through those same two methods, so suppressing them left the player with no battle screen at all and threw `NullReferenceException` out of `SPScoreboardVM.Initialize`. It is now scoped with a `ThreadStatic` flag set in a prefix on `MenuHelper.EncounterLeaveConsequence` and cleared in a finalizer — the same idiom `CoopEmptyTeamDeploymentPatch` uses. The leave-menu crash stays fixed; everything else simulates normally.

### Diagnostics

Side replication is logged at both ends — what the server broadcasts, what index the client binds each side to, and where the local player's own party lands — because a divergence there puts every later party on the wrong side and aborts the battle with "your party's troop reserve was not received". The temporary spawn diagnostics added while tracking the wedge are removed again in this PR.

---

### Independent


garrettluskey:

PR #2758, opened 2026-08-06
Stop silently dropping party state, and a 4,200 line/sec log flood

alikhan866:
Three ways party state was lost or drowned out, none of which said anything.

### A log flood that cost throughput, not just noise

The generated property-set prefix logged the client branch **before** the equality check, so a client wrote an `Error` for every set of a synced property — including the many the engine performs with the value the object already holds.

`Settlement.IsVisible` is rewritten every frame for every settlement, and the cost was not subtle: ~4,200 log lines per **second** while the campaign clock ran — 47,592 lines out of a 47,703-line session log, 99.8% of it — all Serilog work on the game thread. Server throughput went from 38 to 155 party behaviour updates per ten seconds once it stopped. It also stopped the moment the game was paused, which is what made it look like a gameplay problem rather than a logging one.

Moving the check above the client branch fixes it for every synced property rather than that one. The server already skipped no-op writes; this gives the client the same courtesy before it pays for a log line.

### Party behaviour dropped wholesale instead of degrading

`TryCreate` abandoned a party's whole behaviour update when its interactable, target party or target settlement was not registered — so a party pointing at something destroyed had **no** behaviour sent for it, ever. The `MoveTargetParty` branch a few lines below has always dropped an unresolvable reference and carried on; the other three now do the same.

`TryApplyJoinBaseline` was all-or-nothing: one party that failed to resolve, or a client whose party count differed from the server's by a single party, discarded the **entire** baseline. That is worse than it sounds, because ordinary updates are published on *change* — a lord already patrolling when a client joins generates nothing to send — so the baseline is the only way such a party's behaviour ever reaches that client. Losing it leaves the party displaying an order it never acts on. It now applies everything that resolves and reports what it skipped.

Both failures returned `false` without a word, which is what made them invisible: a client could discard 573 KB of authoritative state and log nothing at all.

### A disposed container threw out of a method that reports failure with `false`

`TryGetContainer` only null-checks the scope, and a disposed Autofac scope is still a non-null reference — so `TryResolve` went on to resolve from it and Autofac threw `ObjectDisposedException` straight through a method whose whole contract is to answer with `false` when it cannot resolve.

It escapes into teardown, which is the worst place for it. `ConversationPartyTracker` disposes by releasing every party it still holds, and a party held through `DisableAi` never re-enables itself — `CampaignTime.Never`, and `_isDisabled` is saveable. An exception part-way through that loop leaves the remaining parties frozen for the life of the save.

---

### Independent


garrettluskey:

PR #2759, opened 2026-08-06
Split the methods the analyzer flagged as too complex

alikhan866:
No behaviour change. This is the analyzer clean-up for the methods the PRs below it grew, kept separate so those PRs stay readable as diffs of intent.

Each flagged method was a single body doing validation, pricing and mutation at once; splitting them along those seams is what drops the complexity, and the seams were already there as comment blocks.

- The barter handlers now separate "is this request legitimate" from "what is the offer worth" from "apply it", which also makes the mutation boundary explicit — past that point a failure has to be reported as success, because the server state has already moved.
- `ModOptions`' constructor reads its defaults through a helper rather than seventeen null-coalesces in a row. A run of `??` that long looks like seventeen branches both to an analyzer and to anyone skimming for the one option they care about.
- Fixes the outstanding blank-line and closing-brace findings.

It sits last in the stack because it touches files introduced across most of the PRs below it.

---

### The stack

Split out of #2668 so each PR fixes a single problem. Review and merge bottom-up: every PR targets the branch below it, and the bottom targets `development`.

1. #2748 — Restore lord recruitment by persuasion
2. #2749 — Fix barters being refused, double-applied or misreported
3. #2750 — Let clients grant fiefs, and pay the relation for it
4. #2751 — Make retreating from a battle server-authoritative
5. #2752 — Fix breaking into a besieged settlement, and the capture aftermath menu
6. #2753 — Make besieged settlements sortie, and fix which side a relief force joins
7. #2754 — Add debug commands for driving siege and sortie scenarios
8. #2755 — Free a captor holding a captured player, and replicate army joins
9. #2756 — Pull nearby AI parties into a player's battle, and keep Surrender working
10. #2757 — Fix battle deployment wedging and sides spawning nothing
11. #2758 — Stop silently dropping party state, and a 4,200 line/sec log flood
12. **#2759 — Split the methods the analyzer flagged as too complex** ← this PR

<sub>A thirteenth PR (#2747, kingdom membership replication) was closed as a duplicate of #2607.</sub>


alikhan866:
Closing this as part of flattening the stack.

On review feedback that a PR should fix a single problem, the stack has been restructured so that nine of its PRs target `development` directly, leaving only two genuine dependencies (#2749 on #2748, and #2752 on #2751). This PR cannot survive that change: it refactors methods written by six of the others, so with those now merging in parallel there is no single branch it can sit on.

It was also the weakest member of the stack — no behaviour change, purely analyzer complexity findings — so it is the right one to drop rather than force into an artificial position.

The analyzer findings are worth doing, just not here and not yet. Once the functional PRs have landed, the same clean-up can be redone against `development` in one pass, which will be smaller and easier to review than a refactor spread across six unmerged branches.

Nothing is lost: every functional change stays in its own PR, and none of them depend on this one.

PR #2761, opened 2026-08-06
Allow filtering hero list by name prefix

ShoT-UPfps:
## PR Checklist

## PR Type

`coop.debug.hero.list` always prints every hero.

## What is the new behavior?

`coop.debug.hero.list` accepts an optional case-insensitive name prefix. Multi-word prefixes are supported, and omitting the prefix still lists every hero.

## Bot Changelog Entry

- Added optional name filtering to the hero list debug command.


source/GameInterface/Services/Heroes/Commands/HeroDebugCommand.cs
```diff
@@ -39,7 +39,9 @@ public static string ListHeroes(List<string> args)
             return $"Unable to get {nameof(IObjectManager)}";
         }
 
-        foreach (var hero in Campaign.Current.CampaignObjectManager.GetAllHeroes())
+        string namePrefix = args == null ? string.Empty : string.Join(" ", args).Trim();
+        foreach (var hero in Campaign.Current.CampaignObjectManager.GetAllHeroes()
+                     .Where(hero => NameStartsWithPrefix(hero.Name?.ToString(), namePrefix)))
```
garrettluskey, line 44: when nothing matches this returns an empty string, so you cant tell a bad prefix from zero heroes. want to print something like `FindIds` does (`No hero named '...' was found.`)?

garrettluskey (dismissed):
small clean change, the prefix helper and its tests look right and the no-arg path still lists every hero. one thing though, when the filter matches nothing the command returns an empty string, so you cant tell a typo from zero matches. `FindIds` right below prints `No hero named ... was found.`, so itd be nice to do the same here.

source/GameInterface/Services/Heroes/Commands/HeroDebugCommand.cs
ShoT-UPfps: fixed, it now reports the prefix when no heroes match.

garrettluskey (approved):
Earlier comment is addressed: a prefix that matches nothing now returns `No hero with a name starting with '<prefix>' was found.` instead of an empty string, and the no-prefix path still lists everything. Helper handles null names and matches the OrdinalIgnoreCase pattern used by FindIds/SetGold, with tests for empty prefix, case-insensitivity, and mid-name non-match. One note: only CodeFactor has run on this new head, the build/test/E2E checks haven't triggered, so give those a run before merge.

PR #2722, opened 2026-08-05
15 trade quests

Thaipho:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] All new classes have class-level documentation comments, if there are any at all
- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update
- [ ] Other... Please describe:

## What is the current behavior?
All vanilla Issue-system quests are currently blocked in multiplayer — the Issue system is disabled wholesale as a safety measure, so none of these can be offered, accepted, or completed in a coop session.

## What is the new behavior?
Adds multiplayer sync for 15 trade/economic vanilla quest types: Village Needs Tools, Village Needs Crafting Materials, Village Needs Grain Seeds, Village Needs Draught Animals, Deliver the Herd to Town, Lord Needs Horses, Landlord: The Art of the Trade, Artisan Can't Sell Products At A Fair Price, Artisan Overpriced Goods, Smugglers, Revenue Farming, Army Needs Supplies, Gang Leader Needs Special Weapons, Gang Leader Needs Weapons, Gang Leader Needs to Offload Stolen Goods.

This is a scoped-down slice of the broader multiplayer-quests effort in #2675, pulled onto its own branch so it's reviewable on its own before the rest. Each quest has its own commit. 91 E2E tests pass standalone on this branch; 9 of the 15 quests have dedicated test files, the other 6 don't yet (not hiding that).

## Bot Changelog Entry
Added multiplayer support for 15 trade-related quests (village supply deliveries, caravans, smugglers, tax collection, and more).

## Other information
Split off from #2675, which stays open covering the rest of the vanilla quest roadmap (~20 more non-trade types already implemented, plus a few still in progress). No changes outside the files needed for these 15 quests — development itself isn't touched by anything here beyond what these quests require.

PR #2763, opened 2026-08-06
Feature/generic quest handler

### LINKED ISSUE #2800: Generic quests handler
## Description
Generic handler for quests

## Intended Design
provide a framework for quests

## Location
`source/GameInterface/Services/Issues/`
- `Generic/` — the extensibility core: `QuestTypeDescriptor.cs`, `QuestTypeRegistry.cs`, `QuestTypeBootstrap.cs`, `IssueOwnershipRegistry.cs`, `IssueGenerationRegistry.cs`, `IssueFinalizationSupport.cs`, `AlternativeSolutionCompletion.cs`, `ModuleRescanCompletion.cs`, plus `AcceptMirror/`, `CreationCapture/`, `Dispatch/`
- `Handlers/` — network message handlers (`IssueFinalizationHandler.cs`, `AwaitingAlternativeSolutionTroopsHandler.cs`, `AlternativeSolutionCompletionHandler.cs`, the accept-mirror handler)
- `Messages/`, `Patches/`, `Interfaces/` — wire messages, Harmony patches, and supporting interfaces
- `Generic/Migrated/<QuestName>/` — one folder per quest type that's been ported onto the framework (currently `VillageNeedsTools`, `VillageNeedsCraftingMaterials`)

## Related Issues
None

## Requirements
- Accept is server-authoritative: the server arbitrates races between simultaneous accepts (first valid request wins), and losers are rolled back on every peer regardless of message arrival order.
- Every accept/finalize request is validated server-side against real state (requester identity derived from the authenticated connection, not client-claimed; issue-generation stamp rejects stale/superseded requests; troop selections clamped to the requester's actual party roster).
- A claimed quest-success completion is independently re-derived server-side before being honored, not trusted at face value from the client.
- Broadcast-only (server→client) messages are rejected if a client attempts to send them directly to the server.
- Onboarding a new quest type onto the framework requires only a `QuestTypeDescriptor` registration (via `QuestDescriptorBuilder`) — no new Harmony patch file for creation/accept detection.
- The behavior Allowlist (`DisableAllIssueBehaviorsExceptAllowlist`) only ever contains quest types with real, tested sync code on that branch — never pre-populated ahead of the implementation.

## Definition of Done

Thaipho:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

No shared infrastructure exists for syncing Issue-quests in multiplayer — each quest type would need to reimplement creation-time field capture, accept-time mirroring (including race-arbitration between simultaneous accepts), and party-spawn coordination independently.

## What is the new behavior?
Adds a shared generic quest-handler infrastructure for Issue-quest multiplayer sync

## Bot Changelog Entry
Infrastructure changes related to quests

## Other information
For discussion https://github.com/Bannerlord-Coop-Team/BannerlordCoop/issues/2800


source/GameInterface/Services/Issues/Generic/Dispatch/GenericQuestTypeDispatchPatches.cs
```diff
@@ -0,0 +1,116 @@
...
+{
+    // See GenericQuestTypeCreationTriggerPatch's own Postfix doc comment for why First is required here too.
+    [HarmonyPriority(Priority.First)]
+    [HarmonyPatch(nameof(IssueManager.StartIssueQuest))]
+    [HarmonyPostfix]
+    private static void Postfix(Hero issueOwner, bool __result)
+    {
+        // A mirror replay (server authoritative replay, or another peer applying a received broadcast) runs
+        // under AllowedThread - skip re-publishing so it doesn't loop back through the network again.
+        if (CallOriginalPolicy.IsOriginalAllowed()) return;
```
ShoT-UPfps: `AllowedThread` isnt valid for the server replay because our patches use it to skip registration and replication. can we use a dedicated replay flag here so `StartIssueQuest` still runs on the server with patches live?

source/GameInterface/Services/Issues/Generic/AcceptMirror/IAlternativeAcceptMirrorStrategy.cs
```diff
@@ -0,0 +1,82 @@
...
+/// <see cref="IRaceArbitratedAcceptMirrorStrategy{TFields}"/> - no Quest object exists to replay/force-write
+/// onto for this path (<c>StartIssueWithAlternativeSolution</c> never calls <c>GenerateIssueQuest</c>), so there
+/// is no <c>ReplayQuestAccepted</c> step and no "ensure quest exists" precondition. The target is the parent
+/// <see cref="TaleWorlds.CampaignSystem.Issues.IssueBase"/>'s own state.
+/// </summary>
+public interface IAlternativeAcceptMirrorStrategy<TPayload>
+{
+    /// <summary>Force-writes the generic Issue-state transition common to EVERY real instance of this shape
+    /// (<c>_issueState -&gt; SolvingWithAlternativeSolution</c>, <c>IsTriedToSolveBefore -&gt; true</c>,
+    /// <c>IssueDueTime -&gt; CampaignTime.Never</c>), then applies <paramref name="payload"/> via whatever
```
ShoT-UPfps: this doesnt match vanilla `StartIssueWithAlternativeSolution`. vanilla sets the due time to `AlternativeSolutionReturnTimeForTroops` and also initializes the failure/casualty state, logs, effect-clear time, and companion state, so mirroring only these fields leaves an incomplete alternative solution.

source/GameInterface/Services/Issues/Generic/Dispatch/GenericQuestTypeDispatchPatches.cs
```diff
@@ -0,0 +1,116 @@
...
+/// </summary>
+[HarmonyPatch(typeof(IssueBase), nameof(IssueBase.CompleteIssueWithAlternativeSolution))]
+internal class GenericQuestTypeAlternativeSolutionOwnershipGatePatch
+{
+    [HarmonyPrefix]
+    private static bool Prefix(IssueBase __instance)
+    {
+        if (!QuestTypeRegistry.IsRegistered(__instance?.GetType())) return true;
+
+        return VillageNeedsToolsIssueOwnership.IsLocalPeerOwner(__instance.IssueOwner);
```
ShoT-UPfps: this makes alternative completion client-authoritative because the dedicated server can never match a player controller id. shouldnt the owner send a request and let the server run `CompleteIssueWithAlternativeSolution`?

source/GameInterface/Services/Issues/Patches/VillageNeedsToolsIssueOwnershipGatePatches.cs
```diff
@@ -0,0 +1,38 @@
...
+/// <c>FinishQuestSuccess1()</c> (which deducts items, pays gold/exchange goods, and completes the quest)
+/// becomes unreachable for anyone except the recorded owner.
+/// </summary>
+[HarmonyPatch(typeof(VillageNeedsToolsIssueBehavior.VillageNeedsToolsIssueQuest), "PlayerHasTools")]
+internal class VillageNeedsToolsQuestOwnershipGatePatch
+{
+    [HarmonyPrefix]
+    private static bool Prefix(VillageNeedsToolsIssueBehavior.VillageNeedsToolsIssueQuest __instance, ref bool __result)
+    {
+        if (!VillageNeedsToolsIssueOwnership.IsLocalPeerOwner(__instance.QuestGiver))
```
ShoT-UPfps: nothing in this PR calls `SetOwner` outside save restore, so this returns false for every fresh or old-save quest and hides the valid turn-in. can we add the writer before activating this gate?

source/GameInterface/Services/Issues/Generic/AcceptMirror/IRaceArbitratedAcceptMirrorStrategy.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// issue is still genuinely acceptable (the same "is this the right issue type AND still
+    /// IsOngoingWithoutQuest" check every real handler already does), replays the accept, and captures what the
+    /// replay produced. Returns false (nothing to broadcast, caller should reject the requester) if the issue was
+    /// no longer acceptable or the replay didn't produce a readable quest.</summary>
+    public bool TryArbitrate(Hero owner, System.Func<Hero, bool> canAccept, out TFields fields)
+    {
+        fields = default;
+        if (owner == null || canAccept == null || !canAccept(owner)) return false;
+
+        _strategy.ReplayQuestAccepted(owner);
```
ShoT-UPfps, line 30: `ReplayQuestAccepted` mutates the server before capture can fail. if `TryCaptureQuestFields` returns false we reject the requester but leave the issue accepted with nothing broadcast, so this needs to roll back or make replay and capture one operation.

source/GameInterface/Services/Issues/Generic/CreationCapture/CreationCaptureRunner.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// technique every real replicated-construction call site already uses.
+    ///
+    /// <paramref name="afterRegistered"/> is an escape hatch: optional, runs AFTER
+    /// <see cref="IssueManager.CreateNewIssue"/>'s own bookkeeping (including any <c>AfterIssueCreation()</c>
+    /// override) has already completed, for the rare field that override itself overwrites - forcing it
+    /// beforehand (the normal, <see cref="ICreationCaptureStrategy{TIssue,TFields}.ConstructReplicated"/> order)
+    /// would be silently clobbered. Null (the default) for types with no such hazard.</summary>
+    public TIssue ConstructAndRegisterReplicated(Hero owner, TFields fields, Action<TIssue, TFields> afterRegistered = null)
+    {
+        var issue = _strategy.ConstructReplicated(owner, fields);
```
ShoT-UPfps: `ConstructReplicated` and `afterRegistered` are also part of applying the received creation, but they run outside `AllowedThread` here. any patched code they reach can publish back while we are applying the message.

source/GameInterface/Services/Issues/Generic/QuestTypeDescriptor.cs
```diff
@@ -0,0 +1,196 @@
...
+            return this;
+        }
+
+        public Builder<TIssue, TQuest> WithAlternativeAccept<TPayload>(IAlternativeAcceptMirrorStrategy<TPayload> strategy)
+        {
+            _alternativeAccept = strategy;
+            return this;
+        }
+
+        public Builder<TIssue, TQuest> WithBespokeModule(IQuestBespokeModule<TIssue, TQuest> module)
```
ShoT-UPfps: `WithBespokeModule` only stores this object, nothing reads `BespokeModule` or calls its lifecycle/save hooks. is the generic dispatch for these modules missing from this PR?

source/GameInterface/Services/Issues/Interfaces/VillageNeedsToolsIssueOwnership.cs
```diff
@@ -0,0 +1,90 @@
...
+        if (issueGiver == null || string.IsNullOrEmpty(controllerId)) return;
+
+        OwnerControllerIdByIssueGiver[issueGiver] = controllerId;
+    }
+
+    /// <summary>
+    /// Drains this hero's entry, if any, on a genuine finalize so it can never leak into this hero's NEXT,
+    /// unrelated issue.
+    /// </summary>
+    public static void Clear(Hero issueGiver)
```
ShoT-UPfps: `Clear` has no production caller, so an owner entry survives finalization and can apply to the same hero's next issue. this needs to be called from the finalize path.

source/GameInterface/Services/Issues/Generic/AcceptMirror/IAcceptMirrorStrategy.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+        _writers = writers ?? Array.Empty<(FieldInfo, Func<TFields, object>)>();
+    }
+
+    public void ForceWrite(TQuest quest, TFields fields)
+    {
+        if (quest == null) return;
+        foreach (var (field, selector) in _writers)
+        {
+            field.SetValue(quest, selector(fields));
```
ShoT-UPfps: please pass typed read/write delegates here instead of `FieldInfo`. `TaleWorlds.CampaignSystem` is publicized, so these fields can be bound directly and kept compile-time checked.

source/GameInterface/Services/Issues/Generic/PartySpawn/PartySpawnRunner.cs
```diff
@@ -0,0 +1,56 @@
...
+/// class only factors out the capture/forward/create/force-write calls those shells make, instead of each one
+/// duplicating the same 4-step dance by hand.
+///
+/// <c>TSpawned</c> (added alongside CaravanAmbush/SnareTheWealthy's migration - see
+/// <see cref="PartySpawnSpec{TQuest,TCapture,TSpawned}"/>'s own doc comment for the full derivation): generalizes
+/// <c>CreateOnServer</c>/<c>ApplySpawnResult</c> from a hardcoded single <see cref="MobileParty"/> to whatever
+/// shape a given instance's real spawn method actually force-writes - a bare type-parameter widen for the 3
+/// existing single-party consumers (<c>TSpawned = MobileParty</c>), a 2-tuple for the two 2-party consumers.
+/// </summary>
+public static class PartySpawnRunner
```
ShoT-UPfps: please make this an injected transient service instead of a static helper. same for `ModuleRescanCompletionRunner` and the two instance-resolution helpers.

source/GameInterface/Services/Issues/Generic/PartySpawn/PartySpawnSpec.cs
```diff
@@ -0,0 +1,63 @@
...
+/// local body either way; this only decides whether a request is actually published.
+/// </summary>
+public delegate bool TryCaptureForForwarding<TQuest, TCapture>(TQuest quest, out TCapture captured);
+
+/// <summary>
+/// Declarative record of which <see cref="PartySpawnTrigger"/> shape a given migrated type's spawn gate uses,
+/// plus the per-instance reconciliation hooks that a trigger shape alone doesn't capture (a captured/forwarded
+/// scalar; a bespoke deferred continuation; or neither, for the simplest instances).
+///
+/// <typeparamref name="TSpawned"/> - THE OUTPUT-SIDE GENERALIZATION (added when migrating CaravanAmbush/
```
ShoT-UPfps: these comments are far too long for the codebase and some still reference classes that dont exist, like `VillageNeedsToolsAlternativeSolutionCompletionPatches`. please keep the non-obvious why here and move the rest to the commit message.

ShoT-UPfps (changes requested):
comments

source/GameInterface/Services/Issues/Generic/PartySpawn/PartySpawnSpec.cs
Thaipho: already addressing it

source/GameInterface/Services/Issues/Generic/PartySpawn/PartySpawnRunner.cs
Thaipho: will convert to injected transient services

source/GameInterface/Services/Issues/Generic/AcceptMirror/IAcceptMirrorStrategy.cs
Thaipho: will convert to typed read/write delegates

source/GameInterface/Services/Issues/Patches/VillageNeedsToolsIssueOwnershipGatePatches.cs
Thaipho: yes these have no caller in this diff. This PR is the framework in isolation (per an earlier request to split it from the quest types); SetOwner/Clear/the bespoke-module hooks all get real callers once the first quest-type PR lands on top of this. if i submitted the whole thing it would be alot more code here

source/GameInterface/Services/Issues/Generic/QuestTypeDescriptor.cs
Thaipho: yes these have no caller in this diff. This PR is the framework in isolation (per an earlier request to split it from the quest types); SetOwner/Clear/the bespoke-module hooks all get real callers once the first quest-type PR lands on top of this. if i submitted the whole thing it would be alot more code here

source/GameInterface/Services/Issues/Interfaces/VillageNeedsToolsIssueOwnership.cs
Thaipho: yes these have no caller in this diff. This PR is the framework in isolation (per an earlier request to split it from the quest types); SetOwner/Clear/the bespoke-module hooks all get real callers once the first quest-type PR lands on top of this. if i submitted the whole thing it would be alot more code here

ShoT-UPfps (dismissed):
lgtm

source/GameInterface/Services/Issues/Patches/IssueAcceptancePatches.cs
```diff
@@ -0,0 +1,60 @@
...
+[HarmonyPatch(typeof(IssueManager))]
+internal class IssueQuestAcceptancePatch
+{
+    // Deliberately never blocked (unlike IssueManagerCreateNewIssuePatches' creation gate): this runs inline
+    // inside a live conversation, and the next dialogue line only exists because building the Quest here
+    // registers it with the ConversationManager - blocking it on a client would break their conversation
+    // mid-flow. Only the genuine local accept ever reaches this method for real; a mirror replay on every
+    // other peer force-writes state instead (see VillageNeedsToolsIssueInterface.MirrorQuestAccepted).
+    [HarmonyPatch(nameof(IssueManager.StartIssueQuest))]
+    [HarmonyPostfix]
```
garrettluskey: This postfix makes the initiating client commit `StartIssueQuest` before the server decides anything, so this is not server-authoritative. It also breaks the race rollback: the loser already has a quest when the winner broadcast arrives, so `MirrorQuestAccepted` no-ops, the handler records the winner, and the later rejection no-ops because ownership is now recorded. The loser keeps its independently-created quest. Intercept before persistent mutation and execute the real accept only after the server accepts it under a replay guard. If the conversation needs synchronous dialogue registration, split that presentation setup from authoritative quest creation rather than committing the quest locally.

source/GameInterface/Services/Issues/Handlers/VillageNeedsToolsIssueHandler.cs
```diff
@@ -0,0 +1,303 @@
...
+            // client-claimed value - a client cannot claim a different ControllerId to steal ownership.
+            if (requester == null || !playerManager.TryGetPlayer(requester, out var player))
+            {
+                Logger.Error("Rejecting {Message} from an unregistered/unknown requester for owner {Owner}",
+                    nameof(RequestVillageIssueAcceptQuest), ownerId);
+                if (requester != null) network.Send(requester, new NetworkVillageIssueAcceptRejected(ownerId));
+                return;
+            }
+
+            if (GenericAcceptMirrorIssueTypes.IsQuestSolutionMirrorEligible(owner.Issue) && owner.Issue.IsOngoingWithoutQuest)
```
garrettluskey: Authenticating the requester proves identity, but this check does not authorize the accept. Any registered peer can submit an owner ID and accept any eligible ongoing issue without an active conversation or vanilla eligibility. Because the request identifies only the hero, a delayed request can also accept that hero’s replacement issue. Require a short-lived server-issued capability (or equivalent server conversation state) bound to requester, exact issue generation, and solution path, then revalidate it here before running the accept.

source/GameInterface/Services/Issues/Handlers/VillageNeedsToolsIssueHandler.cs
```diff
@@ -0,0 +1,303 @@
...
+            {
+                Logger.Error("Rejecting {Message} from an unregistered/unknown requester for owner {Owner}",
+                    nameof(RequestVillageIssueAcceptAlternative), ownerId);
+                if (requester != null) network.Send(requester, new NetworkVillageIssueAcceptRejected(ownerId));
+                return;
+            }
+
+            if (GenericAcceptMirrorIssueTypes.IsAlternativeSolutionMirrorEligible(owner.Issue) && owner.Issue.IsOngoingWithoutQuest)
+            {
+                issueInterface.MirrorAlternativeAccepted(owner);
```
garrettluskey: This is not a server-side alternative accept; it only force-writes the bare issue state after the client already ran vanilla locally. The request contains no selected companion/troop proposal, so the server cannot validate or reserve the roster or authoritatively generate return/effect times, failure/casualty rolls, XP, reward skill, and logs. Send stable selection IDs as a proposal, validate them against the requester’s authoritative party, run `StartIssueWithAlternativeSolution` on the server, and broadcast the complete captured result.

source/GameInterface/Services/Issues/Handlers/VillageNeedsToolsIssueHandler.cs
```diff
@@ -0,0 +1,303 @@
...
+                network.SendAll(new NetworkVillageIssueQuestAccepted(ownerId, player.ControllerId));
+            }
+            else
+            {
+                network.Send(requester, new NetworkVillageIssueAcceptRejected(ownerId));
+            }
+        });
+    }
+
+    private void Handle_NetworkVillageIssueQuestAccepted(MessagePayload<NetworkVillageIssueQuestAccepted> payload)
```
garrettluskey: `NetworkVillageIssueQuestAccepted` is a server-to-client broadcast, but this handler also runs on the server and has no source guard. A client can send this `ICommand`, choose `OwnerControllerId`, and bypass request arbitration entirely. The same omission exists for alternative-accepted, accept-rejected, and removed broadcasts below. Reject these messages on the server (preferably enforce message direction centrally) and add forged-broadcast tests.

source/GameInterface/Services/Issues/Handlers/VillageNeedsToolsIssueHandler.cs
```diff
@@ -0,0 +1,303 @@
...
+    {
+        if (ModInformation.IsClient) return;
+
+        var ownerId = payload.What.OwnerId;
+        var reason = payload.What.Reason;
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Hero>(ownerId, out var owner)) return;
+
+            issueInterface.FinalizeMirror(owner, reason);
```
garrettluskey: This finalizes whichever issue currently belongs to `OwnerId` with a client-supplied outcome, without checking `payload.Who`, recorded ownership, issue generation, state, or the actual completion condition. Any connected client can grant success/cancel/failure to another player’s quest. Completion must run on the server: accept only a narrow completion intent, derive identity from the peer, validate owner + exact issue + condition, choose the outcome server-side, execute it, then broadcast the result.

source/GameInterface/Services/Issues/Patches/NewIssueTypesAlternativeSolutionPatches.cs
```diff
@@ -0,0 +1,266 @@
...
+    /// <see cref="IssueManagerQuestCompletedReasonCapture"/>'s fallback instead of
+    /// <c>AlternativeSolutionSuccess</c> - a cosmetic label difference only, harmless).
+    /// </summary>
+    private static void TryTriggerOwnedAlternativeSolutionCompletion(IssueBase issue)
+    {
+        if (!issue.IsSolvingWithAlternative || !issue.AlternativeSolutionReturnTimeForTroops.IsPast) return;
+
+        IssueManagerQuestCompletedReasonCapture.PendingReasons[issue.IssueOwner] = VillageIssueFinalizeReason.AlternativeSolutionSuccess;
+
+        issue.CompleteIssueWithAlternativeSolution(); // genuine call - NOT under AllowedThread, this is the real trigger
```
garrettluskey: This still executes genuine alternative completion on the owning client, including its unseeded RNG, rewards, relationship changes, XP, and casualties. `AlternativeSolutionCompletionRunner` in this PR explicitly documents the correct server-request shape, but it has no call sites. Route the due-time trigger to an authenticated server request, validate owner/state/due time, call `CompleteOnServer`, and broadcast the authoritative result.

source/GameInterface/Services/Issues/Handlers/AwaitingAlternativeSolutionTroopsHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+        if (ModInformation.IsClient) return;
+
+        var data = payload.What;
+        var roster = TroopRoster.CreateDummyTroopRoster();
+        foreach (var element in troopRosterInterface.UnpackTroopRosterData(data.Troops))
+        {
+            roster.AddToCounts(element.Character, element.Number, false, element.WoundedNumber, element.Xp, false);
+        }
+
+        AwaitingAlternativeSolutionTroopsRegistry.Deposit(data.OwnerControllerId, roster);
```
garrettluskey: The server stores a client-supplied controller ID and arbitrary roster without looking at `payload.Who` or validating ownership/contents; the drain handler likewise clears any claimed owner. A client can fabricate returned troops or overwrite/clear another player’s pending return. Derive controller identity from the authenticated peer and retain the server-validated roster from authoritative alternative acceptance rather than accepting roster contents later from the client.

garrettluskey (changes requested):
The generic capture/force-write pieces are useful, and deriving `ControllerId` from the authenticated `NetPeer` in the request handlers is the right direction. I’m requesting changes because the accept lifecycle is still client-optimistic rather than server-authoritative, and several client messages can directly cause authoritative issue mutations.

The contract here should be: a client sends only an intent plus stable selection IDs; the server authenticates the peer, matches an exact issue generation, revalidates eligibility/ownership and the selected roster, runs the genuine vanilla accept/completion (including RNG and consequences), captures the result, and broadcasts it. Clients apply that result only after the server accepts it. The live-conversation constraint needs a presentation/dialogue staging seam rather than allowing the client to commit persistent quest state first. The blockers below are concrete cases where the current flow violates that contract.

source/E2E.Tests/Services/Issues/AwaitingAlternativeSolutionTroopsTests.cs
```diff
@@ -0,0 +1,469 @@
...
+            }
+
+            if (!records.TryGetValue(key, out var value)) return false;
+            data = (T)value;
+            return true;
+        }
+    }
+
+    /// <summary>
+    /// The core reconnect scenario this bug is about (task's own words: "simulate a client-owned alt-solution
```
zzzzzzzzott: reduce the size of these comments

source/E2E.Tests/Services/Issues/AwaitingAlternativeSolutionTroopsTests.cs
```diff
@@ -0,0 +1,469 @@
...
+            {
+                var exception = Record.Exception(() => Campaign.Current.IssueManager.TryToMakeTroopsReturn(owner.Issue));
+                Assert.Null(exception);
+            }
+            finally
+            {
+                Hero.MainHero.ChangeState(previousState);
+            }
+
+            // NOTE on the man count specifically: AwaitingAlternativeSolutionTroopsRegistry (like every other
```
zzzzzzzzott: reduce the size of these comments

source/E2E.Tests/Services/Issues/AwaitingAlternativeSolutionTroopsTests.cs
```diff
@@ -0,0 +1,469 @@
...
+        // on the owner's next reconnect after already being returned once.
+        Assert.Single(Client.NetworkSentMessages.GetMessages<RequestAwaitingAlternativeSolutionTroopsDrain>());
+        Server.Call(() =>
+        {
+            Assert.False(AwaitingAlternativeSolutionTroopsRegistry.TryGet(controllerId, out _));
+        });
+    }
+
+    /// <summary>
+    /// The task's own explicitly-called-out highest-priority immediate fix: vanilla's
```
zzzzzzzzott: reduce the size of these comments

source/GameInterface/Services/Issues/Interfaces/SimpleIssueFactoryRegistry.cs
```diff
@@ -0,0 +1,161 @@
+using Common.Util;
```
zzzzzzzzott, line 1: reduce comments in this file

source/GameInterface/Services/Issues/Patches/IssueFinalizedPatches.cs
```diff
@@ -0,0 +1,112 @@
+using Common.Messaging;
```
zzzzzzzzott, line 1: reduce comments in this file

source/GameInterface/Services/Issues/Patches/IssueManagerCreateNewIssuePatches.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Common;
```
zzzzzzzzott, line 1: reduce comments in this file

source/GameInterface/Services/Issues/Patches/IssueManagerTickPatches.cs
```diff
@@ -0,0 +1,105 @@
+﻿using Common;
```
zzzzzzzzott, line 1: reduce comments in this file

source/GameInterface/Services/Issues/Patches/IssuesCampaignBehaviorGenerationPatches.cs
```diff
@@ -0,0 +1,46 @@
+﻿using Common;
```
zzzzzzzzott, line 1: reduce comments in this file

source/GameInterface/Services/Issues/Generic/Migrated/VillageNeedsTools/VillageNeedsToolsQuestType.cs
```diff
@@ -0,0 +1,88 @@
+using Common.Messaging;
```
zzzzzzzzott: Migrated folder?

source/GameInterface/Services/Issues/Handlers/VillageNeedsToolsIssueHandler.cs
```diff
@@ -0,0 +1,303 @@
+using Common;
```
zzzzzzzzott: These VillageNeedsToolsIssue handlers/interfaces/messages should be their own pr

source/GameInterface/Services/Issues/Patches/DisableAllIssueBehaviorsExceptAllowlist.cs
```diff
@@ -0,0 +1,219 @@
...
+{
+    private static readonly ILogger Logger = LogManager.GetLogger<DisableAllIssueBehaviorsExceptAllowlist>();
+
+    // These targets aren't known at compile time, so they can't use [HarmonyPatch(typeof(...))] directly.
+    private static readonly Harmony DynamicHarmony = new Harmony("GameInterface.Issues.DisableAllIssueBehaviorsExceptAllowlist");
+
+    private static readonly string[] ScannedAssemblyNames = { "TaleWorlds.CampaignSystem", "SandBox" };
+
+    /// <summary>Issue-type behaviors vetted and allowed to actually generate/offer their issue in co-op.</summary>
+    internal static readonly HashSet<Type> Allowlist = new HashSet<Type>
```
zzzzzzzzott: So were allowing everything even though they are not implemented?

source/GameInterface/Services/Issues/Patches/VillageNeedsToolsAlternativeSolutionCompletionPatches.cs
```diff
@@ -0,0 +1,46 @@
+using GameInterface.Services.Issues.Generic;
```
zzzzzzzzott: For future reference, please make folders for each quest under this ../patches/ path to better organize it. For now this should be removed though, its out of scope for this pr

source/GameInterface/Services/Issues/Generic/Compatibility/IsExternalInitPolyfill.cs
```diff
@@ -0,0 +1,7 @@
+// Polyfill so C# 9 `record`/`init` compiles against netstandard2.0, which has no IsExternalInit type.
+namespace System.Runtime.CompilerServices
+{
+    internal static class IsExternalInit
+    {
```
zzzzzzzzott, line 4: whats this

source/GameInterface/Services/Issues/Generic/AcceptMirror/Unit.cs
```diff
@@ -0,0 +1,6 @@
+namespace GameInterface.Services.Issues.Generic.AcceptMirror;
+
+public readonly struct Unit
+{
+    public static readonly Unit Value = default;
```
zzzzzzzzott: whats this

source/GameInterface/GameInterface.cs
```diff
@@ -16,15 +16,11 @@ public interface IGameInterface : IDisposable
 public class GameInterface : IGameInterface
```
zzzzzzzzott, line 16: I think the changes to this file are unneeded

zzzzzzzzott (changes requested):

garrettluskey:
Thanks for splitting the quest work down from the earlier broad PRs. This comment is only about PR organization and review strategy, not the implementation.

Please keep this PR in draft and narrow it to the minimum genuinely shared quest infrastructure. Quest-specific handlers, messages, patches, and other changes that are not required by the shared framework should move to their respective one-quest PRs. That will give reviewers a stable foundation they can evaluate once instead of reviewing framework and quest behavior together.

For the remaining work, please use this sequence:

1. Finish the scope and current review cycle on this foundation PR.
2. Use one quest per PR, as requested on #2722.
3. Keep quest PRs draft while they depend on this branch, and put `Depends on #2763` prominently in each body.
4. Give each quest its own focused E2E coverage and state how the tester feature flag is preserved.
5. Complete the standard PR template, including an issue link or an explicit explanation when no issue exists.
6. After this merges, rebase and retarget each quest PR to `development` before requesting review.

#2770 is much closer to the review shape we need. Please use that as the model rather than opening another multi-quest batch.

source/Common/Messaging/IServerToClientCommand.cs
```diff
@@ -0,0 +1,5 @@
+namespace Common.Messaging;
```
zzzzzzzzott, line 1: not needed right?

source/Common/PacketHandlers/MessagePacketHandler.cs
```diff
@@ -58,6 +59,13 @@ public virtual void PublishEvent(NetPeer peer, IMessage message)
         if (message is null)
             throw new ArgumentNullException(nameof(message));
 
+        if (ModInformation.IsServer && message is IServerToClientCommand)
```
zzzzzzzzott, line 62: Unneeded bcs `IServerToClientCommand` isnt needed

source/GameInterface/Services/MobileParties/Patches/Disable/DisableLordsNeedsTutorIssueBehavior.cs
```diff
@@ -1,11 +0,0 @@
-﻿using HarmonyLib;
```
zzzzzzzzott, line 1: You deleted a bunch of these patches that disable what seem to be quests. I'm not sure, but is this needed? Does the DisableAllIssueBehaviorsExceptAllowlist.cs handle this now?

source/Common/Messaging/IServerToClientCommand.cs
Thaipho: its needed.  it's the marker for the "server-to-client-only" message family, enforced centrally in MessagePacketHandler.PublishEvent (line 62) rather than one-off if (ModInformation.IsServer) return; guards scattered per-handler.

What it prevents: NetworkIssueRemoved implements this interface. The proper path to finalize an issue goes through Handle_RequestIssueRemoved, which validates the completion condition against the requester's real party before broadcasting NetworkIssueRemoved to everyone (IssueFinalizationHandler.cs:95-111). But the receiving side, Handle_NetworkIssueRemoved (line 115), does zero validation, it just finalizes whatever ownerId/reason is in the payload, no ownership check, nothing. That's fine when the message only ever arrives from the server. Without this guard, though, nothing stops a client from constructing a NetworkIssueRemoved for any hero's issue with reason: QuestSuccess and sending it straight to the server, completely skipping the completion-condition check instant free completion of anyone's quest. The IServerToClientCommand check is what makes the server reject that packet instead of processing it.

source/Common/PacketHandlers/MessagePacketHandler.cs
Thaipho: This line is the actual enforcement point. Without it, the interface in the other file would just be an inert marker nobody checks. PublishEvent is the shared entry point for every incoming network message, so this is where we get one central place to reject any server-to-client-only message (like NetworkIssueRemoved) if it ever arrives at the server instead of from it, instead of needing a manual if (ModInformation.IsServer) return; guard added to every individual handler by hand.

Same example as the other thread: Handle_NetworkIssueRemoved does no validation on receipt (no ownership check, no re-derivation of the completion condition) it trusts that if this message arrived, it came from the server, which already validated before broadcasting. This line is what actually keeps that trust assumption true. Remove it and a client can send NetworkIssueRemoved directly to the server and finalize any hero's issue with a fabricated QuestSuccess reason, with nothing to stop it. While building this out i kept running into ordering bugs where a handler ran on the wrong side or in the wrong order (an accept/reject race, ownership registration racing between save-transfer and live-sync). This guard catches that whole class of bug structurally, instead of depending on every handler author remembering to add the check by hand.

source/GameInterface/Services/MobileParties/Patches/Disable/DisableLordsNeedsTutorIssueBehavior.cs
Thaipho: i did delete some, and there were some i forgot to delete,18 more still existing, DisableLordNeedsHorsesIssueBehavior, DisableArmyNeedsSuppliesIssueBehavior, DisableBettingFraudIssueBehavior, and 15 others across Villages/Towns/Settlements/etc. They're all the same shape: a bare Harmony patch no-opping RegisterEvents for one specific vanilla issue type, these were Garrett's own hand-written patches from 2023, one per type he happened to be dealing with at the time. DisableAllIssueBehaviorsExceptAllowlist supersedes that approach entirely rather than just consolidating it: instead of needing a new hand-written file every time a quest type gets added, it uses reflection to find every vanilla/SandBox issue behavior and disables anything not on the allowlist.

i removed the ones i missed and updated the PR

zzzzzzzzott (dismissed):

zzzzzzzzott:
Please let @garrettluskey review before merging

Thaipho:
commands for testing or unstucking
[coop.debug.issues.give <heroId> <questTypeKey>]
Force-gives a specific quest type to a hero and immediately starts it as a live, testable quest. Server-only. Rejects if the hero already has an issue. Now soft-lock-proof: if the quest object itself fails to construct (needs deeper settlement context than expected), the command automatically rolls the issue back via IssueManager.DeactivateIssue so the hero is left clean and immediately retryable, instead of getting permanently stuck.
Example: coop.debug.issues.give lord_2_7 VillageNeedsTools

[coop.debug.issues.complete <heroId> [outcome]
Force-completes a hero's active quest. outcome is optional: success (default), cancel, fail, timeout, or betrayal.
Example: coop.debug.issues.complete lord_2_7 or coop.debug.issues.complete lord_2_7 fail

[coop.debug.issues.list_types]
Lists every valid questTypeKey

Thaipho:
if this gets approved i have the "Trade quests" ready that @garrettluskey told me to do first in another PR which i will list here. I have the rest ready as well so totaling 43 quests across the board and 14 trade quests

1.  VillageNeedsTools - Ready for PR
2.  VillageNeedsCraftingMaterials - PR 2770
3.  GangLeaderNeedsToOffloadStolenGoods - Ready for PR
4.  Smugglers - Ready for PR
5.  HeadmanNeedsGrainPrice - Ready for PR
6.  HeadmanNeedsToDeliverAHerd - Ready for PR
7.  LandLordTheArtOfTheTrade - Ready for PR
8.  ArtisanCantSellProductsAtAFairPrice - Ready for PR
9.  ArtisanOverpricedGoods - Ready for PR
10. RevenueFarming - Ready for PR
11. CaravanAmbush - Ready for PR
12. EscortMerchantCaravan - Ready for PR
13. MerchantArmyOfPoachers - Ready for PR
14. MerchantNeedsHelpWithOutlaws - Ready for PR

ShoT-UPfps (dismissed):
lgtm

source/Common/Messaging/IServerToClientCommand.cs
garrettluskey, line 1: Remove this, adds unnecessary complexity

source/Common/PacketHandlers/MessagePacketHandler.cs
garrettluskey, line 1: Remove this, adds unnecessary complexity

garrettluskey (changes requested):

source/Common/Messaging/IServerToClientCommand.cs
Thaipho: This is what you asked for in your earlier review comment on this handler:

> "NetworkVillageIssueQuestAccepted is a server-to-client broadcast, but this handler also runs on the server and has no source guard. A client can send this ICommand, choose OwnerControllerId, and bypass request arbitration entirely. The same omission exists for alternative-accepted, accept-rejected, and removed broadcasts below. Reject these messages on the server (preferably enforce message direction centrally) and add forged-broadcast tests."

`IServerToClientCommand` + the check in `MessagePacketHandler.PublishEvent` is that centralized enforcement. Do you want me to revert it?

source/GameInterface/Services/Issues/Handlers/GenericAcceptMirrorHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+
+            if (!IssueGenerationRegistry.TryGetGeneration(owner, out var currentGeneration) || currentGeneration != requestedGeneration)
+            {
+                Logger.Error("Rejecting {Message} for a stale/superseded issue generation for owner {Owner}",
+                    nameof(RequestGenericIssueAcceptQuest), ownerId);
+                network.Send(requester, new NetworkGenericIssueAcceptRejected(ownerId));
+                return;
+            }
+
+            if (GenericAcceptMirrorIssueTypes.IsQuestSolutionMirrorEligible(owner.Issue) && owner.Issue.IsOngoingWithoutQuest &&
```
garrettluskey: The generation check identifies an issue version, but it does not authorize this accept. `NetworkSimpleIssueCreated` broadcasts that generation to every peer, so any registered client can submit `RequestGenericIssueAcceptQuest(ownerId, generation)` for any ongoing eligible issue and win ownership without ever entering the conversation or satisfying its location/dialogue eligibility. Bind the request to short-lived server-tracked conversation state (or a server-issued capability) for the requester, exact issue generation, and solution path, then revalidate that here before starting the quest.

source/GameInterface/Services/Issues/Handlers/GenericAcceptMirrorHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+                return;
+            }
+
+            if (GenericAcceptMirrorIssueTypes.IsAlternativeSolutionMirrorEligible(owner.Issue) && owner.Issue.IsOngoingWithoutQuest &&
+                owner.Issue.IssueStayAliveConditions())
+            {
+                try
+                {
+                    ApplyValidatedSentTroops(owner, player, payload.What.SentTroops);
+                    issueInterface.MirrorAlternativeAccepted(owner, payload.What.State);
```
garrettluskey: This is still applying client-generated authoritative state on the server. `AlternativeSolutionVanillaState` contains the return/effect times, failure chance, casualty count, XP amount, and reward skill; clamping the roster does not make those client values trustworthy, and line 214 broadcasts them unchanged. Send only stable companion/troop selections, validate and reserve them against the requester’s authoritative party, run `StartIssueWithAlternativeSolution` on the server under the true-owner scope, capture the resulting state there, and broadcast that result. The descriptor API needs a server start/capture operation rather than a mirror operation over opaque client bytes for the same reason.

source/GameInterface/Services/Issues/Patches/NewIssueTypesAlternativeSolutionPatches.cs
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+[HarmonyPatch]
+internal class NewIssueTypesAlternativeSolutionCompletionPatches
+{
+    [HarmonyPatch(typeof(LordNeedsHorsesIssueBehavior), nameof(LordNeedsHorsesIssueBehavior.RegisterEvents))]
+    [HarmonyPostfix]
+    private static void LordNeedsHorsesRegisterEventsPostfix(LordNeedsHorsesIssueBehavior __instance) =>
+        CampaignEvents.HourlyTickEvent.AddNonSerializedListener(__instance, OnHourlyTick);
```
garrettluskey: Every one of the 23 `RegisterEvents` postfixes installs this same static `OnHourlyTick` callback. Harmony postfixes still run when the allowlist’s prefix skips the original method, so a campaign accumulates many identical listeners; each due issue can send repeated completion requests and the full issue collection is rescanned once per patched behavior every hour. Register one listener once from a shared coordinator (or `IssuesCampaignBehavior`) and add a test proving one due issue emits one request. That also removes the need for a new per-type Harmony edit when onboarding a quest.

source/GameInterface/Services/Issues/Interfaces/GenericAcceptMirrorIssueTypes.cs
```diff
@@ -0,0 +1,70 @@
+using System;
+using System.Collections.Generic;
+using TaleWorlds.CampaignSystem.Issues;
+
+namespace GameInterface.Services.Issues.Interfaces;
+
+internal static class GenericAcceptMirrorIssueTypes
+{
+    internal static readonly HashSet<Type> QuestSolutionMirrorEligible = new HashSet<Type>
```
garrettluskey: This creates another quest-registration system outside `QuestTypeDescriptor`. A quest can now be named here, in `SimpleIssueFactoryRegistry`, in `QuestTypeRegistry`, in the behavior allowlist, and in the per-type completion patches; drift makes a path silently inert, while overlap can drive both accept pipelines. That conflicts with #2800’s requirement that onboarding need only one descriptor registration. Put the key/factory and supported accept/completion capabilities on one injected quest definition, make all shared patches and handlers query that registry, and move the concrete definitions to their one-quest PRs.

garrettluskey (changes requested):
The generation stamp, centralized request handlers, and authenticated `NetPeer` lookup are useful foundations. I’m still requesting changes because the extension point is split across independent policy tables, while the server-authoritative contract stops at identity and roster validation: arbitrary peers can initiate accepts and alternative-solution state is still produced by the client.

```mermaid
flowchart LR
  P[Harmony lifecycle patches] --> D[QuestTypeRegistry]
  P --> A[GenericAcceptMirrorIssueTypes]
  C[Creation handler] --> F[SimpleIssueFactoryRegistry]
  H[Accept/finalize handlers] --> D
  H --> A
  H --> S[static ownership/generation state]
```

The foundation should collapse this to one injected registry of quest definitions and one set of lifecycle coordinators. A definition should own its key/factory, supported solution paths, server execution/capture strategy, and completion validator. The handlers should accept only intent plus stable selection IDs, run vanilla on the server, and broadcast the captured result. Ownership, generation, and pending-return state should be lifetime-scoped services rather than process globals. That gives each one-quest PR one registration seam and one focused test surface instead of several lists and patch files that have to stay synchronized.

source/GameInterface/Services/Issues/Handlers/IssueConversationHandler.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            if (requester == null || !playerManager.TryGetPlayer(requester, out var player))
+            {
+                Logger.Error("Rejecting {Message} from an unregistered/unknown requester for issue-giver {IssueGiver}",
+                    nameof(RequestIssueConversationOpened), issueGiverId);
+                if (requester != null) network.Send(requester, new NetworkIssueConversationDenied(issueGiverId));
+                return;
+            }
+
+            if (TryRegisterConversation(issueGiverId, player.ControllerId) &&
```
garrettluskey, line 90: This grants an acceptance capability solely because the client says it opened a conversation. Any registered peer that knows an issue-giver ID can send `RequestIssueConversationOpened`, pass these issue-state checks, and receive a valid generation without being near the giver or in a server-observed interaction. Establish this authorization from authoritative server state instead of minting it from this request.

source/GameInterface/Services/Issues/Generic/Dispatch/GenericQuestTypeDispatchPatches.cs
```diff
@@ -0,0 +1,92 @@
...
+        descriptor?.OnGenuineCreation?.Invoke(issue);
+    }
+}
+
+[HarmonyPatch(typeof(IssueManager))]
+internal class GenericQuestTypeQuestSolutionAcceptTriggerPatch
+{
+    [HarmonyPriority(Priority.First)]
+    [HarmonyPatch(nameof(IssueManager.StartIssueQuest))]
+    [HarmonyPostfix]
```
garrettluskey, line 34: This postfix runs after the initiating client has already committed `StartIssueQuest`. If the server rejects the request or another peer wins the race, that client keeps a locally created quest and there is no rollback/resync path. Intercept acceptance before the persistent mutation, send intent to the server, and create or mirror the quest only after the server accepts it.

source/GameInterface/Services/Issues/Handlers/GenericQuestTypeAcceptHandler.cs
```diff
@@ -0,0 +1,463 @@
...
+            {
+                if (!ApplyValidatedSentTroops(owner, player, payload.What.SentTroops))
+                {
+                    Logger.Error("Rejecting {Message} for owner {Owner} - requester's validated troop roster is empty",
+                        nameof(RequestQuestTypeAcceptAlternative), ownerId);
+                    network.Send(requester, new NetworkQuestTypeAcceptRejected(ownerId, isAlternative: true));
+                    return;
+                }
+
+                state = AlternativeSolutionStartRunner.StartOnServer(owner, player);
```
garrettluskey: This starts the alternative solution after only clamping the claimed roster. Vanilla separately checks `AlternativeSolutionCondition`, companion availability, required troop health/count/type, applies `AlternativeSolutionStartConsequence`, and removes the selected troops from the party. Those authoritative steps are absent here, so a forged request can start without requirements or costs and regular troops remain in the party while also being recorded for return. Reproduce the complete transaction on the server before broadcasting it.

source/GameInterface/Services/Issues/Handlers/IssueFinalizationHandler.cs
```diff
@@ -0,0 +1,221 @@
...
+                }
+
+                if (reason == IssueFinalizeReason.QuestTimeout && !quest.QuestDueTime.IsPast)
+                {
+                    Logger.Error("Rejecting {Message} claiming QuestTimeout for owner {Owner} - due time has not passed",
+                        nameof(RequestIssueRemoved), ownerId);
+                    return;
+                }
+            }
+            else if (quest != null)
```
garrettluskey: This guard only rejects `IssueOnly` when an ongoing `IssueQuest` exists. An issue being solved through the alternative path has no `IssueQuest`, so its owner can send `RequestIssueRemoved(..., IssueOnly, generation)` and fall through to `IssueFinalized`, clearing the issue and sent troops without authoritative completion. Explicitly reject this reason while `IsSolvingWithAlternative` and allow only the server-owned completion path.

source/GameInterface/Services/Issues/Patches/IssueFinalizedPatches.cs
```diff
@@ -0,0 +1,67 @@
...
+            ContainerProvider.TryResolve<IIssueConversationTracker>(out var conversationTracker) &&
+            objectManager.TryGetIdWithLogging(owner, out var ownerId))
+        {
+            conversationTracker.Clear(ownerId);
+        }
+
+        if (CallOriginalPolicy.IsOriginalAllowed()) return;
+        if (!DisableAllIssueBehaviorsExceptAllowlist.IsAllowlisted(__instance)) return;
+
+        MessageBroker.Instance.Publish(__instance, new IssueFinalizedTriggered(owner, reason));
```
garrettluskey, line 81: This notification is emitted from a postfix, after the initiating client has already finalized its local issue. Every server rejection above simply returns, so a rejected success/cancel/failure leaves that client permanently divergent. Intercept finalization before the persistent mutation or add a rejection path that restores the authoritative issue and quest state.

garrettluskey (changes requested):
The centralized message-direction guard, issue generations, ownership registry, and server-side completion runners are useful foundations. I’m still requesting changes because acceptance and finalization can be forged or leave the initiating client permanently divergent.

source/GameInterface/Services/Issues/Handlers/GenericQuestTypeAcceptHandler.cs
```diff
@@ -0,0 +1,529 @@
...
+            sentTroops.Clear();
+        }
+    }
+
+    private void RollbackFailedAlternativeAcceptStart(Hero owner, string controllerId)
+    {
+        if (owner?.Issue == null) return;
+
+        using (new IssueFinalizeAuthorityGuard())
+        using (new AllowedThread())
```
garrettluskey, line 487: `RollbackFailedAlternativeAcceptStart` is a server path, but this runs `CompleteIssueWithCancel` under `AllowedThread`. after a partial alternative start it removes the server's issue without replicating that cancellation, while the requester receives `NetworkQuestTypeAcceptRejected` and keeps the original issue. restore the server transaction without finalizing the issue, or run a real authoritative cancellation with patches live and broadcast it.

source/GameInterface/Services/Issues/Handlers/IssueFinalizationHandler.cs
```diff
@@ -0,0 +1,339 @@
...
+            objectManager.TryGetObjectWithLogging<MobileParty>(player.MobilePartyId, out ownerParty);
+        }
+
+        objectManager.TryGetObjectWithLogging<Hero>(player.HeroId, out var truePlayerHero);
+
+        try
+        {
+            using (new MainHeroSubstitutionScope(truePlayerHero ?? owner, ownerParty))
+            {
+                IssueFinalizationSupport.FinalizeMirror(owner, reason);
```
garrettluskey: `FinalizeMirror` enters `AllowedThread`, so this server path runs `CompleteQuestWith*` with replication and registry patches disabled, then sends only `NetworkIssueRemoved`. clients rerun the quest consequences independently, including RNG and object/roster side effects. run the authoritative completion with patches live here and keep the `AllowedThread` mirror only in `Handle_NetworkIssueRemoved`.

source/GameInterface/Services/Issues/Handlers/IssueFinalizationHandler.cs
```diff
@@ -0,0 +1,339 @@
...
+                or IssueFinalizeReason.QuestTimeout or IssueFinalizeReason.QuestBetrayal)
+            {
+                if (quest == null)
+                {
+                    Logger.Error("Rejecting {Message} claiming {Reason} for owner {Owner} - no ongoing quest to finalize",
+                        nameof(RequestIssueRemoved), reason, ownerId);
+                    return;
+                }
+
+                if (reason is IssueFinalizeReason.QuestTimeout or IssueFinalizeReason.QuestFail && !quest.QuestDueTime.IsPast)
```
garrettluskey: `QuestFail` isnt a timeout, quests can call `CompleteQuestWithFail` before `QuestDueTime`. this rejects that normal failure after the client has already finalized locally, so the server keeps the quest and no rejection restores it. only gate `QuestTimeout` on the due time, and add a quest-specific failure validator if early failures need revalidation.

garrettluskey (changes requested):
The server-side accept runners and exact generation/ownership checks are much closer to the right shape. changes are still needed because failed alternative starts can delete only the server's issue, quest finalization runs under `AllowedThread`, and normal early quest failures are rejected after the client has already finalized.

garrettluskey (dismissed):
the latest blockers are fixed: failed alternative starts restore the ongoing state instead of finalizing, server finalization keeps replication patches live, and early fail/cancel paths require quest-specific validators. no new blockers found in the changed lifecycle paths.

Thaipho:
just want some clarification based on you last comment. does that mean you approved the whole PR and the stack or just that those problems on your previous review were fixed and you will do a in depth review later? mostly asking to see if i should start work on individual quest PR's. this last update just put the handler on to the latest branch

garrettluskey (approved):

PR #2765, opened 2026-08-06
Fix forming party in companion rescue

### LINKED ISSUE #2680: [COMPANIONS] Companion remains a prisoner after choosing to continue his party
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

The companion cannot be released, talked to, located on the map, or added to the reporter’s party. The reporter was not at war with anyone, and the issue persisted for two real-life days despite leaving and rejoining and searching dungeons and taverns.

## Steps to reproduce

1. Free a companion from captors after a battle.
2. In the prisoner dialogue, select the option to continue his party.
3. Complete the dialogue; no party forms, and the companion remains listed as a prisoner.

## Additional information

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1534171134512333001/1534171134512333001

## Images and videos

[](<https://cdn.discordapp.com/attachments/1534171134512333001/1534171135652925550/image.png?ex=6aa497f6&is=6aa34676&hm=441a2b579b80e4ec815efd958c6b9b0876e91d7341fbd913398a5f380699cb58&>) ([Discord message](https://discord.com/channels/709516043332354119/1534171134512333001/1534171134512333001))

## Logs

Client log: Not provided

Server log: Not provided

## Additional report details

- A follow-up says a possible fix may include an extra daily check to free companions that are in a stuck state. ([Discord message](https://discord.com/channels/709516043332354119/1534171134512333001/1534817652030312560))

jordanbrymora:
## PR Checklist

## PR Type

## What is the new behavior?

Resolves #2680

## Bot Changelog Entry

Fixed companions becoming permanently stuck as prisoners when forming a party after rescuing in a battle.

### DIFF
--- a/source/GameInterface/Services/Companions/Patches/CompanionRolesPatches.cs
+++ b/source/GameInterface/Services/Companions/Patches/CompanionRolesPatches.cs
@@ -115,7 +115,9 @@ internal class CompanionRolesPatches
         // Call original if we call this function
         if (CallOriginalPolicy.IsOriginalAllowed()) return true;
 
-        __instance._partyCreatedAfterRescueForCompanion = false;
+        // Skip CompanionRescued when a party was created
+        if (__instance._partyCreatedAfterRescueForCompanion) return false;
+
         if (Hero.OneToOneConversationHero.IsPrisoner)
         {
             var message = new CompanionRescued(Hero.OneToOneConversationHero);
--- a/source/GameInterface/Services/Companions/Patches/CompanionRolesPatches.cs
+++ b/source/GameInterface/Services/Companions/Patches/CompanionRolesPatches.cs
@@ -97,13 +97,12 @@ internal class CompanionRolesPatches
 
         if (fromCancel) return false;
 
+        __instance._partyCreatedAfterRescueForCompanion = true;
+
         var message = new PartyScreenClosedFromRescuing(
-            leftOwnerParty,
             leftMemberRoster,
             leftPrisonRoster,
-            rightOwnerParty,
-            rightMemberRoster,
-            rightPrisonRoster);
+            rightOwnerParty);
         MessageBroker.Instance.Publish(__instance, message);
 
         return false;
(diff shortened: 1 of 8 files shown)


source/GameInterface/Services/Companions/Patches/CompanionRolesPatches.cs
```diff
@@ -97,13 +97,12 @@ public static bool PartyScreenClosedPrefix(ref CompanionRolesCampaignBehavior __
 
         if (fromCancel) return false;
 
+        __instance._partyCreatedAfterRescueForCompanion = true;
```
ShoT-UPfps, line 100: could we skip `CompanionRescued` when `_partyCreatedAfterRescueForCompanion` was true? otherwise the close-window consequence can send a second release before the server reply arrives, and that second `EndCaptivityAction` makes the new party leader fugitive and destroys the party.

ShoT-UPfps (changes requested):
cmts

source/GameInterface/Services/Companions/Patches/CompanionRolesPatches.cs
```diff
@@ -116,7 +115,9 @@ public static bool EndRescueCompanionPrefix(ref CompanionRolesCampaignBehavior _
         // Call original if we call this function
         if (CallOriginalPolicy.IsOriginalAllowed()) return true;
 
-        __instance._partyCreatedAfterRescueForCompanion = false;
+        // Skip CompanionRescued when a party was created
+        if (__instance._partyCreatedAfterRescueForCompanion) return false;
```
ShoT-UPfps: we still need to reset `_partyCreatedAfterRescueForCompanion` before returning here, otherwise it stays true and the next companion who chooses to go off on their own will skip `CompanionRescued` and remain a prisoner

ShoT-UPfps (dismissed):
cmt

garrettluskey (approved):
Verified the new PartyScreenClosed against decompiled vanilla: same steps, with the server-safe substitutions the dedicated host needs (rightOwnerParty/LeaderHero/LeaderHero.Clan stand in for MainParty/MainHero/PlayerClan), and the CompanionOf != null predicate is the server-side equivalent of IsPlayerCompanion, which reads Hero.MainHero. Sending packed roster data instead of registry ids fixes the actual bug since those party-screen rosters are never registered, so the old TryGetIdWithLogging path failed and the whole transfer silently never happened. The _partyCreatedAfterRescueForCompanion flag now gates end_rescue_companion so the CompanionRescued path doesn't double-release, and the daily fugitive sweep only runs on the server because RegisterEvents is server-gated, which is a clean safety net for saves that already have stuck companions. RunSafe on the apply handlers matches the threading guidance.

PR #2773, opened 2026-08-06
Apply hero damage from battles on server

### LINKED ISSUE #2689: [COMPANIONS] Companion stops healing after being injured
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

After the last fight injured a companion, the companion did not heal beyond 10% health.

## Steps to reproduce

Not provided

## Additional information

Environment: dedicated server. Attached evidence: coop-server-20260804-145529.log.

## Discord source

https://discord.com/channels/709516043332354119/1534226672063352993/1534226672063352993

## Logs

Client log: Not provided

Server log: [Download coop-server-20260804-145529.log](<https://cdn.discordapp.com/attachments/1534226672063352993/1534226672679653587/coop-server-20260804-145529.log?ex=6aa4cbaf&is=6aa37a2f&hm=b5b824956cdd2fe2ef5878a487d62a36dd82d994d6c5f7f29d10f74c0df30db9&>)

jordanbrymora:
## PR Checklist

## PR Type

Hero damage in battles isn't saved on the server. This leaves companions (and other heroes) in player parties wounded only client side. When the tick runs to heal these heroes on the server they are already at full health.

## What is the new behavior?

Resolves #2689

## Bot Changelog Entry

Fixed companions not healing after battles.


source/GameInterface/Services/Heroes/Extensions/HeroExtensions.cs
```diff
@@ -51,4 +52,27 @@ public static bool IsControlledByThisInstance(this Hero hero)
...
+
+        // Hero is this player's character, don't need to check party
+        if (hero.IsControlledByThisInstance()) return true;
+
+        // Don't manage health for other players in this player's party (for future)
+        if (hero.IsPlayerHero()) return false;
+
+        var party = hero.PartyBelongedTo;
+
+        return party != null && party.IsControlledByThisInstance();
```
ShoT-UPfps: this needs to follow current battle agent authority, not the party registry owner. `BattleAuthorityMigrator` transfers the agents to the host when a player disconnects, but the party stays registered to the old controller, so the adopted host drops companion health changes here.

ShoT-UPfps (changes requested):
cmt

garrettluskey (dismissed):
The ownership extension reads right: a client forwards HitPoints for its own hero and for non-player heroes in a party it controls, and the IsPlayerHero early-out keeps one client from owning another player's hero even when they share a party. PartyBelongedTo gives exactly one owning client per companion, so there's no double-forward, and the server guard at the top of the setter patch means server behavior is untouched. That plugs the gap where companion battle wounds only existed client side and the server's heal tick saw them at full health.

source/GameInterface/Services/Heroes/Extensions/HeroExtensions.cs
jordanbrymora: I'm not sure how to implement this. Do I have to do something like get the agent from the hero, get the controller id from the agent and compare it to the client this is called on?

ShoT-UPfps: the network owner is `CoopAgentInfo.CurrentAuthority`, so compare that with ControllerId in the mission layer and send the health request from there instead of using the party registry owner.

ShoT-UPfps (approved):
lgtm

PR #2774, opened 2026-08-06
Map Time UI Implementation

### LINKED ISSUE #2042: [ISSUE] Show map time somewhere when in mission/location (tavern/field battle/village/etc)
We should have an overlay or text somewhere (maybe top right, next to the kill feed?) that is visible when in mission/location and says: "Map Time: Paused/Normal/Fast Forward"

codebydean:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Players cannot see the current campaign map time control state while they are inside a mission or location.

## What is the new behavior?

Resolves #2042

A map time overlay is displayed while the player is inside a location, battle, siege, or tournament mission. It displays one of the following synchronized campaign map time-control states:

- Map Time: Paused
- Map Time: Normal
- Map Time: Fast Forward

The overlay initializes from the current time control mode and updates when the server broadcasts a new mode.

## Bot Changelog Entry

- Added an in-mission overlay showing whether campaign map time is paused, normal, or fast-forwarded.

## Other information

Added ViewModel tests covering all supported map time-control states and state changes.

Also updated the tournament mission behavior-order test to include the map-time overlay.
sprite
battlemapview


source/GameInterface/Services/Time/UI/MissionMapTimeView.cs
```diff
@@ -0,0 +1,78 @@
...
+        movie = null;
+        gauntletLayer = null;
+        dataSource = null;
+        
+        base.OnMissionScreenFinalize();
+    }
+
+    private void HandleTimeControlModeChanged(MessagePayload<NetworkChangeTimeControlMode> payload)
+    {
+        dataSource?.SetTimeControlMode(payload.What.NewControlMode);
```
ShoT-UPfps: this runs on the network poll thread, and `SetTimeControlMode` synchronously updates the Gauntlet binding. please put this in `GameThread.RunSafe` otherwise the poll thread can mutate the widget while the game thread is rendering or releasing it

source/GameInterface/Services/Time/UI/MissionMapTimeView.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        base.OnMissionScreenInitialize();
+        
+        dataSource = new MissionMapTimeVM(timeControlInterface.GetTimeControl());
+        gauntletLayer = new GauntletLayer("MissionMapTime", LayerOrder);
+        movie = gauntletLayer.LoadMovie("MissionMapTime", dataSource);
+        
+        MissionScreen.AddLayer(gauntletLayer);
+        
+        messageBroker.Subscribe<NetworkChangeTimeControlMode>(HandleTimeControlModeChanged);
```
ShoT-UPfps: could we subscribe before taking the initial snapshot? if `NetworkChangeTimeControlMode` arrives after line 38 but before this subscription, `TimeHandler` applies it but this view misses the only notification and keeps showing the old mode until another change

ShoT-UPfps (changes requested):
cmts

codebydean:
> cmts

Applied some fixes, re-check

zzzzzzzzott:
Looks good to me but I think you should add a coop option to disable this in case a player does not want it
(If you do make sure to properly edit the correct xml file then run the generator)

codebydean:
> Looks good to me but I think you should add a coop option to disable this in case a player does not want it (If you do make sure to properly edit the correct xml file then run the generator)

I can definitely check it out. I'll do it right now

codebydean:
@zzzzzzzzott Added that extra tab. Altered some of your tests for the kill feed, as you can see in the commits, as they were testing a "single" tab. 

All tests pass, all new UI is available, coop-options.json is now updated, and the actual options are being saved @ShoT-UPfps

ShoT-UPfps:
@codebydean can you share screenshots of the new config? and in which config file is this config saved?

codebydean:
> @codebydean can you share screenshots of the new config? and in which config file is this config saved? @ShoT-UPfps 

It is saved in the coop_options config file. Attached below are the screenshots of the actual UI.

maptimeuicoopoptions
maptimeuicoopoptions2

codebydean:
image

zzzzzzzzott:
> image

Is there a reason kill feed color does not appear?

codebydean:
> > image
> 
> Is there a reason kill feed color does not appear?

The image you quoted is not available, but if you mean the tab where only Map Time is shown, yes, the reason is that it would be silly to have the kill feed sliders appearing in the Map Time tab, which is something totally different, basically.

zzzzzzzzott:
> > > image
> > 
> > 
> > Is there a reason kill feed color does not appear?
> 
> The image you quoted is not available, but if you mean the tab where only Map Time is shown, yes, the reason is that it would be silly to have the kill feed sliders appearing in the Map Time tab, which is something totally different, basically.

?

ShoT-UPfps:
> > > image
> > 
> > 
> > Is there a reason kill feed color does not appear?
> 
> The image you quoted is not available, but if you mean the tab where only Map Time is shown, yes, the reason is that it would be silly to have the kill feed sliders appearing in the Map Time tab, which is something totally different, basically.

I think he's talking about the settings json file

codebydean:
> > > > image
> > > 
> > > 
> > > Is there a reason kill feed color does not appear?
> > 
> > 
> > The image you quoted is not available, but if you mean the tab where only Map Time is shown, yes, the reason is that it would be silly to have the kill feed sliders appearing in the Map Time tab, which is something totally different, basically.
> 
> I think he's talking about the settings json file

Ahhh apologies then.

image

As you can see in the new screenshot, the Kill Feed Tab is now shown. This happens because the existing Coop Options implementation only saves the selected Tab. Before my addition of `Map Time`, Kill Feed was the only tab. Applying `Map Time` now writes only `MapTimeTab` the `KillFeedTab` is added after applying the kill-feed tab (After clicking apply).

The default of kill-feed still loads to memory, but from what I saw is that is is not persisted to `coop_options.json` until the tab is applied.

This is how the current logic works; I only implemented Map Time on top of it. 

@ShoT-UPfps @zzzzzzzzott

garrettluskey (approved):

ShoT-UPfps (approved):
lgtm

PR #2781, opened 2026-08-06
Chat

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

No chat

## What is the new behavior?

Resolves #

Hello Chat! Welcome to the stream!

## Bot Changelog Entry

Added a chat window.


source/GameInterface/Services/Chat/ChatPlayerName.cs
```diff
@@ -0,0 +1,22 @@
+using GameInterface.Services.ObjectManager;
+using GameInterface.Services.Players.Data;
+using TaleWorlds.CampaignSystem;
+
+namespace GameInterface.Services.Chat;
+
+/// <summary>Resolves the campaign display name for a registered player.</summary>
+public static class ChatPlayerName
```
ShoT-UPfps: please make this an injected transient instead of a static helper so `ServerChatHandler` and `ChatService` can fake the name resolver in tests.

source/GameInterface/Services/Chat/ChatVM.cs
```diff
@@ -0,0 +1,305 @@
...
+
+    private ChatChannelVM selectedChannel;
+    private string writtenText = string.Empty;
+    private string transcriptText = string.Empty;
+    private bool isOpen;
+    private int unreadMessageCount;
+
+    public ChatVM(Action<NetworkSendChatMessage> send, Func<string> getLocalControllerId)
+    {
+        this.send = send ?? throw new ArgumentNullException(nameof(send));
```
ShoT-UPfps: please use the statement form for these null checks, `if (send == null) throw new ArgumentNullException(nameof(send));`, instead of `?? throw`. same for `getLocalControllerId`, `ChatOverlay`, and `ChatChannelVM`.

source/Coop.Core/Server/Services/Chat/Handlers/ServerChatHandler.cs
```diff
@@ -0,0 +1,184 @@
...
+            ChatChannel.Direct,
+            sender.ControllerId,
+            ChatPlayerName.Resolve(objectManager, sender),
+            recipient.ControllerId,
+            ChatPlayerName.Resolve(objectManager, recipient),
+            text);
+
+        network.Send(senderPeer, message);
+        if (!ReferenceEquals(senderPeer, recipientPeer))
+            network.Send(recipientPeer, message);
```
ShoT-UPfps: these chat sends need to use `SendImmediate` because `CoopServer.Send` goes through `ConnectionMessageQueue`. a reconnecting peer in `Dropping` loses this message because chat isnt in the transfer save, while the sender still gets the echo at line 135.

source/GameInterface/Services/Chat/ChatOverlay.cs
```diff
@@ -0,0 +1,178 @@
...
+        bool mouseButtonPressed,
+        bool inputHovered)
+    {
+        return inputFocused && mouseButtonPressed && !inputHovered;
+    }
+
+    internal static void SetPassiveInputRestrictions(InputRestrictions inputRestrictions)
+    {
+        inputRestrictions.SetInputRestrictions(
+            isMouseVisible: false,
```
ShoT-UPfps, line 222: we need a keyboard/controller open path here. vanilla `MissionScreen` keeps mouse visibility off during normal play, so the ribbon is not a usable open path in combat.

source/GameInterface/Services/Chat/ChatOverlay.cs
```diff
@@ -0,0 +1,178 @@
...
+        else if (!isInputFocused && leftMousePressed && inputHovered)
+        {
+            FocusInput();
+        }
+        else if (ShouldReleaseInputFocus(isInputFocused, mouseButtonPressed, inputHovered))
+        {
+            ReleaseInputFocus();
+        }
+
+        if (!isInputFocused) return;
```
ShoT-UPfps, line 91: `Esc` only runs below this return. after any click outside the input releases focus at line 64, `Esc` reaches the map/mission and the chat window stays open.

source/GameInterface/Services/Chat/ChatService.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    public void Dispose()
+    {
+        overlay.Dispose();
+    }
+
+    private void RefreshParticipants()
+    {
+        var participants = new List<(string ControllerId, string Name)>();
+        foreach (var player in playerManager.Players)
```
ShoT-UPfps: we need this list to use live session membership. `Players` includes disconnected registrations, so offline characters stay selectable and every send to one only comes back as not connected.

source/Coop/CoopMod.cs
```diff
@@ -488,6 +489,9 @@ protected override void OnGameStart(Game game, IGameStarter gameStarterObject)
             if (ContainerProvider.TryResolve<IGameInterface>(out var gameInterface))
                 gameInterface.PatchGameStarted();
 
+            if (!isServer && ContainerProvider.TryResolve<IChatService>(out var chatService))
```
ShoT-UPfps: use `ModInformation.IsClient` instead of `!isServer`.

ShoT-UPfps (dismissed):

ShoT-UPfps:
test failing appears related to these changes

garrettluskey (approved):
This is well put together. Sender identity comes from the originating peer on the server instead of being self-reported, text is length-capped and control-char stripped before routing, direct messages validate the recipient and echo to both ends with a peer-dedupe on the global path, and system rejections come back on the channel the client expects. The receive path marshals through RunSafe before touching the VM, the overlay's focus handling guards against stealing input from higher-order layers and other editable widgets, and disposal tears down the movie, layer, and subscriptions. Test coverage across the server handler, VM, overlay logic, and the movie is thorough, and the DI wiring (explicit resolver registration, IChatService through the game-abstraction scan, client-only Initialize in CoopMod) lines up.

PR #2795, opened 2026-08-06
Add client join-baseline diagnostics

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update
- [x] Other... Please describe:

Clients that reject a complete campaign baseline can remain at Finishing synchronization while their logs do not identify the invalid party state or capture the player-party state around player switching.

## What is the new behavior?

Client logs now identify each distinct join-baseline rejection once, including the affected party or dependency, and record bounded player-party state before and after player switching so future stalls can be diagnosed without exposing save contents.

## Bot Changelog Entry

- Added clearer client logs for diagnosing stalls at Finishing synchronization.

### DIFF
--- a/source/GameInterface.Tests/Services/MobileParties/MobilePartyBehaviorSnapshotTests.cs
+++ b/source/GameInterface.Tests/Services/MobileParties/MobilePartyBehaviorSnapshotTests.cs
@@ -3,6 +3,7 @@ using GameInterface.Services.MobileParties.Data;
 using GameInterface.Services.ObjectManager;
 using GameInterface.Tests.Services.SiegeEvents;
 using Moq;
+using System;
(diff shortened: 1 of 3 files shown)


garrettluskey (approved):
Diagnostics-only change and it reads well. The join-baseline failures now each say exactly which state/party/dependency failed, identical retries are deduped and the set clears on success, and the player-switch before/after log covers the registration/collection state we'd need to diagnose a 'Finishing synchronization' stall. The new unit tests match the failure strings and the TryGetObject switch on the baseline path. One heads up: only CodeFactor has run on this branch, no build/test/E2E runs have triggered yet, so worth kicking those off before merge.

PR #2805, opened 2026-08-07
[2799] Add adaptive battle movement rates

### LINKED ISSUE #2799: adaptive movement hz in battles
For performance, we can increase the movement hz when there are fewer troops/players and decrease the movement hz when there are more troops/players

Need to do extensive benchmarking to understand the optimal adaptive tuning at high troop counts (i.e. 1800) and low troop counts (i.e. 10). Need the adaptive tuning to listen to the performance of the host of the battle in realtime and generate metrics so that it can react accordingly? perhaps also the clients need to be able to request the host to adapt as well? (if there's one client that's struggling comparatively)

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Battle movement updates always run at the same 40 Hz rate, regardless of battle size or how well each player's game is keeping up. This means we're missing out on extra Hz we can get from smaller battles and improved performance when there's large battles.

## What is the new behavior?

Resolves #2799

Battles now start at 40 Hz and each player's movement update rate adjusts between 10 and 60 Hz based on their measured game performance. Bulk movement updates are limited separately for each receiving player, so one slow client does not lower the rate for everyone. Locations remain fixed at 40 Hz, and tournaments always run at 60 Hz.

## Bot Changelog Entry

- Battle CPU and Network consumption now adapts to the performance of the players involved in order to keep FPS high


source/Missions/Agents/Handlers/MovementRateController.cs
```diff
@@ -0,0 +1,862 @@
...
+        {
+            Logger.Error(ex, "Failed to publish the movement receiver-cap heartbeat");
+        }
+    }
+
+    private void Handle_ReceiverCap(MessagePayload<NetworkMovementReceiverCap> payload)
+    {
+        NetworkMovementReceiverCap message = payload.What;
+        if (string.IsNullOrEmpty(message.ControllerId) ||
+            message.ControllerId == controllerIdProvider.ControllerId)
```
garrettluskey, line 773: all the other send paths build the advertisement under `gate` and send after releasing it (`Configure`, `AdvanceFrame`, `TrySetForcedReceiverCapHz`, `Handle_PeerEntered`), but this one sends inside the lock. `SendAll` serializes one packet per route, so a slow send holds `gate` and stalls `AdvanceFrame` on the game thread and `Handle_ReceiverCap` on the poller thread until it finishes. capture the advertisement under the lock and send after, like the rest.

source/Missions/Agents/Handlers/MovementRateController.cs
```diff
@@ -0,0 +1,862 @@
...
+            advertisedReceiverCapHz = hz ?? automaticReceiverCapHz;
+            RecomputeEffectiveRate();
+            advertisement = CreateReceiverCapAdvertisement();
+            error = null;
+        }
+
+        if (advertisement.HasValue)
+            network.SendAll(advertisement.Value);
+        return true;
+    }
```
garrettluskey, line 456: Nit: plain `Timer.Dispose()` doesn't wait for a running callback, so a heartbeat that already passed the `disposed` check can still `SendAll` mid-teardown. `Dispose(WaitHandle)` would close that.

garrettluskey (changes requested):
the adaptive-rate design here is good: fixed profiles for locations and tournaments, population ceilings, stepwise recovery with hysteresis, and receiver caps that expire on a heartbeat. coverage is thorough too. requesting changes because the receiver-cap heartbeat does its network send inside the controller lock, unlike every other send path in `MovementRateController`.

for what it's worth, the branch builds clean locally (draft PRs skip CI) and the movement-rate test suites all pass.

source/Missions/Agents/Handlers/MovementRateController.cs
ShoT-UPfps: fixed, `Dispose` now drains the timer callback with `Dispose(WaitHandle)` before unsubscribing.

source/Missions/Agents/Handlers/MovementRateController.cs
ShoT-UPfps: fixed, `PublishReceiverCapHeartbeat` now builds the advertisement under `gate` and calls `SendAll` after releasing it.

garrettluskey (dismissed):
both review comments are fixed. the per-recipient cadence/state refactor also looks correct, and the new tests cover the lock release, timer drain, slow vs fast recipients, non-divisor cadence, and independent route budgets.

one thing left before approve: the PR body and changelog still describe the old design. this no longer adapts from battle population or lets a slow peer lower the shared rate. battles start at 40 Hz, local capture adapts from measured performance, and bulk sends are capped independently per recipient. please update those two sections to match.

i couldn't rerun the suites on this head because Steam removed the local game DLLs during an update, and draft CI skips them. the scoped diff passed static review and `git diff --check`; the previous head built clean and passed the affected suites.

ShoT-UPfps:
updated the new behavior and changelog to match the final design. battles start at 40 Hz, local capture adapts from measured performance, and bulk sends are capped independently per recipient.

garrettluskey (approved):
the heartbeat send now runs outside the controller lock and disposal drains the timer callback. the per-recipient cadence and updated PR description both look good.

PR #2807, opened 2026-08-07
Fix new banner items for players when loading a save game

### LINKED ISSUE #2297: Players somehow get banners when they shouldn't have banners
# For New Contributors

## Location

## Definition of Done

jordanbrymora:
## PR Checklist

## PR Type

`OnGameLoadFinishedEvent` is fired before players are registered. In `BannerCampaignBehavior`, this is means the `IsPlayerClan()` check doesn't succeed in `GiveBannersToHeroes()` for player clan heroes. As a result, when loading a save game on the server, new banners are given to all players and their clan's heroes if they don't already have them.

## What is the new behavior?

Resolves #2297 

Block this event. `DailyTickHero` already fills/replaces invalid banner item slots on heroes over time so `OnGameLoadFinishedEvent` isn't really needed.

## Bot Changelog Entry

Fixed new banners being given to player heroes when loading a save.


garrettluskey (approved):
Verified against vanilla: OnNewGameCreated calls GiveBannersToHeroes synchronously, while the load path reaches it only through the OnGameLoadFinishedEvent subscription, so the prefix/postfix flag gates exactly the load case and new games still get their banners. Clients only ever see the load path, which is what we want since banners come with the server's save. DailyTickHero keeps backfilling invalid banner slots, so nothing is lost by blocking the event.

PR #2834, opened 2026-08-08
Arena masters (perk resetting & tournament locating, not practice fights)

jordanbrymora:
## PR Checklist

## PR Type

Arena masters provide services other than just starting practice fights. They provide two key gameplay mechanics several users have complained about not having access to.
1. Resetting perks for player and companions.
2. Locating nearby tournaments.

## What is the new behavior?

This PR enables the arena masters so players can use these two mechanics. It also lays some minor ground work for a future arena implementation by saving interactions data with arena masters.

## Bot Changelog Entry

Added arena masters for resetting perks and locating tournaments. Practice fights not implemented yet.


source/GameInterface/Services/MobileParties/Interfaces/SessionInteractionsPlayerDataInterface.cs
```diff
@@ -92,6 +93,20 @@ public void SetPlayerPatrolInteraction(string playerHeroId, string settlementId,
         SetPlayerInteraction(playerHeroId, settlementId, interactionTime, InteractionsPlayerData.PlayerInteractedPatrols);
     }
 
+    public void AddMetArenaMaster(string playerHeroId, string settlementId)
+    {
+        if (!IsPlayerHeroIdValid(playerHeroId)) return;
+
+        InteractionsPlayerData.PlayerMetArenaMasters[playerHeroId]?.Add(settlementId);
```
garrettluskey, line 104: Nit: this appends without a contains check, so a re-publish for a settlement the client lost locally would store the id twice. harmless for the contains-based checks on restore, but a `?.Add` guarded by Contains would keep the list clean.

garrettluskey (dismissed):
The split from the blanket RegisterEvents disable into targeted patches reads well: practice fights stay gated, the meet/tournament-discovery conversations persist through the server's CoopSession and restore per player on join like the other interaction data, and the perk reset is faithful to vanilla (verified the gold+reset application and the RemoveACompanionFromPlayerParty loop against the decompiled original). HeroSync picking up _powerModifier covers the stat recalculation the reset triggers, and gating the companion-limit DailyTick to the clan leader's client avoids duplicate warning timers.

source/GameInterface/Services/MobileParties/Interfaces/SessionInteractionsPlayerDataInterface.cs
jordanbrymora: Added

source/GameInterface/Services/Arenas/Patches/ArenaMasterCampaignBehaviorPatches.cs
```diff
@@ -0,0 +1,184 @@
...
+using TaleWorlds.CampaignSystem.GameMenus;
+using TaleWorlds.CampaignSystem.Settlements;
+using TaleWorlds.CampaignSystem.Settlements.Locations;
+using TaleWorlds.CampaignSystem.TournamentGames;
+using TaleWorlds.Localization;
+using TaleWorlds.MountAndBlade;
+
+namespace GameInterface.Services.Arenas.Patches;
+
+[HarmonyPatch(typeof(ArenaMasterCampaignBehavior))]
```
ShoT-UPfps, line 19: this re-enables `ArenaMasterCampaignBehavior.OnGameLoadFinished` on the server. `Settlement.CurrentSettlement` dereferences `MobileParty.MainParty`, so this will throw when the dedicated server loads because it doesnt have a main party. please make `OnGameLoadFinished` client-only.

source/GameInterface/Services/Companions/Handlers/PerkResetCampaignBehaviorHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+    private void Handle_NetworkRemoveACompanionFromPlayerParty(MessagePayload<NetworkRemoveACompanionFromPlayerParty> obj)
+    {
+        var data = obj.What;
+
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Clan>(data.PlayerClanId, out var playerClan)) return;
+
+            // Re-implement vanilla to not use Clan.PlayerClan
+            int companionsCount = playerClan.Companions.Count;
```
ShoT-UPfps, line 141: recheck `playerClan.Companions.Count > playerClan.CompanionLimit` here before choosing someone. otherwise a request that becomes stale while waiting on the game thread can permanently remove another companion after the clan is already back under the limit.

source/GameInterface/Services/Companions/Patches/PerkResetCampaignBehaviorPatches.cs
```diff
@@ -1,12 +1,66 @@
...
-    [HarmonyPatch(typeof(PerkResetCampaignBehavior))]
-    internal class DisablePerkResetCampaignBehavior
+    [HarmonyPatch(nameof(PerkResetCampaignBehavior.DailyTick))]
+    [HarmonyPrefix]
+    public static bool DailyTickPrefix()
+    {
+        // Server has no main hero/player clan
+        // Manage warning time client side and broadcast result to remove companion on server
+        // If the client isn't the clan leader, don't run the tick to avoid duplication
+        return ModInformation.IsClient && Hero.MainHero.IsClanLeader;
```
ShoT-UPfps, line 32: we need to persist `_warningTime` per player because `DailyTick` now only mutates it on the client and clients cant save. reconnecting reloads the server's unchanged value and restarts the six-day timer.

source/GameInterface/Services/Companions/Patches/Disable/DisablePerkResetCampaignBehavior.cs
```diff
@@ -1,12 +1,66 @@
...
+using TaleWorlds.CampaignSystem;
 using TaleWorlds.CampaignSystem.CampaignBehaviors;
 
-namespace GameInterface.Services.Companions.Patches.Disable
+namespace GameInterface.Services.Companions.Patches.Disable;
+
+[HarmonyPatch(typeof(PerkResetCampaignBehavior))]
+internal class DisablePerkResetCampaignBehavior
+{
+    private static IEnumerable<MethodBase> TargetMethods() => new MethodBase[]
```
ShoT-UPfps: please use direct harmony patches for these methods. `TaleWorlds.CampaignSystem` is publicized in `GameInterface`, so we shouldnt use `AccessTools.Method` here.

ShoT-UPfps (changes requested):
cmt

source/GameInterface/Services/Companions/Patches/Disable/DisablePerkResetCampaignBehavior.cs
jordanbrymora: Changed to use direct patches

source/GameInterface/Services/Companions/Patches/PerkResetCampaignBehaviorPatches.cs
jordanbrymora: Added to CoopSession

source/GameInterface/Services/Companions/Handlers/PerkResetCampaignBehaviorHandler.cs
jordanbrymora: Added

source/GameInterface/Services/Arenas/Patches/ArenaMasterCampaignBehaviorPatches.cs
jordanbrymora: Added patch for `OnGameLoadFinished`

ShoT-UPfps (approved):
lgtm

PR #2837, opened 2026-08-08
Wars non auto-resolve

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Wars against player kingdoms get auto resolved as the player is not taken into account

## What is the new behavior?

Resolves #

## Bot Changelog Entry

Players will not be taken into account when AI tries to propose peace.

### DIFF
--- a/source/E2E.Tests/Services/Kingdoms/PlayerKingdomCreationFlowTests.cs
+++ b/source/E2E.Tests/Services/Kingdoms/PlayerKingdomCreationFlowTests.cs
@@ -14,6 +14,7 @@ using GameInterface.Services.GameDebug.Messages;
 using GameInterface.Services.Kingdoms;
 using GameInterface.Services.Kingdoms.Commands;
 using GameInterface.Services.Kingdoms.Data;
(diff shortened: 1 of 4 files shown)


garrettluskey (approved):
Verified both vanilla reproductions against the decompiled originals. The DetermineSupport prefix is the coop form of the ruling-clan block (Clan.PlayerClan.MapFaction == Kingdom becomes 'kingdom contains a player clan') and only intercepts the proposed-by-opponent offers, so other clans still score through the DiplomacyModel. The ShouldBeCancelled replacement matches vanilla's first four cancel conditions and deliberately drops the PlayerClan exemption and the support re-narrowing, which is what keeps the offer pending until the vote resolves. Expiry removing the decision as cancelled instead of force-resolving is the right call, the mirrored offer flips the tribute sign for the reversed direction, and the dedupe against an already-pending offer plus the server-only authoring of the mirror line up with the replication model. Tests cover accept, reject, and expiry.

PR #2838, opened 2026-08-08
More bandit party config

### LINKED ISSUE #2771: [BANDITS] Add other bandit parties to looter spawn multiplier
# For New Contributors

Not provided

## Description

Add a multiplier option for all bandit troop groups (such as forest, desert, and deserter bandits), similar to the looter multiplier behavior.

## Intended Design

Not provided

## Location

Not provided

## Related Issues

Not provided

## Additional information

Not provided

## Definition of Done

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1534838980263018599/1534838980263018599

### LINKED ISSUE #2827: [BANDITS] Add a looter and bandit party-size option instead of increasing party count
# For New Contributors

Not provided

## Description

Add a setting to increase looter and bandit party sizes (for example 40–100 men) rather than spawning more looter parties, since high multipliers like 10x/15x can corrupt saves while only smaller multipliers remain stable.

## Intended Design

Not provided

## Location

Not provided

## Related Issues

Not provided

## Requirements

Not provided

## Additional information

There is already a configuration option for the number of looter parties (`maximumLootersMultiplier`); the request is for a separate party-size configuration to avoid many tiny bands and improve early/mid-game grinding.

## Definition of Done

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1535046862338396160/1535046862338396160

jordanbrymora:
## PR Checklist

## PR Type

## What is the new behavior?

Resolves #2771
Resolves #2827

## Bot Changelog Entry

MaximumLootersMultiplier now scales number of bandit parties (not just looters).
Added a new config option to scale the size of new bandit parties.


source/GameInterface/Services/PartyBases/Patches/DefaultPartySizeLimitModelPatches.cs
```diff
@@ -59,4 +62,28 @@ public static bool GetInitialPartySizeRatioForMobilePartyPrefix(ref float __resu
...
     }
+
+    [HarmonyPatch(nameof(DefaultPartySizeLimitModel.FindAppropriateInitialRosterForMobileParty))]
+    [HarmonyPostfix]
+    public static void GetInitialPartySizeRatioForMobilePartyPostfix(TroopRoster __result, MobileParty party, PartyTemplateObject partyTemplate)
+    {
+        // Increase/decrease size of new bandit parties based on config
+        if (party.IsBandit)
+        {
+            foreach (var troopRosterElement in __result.data)
```
garrettluskey: `data` is the capacity-sized backing array, and `AddToCounts(..., removeDepleted: true)` shifts it when a scaled stack reaches zero. this skips later troop stacks on reductions, and the unused slots call `AddToCounts(null, 0)` even at the default multiplier. iterate the snapshot from `GetTroopRoster()` or walk the valid indices backwards instead.

source/GameInterface/Services/PartyBases/Patches/DefaultPartySizeLimitModelPatches.cs
```diff
@@ -59,4 +62,28 @@ public static bool GetInitialPartySizeRatioForMobilePartyPrefix(ref float __resu
...
+            {
+                var newCharacterCount = (int)(troopRosterElement.Number * ModConfigProvider.ModOptions.LooterPartySizeMultiplier);
```
garrettluskey: `looterPartySizeMultiplier` isnt bounded. a negative value makes `newCharacterCount` negative, and `AddToCounts` only removes a depleted stack when its count is exactly zero, so this leaves negative troop counts. reject or clamp values below zero before applying them.

source/GameInterface/Configuration/ModConfigProvider.cs
```diff
@@ -49,10 +49,12 @@ public readonly struct ModOptions
     [ProtoMember(14)]
     public readonly float MaximumLootersMultiplier { get; } = 1f;
     [ProtoMember(15)]
+    public readonly float LooterPartySizeMultiplier { get; } = 1f;
+    [ProtoMember(16)]
```
garrettluskey, line 53: `LordDefectionRetries`, `EnableHeroExecutions`, and `EnablePlayerClanMemberExecutions` already use tags 15-17 on `development`. dont renumber protobuf fields, put the new option at 18 so existing `NetworkLoadModConfig` tags keep their meanings.

source/GameInterface/Configuration/ModConfigProvider.cs
```diff
@@ -49,10 +49,12 @@ public readonly struct ModOptions
     [ProtoMember(14)]
     public readonly float MaximumLootersMultiplier { get; } = 1f;
     [ProtoMember(15)]
+    public readonly float LooterPartySizeMultiplier { get; } = 1f;
```
garrettluskey, line 52: please add this to `ModConfigNetworkMessageSerializationTest`, especially `0f` in `AllOptionsOff` and a non-default configured round trip. the new network option currently isnt covered by the wire-contract tests.

garrettluskey (changes requested):
the spawn-count side now covers looters and the other bandit factions, and the config is included in the shipped template. requesting changes because the party-size postfix can skip or corrupt roster entries, and the network contract needs to keep its existing tags.

source/GameInterface/Configuration/ModConfigProvider.cs
jordanbrymora: These were re-numbered to keep in the same order as the `mod-config.json`. As we add more config options this list is going to get quite large and not keeping them in the same order will make it harder to track.

source/GameInterface/Services/PartyBases/Patches/DefaultPartySizeLimitModelPatches.cs
jordanbrymora: Changed to use snapshot

source/GameInterface/Services/PartyBases/Patches/DefaultPartySizeLimitModelPatches.cs
jordanbrymora: Changed to default back to `1f` if multiplier is negative.

source/GameInterface/Configuration/ModConfigProvider.cs
jordanbrymora: Added

source/GameInterface/Configuration/ModConfigProvider.cs
garrettluskey, line 53: the field order here doesnt need to match the json. tags 15-17 are already part of the network contract, so keep those assignments and put `LooterPartySizeMultiplier` at 18, otherwise mixed-version clients deserialize different settings.

source/GameInterface/Services/Bandits/Patches/DisableBanditsCampaignBehavior.cs
```diff
@@ -19,6 +19,23 @@ internal class BanditSpawnCampaignBehaviorPatches
...
-        __result = (int)(__result * ModConfigProvider.ModOptions.MaximumLootersMultiplier);
+        // Use default if provided with a negative value
+        var multiplier = 1f;
+        if (ModConfigProvider.ModOptions.MaximumLootersMultiplier >= 0)
+            multiplier = ModConfigProvider.ModOptions.MaximumLootersMultiplier;
+
+        __result = (int)(__result * multiplier);
+    }
+
+    [HarmonyPatch(nameof(BanditSpawnCampaignBehavior._numberOfMaxBanditCountPerClanHideout), MethodType.Getter)]
```
garrettluskey, line 30: `BanditSpawnCampaignBehavior` excludes the `deserters` clan, and `DesertersCampaignBehavior` reads its cap directly from `BanditDensityModel`, so this still doesnt scale deserter party counts from #2771.

source/GameInterface/Services/PartyBases/Patches/DefaultPartySizeLimitModelPatches.cs
```diff
@@ -59,4 +62,33 @@ public static bool GetInitialPartySizeRatioForMobilePartyPrefix(ref float __resu
 
         return true;
     }
+
+    [HarmonyPatch(nameof(DefaultPartySizeLimitModel.FindAppropriateInitialRosterForMobileParty))]
```
garrettluskey, line 66: `DesertersCampaignBehavior` creates these parties with a null template and adds the routed roster afterward, so deserter party sizes never pass through this method.

deploy/mod-config.default.json
```diff
@@ -95,12 +95,20 @@
     // Stamina is stored as an integer and ticks are rounded up to ensure heroes always recover some stamina.
     "smithingStaminaRecoveryMultiplier": 0.1,
 
-    // Set a multiplier for increasing/decreasing the number of looter parties.
+    // Set a multiplier for increasing/decreasing the number of looter/bandit parties.
     // The higher the multiplier, the more looter parties can exist on the map at a time.
     // WARNING: Setting this value too high will cause performance issues.
     // The default is 1, which is equivalent to the vanilla spawn limit.
+    // Negative values will default back to 0
```
garrettluskey: both implementations fall back to 1, not 0. please make these comments and the behavior agree.

garrettluskey (changes requested):
the roster snapshot and negative-value guards fix the previous mutation problems, and the serialization test now covers the new field. deserters are still outside both scaling paths, and the existing protobuf tags are still shifted.

source/GameInterface/Configuration/ModConfigProvider.cs
jordanbrymora: Clients can't be on mismatched versions because we already block this. Keeping all locations of 20+ config options (and more in the future) in the same order is far better for maintainability.

source/GameInterface/Services/Bandits/Patches/DisableBanditsCampaignBehavior.cs
jordanbrymora: Added

source/GameInterface/Services/PartyBases/Patches/DefaultPartySizeLimitModelPatches.cs
jordanbrymora: Added

deploy/mod-config.default.json
jordanbrymora: Fixed

source/GameInterface/Services/Bandits/Patches/DisableDesertersCampaignBehavior.cs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+[HarmonyPatch(typeof(DesertersCampaignBehavior))]
+internal class DesertersCampaignBehaviorPatches
+{
+    [HarmonyPatch(nameof(DesertersCampaignBehavior.SpawnDesertersParty))]
+    [HarmonyPrefix]
+    public static bool SpawnDesertersPartyPrefix(DesertersCampaignBehavior __instance, MapEvent mapEvent, TroopRoster troops, Settlement settlement)
+    {
+        CampaignVec2 deserterSpawnPosition = __instance.GetDeserterSpawnPosition(settlement);
```
garrettluskey, line 29: `GetDeserterSpawnPosition` dereferences `MobileParty.MainParty` for seeing range and position. this runs on the dedicated server after `RemoveMainParty`, so the first eligible deserter spawn will NRE here. calculate the spawn position without a host party.

source/GameInterface/Services/Bandits/Patches/DisableBanditsCampaignBehavior.cs
```diff
@@ -19,6 +19,23 @@ internal class BanditSpawnCampaignBehaviorPatches
...
+    [HarmonyPostfix]
+    public static void NumberOfMaxBanditCountPerClanHideoutGetterPostfix(ref int __result)
+    {
+        // Use default if provided with a negative value
+        var multiplier = 1f;
+        if (ModConfigProvider.ModOptions.MaximumLootersMultiplier >= 0)
+            multiplier = ModConfigProvider.ModOptions.MaximumLootersMultiplier;
+
+        __result = (int)(__result * multiplier);
```
garrettluskey, line 39: `FillANewHideoutWithBandits` creates the minimum parties without reading this cap, both when initial hideouts are filled and when new hideouts are added. at `0f` this getter returns zero but those paths still create bandit parties.

source/GameInterface/Services/PartyBases/Patches/DefaultPartySizeLimitModelPatches.cs
garrettluskey, line 66: `FindAppropriateInitialRosterForMobileParty` only runs when the bandit party has a template. the null-template paths fill `MemberRoster` after creation, so those new bandit parties never use this multiplier; cover those paths or narrow the config text.

garrettluskey (changes requested):
the roster snapshot and value guards are fixed, and deserters now enter both scaling paths. the existing protobuf-tag comment is still unresolved, and the new deserter path crashes on the dedicated host.

source/GameInterface/Services/PartyBases/Patches/DefaultPartySizeLimitModelPatches.cs
jordanbrymora: The null party template is only used by quests. To avoid interfering with the balance of quests from the config, we shouldn't cover those paths. Updated the config to specify this.

source/GameInterface/Services/Bandits/Patches/DisableBanditsCampaignBehavior.cs
jordanbrymora: Added patch for `_numberOfMinimumBanditPartiesInAHideoutToInfestIt` to also use multiplier.

source/GameInterface/Services/Bandits/Patches/DisableDesertersCampaignBehavior.cs
jordanbrymora: Added patch for `GetDeserterSpawnPosition`.

source/GameInterface/Configuration/ModConfigProvider.cs
garrettluskey, line 52: `LordDefectionRetries`, `EnableHeroExecutions`, and `EnablePlayerClanMemberExecutions` already use tags 15-17 on `development`. keep those assignments and put `LooterPartySizeMultiplier` at 18 instead of changing the existing wire contract.

garrettluskey (changes requested):
the dedicated-host deserter path, hideout minimum, and roster scaling now use the configured multipliers. the protobuf field-number change is still unresolved.

garrettluskey (approved):

PR #2854, opened 2026-08-09
Velocity fix

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Velocity is seemingly being calculated wrong with the puppets

## What is the new behavior?

Resolves #

## Bot Changelog Entry

Projectile damage should now scale correctly.

### DIFF
--- a/source/Missions/Agents/AgentPositionInterpolator.cs
+++ b/source/Missions/Agents/AgentPositionInterpolator.cs
@@ -32,6 +32,9 @@ public interface IAgentPositionInterpolator
         out Vec3 lookDirection,
         out long updateSequence);
 
+    /// <summary>Read a fresh horizontal world-velocity sample reported by this puppet's owner.</summary>
+    bool TryGetAuthoritativeGlobalVelocity(Agent agent, out Vec2 globalVelocity);
+
     /// <summary>[Game thread] Apply each tracked agent's latest native target frame.</summary>
     void Tick(float dt);
 
@@ -59,6 +62,9 @@ public class AgentPositionInterpolator : IAgentPositionInterpolator
     private const float RiderSnapDistance = 6f;
     private const float MountSnapDistance = 12f;
     private const float StaleTargetSeconds = 1f;
+    // Movement heartbeats are sent every 0.25 seconds. Allow one missed heartbeat, then let combat fall back
+    // to the victim's local native velocity rather than applying an old owner sample.
+    private const float CombatVelocityStaleSeconds = 0.5f;
     // Exponential ease rate for the mounted-puppet position follow: fraction MountedFollowRate*dt of the gap is
     // closed each frame, so it tracks the owner with a small lag and settles when the owner stops.
     private const float MountedFollowRate = 12f;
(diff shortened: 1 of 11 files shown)


EgardA:
Gotta compare damage from before and now, should be better.

source/Missions/Battles/BattleDamageRouter.cs
```diff
@@ -168,6 +173,30 @@ public void Dispose()
         return info.CurrentAuthority != session.OwnControllerId;
     }
 
+    private Vec2? ProbeRemoteGlobalVelocity(Agent agent)
+    {
+        if (agent == null)
+            return null;
+
+        if (!coopMissionComponent.AgentRegistry.TryGetAgentInfo(agent, out var info))
```
ShoT-UPfps, line 181: this uses the horse's registry authority, but a ridden horse's movement comes from its rider's `MountData`. if B takes A's registered loose horse, A returns null here even though B owns the velocity sample, so shouldnt we check the live rider's authority first?

source/Missions/Agents/AgentPositionInterpolator.cs
```diff
@@ -202,10 +218,47 @@ public bool TryGetTargetFrame(
...
+            float.IsNaN(globalVelocity.X) ||
+            float.IsInfinity(globalVelocity.X) ||
+            float.IsNaN(globalVelocity.Y) ||
+            float.IsInfinity(globalVelocity.Y))
+        {
+            _globalVelocities.Remove(agent);
+            return;
+        }
+
+        _globalVelocities[agent] = new VelocitySample(globalVelocity, elapsed);
```
ShoT-UPfps, line 261: these packets are `Unreliable` and have no sender sequence, so an older moving snapshot can arrive after a newer stop and become fresh again here. can we reject out-of-order samples before using them for damage?

source/Missions/Agents/AgentPositionInterpolator.cs
```diff
@@ -68,6 +74,8 @@ public class AgentPositionInterpolator : IAgentPositionInterpolator
     private readonly Dictionary<Agent, TargetFrame> _targets = new Dictionary<Agent, TargetFrame>();
     private readonly Dictionary<Agent, long> _mountedGuardProcessedSequences =
         new Dictionary<Agent, long>();
+    private readonly Dictionary<Agent, VelocitySample> _globalVelocities =
```
ShoT-UPfps, line 77: `Tick` evicts dead, inactive, and stale agents from `_targets` but doesn't remove them from `_globalVelocities`, and the battle death paths don't call `Forget`. this keeps every dead remote `Agent` referenced until mission end.

source/Missions/Agents/Packets/AgentData.cs
```diff
@@ -140,5 +142,11 @@ public void Apply(Agent agent)
         /// <summary>The owner's current translation and turn inputs.</summary>
         [ProtoMember(9)]
         public uint MovementFlag { get; }
+        /// <summary>The owner's horizontal world velocity used by native relative-speed damage.</summary>
+        [ProtoMember(10)]
+        public Vec2 GlobalVelocity { get; }
+        /// <summary>Distinguishes a valid stationary sample from a packet sent by an older peer.</summary>
```
ShoT-UPfps, line 148: we don't support mixed peer versions, so i don't think we need `HasGlobalVelocity` in either packet type. current constructors always set it, and `Vec2.Zero` already represents a valid stop.

ShoT-UPfps (changes requested):
cmt

EgardA:
Not sure anymore that velocity is an issue, think the high damage is from elsewhere.

PR #2855, opened 2026-08-09
Prevent null-party map event result crashes

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

An AI battle can retain an invalid participant. When that battle ends, the server throws a null-reference error while applying renown and stops campaign ticks.

## What is the new behavior?

The server removes invalid battle participants before calculating results, so valid rewards are applied and the map event can finish without stopping campaign ticks. 
image

## Bot Changelog Entry

- Fixed a server crash when an AI battle ended with an invalid party reference.


source/GameInterface/Services/MapEvents/Patches/MapEventPatches.cs
```diff
@@ -195,14 +227,77 @@ private static bool Prefix_OnBattleWon(MapEvent __instance)
...
+        }
+
+        return false;
+    }
+
+    internal static int CommitCalculatedMapEventResults(
+        MapEvent mapEvent,
+        IReadOnlyList<Action<MapEventParty>> commitPhases)
+    {
+        int removedPartyCount = RemovePartiesWithoutParty(mapEvent);
```
garrettluskey, line 280: `RemovePartiesWithoutParty` also needs to run before `CalculateMapEventResults` at `MapEventPatches.cs:206`. vanilla dereferences each `defeatedParty.Party` while calculating plunder before this commit prefix is reached, so an already-invalid participant still crashes the tick.

garrettluskey (dismissed):
the per-phase validity checks preserve the vanilla commit order. cleanup also needs to cover the earlier result-calculation phase.

source/GameInterface/Services/MapEvents/Patches/MapEventPatches.cs
ShoT-UPfps: removed invalid parties before `CalculateMapEventResults`, otherwise vanilla can dereference `defeatedParty.Party` during plunder calculation. fixed in `b778b977`

garrettluskey (approved):
the cleanup now covers pre-calculation invalid parties and preserves vanilla phase order while rechecking each entry. looks good.

PR #2866, opened 2026-08-09
[2794] Fall back when final kingdom vote publication fails

### LINKED ISSUE #2794: [KINGDOMS] Failed final vote publication leaves the kingdom decision screen stuck
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

In a dedicated server session, attempting to enact a law in a self-founded kingdom leaves the player stuck on the voting screen forever, the law never gets enacted, all inputs repeat the same vote state, and the menu cannot be exited.

## Steps to reproduce

Repro steps: in a dedicated server game, start from a self-founded kingdom, attempt to enact a law, and observe that the voting menu does not close and the law does not resolve.

## Additional information

Attached logs include `coop-server-20260806-232926.log` and `Coop_client.log`; the reporter notes the client log data is empty.

## Discord source

https://discord.com/channels/709516043332354119/1535043388900909118/1535043388900909118

## Logs

Client log: [0.1.1+324d59795711727d40bc1aa3eb56a652ae656306 - Download Coop_client.log](<https://cdn.discordapp.com/attachments/1535043388900909118/1535043390268252240/Coop_client.log?ex=6aa4788f&is=6aa3270f&hm=e232fc4b82057ecae7ea9c7b36bcdd917044c81e7ac92425260e6deb8fd075a8&>)

Server log: [Download coop-server-20260806-232926.log](<https://cdn.discordapp.com/attachments/1535043388900909118/1535043389530177596/coop-server-20260806-232926.log?ex=6aa4788f&is=6aa3270f&hm=ed25a187d25d77daea9f43bc6dfdb9022fddbac5acf8b6e9acc8bbaac52bea83&>)

## Confirmed failure boundary and consolidated report (#2625)

The shared live client evidence repeatedly logs `Unable to publish final kingdom decision vote from the local decision UI.` `DecisionItemBaseVM.ExecuteFinalSelection` is patched to suppress vanilla finalization whenever local resolution is blocked, but the prefix still returns `false` when `TryPublishFinalVote` fails. That failed path does not mark the local vote submitted, so the decision UI remains active indefinitely. The evidence does not identify which inner `TryCreateVoteData` guard rejected the vote.

[#2625](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/issues/2625) is the same policy-finalization failure and includes the same support ticket (#1314) and six source attachments. It adds the French-UI reproduction: clicking `Terminé` plays the button sound but does not validate or dismiss the decision. The report also records temporary recovery after restarting Steam/Bannerlord, policies applying one or two in-game days late or only partially despite influence being spent, recurrence on `Done`, a screenshot, client/server logs, three save pairs, and the transfer JSON.

This issue tracks failed policy-decision final vote publication and dismissal. War-declaration and forced-peace decision failures remain separate unless they produce this same publication failure.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Attempting to enact a kingdom policy can leave the final vote screen stuck when the local vote cannot be published.

## What is the new behavior?

Resolves #2794

The vote falls back to normal finalization, closes, and resolves the policy instead of remaining stuck.

## Bot Changelog Entry

- Fixed an issue where kingdom policy votes would sometimes get stuck on the screen.


source/GameInterface/Services/Kingdoms/Patches/KingdomDecisionVMPatches.cs
```diff
@@ -72,8 +72,7 @@ private static bool ExecuteFinalSelectionPrefix(DecisionItemBaseVM __instance)
             if (!KingdomDecisionsVMPatches.TryGetVoteManager(out var voteManager)) return true;
             if (!voteManager.ShouldBlockLocalResolution(__instance)) return true;
 
-            voteManager.TryPublishFinalVote(__instance);
-            return false;
+            return !voteManager.TryPublishFinalVote(__instance);
```
garrettluskey: please dont fall through to vanilla here. if both vote-data attempts fail, `KingdomElection.ApplyChosenOutcome` runs on the client, but the client patches block the policy and decision removal while still applying influence and secondary effects, so the server keeps the decision unresolved and can resolve it again later. keep local resolution suppressed and recover the vote data or close the VM without calling `ApplySelection`.

garrettluskey (dismissed):
the failed-publication path does close the stuck screen, but it can resolve different state on the client and server.

source/GameInterface/Services/Kingdoms/Patches/KingdomDecisionVMPatches.cs
ShoT-UPfps: fixed, failed publication now closes the decision VM and always suppresses vanilla ApplySelection.

source/GameInterface/Services/Kingdoms/Patches/KingdomDecisionVMPatches.cs
```diff
@@ -72,7 +72,10 @@ private static bool ExecuteFinalSelectionPrefix(DecisionItemBaseVM __instance)
             if (!KingdomDecisionsVMPatches.TryGetVoteManager(out var voteManager)) return true;
             if (!voteManager.ShouldBlockLocalResolution(__instance)) return true;
 
-            voteManager.TryPublishFinalVote(__instance);
+            if (!voteManager.TryPublishFinalVote(__instance))
+            {
+                voteManager.CloseDecisionItem(__instance);
```
garrettluskey: `CloseDecisionItem` only dismisses the client VM, no final vote is sent. `HasAllVotes` stays false for this clan, so the server only resolves it in the forced `TriggerTime.IsPast` sweep. recover and publish the final vote instead, otherwise the selected policy still isnt resolved when Done is pressed.

garrettluskey (dismissed):
closing the client VM avoids the unsafe local vanilla resolution path. one problem remains, the failed publication still doesnt resolve the selection on the server.

source/GameInterface/Services/Kingdoms/Patches/KingdomDecisionVMPatches.cs
ShoT-UPfps: fixed, equivalent selected outcomes are recovered by OutcomeKey and published as the final vote. the regression now verifies immediate server policy resolution.

garrettluskey: `TryPublishFinalVote` can still fail before sending anything. closing here leaves `HasAllVotes` false and the selected policy unresolved until the expiry sweep, so recover or retry these cases instead of dismissing the only UI.

garrettluskey (dismissed):
matching detached outcomes by their stable key fixes the reproduced policy path and sends an equivalent final vote. the remaining pre-send failures still dismiss the only UI without resolving the selection.

source/GameInterface/Services/Kingdoms/Patches/KingdomDecisionVMPatches.cs
ShoT-UPfps: fixed, failed pre-send publication now leaves the decision VM active so Done can retry. the regression removes and restores the kingdom registry id, then verifies the retry resolves on the server.

source/GameInterface/Services/Kingdoms/KingdomDecisionVoteManager.cs
```diff
@@ -1128,6 +1130,20 @@ private static int GetOutcomeIndex(DecisionOutcome decisionOutcome, KingdomElect
             return -1;
         }
 
+        private int GetOutcomeIndex(DecisionOutcome decisionOutcome, KingdomElection election, string outcomeKey)
+        {
+            int outcomeIndex = GetOutcomeIndex(decisionOutcome, election);
```
garrettluskey: `PolicyDecisionItemVM` doesnt detach this outcome: vanilla builds every option from the same election's `PossibleOutcomes`, so the reference lookup above succeeds. what production path makes this fallback run? the regression only creates one by assigning the private setter, while the reported failed guard is still unknown.

source/E2E.Tests/Services/Kingdoms/PlayerKingdomCreationFlowTests.cs
```diff
@@ -1518,6 +1518,138 @@ public void KingdomDecisionVoteData_CreatedFromDecisionVmUsesRegisteredKingdomId
...
+            var detachedOutcome = new KingdomPolicyDecision.PolicyDecisionOutcome(shouldBeEnforced: true);
+            DecisionOptionVM option = decisionItem.DecisionOptionsList.Single(candidate =>
+                candidate.Option is KingdomPolicyDecision.PolicyDecisionOutcome outcome &&
+                outcome.ShouldDecisionBeEnforced);
+            option.Option = detachedOutcome;
+            option.CurrentSupportWeight = Supporter.SupportWeights.FullyPush;
+            decisionItem._currentSelectedOption = option;
+
+            Assert.DoesNotContain(detachedOutcome, decisionItem.KingdomDecisionMaker._possibleOutcomes);
+            Assert.True(GetVoteManager(client).TryPublishFinalVote(decisionItem));
```
garrettluskey: please drive this through `ExecuteFinalSelection`. calling the manager directly with an empty callback skips the real prefix and `KingdomDecisionsVM` closure path this regression is meant to prove.

garrettluskey (dismissed):
the failure path now keeps vanilla client resolution suppressed and leaves the UI retryable. the recovery regression still manufactures an outcome state the policy VM does not create, so it doesnt establish that the reported publication failure is fixed.

source/GameInterface/Services/Kingdoms/KingdomDecisionVoteManager.cs
ShoT-UPfps: removed the fallback, the policy VM has no production path that detaches this outcome.

source/E2E.Tests/Services/Kingdoms/PlayerKingdomCreationFlowTests.cs
ShoT-UPfps: fixed, the regression now uses the real two-clan `KingdomDecisionsVM` and `PolicyDecisionItemVM`, calls `ExecuteFinalSelection`, and verifies the real callback keeps the failed item open then clears it after retry.

garrettluskey (approved):
The unsupported detached-outcome fallback is gone, and the regression now drives the real policy decision VM through ExecuteFinalSelection, proving that a pre-send failure leaves the item open and the retry resolves it. The checks are green. Approved.

PR #2901, opened 2026-08-10
IsReady Removed

EgardA:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Sync of local variables IsReady

## What is the new behavior?

Resolves #

Removed sync of local variable IsReady

## Bot Changelog Entry

Removed sync of unnecessary variable


source/GameInterface/Services/MobileParties/MobilePartySync.cs
```diff
@@ -24,6 +24,8 @@ public MobilePartySync(AutoSyncRegistry autoSyncBuilder)
 
         // Movement is replicated atomically by NetworkUpdatePartyBehavior. These members are
         // interdependent, so independent AutoSync messages can apply a behavior before its target.
+        // EventPositionAdder is a locally computed map-event visual offset and is also carried by
```
ShoT-UPfps: `EventPositionAdder` isnt computed locally for replicated map events. `MapEventInitializationBarrier` attaches received parties through `_battleParties` and `_mapEventSide` instead of `AddInvolvedPartyInternal`, so already-connected clients keep the old offset when a field battle starts. please keep this synced or include it in the live map-event state.

ShoT-UPfps (dismissed):
cmt

source/GameInterface/Services/MobileParties/MobilePartySync.cs
garrettluskey: `EventPositionAdder` isn't computed locally when replicated map events attach a party. `MapEventInitializationBarrier.AttachClient` only updates `_battleParties` and `_mapEventSide`, so an already-connected client keeps its old offset when `MapEvent.AddInvolvedPartyInternal` computes a field-battle position on the server. please keep this synced or include it in the live map-event state.

garrettluskey (changes requested):
`IsReady` is local registration state, and limiting the constructor logger to debug does not affect the auto-registry or destruction patches. `EventPositionAdder` still needs a live update path for clients that are already connected.

garrettluskey (changes requested):
The `IsReady` removal and debug-only constructor logger look correct. The existing unresolved `EventPositionAdder` thread still blocks this; the head has not changed and there is still no live update path for already-connected clients.

gamingmenacers-svg (dismissed):
reviewed exact head `f3a0132ec` against current `development`. removing `Hero.IsReady` is correct and the debug-only constructor logger is fine, but `EventPositionAdder` still has no live update for connected clients. replicated map-event attachment bypasses vanilla `AddInvolvedPartyInternal`, so the server computes a field-battle offset that peers never receive. keep this sync or carry the value in the live map-event attachment, then add a connected-client test. at 918 commits behind, this should be rebased before any further merge decision.

ShoT-UPfps:
does the title need to be updated? i see a commit saying "Restored eventpositionadder" but the title still says "eventpositioonadder removed"

EgardA:
> does the title need to be updated? i see a commit saying "Restored eventpositionadder" but the title still says "eventpositioonadder removed"

Fixed

garrettluskey (approved):
The `EventPositionAdder` blocker is fixed: live autosync is present and the connected-client reinforcement regression passed in CI. The `IsReady` removal and debug-only constructor logger still look correct.

PR #2909, opened 2026-08-10
[2762] Prevent repeated companion rescue completion from duplicating roster counts

### LINKED ISSUE #2762: [ISSUE] Prevent repeated companion rescue completion from duplicating roster counts
# For New Contributors

## Description

Companion rescue completion can add the same hero to a player party more than once.

`Handle_DoCompanionJoinedPartyByRescue` runs these authoritative operations for every received `DoCompanionJoinedPartyByRescue` message:

1. `EndCaptivityAction.ApplyByReleasedAfterBattle(...)`
2. Change the companion state to `Active`
3. `mainParty.AddElementToMemberRoster(..., 1, false)`

The message contains only the companion and party ids. It has no request correlation or duplicate protection, and the handler does not verify that the companion is still a prisoner or already belongs to the target roster. A repeated rescue consequence therefore increments the same hero roster element from one to two, three, and so on. The rescue path that creates a companion-led party also applies its completion through `CompanionRolesCampaignBehavior.PartyScreenClosed` without duplicate protection.

This matches #2702's post-fight prisoner-screen report, which occurred after PR #2288. PR #2288 synchronized dismissal and reconciled an already duplicated roster; its count-six test creates that state directly and does not reproduce the operation that created the extra copies.

A supplied dedicated-server save contains the downstream result:

- Hero: `Suvys the Scholar` (`CharacterObject_1585`)
- Hero state: `Dead`
- The player member roster contains the hero with `Number = 5`

The save proves that duplicate hero counts persist, but it contains no mutation history and cannot identify which rescue request created each copy. Vanilla removal and death operations subtract one hero, so later dismissal or death exposes the remaining copies as inaccessible ghost entries.

Original save SHA256: `aa17fce46f99c247123252ba9bd2c294b37213200a96c21ec994149cdfc31fe3`

## Intended Design

Authoritative companion rescue completion is idempotent. Processing the same logical rescue more than once must leave exactly one companion in exactly one party and must not repeat captivity-release or party-creation side effects.

Fix the operation that adds the copies. Do not add a startup repair pass, save migration, or death-time duplicate cleanup as the solution to this issue.

## Location

- `source/GameInterface/Services/Companions/Handlers/CompanionRolesHandler.cs`
- `source/GameInterface/Services/Companions/Messages/NetworkCompanionRolesMessages.cs`
- `source/GameInterface/Services/Companions/Patches/CompanionRolesPatches.cs`
- `source/E2E.Tests/Services/Companions/`

Create a branch based from [development](https://github.com/Bannerlord-Coop-Team/BannerlordCoop).

## Related Issues

- #2702 reports duplicate companions after the post-fight prisoner flow
- #2680 reports the companion rescue/party-creation flow remaining incomplete
- #2280 and PR #2288 fixed dismissal ordering and cleanup of an already duplicated roster

## Requirements

- Correlate or otherwise make `DoCompanionJoinedPartyByRescue` idempotent on the server.
- Validate the companion's current captivity, clan, party, and target-roster state before applying rescue completion.
- Do not call `AddElementToMemberRoster(..., 1)` when the companion is already present in the target member roster.
- Make the companion-led-party rescue completion idempotent as well; a repeated `DoPartyScreenClosedFromRescuing` must not add the hero again or create another party.
- Apply the authoritative rescue with patches live so normal roster, hero-state, and party messages are sent.
- Return a terminal completion or rejection to the requesting client so a delayed or repeated UI consequence cannot be treated as a new rescue.
- Do not change normal companion hiring, rejoining after imprisonment, dismissal, or unrelated prisoner transfers.
- Do not add save repair, startup migration, or death cleanup.
- Add focused E2E coverage that sends each rescue completion twice and proves the companion count remains one and side effects occur once.
- Add a server-only debug fixture that starts with a captive companion and exercises the real repeated rescue request. Do not seed an invalid count directly.

## Additional information

Vanilla `EndCaptivityAction.ApplyByReleasedAfterBattle` can be called again after `PartyBelongedToAsPrisoner` is already null; it does not make the following raw roster addition idempotent. Vanilla `TroopRoster.AddToCountsAtIndex` also calls the hero-added callback on every positive delta, so each repeated `+1` updates `PartyBelongedTo` while increasing the stored count.

The corrected support save is a workaround for that player only and is not part of this implementation.

## Definition of Done

- [ ] One rescue completion adds the companion once and completes the expected release flow.
- [ ] Replaying the same join-party rescue completion leaves the roster count at one.
- [ ] Replaying the same companion-led-party completion creates only one party and one hero membership.
- [ ] Server and clients converge on the same companion roster count and party reference.
- [ ] Captivity-release and companion-join events occur once per logical rescue.
- [ ] The client receives a terminal result for accepted, already-completed, and stale rescue requests.
- [ ] Existing companion dismissal and ordinary prisoner-transfer tests remain green.
- [ ] The dedicated-server debug fixture passes with one client and after a server restart.

## Consolidated field report (#2702)

[#2702](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/issues/2702) is the player-facing report mapped to the repeated rescue-completion path above. After the post-fight prisoner flow, three visible copies of the companion appeared; a restart reduced the visible count from three to two, and the remaining companion could become inaccessible. Dismissing the companion again from a later post-fight prisoner-taking screen removed one copy and it had not reappeared when the report was closed.

The report supplies three screenshots, `coop-server-20260803-155607.log`, and `Coop_client.log`. Its persisted overcount is the state explained by the unguarded repeated `AddElementToMemberRoster(..., 1)` operation. As noted above, the supplied save proves the downstream duplicate count rather than preserving the mutation history, which is why the fix and test must make the rescue operation itself idempotent.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Completing the same companion rescue more than once can add duplicate copies of the companion to the player's party or create multiple companion-led parties.

## What is the new behavior?

Resolves #2762

Rescue completion is applied once, repeated or delayed completion is reported without adding another companion or creating another party, and stale requests are rejected.

## Bot Changelog Entry

- Fixed an issue where rescuing a companion could duplicate the companion


source/GameInterface/Services/Companions/Handlers/CompanionRolesHandler.cs
```diff
@@ -438,21 +522,213 @@ private void Handle_PartyScreenClosedFromRescuing(MessagePayload<PartyScreenClos
...
+        return true;
+    }
+
+    private static void ValidateRequestId(string requestId)
+    {
+        if (string.IsNullOrWhiteSpace(requestId))
+            throw new InvalidOperationException("The companion rescue request has no correlation id.");
+    }
+
+    private Clan ValidateRescueOwnership(Hero companion, MobileParty targetParty,
```
garrettluskey: This validates that the companion and target party belong to the same clan, but it never authorizes the peer that sent the request. `requester` is only used for the reply/cache, so any connected client that knows the synchronized companion, clan, captor, and party IDs can rescue another clan’s companion or create its war party. Resolve the player from `payload.Who` and require that player’s authoritative clan/party to own the rescue before either mutation path runs.

garrettluskey (dismissed):
The correlated completion response and terminal-state checks make retries much safer and address the duplicate roster count directly. I’m requesting changes because the new server mutation still is not authorized to the peer that submitted it.

jordanbrymora:
Wouldn't it be simpler to just reject rescuing the companion if they are no longer a prisoner and/or already in the player's or their own new party?

ShoT-UPfps:
> Wouldn't it be simpler to just reject rescuing the companion if they are no longer a prisoner and/or already in the player's or their own new party?

This is already happening with the current PR, but the current PR was also doing extra defensive checks that are probably not worth the complexity. will simplify

ShoT-UPfps:
yeah, removed the request cache. each callback creates a fresh request id, the channel is reliable ordered, and the terminal-state checks handle repeated completion.

ShoT-UPfps:
pulled the request ids and expected clan/captor snapshots out too. the server now uses the companion's current prisoner/party state for idempotency and still checks that the requesting peer owns the clan and target party.

garrettluskey (approved):
The requester is now bound to the authoritative clan and party, and the terminal-state checks make repeated join/lead completion idempotent while returning both results in order. looks good.

PR #2912, opened 2026-08-10
[1547] Sync town garrison lifecycle changes

### LINKED ISSUE #1547: [ISSUE] AutoSync field Fief.GarrisonPartyComponent is registered on the base but never stored there (dead transpiler)
# For New Contributors

## Description

`FiefSync` registers `Fief.GarrisonPartyComponent` as a synced **field**, which generates a field-set
transpiler (`GarrisonPartyComponent_SetTranspiler`) that matches a `stfld` of that field. But **no `Fief`
declared method or constructor stores `GarrisonPartyComponent`** — it's set in the derived `Town`/`Village`
types, not on `Fief` itself. So the transpiler matches nothing in `Fief`, i.e. it's a **dead transpiler**:
`GarrisonPartyComponent` is not actually intercepted or synced through this path.

Before #1542 this was a silent no-op (the transpiler was attached to every `Fief` method via the
unconditional `TargetMethods()` and matched none of them). #1542's narrowing surfaced it: `TargetMethods()`
narrowed to **empty**, and Harmony throws `ArgumentException: Undefined target method` for a patch class
that has a transpiler but zero targets, crashing `PatchAllUncategorized` on host start ("Host Coop
Campaign"). #1542 added a fall-back (`GenericPatchHelpers.SelectTargetMethods` reverts to the declared
methods when narrowing matches nothing) so the host no longer crashes, but the fall-back only restores the
old no-op, the underlying registration is still dead.

An enumeration over all 38 AutoSync-registered types (against the real game assemblies) found **`Fief` is
the only type whose narrowing empties out**, so this is the one fully-dead field transpiler.

## Intended Design

Decide how `GarrisonPartyComponent` should actually replicate and fix the registration accordingly:

- If it's meant to sync as a field, register/intercept it where it's actually stored (the derived
  `Town`/`Village` that set it), so a real method becomes the transpiler target.
- Or sync it via the mechanism that already handles party components (it's a `PartyComponent` reference),
  and drop the dead `Fief.GarrisonPartyComponent` field registration.

First confirm whether `GarrisonPartyComponent` changes replicate server -> client today (they likely do
**not** via this transpiler) using a co-op session, then pick the fix.

## Location

- `source/GameInterface/Services/Fiefs/FiefSync.cs` (the `AddField(... Fief.GarrisonPartyComponent ...)` registration).
- `source/GameInterface/AutoSync/Templates/Patches/DynamicPatchTemplate.txt` + `GenericPatchHelpers.SelectTargetMethods` (the #1542 fall-back that currently masks it).

## Related Issues
Surfaced by #1542 (the AutoSync `TargetMethods` narrowing); child of #1539 (the root-cause deadlock work).

## Requirements
- Confirm whether `Fief.GarrisonPartyComponent` replicates today, and make it replicate server -> client (or remove the dead registration if it's synced elsewhere).
- No host crash and no fall-back to over-broad detouring for `Fief`.

## Additional information

The same class of latent bug (a field registered on a type whose own methods never store it) could exist
*partially* on other types, a field whose store is in a derived/base type but the type has other live
members, so it narrows fine and doesn't crash but that one field still isn't intercepted. A sweep that
checks, per registered field, that at least one declared method of the registered type actually stores it
would catch those.

## Definition of Done
- [ ] `Fief.GarrisonPartyComponent` either replicates server -> client or its dead registration is removed.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Initializing or finalizing a town garrison on the server can leave clients with a missing or stale town-to-garrison link.

## What is the new behavior?

Resolves #1547

Town garrison initialization and finalization now update every connected client's town state, so all peers agree on the active garrison.

## Bot Changelog Entry

- Fixed town garrison creation and removal leaving clients with stale garrison state.

### DIFF
--- a/source/Coop/LiveTesting/LiveTestControlServer.cs
+++ b/source/Coop/LiveTesting/LiveTestControlServer.cs
@@ -33,6 +33,9 @@ namespace Coop.LiveTesting
         private readonly LiveTestProcessInfo processInfo;
         private readonly DateTime processStartedUtc;
         private readonly NamedPipeLiveTestServer pipeServer;
+        private readonly object screenshotGate = new object();
+        private readonly Dictionary<string, ScreenshotCapture> screenshotCaptures =
+            new Dictionary<string, ScreenshotCapture>(StringComparer.Ordinal);
         private int shutdownScheduled;
 
         public LiveTestControlServer(bool isServer, string logFilePath)
@@ -85,10 +88,14 @@ namespace Coop.LiveTesting
                         request,
                         () => CreateStatusResponse(request.Id),
                         false);
+                case "command-catalog":
+                    return HandleCommandCatalog(request);
                 case "command":
                     return HandleCommand(request);
                 case "screenshot":
                     return HandleScreenshot(request);
+                case "screenshot-status":
+                    return HandleScreenshotStatus(request);
                 case "shutdown":
                     return HandleShutdown(request);
                 default:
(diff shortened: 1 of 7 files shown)


garrettluskey (approved):
the target methods are the vanilla assignments for `Fief.GarrisonPartyComponent`, so the existing field sync now publishes initialize and finalize changes. creation ordering and game-thread receive handling preserve the component reference, and the lifecycle test covers both clients.

PR #2913, opened 2026-08-10
[1570] Coalesce non-lord party trade gold updates per tick

### LINKED ISSUE #1570: Per-tick coalescing layer at the AutoSync publish to SendAll seam
this issue came out of the investigation into #1557 (sustained campaign-map AI-sync bandwidth, ~250-380 KB/s, almost all world-wide AI): we profiled the server packet stream, traced the worst offenders, and split the durable fixes into separate issues.

part of #1557, the keystone coalescing win. the AutoSync generated handlers each call `network.SendAll(...)` once per field change (`Templates/Handlers/*.txt`, e.g. `SubscribePropertyOwnerReferenceTemplate.txt:4`, `SubscribeArrayValueTemplate.txt:9/32`). the server AI writes DefaultBehavior/ShortTermBehavior/TargetParty/MoveTargetPoint/_targetSettlement/PartyTradeGold plus roster deltas many times per party per tick, so this seam is the bulk of the spike, each change its own ReliableOrdered packet.

durable fix: insert a per-tick coalescing buffer between the generated handlers and the wire, keyed by (messageType, instanceId, member):
- latest-wins for absolute-value SETs (PartyTradeGold, TargetParty, DefaultBehavior/ShortTermBehavior, position) because they carry the full new value, so only the last one per tick matters
- summed for additive deltas (troop XP add, roster count add)
- one merged snapshot per roster per tick

flush exactly once per server network tick at the existing boundary (`CoopServer.Update`, `CoopServer.cs:121`), emitting through `CoopNetworkBase.SendAll` as today.

obligations otherwise it desyncs: never reorder a field update ahead of that object's create/registration message or after its destroy; flush an object's pending coalesced state before its destroy is sent; respect the ReliableOrdered per-channel FIFO.

durable because its information-preserving: the client only ever observes the final per-tick value, so it stays correct no matter which of the disabled behaviors get re-enabled (and it caps per-party-per-tick traffic at one merged send regardless of churn). subsumes the per-field behavior cluster, hero skill/raw XP rebroadcast, troop-XP volume, and non-lord PartyTradeGold. its also the PR-sized first step toward the area-of-interest epic.

depends on #1576 (the reusable per-tick coalescer). this is its primary consumer: route the AutoSync generated-handler `SendAll` through the shared coalescer rather than building the buffer/flush/ordering here.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Rapid trade-gold changes for non-lord parties send every intermediate value during the same server tick, increasing campaign-map network traffic.

## What is the new behavior?

Resolves #1570

Non-lord party trade-gold changes now send only the final value once per server tick, and pending updates are discarded when the party is destroyed.

| Metric (same save, one client, 180 seconds wall-clock at `Play_2x`) | Development | This change |
| --- | ---: | ---: |
| PartyTradeGold packets | 16,035 | 7,913 (50.7% fewer) |
| PartyTradeGold logical payload bytes | 609,186 B | 281,965 B (53.7% fewer) |

## Bot Changelog Entry

- Reduced campaign-map network traffic from non-lord party trade-gold updates.


garrettluskey (approved):
the existing coalescer keeps the last absolute value per tick, and destruction now drops both full and compact party keys before the destroy message. the E2E tests cover aggregation and destruction before flush.

PR #2914, opened 2026-08-10
[2803] Fix duplicate player-family save keys

### LINKED ISSUE #2803: [SAVES] Duplicate player-family object key prevents dedicated-server save loading
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

A Bannerlord Co-op dedicated server cannot load an affected save because the saved container data contains a duplicate composite object key associated with a player's family name.

During `LoadContext::Load Container Datas`, the engine reports:

```text
An item with the same key has already been added. Key: ((<object-id-a>) , (<object-id-b>) <player-family-name>)
Error: Could not load the game!
```

The failure happens before the dedicated server reaches `SERVING`. This is not the existing `crafted_item_<number>` smithing-key form; neither player had used smithing in the affected save.

## Steps to reproduce

1. Start the dedicated server on build `117131` with the affected matching `.sav` and `.json` pair.
2. Let the server host the configured save.
3. Observe that object and string resolution completes, then container-data loading throws the duplicate composite-key exception and the save fails to load.
4. Run the current bounded save-repair routine, replace both files with its output, and start the dedicated server again.
5. Observe the identical composite key and identical container-load failure; the server still never reaches `SERVING`.

## Additional information

The current repair routine changed unrelated stale join, AI-interactable, and player-registration entries, but it did not remove this duplicate. The post-repair server log proves the same key remains, so a successful repair must not be inferred from the save archive merely changing.

Suggested root-cause scope:

- Identify the save/serialization path that writes duplicate composite object keys for a player-family record.
- Prevent new saves from emitting duplicate keys.
- Add a backward-compatible loader or migration for existing affected saves that removes only the invalid duplicate while preserving valid object references.
- Add a regression covering both creation of a clean save and loading/migrating an already-affected save.

Evidence:

- [Original server log](https://cdn.discordapp.com/attachments/1535117780305584128/1535120089488302140/coop-server-20260806-214825.log?ex=6a769b7e&is=6a7549fe&hm=d6f9c440c965a88111c5cb85107fe647f7528f9343fa05a8baecbf6b675d351b&)
- [Post-repair server log](https://cdn.discordapp.com/attachments/1535117780305584128/1535123934054449192/coop-server-20260806-221147.log?ex=6a769f13&is=6a754d93&hm=d915c1761043ac5be62c9dcfce194a5b71fc4e29a3a6f36378dc4486fb0ca3ad&)
- [Affected save](https://cdn.discordapp.com/attachments/1535117780305584128/1535121071508819988/Jardlav.sav?ex=6a769c68&is=6a754ae8&hm=8157825b9b9fed307584693bdabea73b2974a05d81631026e7843bcc2561e835&) and [matching sidecar](https://cdn.discordapp.com/attachments/1535117780305584128/1535121071127134298/Jardlav.json?ex=6a769c68&is=6a754ae8&hm=76baaae52f914bb47d8c69386ff3ce5b8e62ea6feaaf03cbc4a2d91bfcbff116&)
- [Original support ticket](https://discord.com/channels/709516043332354119/1535117780305584128)

## Discord source

https://discord.com/channels/709516043332354119/1535124261587517500/1535124261587517500

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Dedicated servers stop while loading affected saves with duplicate player-family relationship data and never become ready for players.

## What is the new behavior?

Resolves #2803

Affected saves now finish loading on dedicated servers, and newly joined player families no longer create duplicate relationship data in later saves.

## Bot Changelog Entry

- Fixed an issue where the server would sometimes fail to load a save when a player's relationship data contained duplicates.


source/Coop.Core/Server/Connections/States/CreateCharacterState.cs
```diff
@@ -91,6 +91,22 @@ internal void Handle_NetworkTransferNewHero(MessagePayload<NetworkTransferNewHer
...
+        {
+            heroInterface.SetupServerHero(hero);
+        }
+        catch (System.Exception exception)
+        {
+            Logger.Error(
+                exception,
+                "Failed to set up hero for {ControllerId}; disconnecting the joining peer",
+                controllerId);
+            ConnectionLogic.Peer.Disconnect();
```
garrettluskey, line 124: At this point the player has already been added, bound to the peer, and announced to every existing client. `PlayerDisconnected` only clears the peer and parks the party; it does not call `RemovePlayer`, so a setup exception leaves this partially initialized player registered on the server and created on clients. A retry for the same controller then resolves against that poisoned graph. Roll back the player/object registration and notify the clients, or delay `AddPlayer` and the creation broadcast until setup can no longer fail, before disconnecting. Extend the failure test to assert that no player or remote creation remains.

garrettluskey (dismissed):
The save migration is narrowly scoped and the registration-order changes address the duplicate stance key. One failure path still leaves a partially initialized player graph registered and replicated, so changes are needed before this is safe to merge.

source/Coop.Core/Server/Connections/States/CreateCharacterState.cs
ShoT-UPfps: fixed, the setup failure now removes the exact player registration and sends `NetworkPlayerRemoved` after the creation message before disconnecting. the failure test checks the server removal and the ordered client create/remove pair.

source/Coop.Core/Server/Connections/States/CreateCharacterState.cs
```diff
@@ -91,6 +92,28 @@ internal void Handle_NetworkTransferNewHero(MessagePayload<NetworkTransferNewHer
...
+            Logger.Error(
+                exception,
+                "Failed to set up hero for {ControllerId}; disconnecting the joining peer",
+                controllerId);
+
+            if (!playerManager.RemovePlayer(player))
+                Logger.Error("Failed to roll back player registration for {ControllerId}", controllerId);
+
+            // Existing clients created this graph before setup ran, so remove their registration on the same channel.
+            network.SendAllBut(netPeer, new NetworkPlayerRemoved(controllerId, player.HeroId));
```
garrettluskey: `NetworkPlayerRemoved` only drops the client registration and marks its hero dead; this path never removes the server's imported hero/party/clan or sends the party-destroy cleanup that message expects. if `SetupServerHero` throws after partially registering the graph, the server and existing clients keep orphan objects and a retry creates another graph. roll back the created objects on both sides, or delay the creation broadcast until setup can no longer fail.

garrettluskey (dismissed):
the save-key repair and setup ordering look right. changes are still needed because setup failure only removes the player registration and leaves the created object graph behind.

source/Coop.Core/Server/Connections/States/CreateCharacterState.cs
ShoT-UPfps: fixed, setup failure now rolls back the exact imported graph on the game thread and carries every captured registration id to existing clients, including roster ids after the party destroy. clan rollback also removes its stance entries before the clan leaves the campaign.

garrettluskey (approved):
the failed setup path now captures every imported registration, removes the player and campaign graph on the game thread, drops pending coalesced updates, and sends one final ordered rollback for existing clients. the server/client rollback tests cover the previous orphan-graph blocker.

PR #2918, opened 2026-08-10
[1539] Prevent server deadlocks from trivial AutoSync detours

### LINKED ISSUE #1539: [ISSUE] Server GC deadlock from AutoSync detouring tiny methods and corrupting x64 unwind info

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Dedicated servers still install unnecessary synchronization hooks on empty or constant-only game routines, leaving the original large-battle freeze path dependent on a separate safety guard.

## What is the new behavior?

Resolves #1539

Dedicated servers skip those unnecessary hooks while keeping existing campaign synchronization working normally.

## Bot Changelog Entry

- Stopped dedicated servers from applying unnecessary synchronization hooks to empty or constant-only game routines.

### DIFF
--- a/source/GameInterface.Tests/DynamicSync/AutoSyncTargetSelectionTests.cs
+++ b/source/GameInterface.Tests/DynamicSync/AutoSyncTargetSelectionTests.cs
@@ -1,0 +1,264 @@
+﻿using GameInterface.Utils;
+using GameInterface.Utils.LocalEvents;
+using HarmonyLib;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.Party.PartyComponents;
+using TaleWorlds.CampaignSystem.Settlements;
+using Xunit;
+
+namespace GameInterface.Tests.AutoSync;
+
+/// <summary>
+/// Verifies that AutoSync drops dangerous trivial stubs while preserving nontrivial methods whose
+/// existing Harmony detours can be required JIT anti-inlining boundaries.
+/// </summary>
+public class AutoSyncTargetSelectionTests
+{
+    [Fact]
+    public void Keeps_nontrivial_methods_but_drops_empty_and_constant_stubs()
+    {
+        var targets = GenericPatches<SelectionPatch, SelectionSample>
+            .TranspilerTargets(true)
+            .ToArray();
+
+        Assert.Contains(typeof(SelectionSample).GetConstructor(Type.EmptyTypes)!, targets);
+        Assert.Contains(SampleMethod(nameof(SelectionSample.StoreSyncedScalar)), targets);
+        Assert.Contains(SampleMethod(nameof(SelectionSample.MutateSyncedList)), targets);
+        Assert.Contains(SampleMethod(nameof(SelectionSample.MutateSyncedArray)), targets);
+        Assert.Contains(SampleMethod(nameof(SelectionSample.MutateSyncedProperty)), targets);
+        Assert.Contains(SampleMethod(nameof(SelectionSample.AssignSyncedProperty)), targets);
+        Assert.Contains(SampleMethod(nameof(SelectionSample.ReadSyncedList)), targets);
+        Assert.Contains(SampleMethod(nameof(SelectionSample.ReadSyncedProperty)), targets);
+        Assert.Contains(SampleMethod(nameof(SelectionSample.ForwardToScalarStore)), targets);
+        Assert.Contains(SampleMethod(nameof(SelectionSample.ReadSyncedScalar)), targets);
+        Assert.Contains(SampleMethod(nameof(SelectionSample.ReadSyncedArray)), targets);
+        Assert.Contains(SampleMethod(nameof(SelectionSample.StoreOtherScalar)), targets);
+
+        Assert.DoesNotContain(SampleMethod(nameof(SelectionSample.ConstantFalse)), targets);
+        Assert.DoesNotContain(SampleMethod(nameof(SelectionSample.ConstantLong)), targets);
+        Assert.DoesNotContain(SampleMethod(nameof(SelectionSample.ConstantNull)), targets);
+        Assert.DoesNotContain(SampleMethod(nameof(SelectionSample.Empty)), targets);
+    }
+
+    [Fact]
+    public void Drops_the_reported_clan_interface_getter()
+    {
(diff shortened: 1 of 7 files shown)


ShoT-UPfps:
closing this since this hasnt reproduced in a long time.

PR #2928, opened 2026-08-10
[1574] Relevance-gate world AI troop XP to the controlling client

### LINKED ISSUE #1574: Relevance-gate world AI troop XP to the controlling client
this issue came out of the investigation into #1557 (sustained campaign-map AI-sync bandwidth, ~250-380 KB/s, almost all world-wide AI): we profiled the server packet stream, traced the worst offenders, and split the durable fixes into separate issues. `NetworkTroopRosterSetXp` is the #2 profiler line: per-troop XP for every AI party, broadcast to all clients.

durability: clients never read AI-party troop XP, and this rests on a PERMANENT invariant, not on a temp disable. even with auto-upgrade/training re-enabled, the periodic ticks that drive them run server-only (`PartyTickPatch.cs:13` TickPeriodicEvents => IsServer, `HourlyPartyTickServerPatch.cs:13`), and the client is a passive AI mirror (`MobilePartyAIDisablePatches.cs:19`) so it never computes upgrades; the resulting roster composition arrives separately as count deltas. no UI shows an enemy party's per-troop XP.

durable fix (a targeted area-of-interest slice, NOT a static drop):
- gate the SetXp send to the controlling client: in `TroopRosterDeltaHandler.Handle_ElementXpSet` (`TroopRosterDeltaHandler.cs:97-102`) skip/relevance-route when the roster's owning party isnt player-controlled (resolve owner, then `MobilePartyExtensions.IsPlayerParty`), ideally `network.Send(peer,...)` the controlling client instead of SendAll
- AOI baseline: when a party transitions AI to player-controlled, re-sync that roster's per-element XP once (bounded snapshot) otherwise newly controlled parties show stale XP bars
- the same relevance applies to the XpChange rider inside `NetworkTroopRosterAddCounts` (`TroopRosterDeltaHandler.cs:80`) where re-enabled training/upgrade XP will actually flow. dont drop AddCounts (the count change IS roster composition the client needs), just gate the Xp component for AI parties
- for retained player parties, coalesce SetXp latest-wins per (roster,element) per tick (SetElementXp is an absolute set)

this is the first concrete slice of the area-of-interest epic and could be folded into it.

depends on #1576 (the reusable per-tick coalescer): the retained-player-party SetXp coalescing should use the shared coalescer (per-(roster,element) latest-wins key). the relevance gate itself is independent.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Troop XP changes for every AI-controlled mobile party are sent to every connected client, even though clients do not use or display that XP. A player who joins or reconnects also has no bounded way to receive the current troop XP for their own party.

## What is the new behavior?

Resolves #1574

World AI troop XP is no longer sent to unrelated clients. Each player receives full troop XP for their own party and the non-player-controlled parties in their clan, while all clients still receive roster composition changes. Joining or reconnecting restores the current member and prisoner troop XP for those relevant parties.

Over the same 120-second maximum-fast-forward window with one connected client:

| Metric | Development | This PR | Reduction |
| --- | ---: | ---: | ---: |
| Packets received | 461,654 | 258,949 | 43.91% |
| Serialized bytes received | 24,526,581 | 12,858,935 | 47.57% |

## Bot Changelog Entry

- Stop sending world AI troop XP to unrelated players while keeping each player's own and same-clan party XP up to date.


garrettluskey (approved):
The routing keeps roster composition broadcast while sending regular-troop XP only to the controlling or relevant clan peers, preserves per-instance ordering across targeted coalesced sends, and restores bounded relevant XP during join. The controller, observer, disconnected-player, multi-clan-peer, and baseline paths have focused coverage; the build and E2E checks pass, with CodeFactor reporting only fixture/provider complexity notices. Approved.

PR #2932, opened 2026-08-11
[2799] Improve adaptive battle movement recovery

### LINKED ISSUE #2799: adaptive movement hz in battles

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Battle movement can remain at 10 Hz after temporary load clears, even when the client is again meeting its configured frame limit. A deliberate 30 FPS cap is treated as overload, and overly strict recovery thresholds can prevent the rate from climbing back toward the highest stable tier.

## What is the new behavior?

Resolves #2799

Battle movement still drops immediately when measured load requires it, then recovers one tier at a time toward the highest rate supported by the client's configured frame limit and current movement work. Clients capped at 30 FPS can recover to 30 Hz, clients with more headroom can climb higher, failed higher-rate probes are learned without leaving the rate permanently low, and one struggling receiver still does not lower movement cadence for other clients.

## Bot Changelog Entry

- Improved the adaptive networking algorithm during battles which should improve syncing and FPS.


source/E2E.Tests/Services/Missions/MovementRateControllerTests.cs
```diff
@@ -408,8 +783,40 @@ public void AdvanceClock(float seconds)
...
+        public void AdvanceWindow(
+            int framesPerSecond,
+            double senderMilliseconds = 0d,
+            double prioritySenderMilliseconds = 0d,
+            double receiverApplyMilliseconds = 0d,
+            double maximumReceiverQueueMilliseconds = 0d)
         {
+            if (senderMilliseconds > 0d)
+            {
+                int bulkReports = Controller.Snapshot.BulkHz;
```
garrettluskey: this reports `BulkHz` polls even when `framesPerSecond` is lower, but production can only report once per mission tick. drive this through the real `PollMovement` cadence so the retry baseline matches runtime.

garrettluskey (dismissed):
the controller now separates transient overload from the configured frame ceiling and recovers one tier at a time. the frame-rate recovery regression feeds it a send cadence production cannot produce, so it doesnt verify the runtime path yet.

source/E2E.Tests/Services/Missions/MovementRateControllerTests.cs
ShoT-UPfps: fixed. the regression now runs through `PollMovement`; only polls that include authoritative bulk movement receive the modeled send cost.

source/Missions/Agents/Handlers/MovementRateController.cs
```diff
@@ -550,19 +604,127 @@ private void EvaluateBattleRate()
...
+        int currentFrameCeilingHz = CalculateSenderCeiling(
+            lastFramesPerSecond,
+            0d,
+            effectiveFrameLimitHz);
+        bool frameRateImproved =
+            (baselineFrameCeilingHz < candidate && currentFrameCeilingHz >= candidate) ||
+            (rejectedRate.LowerTierFramesPerSecond > 0f &&
+                lastFramesPerSecond >=
+                    rejectedRate.LowerTierFramesPerSecond / RejectedRateRetryCostRatio);
+        if (!bulkCostImproved && !senderCostImproved && !frameRateImproved)
```
garrettluskey, line 660: this can blacklist a tier forever after repeated transient failures. `BattleProfile_FrameRateRecoveryRetriesAfterConfirmationFailure` returns to 60 FPS / 70 ms for 12 windows but remains at 40 Hz because the baseline was captured before the failed probe and none of these improvement checks can pass. expire the rejection or compare against the failed probe conditions.

source/Missions/Agents/Handlers/MovementRateController.cs
```diff
@@ -289,9 +320,11 @@ public void Configure(MovementCadenceProfile profile)
                     localReason = "tournament-fixed";
                     break;
                 case MovementCadenceProfile.Battle:
-                    performanceCeilingHz = 60;
-                    localAdaptiveHz = 40;
-                    automaticReceiverCapHz = 40;
+                    // Start every battle from the same baseline, then let measured windows adapt it.
+                    const int startingBattleHz = 40;
+                    performanceCeilingHz = NormalizeRate(effectiveFrameLimitHz);
+                    localAdaptiveHz = startingBattleHz;
```
garrettluskey: cap the starting rate and advertised receiver cap to `performanceCeilingHz`. with a configured 30 FPS limit this computes a 30 Hz ceiling above but runs and advertises 40 Hz until the first window.

source/E2E.Tests/Services/Missions/MovementRateControllerTests.cs
```diff
@@ -408,8 +758,40 @@ public void AdvanceClock(float seconds)
...
```
garrettluskey: this still fabricates `BulkHz` polls when mission FPS is lower. drive a repeated probe failure and recovery through `PollMovement` so the `Failures > 1` behavior is tested with production cadence.

garrettluskey (dismissed):
the controller now separates sender and receiver pressure and recovers one tier at a time, but repeated probe failures can still recreate the permanent-low-rate behavior this is meant to fix.

source/Missions/Agents/Handlers/MovementRateController.cs
ShoT-UPfps: fixed. retries now compare against the conditions that actually rejected the probe, so later healthy conditions can retry the tier.

source/Missions/Agents/Handlers/MovementRateController.cs
ShoT-UPfps: fixed. battle startup and the initial receiver-cap advertisement are now capped to the configured performance ceiling.

source/E2E.Tests/Services/Missions/MovementRateControllerTests.cs
ShoT-UPfps: fixed. repeated failures and recovery now run through `AgentMovementHandler.PollMovement`, and the test counts completed controller windows from production advertisements.

garrettluskey (approved):
The controller now caps its initial advertisement to the configured frame ceiling and retries rejected tiers against the conditions that actually failed. The repeated-failure regression runs through the production PollMovement cadence, and the checks are green. Approved.

PR #2776, opened 2026-08-06
[2704][2933] Release parties stuck in abandoned map events

ShoT-UPfps:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] All new classes have class-level documentation comments, if there are any at all
- [x] Tests for the changes have been added (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update
- [ ] Other... Please describe:

## What is the current behavior?

Parties can remain grouped at an old battle location after the server reloads a save. This can persist when the interrupted battle belongs to an offline player, leaving the affected parties stuck near Lageta instead of resuming normal campaign movement.

## What is the new behavior?

Resolves #2704
Resolves #2933

Server startup now clears interrupted player battles even when only a saved player party still references them, releases and parks offline player parties, and lets the other affected parties resume or disband while ongoing AI-only battles continue normally.
image

## Bot Changelog Entry

- Fixed an issue where AI parties would get permanently stuck in a map event.

PR #2967, opened 2026-08-12
Fix MapEvent AI freeze: repopulate fallback TroopUpgradeTracker on load

Thaipho:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

`MapEvent.TroopUpgradeTracker` is not saved, so after loading a save it comes back `null` for any battle that was still in progress at save time — this is normal, expected vanilla behavior. Coop's own fallback (a Harmony postfix on the `TroopUpgradeTracker` getter, in `MapEventRobustnessPatches.cs`) replaces that `null` with a brand-new, empty `TroopUpgradeTracker` — but unlike vanilla's own `MapEvent.OnAfterLoad()` self-heal, it never actually populates the new tracker with the battle's parties

Any later call into `TroopUpgradeTracker.CalculateReadyToUpgradeSafe` (reached via `CheckUpgradedCount`, called every `MapEvent.Update()`) looks up the owning party in the tracker's internal party list and dereferences the result without a null check. Against the empty fallback tracker that lookup always misses, so it throws a `NullReferenceException` on every tick for that MapEvent. Since the exception fires before `Update()` can ever progress the battle to completion, the MapEvent can never resolve — leaving whichever parties were actually fighting in it stuck in that battle indefinitely (observed in-game as AI lords going permanently immobile after loading a save with in-progress battles, while unrelated AI/villages/caravans kept working normally).

## What is the new behavior?
The fallback in `MapEventRobustnessPatches.cs` now repopulates the replacement `TroopUpgradeTracker` the same way vanilla's `OnAfterLoad()` does: it calls `AddParty` for every party on both the attacker and defender sides before assigning it back to the `MapEvent`. This keeps the crash from ever happening in the first place, so affected battles can resolve normally instead of freezing forever.

Added `MapEventRobustnessPatchesTests.cs` with a regression test that forces `TroopUpgradeTracker` to `null` on a live `MapEvent`, triggers the fallback through the real getter, and asserts every involved party was re-registered in the restored tracker.

## Bot Changelog Entry
Fixed a bug where some AI lords could get stuck frozen in battle after loading a save, because their fight could never resolve.

## Other information
decided to patch this after running into it in a live game with my friend preventing me and my friend from playing that save until this is patched

### DIFF
--- a/source/E2E.Tests/Services/MapEvents/MapEventRobustnessPatchesTests.cs
+++ b/source/E2E.Tests/Services/MapEvents/MapEventRobustnessPatchesTests.cs
@@ -1,0 +1,45 @@
+using HarmonyLib;
+using System.Collections;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.MapEvents;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace E2E.Tests.Services.MapEvents;
+
+public class MapEventRobustnessPatchesTests : MapEventTestBase
+{
+    public MapEventRobustnessPatchesTests(ITestOutputHelper output) : base(output) { }
+
+    [Fact]
+    public void TroopUpgradeTracker_WhenNullAfterLoad_IsRepopulatedWithEveryInvolvedParty()
+    {
+        var context = CreateServerMapEvent();
+
+        Server.Call(() =>
+        {
+            Assert.True(Server.ObjectManager.TryGetObject<MapEvent>(context.MapEventId, out var mapEvent));
+
+            var trackerField = AccessTools.Field(typeof(MapEvent), "<TroopUpgradeTracker>k__BackingField");
+            Assert.NotNull(trackerField);
+            trackerField.SetValue(mapEvent, null);
+
+            var restored = mapEvent.TroopUpgradeTracker;
+            Assert.NotNull(restored);
+
+            var mapEventPartiesField = AccessTools.Field(typeof(TroopUpgradeTracker), "_mapEventParties");
+            Assert.NotNull(mapEventPartiesField);
+            var mapEventParties = (IList)mapEventPartiesField.GetValue(restored)!;
+
+            var involvedParties = new List<MapEventParty>();
+            involvedParties.AddRange(mapEvent.AttackerSide.Parties);
+            involvedParties.AddRange(mapEvent.DefenderSide.Parties);
+
+            Assert.Equal(involvedParties.Count, mapEventParties.Count);
+            foreach (var party in involvedParties)
+            {
+                Assert.Contains(party, mapEventParties.Cast<MapEventParty>());
+            }
+        }, MapEventDisabledMethods);
+    }
+}
--- a/source/GameInterface/Services/MapEvents/Patches/MapEventRobustnessPatches.cs
+++ b/source/GameInterface/Services/MapEvents/Patches/MapEventRobustnessPatches.cs
@@ -22,12 +22,8 @@ internal class MapEventRobustnessPatches
     {
         if (__result is null)
         {
-            // The assignment below runs the generated AutoSync setter prefix, which reads this getter
-            // to compare values. Let that nested read observe null instead of recursively restoring again.
             if (restoringTroopUpgradeTracker) return;
 
-            // Pending graphs are incomplete by design. A fallback here would replace the registered
-            // tracker before its queued reference apply reaches the game thread.
             if (ContainerProvider.TryResolve<IMapEventInitializationBarrier>(out var barrier) &&
                 barrier.IsPending(__instance))
             {
@@ -38,7 +34,17 @@ internal class MapEventRobustnessPatches
             restoringTroopUpgradeTracker = true;
             try
             {
-                __result = new TroopUpgradeTracker();
+                var restoredTracker = new TroopUpgradeTracker();
+                foreach (var party in __instance.AttackerSide.Parties)
+                {
+                    restoredTracker.AddParty(party);
+                }
+                foreach (var party in __instance.DefenderSide.Parties)
+                {
+                    restoredTracker.AddParty(party);
+                }
+
+                __result = restoredTracker;
                 __instance.TroopUpgradeTracker = __result;
             }
             finally


ShoT-UPfps:
is this already fixed by https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/2776?

ShoT-UPfps (changes requested):
cmt

Thaipho:
@ShoT-UPfps 

This PR fixes a crash in `MapEventRobustnessPatches.cs`: when `MapEvent.TroopUpgradeTracker` comes back null after a save load (normal/expected for any in-progress battle, since it isn't a saved field), our fallback replaced it with an *empty* tracker instead of populating it like vanilla's own `OnAfterLoad()` does. Any subsequent `CheckUpgradedCount` call against that empty tracker throws an NRE on every tick, which permanently blocks `MapEvent.Update()` for that battle — the fight can never resolve through the normal simulation loop, regardless of whether a player is involved.

Here's the actual crash from a live session — the warning is our fallback firing (confirming the tracker came back empty), immediately followed by the NRE:

03:44:37.964 [Coop] "TroopUpgradeTracker" was not properly set for MapEvent "Created_88651"
03:44:37.964 [Coop] "TroopUpgradeTracker" was not properly set for MapEvent "Created_88763"
03:44:37.964 [Coop] "TroopUpgradeTracker" was not properly set for MapEvent "Created_88803"
...(12 distinct MapEvents total)...
03:44:38.104 [Coop] Failed to run "Game.OnTick"
03:44:38.104 System.NullReferenceException: Object reference not set to an instance of an object.
03:44:38.104 at TaleWorlds.CampaignSystem.MapEvents.MapEvent.Update_Patch2(MapEvent this)
03:44:38.104 at TaleWorlds.CampaignSystem.MapEvents.MapEventManager.Tick()
03:44:38.104 at TaleWorlds.CampaignSystem.Campaign.Tick_Patch1(Campaign this)
03:44:38.104 at TaleWorlds.CampaignSystem.GameState.MapState.OnMapModeTick_Patch1(MapState this, Single dt)

That last block repeats identically thousands of times per session (32,416 / 29,398 occurrences across two separate log captures) — one entry per broken MapEvent, every single tick, forever, until the process restarts.

#2776's `MapEventLoadCleaner` solves a different problem: a MapEvent tied to a *disconnected* player has no natural way to resolve even without any crash, since nobody's there to drive the player-side decision point. That's a real structural gap on its own, handled with an explicit startup cleanup/repair pass, it doesn't touch `TroopUpgradeTracker` or the NRE at all.

Checked the two issues #2776 references too: #2704 (8 lords sortieing out and returning to a rally point together) doesn't match our crash's signature — that looks like a separate AI holding-pattern bug. #2933 is plausibly related, but has no server log attached, so I can't confirm it's actually this NRE versus the offline-player scenario #2776 targets.

If anything the two are complementary rather than redundant: #2776's `FinalizeOrRepairEvent` path can hit the same `TroopUpgradeTracker` getter our postfix patches, so once this merges it makes their cleanup path more crash-safe too.

source/GameInterface/Services/MapEvents/Patches/MapEventRobustnessPatches.cs
```diff
@@ -38,7 +34,17 @@ private static void PostfixTroopUpgradeTracker(MapEvent __instance, ref TroopUpg
             restoringTroopUpgradeTracker = true;
             try
             {
-                __result = new TroopUpgradeTracker();
+                var restoredTracker = new TroopUpgradeTracker();
+                foreach (var party in __instance.AttackerSide.Parties)
+                {
+                    restoredTracker.AddParty(party);
```
ShoT-UPfps, line 40: these `AddParty` calls only update the server tracker. connected clients receive the replacement object and property reference, but `TroopUpgradeTrackerRegistry.OnClientCreated` leaves `_mapEventParties` empty and `AddParty` isnt synced

source/E2E.Tests/Services/MapEvents/MapEventRobustnessPatchesTests.cs
```diff
@@ -0,0 +1,45 @@
...
+using Xunit.Abstractions;
+
+namespace E2E.Tests.Services.MapEvents;
+
+public class MapEventRobustnessPatchesTests : MapEventTestBase
+{
+    public MapEventRobustnessPatchesTests(ITestOutputHelper output) : base(output) { }
+
+    [Fact]
+    public void TroopUpgradeTracker_WhenNullAfterLoad_IsRepopulatedWithEveryInvolvedParty()
```
ShoT-UPfps, line 15: this doesnt reproduce the reported freeze. `TroopUpgradeTrackerUntrackedPartyPatch` already returns 0 for owners missing from `_mapEventParties`, and vanilla `MapEvent.Update` doesnt call `CheckUpgradedCount`; can this exercise the actual stuck-battle path before asserting this fixes it?

source/E2E.Tests/Services/MapEvents/MapEventRobustnessPatchesTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            Assert.True(Server.ObjectManager.TryGetObject<MapEvent>(context.MapEventId, out var mapEvent));
+
+            var trackerField = AccessTools.Field(typeof(MapEvent), "<TroopUpgradeTracker>k__BackingField");
+            Assert.NotNull(trackerField);
+            trackerField.SetValue(mapEvent, null);
+
+            var restored = mapEvent.TroopUpgradeTracker;
+            Assert.NotNull(restored);
+
+            var mapEventPartiesField = AccessTools.Field(typeof(TroopUpgradeTracker), "_mapEventParties");
```
ShoT-UPfps, line 30: use the publicized `restored._mapEventParties` directly here

source/GameInterface/Services/MapEvents/Patches/MapEventRobustnessPatches.cs
```diff
@@ -22,12 +22,8 @@ private static void PostfixTroopUpgradeTracker(MapEvent __instance, ref TroopUpg
     {
         if (__result is null)
         {
-            // The assignment below runs the generated AutoSync setter prefix, which reads this getter
-            // to compare values. Let that nested read observe null instead of recursively restoring again.
             if (restoringTroopUpgradeTracker) return;
```
ShoT-UPfps, line 25: please keep the comments explaining these two guards, neither one is obvious from the code

ShoT-UPfps (changes requested):
cmts

source/E2E.Tests/Services/MapEvents/MapEventRobustnessPatchesTests.cs
Thaipho: your right thought i had it but will close this PR and open a issue bug report with my logs so someone else can find it

PR #2975, opened 2026-08-12
Potential Fix to the Battle Simulation Cancel and Restart

### LINKED ISSUE #2407: [BATTLE SIMULATION] Send Troops button does nothing after canceling a simulation battle
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

After canceling a simulation battle, the player gets stuck because clicking Send Troops does nothing.

## Steps to reproduce

1. Cancel a simulation battle.
2. Click Send Troops.
3. The action produces no response.

## Additional information

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1530970137975590952/1530970137975590952

## Images and videos

[](<https://cdn.discordapp.com/attachments/1530970137975590952/1530970138877104331/image.png?ex=6aa4d04d&is=6aa37ecd&hm=f11a6ead0a05d8fd5b257593c45441b695320d769699547bb4ebe5cbb0a295e3&>) ([Discord message](https://discord.com/channels/709516043332354119/1530970137975590952/1530970137975590952))

## Logs

Client log: [0.0.3+7003060dde3e44abc1d0e243c45ba9a6627516cb - Download Coop_client_-_2.log](<https://cdn.discordapp.com/attachments/1530970137975590952/1530970139745456149/Coop_client_-_2.log?ex=6aa4d04d&is=6aa37ecd&hm=68fb76136bc9f7978878c6eed98aa98ae585da6e73afbb985676fe5faad73352&>)

Server log: Not provided

codebydean:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Cancelling an unfinished Send Troops battle simulation leaves the simulation active on the server. Trying to use Send Troops again for the same battle is rejected and appears to do nothing.

## What is the new behavior?

Resolves #2407

Cancelling an unfinished battle simulation now notifies the server. The server cleans up the simulation, releases the battle mode, and allows Send Troops to be used again for the same battle.

Tests were added for cancellation message serialization, simulation claim release, and restarting a cancelled simulation.

## Bot Changelog Entry

- Fixed Send Troops doing nothing after cancelling a battle simulation

## Other information

Tested locally in-game by cancelling an unfinished Send Troops simulation and starting it again for the same battle.


source/GameInterface/Services/MapEvents/BattleSimulationReplay.cs
```diff
@@ -183,6 +183,19 @@ private static void ApplyRound(BattleSimulation simulation, ResolvedChange[] rou
...
+    // Ends an unfinished local playback and returns its MapEventId;
+    // Spectators and simulations already completed by the server do not request cancellation.
+    public static bool TryCancel(out string cancelledMapEventId)
+    {
+        cancelledMapEventId = null;
+
+        if (mapEventId == null || finishRequested || isSpectator) return false;
+        
+        cancelledMapEventId = mapEventId;
+        Reset();
```
ShoT-UPfps, line 209: could this wait for the server cancellation result before calling `Reset()`? if the preceding advance finishes the battle, the server sends loot and finish then ignores the cancellation, but this client has already cleared `mapEventId` so it drops both messages and misses the completed result

source/GameInterface/Services/MapEvents/Handlers/BattleSimulationRunHandler.cs
```diff
@@ -218,6 +230,44 @@ private void Handle_NetworkBattleStartRequest(MessagePayload<NetworkBattleStartR
...
+            }
+            activeSimulations.Remove(mapEventId);
+        }
+        
+        GameThread.RunSafe(() =>
+        { EndSimulationSession(simulation); }, blocking: true, context: nameof(Handle_NetworkCancelBattleSimulation));
+
+        if (ServerBattleModeArbiter.ReleaseSimulation(mapEventId))
+        {
+            network.SendAll(new NetworkBattleSimulationFinished(mapEventId));
```
ShoT-UPfps: should cancellation use a separate message from `NetworkBattleSimulationFinished`? spectators still have an active replay, so this calls `BattleResultsReady` with no winner and shows them a defeat instead of closing the cancelled simulation

ShoT-UPfps (changes requested):
cmts

source/GameInterface/Services/MapEvents/BattleSimulationReplay.cs
garrettluskey, line 209: Resetting here makes cancellation fire-and-forget. If an advance has already completed on the server, its loot and finish packets are valid, but this client now drops them because the replay ID is gone; the server then ignores the late cancellation because the active simulation was already removed. Keep the replay in a cancelling state until the server acknowledges whether cancellation won, then reset on that acknowledgement—or process normal completion if it won the race.

source/GameInterface/Services/MapEvents/Handlers/BattleSimulationRunHandler.cs
garrettluskey: `NetworkBattleSimulationFinished` means a real battle result everywhere else. Spectators still have an active replay, so this drives `RequestFinish()` and eventually `BattleResultsReady()` with no winner, presenting a cancellation as a completed/defeated battle. Send a distinct cancellation response and have clients close/reset playback without invoking result handling; that response can also acknowledge the initiator’s request.

garrettluskey (dismissed):
The ownership check, server-side cleanup, and regression coverage are solid. I’m requesting changes because the cancellation protocol currently races normal completion and reuses the normal-finish signal for spectators; both paths can produce incorrect client state.

codebydean:
Currently, the cancellation is updated to wait for the server's response instead of immediately clearing the client replay state like before, allowing normal loot and completion handling to win if it races cancellation.

Added a `NetworkBattleSimulationCancelled` response so initiating clients and spectators reset canceled playback without invoking normal battle result handling, preventing duplicate cancellation requests and further advances while cancellation is pending.

The focused serialization and cancellation tests passed.

source/GameInterface/Services/MapEvents/Handlers/BattleSimulationRunHandler.cs
ShoT-UPfps: this still leaves spectators stuck in the simulation scoreboard. `ConfirmCancellation` only resets the replay, but spectators never called `MapState.EndBattleSimulation`, so their active simulation view stays open

ShoT-UPfps (changes requested):
cmt

source/GameInterface/Services/MapEvents/Handlers/BattleSimulationRunHandler.cs
codebydean: > this still leaves spectators stuck in the simulation scoreboard. `ConfirmCancellation` only resets the replay, but spectators never called `MapState.EndBattleSimulation`, so their active simulation view stays open

With the new commit it checks if the client is the spectator and if they are in the MapState. If both are true then it calls `MapState.EndBattleSimulation` to close the actual scoreboard @ShoT-UPfps

source/GameInterface/Services/MapEvents/Handlers/BattleSimulationRunHandler.cs
```diff
@@ -641,6 +693,29 @@ private void Handle_NetworkBattleSimulationFinished(MessagePayload<NetworkBattle
...
+        {
+            if (!BattleSimulationReplay.IsActiveFor(payload.What.MapEventId)) return;
+
+            bool isSpectator = BattleSimulationReplay.IsSpectator;
+            
+            BattleSimulationReplay.ConfirmCancellation(payload.What.MapEventId);
+
+            if (isSpectator && Game.Current?.GameStateManager?.ActiveState is MapState mapState )
+            {
+                mapState.EndBattleSimulation();
```
ShoT-UPfps, line 717: should this also clear `PlayerEncounter.Current.BattleSimulation`? `EndBattleSimulation()` only clears the map state simulation, so the next spectator open returns at line 745 because the old simulation is still non-null

ShoT-UPfps (changes requested):
cmt

ShoT-UPfps (approved):
lgtm

codebydean:
@garrettluskey need your approval for this so we can merge

PR #2992, opened 2026-08-13
[2938] Grow the mobile-party visual buffer

### LINKED ISSUE #2938: [WORLD MAP] Campaign load crashes when visible mobile parties overflow the fixed visual buffer
## Severity of issue/bug

- [x] High
- [ ] Average
- [ ] Low

Loading the affected existing co-op campaign crashes the graphical host/client while `MapScreen` initializes its mobile-party visuals. The dedicated server can reach its normal serving state, but the client cannot finish entering the campaign map.

The crash dump contains the managed inner exception:

```text
System.IndexOutOfRangeException
  SandBox.View.Map.Visuals.MobilePartyVisual.Tick(...)
  SandBox.View.Map.Managers.MobilePartyVisualManager.AddNewPartyVisualForParty(...)
  SandBox.View.Map.Managers.MobilePartyVisualManager.OnInitialize()
  SandBox.View.Map.MapScreen..ctor(...)
```

This is a deterministic overflow of Bannerlord's fixed `MobilePartyVisual[2500] _dirtyPartiesList`. Vanilla `OnInitialize()` does not reset `_dirtyPartyVisualCount` from its default `0`; each initially dirty visible party increments the count before indexing the array. The effective safe startup capacity is therefore 2,499 visuals.

The supplied save contains:

| Saved state | Count |
| --- | ---: |
| Campaign mobile parties | 3,135 |
| Saved visible mobile parties | 2,759 |
| Visible parties receiving visuals after excluding garrisons and militias | **2,556** |
| Effective vanilla startup capacity | **2,499** |

The 2,500th visible visual writes index `2500` and throws.

## Steps to reproduce

1. Load a campaign containing at least 2,500 saved-visible mobile parties after excluding garrison and militia parties.
2. Enter the graphical campaign map so `MobilePartyVisualManager.OnInitialize()` enumerates `MobileParty.All` with `shouldTick: true`.
3. Observe `IndexOutOfRangeException` in `MobilePartyVisual.Tick` while it appends to `_dirtyPartiesList`.

The reporter reproduces this consistently near the end of hosting/loading the supplied existing campaign on Bannerlord `v1.4.8.119303` with Coop `v0.1.2.0`.

## Expected behavior

Party-visual initialization and normal ticks must size or grow the dirty-visual buffer for the number of visuals being processed. Large long-running campaigns must not become permanently unloadable after crossing a fixed party-count threshold.

## Additional information

- Coop already replaces `MobilePartyVisualManager.OnTick` in `MobilePartyVisualManagerPatches`, but that patch continues using the vanilla 2,500-entry array and does not cover vanilla `OnInitialize`.
- A root fix should cover both initialization and later ticks, reset the dirty count correctly, and avoid a parallel write beyond the current array capacity.
- A bounded save workaround was produced without deleting parties, troops, heroes, or registrations. It cleared only the persisted `IsVisible` flag on the final 57 visible bandit parties, reducing startup visuals from 2,556 to 2,499. Bannerlord's normal `PartyBase.UpdateVisibilityAndInspected` path can recalculate those transient visibility flags after the map loads.
- #2943 is a duplicate join-load report from the same alleged capacity failure. That reporter raised the bandit/looter multiplier to `10` and then `5`, the client completed the 51.6 MB campaign transfer and entered `GameLoadingState`, and the crash recurred after wiping/reinstalling. The attached screenshot shows a captured ~1 GB `dump.dmp`; the reporter also used this issue's diagnosis as their reference. Those facts fit the same excess-mobile-party startup path even though the attached short client log ends before `MapScreen.OnInitialize` and does not contain the final stack.
- #1701 and #1392 are related manifestations in `MobilePartyVisual.Tick` using the same fixed 2,500-entry dirty-visual buffer during Coop's patched runtime tick. They were observed as `AccessViolationException` rather than this deterministic managed initialization overflow, so they remain separate until a common failing index/race is established.

## Discord source

https://discord.com/channels/709516043332354119/1536532171043381398/1536532171043381398

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Loading a long-running campaign with at least 2,500 visible mobile parties crashes the graphical client before the campaign map opens.

## What is the new behavior?

Resolves #2938

Large campaigns now load into the campaign map without overflowing the mobile-party visual buffer, and the buffer remains large enough as party counts change.

## Bot Changelog Entry

- Fixed a crash when loading campaigns with very large numbers of mobile parties.

### DIFF
--- a/source/GameInterface/Services/PartyVisuals/Patches/MobilePartyVisualManagerPatches.cs
+++ b/source/GameInterface/Services/PartyVisuals/Patches/MobilePartyVisualManagerPatches.cs
@@ -4,6 +4,9 @@ using SandBox.View.Map.Managers;
 using SandBox.View.Map.Visuals;
 using Serilog;
 using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Reflection;
 using TaleWorlds.CampaignSystem.Party;
 using TaleWorlds.Library;
 using TaleWorlds.MountAndBlade;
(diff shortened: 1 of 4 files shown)


source/GameInterface/Services/PartyVisuals/Patches/MobilePartyVisualManagerPatches.cs
```diff
@@ -13,13 +15,26 @@ public class MobilePartyVisualManagerPatches
     {
         private static ILogger Logger = LogManager.GetLogger<MobilePartyVisualManagerPatches>();
 
+        [HarmonyPatch(nameof(MobilePartyVisualManager.OnInitialize))]
```
garrettluskey, line 21: This only patches `SandBox.View.Map.Managers.MobilePartyVisualManager`. With the naval DLC active, `NavalMobilePartyVisualManager` independently creates a visual for every non-garrison/non-militia party, overrides both `OnInitialize` and `OnTick`, and owns another fixed `NavalMobilePartyVisual[2500]`; it does not call these methods. A 2,500+ party save can therefore overflow that buffer during its initialization even though this one was grown. Apply the same capacity guard to the naval manager (without making the DLC a mandatory runtime dependency if it is optional), and make the live fixture assert both active managers.

garrettluskey (dismissed):
Growing the land-party visual buffer addresses the vanilla manager overflow and the resize logic is sound. The naval visual manager has the same independent 2,500-entry buffer, so DLC clients are still exposed to the same load crash.

source/GameInterface/Services/PartyVisuals/Patches/MobilePartyVisualManagerPatches.cs
ShoT-UPfps: fixed, the optional naval manager now grows its independent buffer on initialize and tick, and the DEBUG buffer state checks both managers when NavalDLC is active

garrettluskey (approved):
Both land and naval managers now grow before initialization/tick writes, while preserving the existing buffers and optional DLC boundary. looks good.

PR #3034, opened 2026-08-14
Fix AutoSync interception during parallel client save loads

### LINKED ISSUE #2999: [Report] Join stops after host save data arrives, client crashes when it loads the save (Linux/Mono)
## Severity of issue/bug

- [ ] High
- [x] Average
- [ ] Low

A Linux client stops on the loading screen after it receives the host save data. The join does not continue. The UI shows no timeout and no error.

The client does not wait for the network. The client crashes when it loads the save. The sequence is:

1. `GameStateInterface.InteralLoadSaveGame` calls `SaveManager.Load`.
2. The save system fills objects on `NativeParallelDriver` worker threads.
3. AutoSync-patched property setters throw `NullReferenceException` on those threads.
4. The exceptions leave the parallel loop. The load does not return a root object.
5. `InteralLoadSaveGame` then throws `NullReferenceException` at offset `[0x00056]`.
6. The join state machine stops. The loading screen stays.

Each failing property gives this stack:

```
System.Reflection.TargetInvocationException ---> System.NullReferenceException
  at TaleWorlds.CampaignSystem.Hero.set_Weight (System.Single value)
  at System.Reflection.RuntimeMethodInfo.Invoke (...)
  at TaleWorlds.SaveSystem.Load.PropertyLoadData.FillObject ()
  at TaleWorlds.SaveSystem.Load.ObjectLoadData.FillObject ()
  at TaleWorlds.SaveSystem.Load.LoadContext.CreateLoadData (...)
  at TaleWorlds.SaveSystem.Load.LoadContext+<>c__DisplayClass22_0.<Load>b__2 (...)
  at TaleWorlds.Engine.NativeParallelDriver.ParalelForLoopBodyCaller (...)
```

The property that throws changes between attempts. One bad field in the save is therefore not a likely cause. Two joins to the same host used the same build and the same save:

| | 12:59 attempt | 13:09 attempt |
|---|---|---|
| Unhandled exceptions | 12 | 30 |
| Throwing setters | `PartyBase.set_ItemRoster`, `Hero.set_PartyBelongedToAsPrisoner` | `Hero.set_Weight` (27x), `PartyBase.set_ItemRoster` (1x) |
| Preceding `Failed to get` ids | `village_K4_3`, `MemberRoster_Created_3876` | `town_EW6`, `caravan_template_sturgia_414`, `Created_21401`, `SiegeEngineConstructionProgress_Created_76` |

Both attempts end with the same `InteralLoadSaveGame` error at the same offset.

Each attempt also writes many `[ObjectManager] Failed to get ...` errors. These errors occur approximately 12 seconds before the crash. They name objects that the host referenced. The ids are different in each attempt. These missing objects may be the primary problem instead of the save load.

The logs show these facts:

- The generated AutoSync setter prefix calls `CallOriginalPolicy.IsOriginalAllowed()` first.
- `CallOriginalPolicy.IsOriginalAllowed()` checks `AllowedThread.IsThisThreadAllowed()`. This check applies to one thread only.
- The same method then resolves `ISyncPolicy` from `ContainerProvider`.
- `GameStateInterface.InteralLoadSaveGame` does not start an `AllowedThread` scope.

A flag that applies to one thread also cannot reach the worker threads of the save loader. The patches therefore stay active during the load.

The logs do not show which dereference gives the null value. This part is not confirmed.

## Steps to reproduce

Both attempts gave the same result.

1. Host a coop session.
2. Join from a Linux client through Steam. The client runs the mod with Proton.
3. The join reaches "receiving host save data". The join then stops permanently.

## Additional information

Client build: `0.1.2+86c511b53b2edc5c83fc836f54b9e6a3e9dbb0dc`. This build is the Steam Workshop release.

Client runtime: Linux with Proton. The log shows `MonoRuntime=true` and `CLRVersion=4.0.30319.42000`.

This problem may occur only on Linux. The failure is in the parallel save-load path. Mono schedules that path differently than the .NET Framework CLR that a Windows client uses. A Windows client that joins the same host was not tested.

Issue #2095 may be related. That issue is closed. It also describes a Linux join hang, but it stops earlier, at "Sending your Character to the host".

Client logs from both attempts are attached.

garrettluskey:
## Summary

- add a narrowly scoped, process-wide override to `CallOriginalPolicy`
- enable it only while the client synchronously runs `SaveManager.Load`
- make the override thread-safe, reference-counted, and safe against repeated disposal
- cover worker-thread visibility, nested scopes, and disposal behavior with tests

## Root cause

`AllowedThread` is intentionally thread-local, but Bannerlord's save system fills loaded objects through `NativeParallelDriver` worker callbacks. Those workers do not inherit the game thread's allowance, so AutoSync Harmony prefixes remain active while `PropertyLoadData.FillObject` invokes saveable property setters. On the reported Linux/Mono client, that path throws inside patched setters, `LoadContext` returns a failed load, and the join never leaves the loading screen.

The new scope bypasses `ISyncPolicy`/AutoSync interception across those workers only for the synchronous `SaveManager.Load` window. It is disposed before normal campaign startup continues.

## Validation

- `dotnet test source/GameInterface.Tests/GameInterface.Tests.csproj --configuration Debug --verbosity minimal -p:NuGetAudit=false`
- 928 tests passed

A default restore is currently blocked by the repository's existing warnings-as-errors treatment of known Scriban 7.2.0 advisories, so validation disabled NuGet audit rather than changing dependencies in this fix.

## Remaining validation

I could not run the original Steam/Proton join flow locally. This targets the confirmed parallel setter crash. The separate missing-object/repeated registry log noise reported after successful loading is not changed here.

Fixes #2999

### DIFF
--- a/source/GameInterface.Tests/Policies/CallOriginalPolicyTests.cs
+++ b/source/GameInterface.Tests/Policies/CallOriginalPolicyTests.cs
@@ -1,0 +1,53 @@
+using GameInterface.Policies;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace GameInterface.Tests.Policies;
+
+public class CallOriginalPolicyTests
+{
+    [Fact]
+    public async Task AllThreadsScope_IsVisibleOnWorkerThreadUntilDisposed()
+    {
+        Assert.False(CallOriginalPolicy.AreOriginalsAllowedOnAllThreads);
+
+        using (CallOriginalPolicy.AllowOriginalsOnAllThreads())
+        {
+            Assert.True(CallOriginalPolicy.AreOriginalsAllowedOnAllThreads);
+            Assert.True(await Task.Run(CallOriginalPolicy.IsOriginalAllowed));
+        }
+
+        Assert.False(CallOriginalPolicy.AreOriginalsAllowedOnAllThreads);
+    }
+
+    [Fact]
+    public void NestedAllThreadsScope_DoesNotRevokeOuterScope()
+    {
+        using (CallOriginalPolicy.AllowOriginalsOnAllThreads())
+        {
+            using (CallOriginalPolicy.AllowOriginalsOnAllThreads())
+            {
+                Assert.True(CallOriginalPolicy.AreOriginalsAllowedOnAllThreads);
+            }
+
+            Assert.True(CallOriginalPolicy.AreOriginalsAllowedOnAllThreads);
+        }
+
+        Assert.False(CallOriginalPolicy.AreOriginalsAllowedOnAllThreads);
+    }
+
+    [Fact]
+    public void AllThreadsScope_DoubleDispose_DoesNotRevokeAnotherScope()
+    {
+        using (CallOriginalPolicy.AllowOriginalsOnAllThreads())
+        {
+            var scope = CallOriginalPolicy.AllowOriginalsOnAllThreads();
+            scope.Dispose();
+            scope.Dispose();
+
+            Assert.True(CallOriginalPolicy.AreOriginalsAllowedOnAllThreads);
+        }
+
+        Assert.False(CallOriginalPolicy.AreOriginalsAllowedOnAllThreads);
+    }
+}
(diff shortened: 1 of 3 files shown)


ShoT-UPfps (approved):
lgtm

PR #3036, opened 2026-08-14
Reject incompatible co-op builds during validation

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Different client and server builds can be accepted when they share the same displayed Coop version, then fail campaign loading and leave the player on an infinite loading screen.

## What is the new behavior?

Client and server builds must now match exactly. Different or older builds are rejected while joining, and the player sees an incompatible co-op build message instructing them to update both sides.

## Bot Changelog Entry

- Fixed an issue where old clients could connect to newer servers


garrettluskey (approved):
The exact build handshake is checked on both sides, older peers fail with a visible reason, and the round-trip/state tests cover the changed protocol. looks good.

PR #3069, opened 2026-08-15
Fix for static log access before configuration was completed

casey-johnson:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

A static method LiveTestControlServer.IsEnabled was called, triggering static initialization of the Logger field in LiveTestControlServer, which got the logger from LogManager before the logger configuration was completely setup. 

## What is the new behavior?

The Coop_client.log and Coop_server.log fill out correctly (tested on both Windows and Linux). Since it works for most people in the first place, this seems to be a niche bug due to differences in sdk versions used, or some other unknown factor.

### DIFF
--- a/source/Coop/CoopMod.cs
+++ b/source/Coop/CoopMod.cs
@@ -108,12 +108,6 @@ namespace Coop
             }
 
             isAutoConnect = args.Any(a => a.Equals("/autoconnect", StringComparison.OrdinalIgnoreCase));
-#if DEBUG
-            isDeferredClientJoin = args.Any(a =>
-                    a.Equals("/cooptestmanualjoin", StringComparison.OrdinalIgnoreCase)) &&
-                LiveTestControlServer.IsEnabled(Environment.GetCommandLineArgs());
-            isLiveTestRun = LiveTestControlServer.IsEnabled(Environment.GetCommandLineArgs());
-#endif
 
             // GetFullCommandLineString splits on spaces, which would cut a quoted save
             // name apart; the managed-server arguments need real Windows arg parsing.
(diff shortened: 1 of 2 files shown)


source/Coop/CoopMod.cs
```diff
@@ -128,6 +122,13 @@ public override void NoHarmonyInit()
 
             SetupLogging();
             InitializeCrashReporting();
+            
+#if DEBUG
+            isDeferredClientJoin = args.Any(a =>
+                                       a.Equals("/cooptestmanualjoin", StringComparison.OrdinalIgnoreCase)) &&
+                                   LiveTestControlServer.IsEnabled(Environment.GetCommandLineArgs());
```
ShoT-UPfps, line 129: `LiveTestControlServer` has no explicit static constructor, so its logger initializer is emitted as `beforefieldinit` and the CLR may still run it before `SetupLogging()`. can this make the logger initialization explicit or lazy? otherwise this can still create `LogManager._logger.Value` before the file sinks are configured.

ShoT-UPfps (changes requested):
cmt

source/Coop/CoopMod.cs
casey-johnson: This made me do a bit more reading on static field initialization and you're right. Here is a good write-up: https://csharpindepth.com/Articles/BeforeFieldInit
The static constructor seems like what we want, here is the excerpt: 
"The static constructor for a class executes at most once in a given application domain. The execution of a static constructor is triggered by the first of the following events to occur within an application domain:
- An instance of the class is created.
- Any of the static members of the class are referenced."

ShoT-UPfps (approved):
lgtm

PR #3127, opened 2026-08-17
Budget and prioritize agent movement traffic

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Movement bandwidth is limited per recipient, so total outgoing traffic increases with the number of players. When the limit is reached, agents are rotated without considering their distance from the receiving player.

## What is the new behavior?

Resolves #

Movement packets share one configurable total outgoing bandwidth limit. Each client advertises its incoming allowance and active player agent so senders can limit traffic for that route.

When bandwidth is constrained, current movement snapshots are prioritized using their distance from the receiving player and the time since they were last successfully sent. The local player agent remains the highest priority.

## Bot Changelog Entry

- Added configurable movement bandwidth limits and prioritized nearby agent movement when bandwidth is constrained.


source/Missions/Agents/Handlers/AgentMovementHandler.cs
```diff
@@ -686,6 +734,9 @@ private MovementTrafficFrame PollMovementAgents(
                 includePriorityAgent));
         }
 
+        if (recipientIds.Count > 0)
+            recipientSendOffset = (startRecipient + 1) % recipientIds.Count;
```
ShoT-UPfps: this also advances on priority-only polls. with a 10/40 hz cadence and 2 or 4 recipients it wraps back to the same route before every bulk poll, so use a separate bulk cursor otherwise another recipient can stop receiving formation or mount movement under budget pressure.

source/Missions/Agents/Handlers/MovementNetworkSettings.cs
```diff
@@ -0,0 +1,75 @@
...
+                "mod-config.json: network value for '{Setting}' must be greater than zero and no more than {Maximum}; using {Default}",
+                name,
+                MaximumMiBPerSecond,
+                DefaultMiBPerSecond);
+        }
+        return DefaultMiBPerSecond;
+    }
+
+    private static int ToBytes(double mebibytesPerSecond) =>
+        checked((int)Math.Round(mebibytesPerSecond * BytesPerMiB));
```
ShoT-UPfps: a value like `0.0000001` passes validation but rounds to 0 bytes/s here, then `MovementTrafficBudget` throws during mission construction. clamp this to at least 1 or reject values below the supported minimum.

source/Missions/Agents/Handlers/MovementBatchSender.cs
```diff
@@ -163,6 +243,48 @@ public MovementSendResult Send<T>(
...
+                onSent,
+                maximumPackets: 1,
+                allowProbeForGrowth: probedScopes.Add(batch.IdentityScopeId));
+            if (result.ProcessedCount <= 0) break;
+
+            sentSnapshots += result.SentCount;
+            if (result.ProcessedCount >= batch.Data.Count)
+                pending.RemoveAt(0);
+            else
+                pending[0] = Slice(batch, result.ProcessedCount);
```
ShoT-UPfps: this copies the entire unsent tail after every packet, including after the shared budget was just exhausted. keep a start index into the existing batch instead, otherwise large agent sets are repeatedly allocated and copied on the movement hot path.

ShoT-UPfps (changes requested):
comments + conflicts

source/Missions/Agents/Handlers/AgentMovementHandler.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/MovementNetworkSettings.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/MovementBatchSender.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/AgentMovementHandler.cs
```diff
@@ -796,6 +848,14 @@ private MovementTrafficFrame PollMovementAgents(
                 includePriorityAgent));
         }
 
+        if (recipientIds.Count > 0)
+        {
+            if (includeAuthoritativeAgents)
+                bulkRecipientSendOffset = (startRecipient + 1) % recipientIds.Count;
```
ShoT-UPfps: these cursors still advance when no route sends anything. with four recipients and a bucket that only refills enough for one packet every fourth poll, the cursor wraps back to the same recipient on each successful send, so advance from the route that actually consumed the shared budget.

source/Missions/Agents/Handlers/AgentMovementHandler.cs
```diff
@@ -907,7 +1013,13 @@ private MovementTrafficFrame SendMovementToRecipient(
...
-        if (recipient.SendMountMovementFirst)
+        bool sendMountMovementFirst = ShouldSendMountMovementFirst(
+            recipient,
+            movement.Compact.Values,
+            movement.Legacy,
+            mountMovement.Compact.Values,
+            mountMovement.Legacy);
+        if (sendMountMovementFirst)
         {
             mountResult = movementBatchSender.Send(
```
ShoT-UPfps: this only compares the two stream heads once. after one stream wins, `Send` can drain lower-priority packets from it before the other stream is tried, so compare the movement and mount heads again after each packet.

source/Missions/Agents/Handlers/MovementRateController.cs
```diff
@@ -802,12 +896,22 @@ private MovementRateSnapshot CreateSnapshot()
...
+            focusAgentId);
     }
 
     private int GetControllerCount() =>
         1 + (missionContext?.ControllersInMission.Count ?? 0);
 
+    private int CalculateIncomingBytesPerSender()
+    {
+        int senders = Math.Max(1, missionContext?.ControllersInMission.Count ?? 0);
+        return Math.Max(1, networkSettings.IncomingBytesPerSecond / senders);
```
ShoT-UPfps: this can advertise more than the configured total when `IncomingBytesPerSecond` is smaller than the sender count. a 1 byte/s total with four senders becomes 1 byte/s per sender, so reject that configuration or distribute sub-byte shares without raising the aggregate cap.

ShoT-UPfps (changes requested):
new comments after rereview

source/Missions/Agents/Handlers/AgentMovementHandler.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/AgentMovementHandler.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/MovementRateController.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/AgentMovementHandler.cs
```diff
@@ -865,89 +943,96 @@ private MovementTrafficFrame SendMovementToRecipient(
...
-            mountResult = movementBatchSender.Send(
-                controllerId,
-                mountMovement.Compact.Values,
-                mountMovement.Legacy,
-                maxPayloadBytes,
-                CreateMountMovementPacket,
-                (agentId, data) => RecordSentMovement(recipient, agentId, data));
-        }
-        recipient.SendMountMovementFirst = !recipient.SendMountMovementFirst;
+        MovementSendPairResult movementResult = movementBatchSender.SendInterleaved(
```
ShoT-UPfps: `priorityResult` can send 0 because the main-agent packet is larger than the current token balance, but `SendInterleaved` still spends those tokens on lower-tier movement. with a 150-byte player packet and a 100-byte troop packet at a 100-byte refill, the player packet never sends, so reserve the balance or schedule all three streams together.

source/Missions/Agents/Handlers/MovementBatchSender.cs
```diff
@@ -229,22 +519,30 @@ private MovementSendResult SendBatch<T>(
...
 
-        for (int start = 0; start < batch.Data.Count;)
+        for (int start = startOffset;
+            start < batch.Data.Count && sentPackets < maximumPackets;)
         {
             int availablePayloadBytes = Math.Min(
                 maxPayloadBytes,
-                recipient.TrafficBudget.AvailableBytes);
+                Math.Min(
+                    outgoingTrafficBudget.AvailableBytes,
```
ShoT-UPfps: `outgoingTrafficBudget` is shared, so a larger route can starve behind a smaller one. if A needs 150 bytes and B spends the bucket whenever it reaches 100, A is visited first every poll but never accumulates enough, so later routes need to stop spending when the first valid packet is only waiting for tokens.

source/Missions/Agents/Handlers/AgentMovementHandler.cs
```diff
@@ -775,25 +813,52 @@ private MovementTrafficFrame PollMovementAgents(
...
-            traffic = traffic.Add(SendMovementToRecipient(
+            MovementTrafficFrame recipientTraffic = SendMovementToRecipient(
                 recipientId,
                 recipient,
                 capturedMovements,
                 recipientBulkDue,
-                includePriorityAgent));
+                includePriorityAgent);
+            traffic = traffic.Add(recipientTraffic);
+            if (recipientTraffic.SentBytes > 0)
```
ShoT-UPfps: `SentBytes` includes the main-agent packet. if every recipient gets that packet but only the first route has room for bulk movement, the last priority recipient overwrites `lastSendingRecipient` and wraps the bulk cursor back to the same start, so advance the bulk cursor from a route that actually sent bulk data.

source/Missions/Agents/Handlers/MovementRateController.cs
```diff
@@ -802,12 +896,29 @@ private MovementRateSnapshot CreateSnapshot()
...
+            focusAgentId);
     }
 
     private int GetControllerCount() =>
         1 + (missionContext?.ControllersInMission.Count ?? 0);
 
+    private double CalculateIncomingBytesPerSender()
+    {
+        int senders = Math.Max(1, missionContext?.ControllersInMission.Count ?? 0);
+        return (double)networkSettings.IncomingBytesPerSecond / senders;
```
ShoT-UPfps, line 912: this split changes for every sender when membership grows, but `Handle_PeerEntered` only sends the new value to the entrant. existing senders keep the old allowance until the 1-second heartbeat, so broadcast the updated split to every current sender.

ShoT-UPfps (changes requested):
new comments after rereview

source/Missions/Agents/Handlers/AgentMovementHandler.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/MovementBatchSender.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/AgentMovementHandler.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/MovementRateController.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/MovementBatchSender.cs
ShoT-UPfps: gap remains at `AgentMovementHandler.cs:827`: a priority-only block only updates `priorityRecipientSendOffset`, so the next bulk poll can still spend the shared balance through `bulkRecipientSendOffset` before the blocked route can accumulate enough.

source/Missions/Agents/Handlers/MovementBatchSender.cs
```diff
@@ -170,6 +330,169 @@ public MovementSendResult Send<T>(
...
+        PrioritizedBatchCursor<T> cursor = pending[0];
+        MovementBatch<T> batch = cursor.Batch;
+        MovementSendResult result = SendBatch(
+            controllerId,
+            recipient,
+            batch,
+            maxPayloadBytes,
+            createPacket,
+            onSent,
+            maximumPackets: 1,
```
ShoT-UPfps, line 499: `maximumPackets: 1` still lets `SendBatch` put multiple snapshots from this batch in that packet. priorities 1 and 100 can be sent before another batch's priority 2 is reconsidered, so limit the packet at the next competing priority rather than only at one packet.

source/Missions/Agents/Handlers/MovementBatchSender.cs
```diff
@@ -268,6 +610,7 @@ private MovementSendResult SendBatch<T>(
                 !candidate.Fits(availablePayloadBytes) &&
                 candidate.Fits(maxPayloadBytes))
             {
+                blockedBySharedBudget = candidate.Payload.Length > sharedAvailableBytes;
```
ShoT-UPfps: this marks the route shared-blocked even when its recipient bucket is also short. a slow route can then stop every later recipient while it waits on its own allowance, so reserve globally only when the recipient bucket already covers the required packet.

ShoT-UPfps (changes requested):
new comments after rereview

source/Missions/Agents/Handlers/MovementBatchSender.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/MovementBatchSender.cs
garrettluskey: Updated

source/Missions/Agents/Handlers/MovementBatchSender.cs
garrettluskey: Updated

ShoT-UPfps (approved):

PR #3135, opened 2026-08-18
Fix player caravans member limit & speed

### LINKED ISSUE #2623: [CARAVANS] Caravans Move More Slowly Than Normal
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

Caravans are reported to move significantly slower than normal, and the slowdown is described as affecting caravans only.

## Steps to reproduce

Not provided

## Additional information

Supporting context was provided with attachments: `image.png`, `coop-server-20260802-144810.log`, and `Win64_Shipping_Client.7z`.

## Discord source

https://discord.com/channels/709516043332354119/1533595667295899833/1533595667295899833

## Images and videos

[](<https://cdn.discordapp.com/attachments/1533595667295899833/1533595668403323003/image.png?ex=6aa47a44&is=6aa328c4&hm=cf3f70ffc6b0d6367618ebf798060e6399a4de627534c0d988c6c34aefd0a71c&>) ([Discord message](https://discord.com/channels/709516043332354119/1533595667295899833/1533595667295899833))

## Logs

Client log: Not provided

Server log: [Download coop-server-20260802-144810.log](<https://cdn.discordapp.com/attachments/1533595667295899833/1533595668785008740/coop-server-20260802-144810.log?ex=6aa47a44&is=6aa328c4&hm=dc728364db91281f48dd3f7e28935297d5481613f78a31834b51af9787019abc&>)

Diagnostic logs: [Download Win64_Shipping_Client.7z](<https://cdn.discordapp.com/attachments/1533595667295899833/1533595669393051941/Win64_Shipping_Client.7z?ex=6aa47a44&is=6aa328c4&hm=6a79ba1470d45c6ada7e4283886925670729a43a0560b4aa675351f60f5c7010&>)

jordanbrymora:
## PR Checklist

## PR Type

The member limit of a player caravan is determined by a patch that uses `IsPlayerHero()`. Because this extension method relies on the player registry, existing player caravans on a loaded game get cached at a limit of 20 on clients. This causes the party speed to appear slower on clients and the `PartyVM` to show limits of `30/20` (regular) and `50/20` (elite) for player caravans.

## What is the new behavior?

Resolves #2623 

Invalidate player caravan member limit caches when players are loaded to re-calculate with `IsPlayerHero()` correctly.

## Bot Changelog Entry

Fixed player caravans being slower than expected and displaying a party limit of 20.


source/GameInterface/Services/Players/PlayerManager.cs
```diff
@@ -236,6 +241,19 @@ private void InvalidatePlayerPartySpeedCache(MobileParty mobileParty)
...
 
+    private void InvalidatePlayerCaravanMemberLimitCaches(Hero hero)
+    {
+        GameThread.RunSafe(() =>
+        {
+            if (hero.OwnedCaravans == null) return;
+
+            foreach (var ownedCaravan in hero.OwnedCaravans)
+            {
+                ownedCaravan.Party._partyMemberSizeLastCheckVersion = -1;
```
garrettluskey, line 252: this only expires `PartyBase.PartySizeLimit`. `MobileParty` separately caches `_lastCalculatedBaseSpeedExplained` behind `_partyPureSpeedLastCheckVersion`, and vanilla base speed reads `PartySizeLimit` when applying the over-party-size penalty. if speed was calculated before player registration, it keeps the old penalty after this runs. invalidate the caravan mobile party speed cache too and add a test that reads speed before and after registration.

garrettluskey (changes requested):
the player-registration hook is the right place to expire loaded caravan limits. changes are still needed because the caravan speed uses a separate cache that this leaves stale.

source/GameInterface/Services/Players/PlayerManager.cs
jordanbrymora: Added invalidation in `InvalidatePlayerCaravanMemberLimitCaches`.

source/GameInterface/Services/Players/PlayerManager.cs
```diff
@@ -236,6 +241,20 @@ private void InvalidatePlayerPartySpeedCache(MobileParty mobileParty)
...
+                ownedCaravan.Party.MobileParty._partyPureSpeedLastCheckVersion = -1;
```
garrettluskey, line 253: the speed invalidation fixes the previous blocker, but the requested test is still missing. `PlayerManagerTests.AddPlayer_MobileParty_InvalidatesBaseSpeedCache` already has the game-thread setup for this. add the matching owned-caravan test that seeds both cache versions, registers its owner hero, and asserts both are reset.

garrettluskey (changes requested):
the member-limit and base-speed caches are both expired now, so the runtime fix looks correct. the requested regression coverage is still missing.

source/GameInterface/Services/Players/PlayerManager.cs
jordanbrymora: Added test

garrettluskey (approved):
the owned-caravan regression test now covers both cache versions, and the registration hook expires both member-limit and speed caches.

PR #2964, opened 2026-08-12
Feature/quest gangleader needs to offload stolen goods

Thaipho:
Depends on #2763 

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] All new classes have class-level documentation comments, if there are any at all
- [x] Tests for the changes have been added (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update
- [ ] Other... Please describe:

## What is the current behavior?
GangLeaderNeedsToOffloadStolenGoods issues are not synced in multiplayer.

## What is the new behavior?
Adds full coop sync for GangLeaderNeedsToOffloadStolenGoods on top of the shared
framework from #2763 

## Bot Changelog Entry
Added multiplayer sync for the GangLeaderNeedsToOffloadStolenGoods quest

## Other information
Stacked on #2763 (feature/generic-quest-handler) — depends on it merging first

PR #3155, opened 2026-08-18
Fix: restore Issues dialogue gate, allowlist-scoped

Thaipho:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Every "ask about a problem" / issue-quest dialogue option (`LordConversationsCampaignBehavior.conversation_hero_main_options_have_issue_on_condition`, `conversation_lord_task_given_on_condition`, `conversation_lord_task_given_alternative_on_condition`) is currently unconditionally visible for any hero holding an `IssueBase`, regardless of whether that quest type has any working multiplayer sync. A blanket patch (`DisableLordConversationIssueDialogs.cs`) used to hide these until quests were ready, but was deleted three times across this project's history (`d60d70320`, `511fa7e4f`, `4364d5bfb`) on the mistaken assumption that `DisableAllIssueBehaviorsExceptAllowlist` already covers it — it doesn't; that mechanism only gates a vanilla issue type's own `RegisterEvents`, not these three conversation conditions. The third deletion shipped and produced a live player-facing softlock (reported via Discord: picking the option dead-ends the conversation).

## What is the new behavior?

Resolves #

Re-adds the gate, but scoped to `DisableAllIssueBehaviorsExceptAllowlist.IsAllowlisted(...)` instead of hardcoded `false`, and declared next to `Allowlist` in the same file this time instead of as a separate, easy-to-mistake-for-orphaned patch. Each condition now runs vanilla's own check first, then ANDs the result with an allowlist check on the hero's actual `Issue`. Effect: hidden today (Allowlist is currently empty on `development`), and self-clears per quest type the moment that type's own branch merges — no follow-up PR needed to manually lift the block, since every open quest branch already adds itself to `Allowlist` as part of its own diff.

Also ports the Fail/Betrayal proof-validation framework fixes discovered while building GangLeaderNeedsToOffloadStolenGoods (#2964) onto `development`, verbatim from that branch (framework-only, no GangLeader-specific code):
- `QuestTypeDescriptor` gains `CaptureQuestFailProof`/`CaptureQuestBetrayalProof` and `ApplyQuestFailConsequence`/`ApplyQuestBetrayalConsequence`, mirroring the existing Success-only versions, with matching builder methods.
- `IssueFinalizationHandler` now captures and server-side-validates proof for Fail/Betrayal the same way it already did for Success, instead of trusting a client-claimed finalize reason outright for those two.
- `NetworkIssueRemoved` gains the `Proof` field it was missing (computed server-side but never actually broadcast); `RequestIssueRemoved.SuccessProof` is renamed to `Proof` since it now carries proof for any of the three validated reasons.
- `IssueFinalizationSupport`'s consequence dispatch is extracted into `ApplyOngoingQuestFinalize`/`RunConsequenceOrFallback` to stay readable now that all four finalize reasons carry a consequence hook.

Every other open quest branch still needs the equivalent port to pick this up; #2964 already has it natively.

## Bot Changelog Entry

Fixed a bug where talking to a lord or notable about "a problem" could softlock the conversation when no multiplayer-ready quest was actually available.

## Other information
Verified: 0 build errors, 39/39 Issues E2E tests pass. Full solution test run in progress.


source/GameInterface/Services/Issues/Handlers/IssueFinalizationHandler.cs
```diff
@@ -340,9 +423,19 @@ private void Handle_NetworkIssueRemoved(MessagePayload<NetworkIssueRemoved> payl
...
 
-            using (new MainHeroSubstitutionScope(truePlayerHero ?? owner, ownerParty))
+            var isLocalPeerOwner = ownershipRegistry.IsLocalPeerOwner(owner);
+
+            SetProofContext(reason, proof);
+            try
+            {
+                using (new MainHeroSubstitutionScope(truePlayerHero ?? owner, ownerParty))
+                {
+                    IssueFinalizationSupport.FinalizeMirror(owner, reason, suppressReplicationPatches: true, skipConsequenceReapplication: isLocalPeerOwner);
```
garrettluskey: `FinalizeAndBroadcast` runs this consequence on the server with patches live, so `ItemRosterPatch.AddToCountsPostfix` already broadcasts registered roster deltas. this reruns the same consequence on every non-owner under `AllowedThread`, then `NetworkItemRosterUpdate` applies the server delta too, so those clients get the item change twice. dont replay the full consequence on non-owners.

garrettluskey (dismissed):
the dialogue gate now tracks the issue allowlist, and fail/betrayal proof is carried through the finalize messages. changes are still needed because the consequence replay can apply server-replicated state twice on clients.

jordanbrymora:
@garrettluskey @ShoT-UPfps Could we please get this reviewed again before the next update? A lot of people will be soft-locked if we don't include this fix.

ShoT-UPfps (approved):

PR #3158, opened 2026-08-18
tournaments grant xp now

### LINKED ISSUE #2586: [TOURNAMENTS] Tournaments grant no skill experience after repeated participation
# For New Contributors

Not provided

## Description

After roughly 20 tournaments, the player reports gaining 0 skill XP, with One-Handed still at 10 and other fighting stats still very low.

## Intended Design

Not provided

## Location

Not provided

## Related Issues

Not provided

## Requirements

Not provided

## Additional information

Not provided

## Definition of Done

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1533285678509461524/1533285678509461524

zzzzzzzzott:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

tournaments dont give xp

## What is the new behavior?

Resolves #2586 

tournaments give xp

## Bot Changelog Entry

Tournaments give xp now


source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -531,18 +533,19 @@ private void CaptureHitProgression(
         in AttackCollisionData collisionData,
         float shotDifficulty)
     {
-        NetworkApplyTournamentDamage damage = activeDamageMessage;
-        if (!session.IsLocalHost || damage == null ||
+        if (!session.IsLocalHost ||
```
ShoT-UPfps: this no longer proves the hit was accepted. `OnScoreHit` invokes this before `ProcessPendingLocalDamage` resolves the guard candidate, so a blocked hit with positive `blow.InflictedDamage` can grant xp even though no damage is sent.

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -531,18 +533,19 @@ private void CaptureHitProgression(
...
             snapshot?.Phase != TournamentSessionPhase.LiveMatch ||
-            affectedAgent == null || affectorAgent == null ||
-            damage.MatchId != snapshot.CurrentMatchId) return;
-        TournamentAgentSpawnData attackerSpawn = FindManifestAgent(damage.AttackerAgentId);
-        TournamentAgentSpawnData victimSpawn = FindManifestAgent(damage.VictimAgentId);
+            affectedAgent == null || affectorAgent == null) return;
+        var registry = coopMissionComponent.AgentRegistry;
+        if (!registry.TryGetAgentInfo(affectorAgent, out var attackerInfo) ||
```
ShoT-UPfps, line 629: `affectedAgent.Health` is still the pre-hit value on this path because the accepted damage is applied later by `ProcessPendingLocalDamage`. killing blows are therefore submitted with `Fatal == false`.

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -557,10 +560,10 @@ private void CaptureHitProgression(
             snapshot.CurrentMatchId,
             snapshot.Revision,
             snapshot.BracketRevision,
-            damage.OriginControllerId,
-            damage.Sequence,
-            damage.AttackerAgentId,
-            damage.VictimAgentId,
+            attacker.ControllerId,
+            ++hitProgressionSequence,
```
ShoT-UPfps: `hitProgressionSequence` is local to the current host, but the server dedupes it under `DamageOriginControllerId`. after `TransferHostAuthority`, the new host can restart this at 1 and reuse an earlier player/sequence key, so valid hits are dropped as duplicates.

ShoT-UPfps (dismissed):
three correctness issues need to be fixed before this is ready.

source/Missions/Tournaments/CoopTournamentController.cs
zzzzzzzzott: addressed all cmts

ShoT-UPfps (commented):
rereviewed `7b99b96f`. the three earlier comments are fixed, but three correctness issues still leave tournament xp missing: host-origin hits never submit, remote missile replays lose weapon metadata, and melee's `-1f` shot difficulty is rejected.

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -531,18 +547,74 @@ private void CaptureHitProgression(
...
+        {
+            RecordPendingHitProgression(
+                affectedAgent,
+                affectorAgent,
+                attackerWeapon,
+                shotDifficulty);
+            return;
+        }
+
+        if (damage.OriginControllerId == session.OwnControllerId)
```
ShoT-UPfps: `HasHitProgression` is never set on the real local-hit path. the first `RegisterBlow` is intercepted before `OnScoreHit`, then the accepted replay reaches this local-origin return, so every host-origin hit is dropped.

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -531,18 +547,74 @@ private void CaptureHitProgression(
...
+            return;
+
+        SubmitHitProgression(
+            affectedAgent,
+            affectorAgent,
+            attackerWeapon,
```
ShoT-UPfps: `ApplyTournamentDamage` clears `IsMissile` and the weapon slot before this replay callback, so remote ranged hits reach here with no `attackerWeapon` and `shotDifficulty == -1f`. even after accepting the melee sentinel, bow/crossbow/throwing hits would grant Athletics/Riding instead of their weapon skill.

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -531,18 +547,74 @@ private void CaptureHitProgression(
...
+            in blow,
+            in collisionData,
+            shotDifficulty,
```
ShoT-UPfps: v1.4.8 passes `-1f` for melee `shotDifficulty`, but `IsValidProgressionData` rejects values below zero. this forwards that sentinel unchanged, so one-handed/melee progression is still dropped.

source/GameInterface/Services/Tournaments/Handlers/TournamentHitProgressionHandler.cs
```diff
@@ -27,7 +27,7 @@ private void Handle_HitProgression(MessagePayload<NetworkSubmitTournamentHitProg
             if (!IsValidProgressionData(data) ||
                 !sessionRegistry.TryGet(data.SessionId, out snapshot) ||
                 snapshot.Phase != TournamentSessionPhase.LiveMatch ||
-                snapshot.HostControllerId != player.ControllerId ||
+                data.DamageOriginControllerId != player.ControllerId ||
```
garrettluskey, line 30: `snapshot.Revision != data.Revision` below drops a valid hit whenever a spectator request or mission entry advances the session before this game-thread action runs. `TryRequestSpectate` and `TryEnterMission` both increment `Revision` without changing the live match, while the sender stamps its older snapshot, so accept an older revision when the current match, bracket, manifest, and contestant still match.

garrettluskey (changes requested):
the accepted-hit gating, missile-origin metadata preservation, and game-thread registry resolution look right. one stale-revision path still drops valid progression during routine spectator joins.

garrettluskey (changes requested):
the accepted-hit replay now preserves the origin's weapon metadata and melee sentinel, and the dedupe key handles authority migration. the stale session revision check in `TournamentHitProgressionHandler.cs:32` still drops valid hits when spectators enter during a live match, so that existing thread still needs to be addressed.

gamingmenacers-svg (dismissed):
reviewed exact head `22e083c6e` against current `development`. the accepted-hit replay and origin/weapon fixes are good, but the open revision blocker remains at `TournamentHitProgressionHandler.cs:32`: spectator entry increments the session revision without changing the live match, so a valid hit stamped just before that update is rejected on the game thread and grants no xp. accept an older session revision when the match, bracket, manifest, and contestant identity still match, and add that race to the tests. this is also 266 commits behind and needs a rebase.

ShoT-UPfps (commented):
rereviewed `46ce9f65` against `development` at `c6daab28`; Build and Test #5252 is green. the host-origin path, melee sentinel, and stale-session-revision findings are fixed. one ranged-hit blocker remains: the damage-origin client preserves the original engine missile index across the guard-window delay, after the collision path can remove that missile, so the accepted replay can fail before damage and progression are captured.

source/Missions/Tournaments/CoopTournamentController.cs
ShoT-UPfps: the origin-only preservation still replays an engine missile index after the guard-window delay. the original v1.4.8 `MissileHitCallback` continues past the intercepted `Agent.RegisterBlow` into `HandleMissileCollisionReaction` in the same callback; non-pass-through collisions remove or convert the source missile, and `OnMissileRemoved` deletes its dictionary entry. when `ProcessPendingLocalDamage` later replays this blow, `Mission.OnAgentHit` indexes `_missilesDictionary[index]` directly, so it can throw before `OnScoreHit` and lose both damage and ranged progression. snapshot the weapon and shot data before queuing, clear missile metadata for replay, and cover the queued path after removing the source missile. `MissileReplay_PreservesMetadataOnlyOnDamageOrigin` currently hides the failure by registering missile 99 immediately before direct replay.

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -467,12 +483,38 @@ public bool InterceptBlow(Agent victim, Blow blow, AttackCollisionData collision
...
+        out WeaponComponentData attackerWeapon,
+        out float? shotDifficulty)
+    {
+        attackerWeapon = null;
+        shotDifficulty = null;
+        if (!blow.IsMissile || attacker == null) return;
+
+        int missileIndex = blow.WeaponRecord.AffectorWeaponSlotOrMissileIndex;
+        if (Mission.Current._missilesDictionary.TryGetValue(missileIndex, out var missile))
+            attackerWeapon = missile.Weapon.CurrentUsageItem;
```
garrettluskey: snapshot `missile.Weapon.Item` and `CurrentUsageIndex` here instead of resolving this component from the attacker's current equipment later. vanilla passes `missile.Weapon` into `OnScoreHit`, but if the ammo or thrown-weapon slot was dropped or replaced while the projectile was in flight, `TryResolveHitWeapon` returns false and this valid hit grants no ranged xp.

source/Missions/Tournaments/CoopTournamentController.cs
```diff
@@ -511,14 +559,23 @@ private void ApplyTournamentDamage(NetworkApplyTournamentDamage message)
 
         applyingDamage.Add(message.VictimAgentId);
         NetworkApplyTournamentDamage previousDamageMessage = activeDamageMessage;
+        WeaponComponentData previousDamageAttackerWeapon = activeDamageAttackerWeapon;
+        float? previousDamageShotDifficulty = activeDamageShotDifficulty;
+        bool previousDamageFatal = activeDamageFatal;
         activeDamageMessage = message;
+        activeDamageAttackerWeapon = originAttackerWeapon;
+        activeDamageShotDifficulty = originShotDifficulty;
+        activeDamageFatal = blow.InflictedDamage >= victim.Health;
```
garrettluskey: vanilla marks the hit fatal when post-hit health is `< 1f`; this only does so when damage reaches the full current health. use `victim.Health - blow.InflictedDamage < 1f`, otherwise a knockout that leaves fractional health between 0 and 1 is submitted as nonfatal and gets less xp.

source/E2E.Tests/Services/Missions/Tournaments/TournamentDamageReplayTests.cs
```diff
@@ -222,6 +224,332 @@ public void UnguardedStrikeAgentCandidate_BroadcastsAfterDecisionWindow()
...
+            if (blocked)
+            {
+                Assert.Empty(
+                    network.NetworkSentMessages.GetMessages<NetworkSubmitTournamentHitProgression>());
+                return;
+            }
+
+            NetworkApplyTournamentDamage acceptedDamage = Assert.Single(
+                battleNetwork.NetworkSentMessages.GetMessages<NetworkApplyTournamentDamage>());
+            SetField(controller, "activeDamageMessage", acceptedDamage);
```
garrettluskey: this sets the replay internals after `ProcessPendingLocalDamage` has already finished, then calls `CaptureHitProgression` directly. wire `mock.RegisteredBlow` before processing as the missile test does, otherwise this test does not exercise the active damage/fatal scope used in production.

garrettluskey (changes requested):
the authority/origin split, stale-revision handling, and transient missile-index cleanup look right. ranged weapon identity and the fatal cutoff still differ from vanilla, and the melee test bypasses production sequencing.

garrettluskey (approved):
the earlier blockers are addressed: ranged hits snapshot the missile's item and usage before replay, fractional-health knockouts use the vanilla cutoff, and the melee test now exercises the active replay scope. progression stays tied to accepted damage on its origin client.

the added state is justified by the delayed replay and stays within the existing tournament path. no remaining blocking findings. build, unit tests, and all eight E2E shards are green on this head; no live game run was performed for this review.

Issue #3129, opened 2026-08-17
[SIEGES] Orphaned siege-assault MapEvent leaves AI lords immobilized at 0 troops after peace/disconnect


ShoT-UPfps:
### Summary

AI lord parties can remain permanently immobilized beside a castle after a siege or siege assault should have ended. The map still renders what appears to be an empty battle/map-event cluster, with both sides displaying `0`, and affected lords remain in place for many campaign hours.

![Stuck AI lords beside a castle with a 0/0 event](https://gist.githubusercontent.com/ShoT-UPfps/379116b9afa6cade641242384a32b49e/raw/starving.png)

The screenshot is consistent with the same **stale siege/map-event lifecycle family** already reported in #2922 and #2979: the siege or visible fighting may be over, but the authoritative `MapEvent` graph still owns the parties. Those parties therefore never return to ordinary campaign AI movement.

The `0` values do **not by themselves prove starvation**. They show that the visible party/event strength has reached zero. Starvation is a plausible downstream effect if the parties remain active and continue consuming food while immobilized, but the cause must be confirmed from party food and roster state in a save or DEBUG dump. The parties could also have reached zero through casualties, wounds, invalid roster state, or stale UI data.

### Consolidated reports and artifacts

The canonical description should retain the following report-specific evidence rather than merely linking to the older issues. GitHub-hosted attachments should be re-uploaded or preserved by direct reference where possible.

#### Current report — AI lords frozen beside a castle with a visible `0/0` event

- Multiple lord parties appear stationary around one castle and map-event marker.
- The visible strengths have reached `0`, and users report that the parties have remained there for a long time.
- The current screenshot must be attached to this issue.
- Starvation is suspected but not yet established; capture party food, healthy/wounded roster totals, and elapsed campaign time before asserting it.

#### Folded from #2979 — ghost siege persists with no enemies

- A dedicated-server city remained in an apparently old siege state even though no enemies were present.
- Caravans remained stuck in front of the gate.
- Changing the city's owner and running `coop.unstuck` did not repair the state.
- Original artifacts: `coop-server-20260812-190731.log`, `saveauto1.sav`, and `saveauto1.json`, plus the original screenshot and Discord source.
[...]

PR #3159, opened 2026-08-18
Support land siege interaction battle missions

garrettluskey:
## PR Checklist

## PR Type

Siege ambushes are routed through the field-battle start path.

## What is the new behavior?

Siege ambushes open the settlement siege scene with the native sally-out configuration. Naval battle mission requests are rejected until Warsails support is enabled.

The E2E suite covers siege assaults, sally outs, siege-outside battles, and siege ambushes across mission routing, scenes, troop ownership, side counts, and late AI reserve expansion.

## How to test

```powershell
dotnet test source/E2E.Tests/E2E.Tests.csproj -c Release --filter "FullyQualifiedName~SiegeInteractionBattleTests|FullyQualifiedName~BattleMissionStartHandlerTests|FullyQualifiedName~SiegeAssaultLeaveTests|FullyQualifiedName~BattleTroopReserveBuilderTests"
dotnet test source/GameInterface.Tests/GameInterface.Tests.csproj -c Release
```

## Bot Changelog Entry

- Siege ambushes now open the correct battle mission in co-op.


source/GameInterface/Services/MapEvents/Messages/Start/NetworkStartSiegeMission.cs
```diff
@@ -4,7 +4,7 @@
 namespace GameInterface.Services.MapEvents.Messages.Start;
 
 /// <summary>
-/// Opens the walls-assault siege mission on an authoritative participant. Carries the mission-defining siege
+/// Opens a walls-assault or siege-ambush mission on an authoritative participant. Carries the mission-defining siege
```
ShoT-UPfps, line 7: should `DisableSiegeAmbushMenuPatch` be removed with this? otherwise the siege menu still forces Ambush disabled, so this path cant be reached in normal gameplay

source/Missions/Battles/CoopSiegeBattleLauncher.cs
```diff
@@ -131,6 +135,9 @@ private Mission CreateCoopSiegeBattle(MissionInitializerRecord rec, string mapEv
                 behaviors.Add(new WorkshopMissionHandler(currentTown));
             }
 
+            if (isSallyOut)
+                behaviors.Add(new SandBoxSallyOutMissionController(isSallyOutAmbush: true));
```
ShoT-UPfps, line 139: should this use a side-stable ambush result? `SallyOutEndLogic` uses `MissionResult.CreateSuccessful(Mission)`, so an attacker-side client elected host reports `AttackerVictory` for the same return-to-castle outcome that a defender reports as `DefenderVictory`

source/GameInterface/Services/MapEvents/Handlers/BattleMissionStartHandler.cs
```diff
@@ -209,7 +217,7 @@ private void Handle_NetworkBattleStartRequest(MessagePayload<NetworkBattleStartR
                 operation = "send battle start reply";
                 network.Send(requester, new NetworkBattleStartReply(payload.What.RequestId, true));
 
-                if (mapEvent.IsSiegeAssault)
+                if (mapEvent.IsSiegeAssault || mapEvent.IsSiegeAmbush)
```
ShoT-UPfps, line 216: should `SiegeAmbushCampaignBehavior.OnMissionEnded` be patched for coop missions? it calls `PlayerEncounter.FinalizeBattle()`, so the elected host can finalize the map event before the result-ready barrier reconciles the other members

source/Missions/Battles/CoopSiegeBattleLauncher.cs
```diff
@@ -97,16 +99,18 @@ private Mission CreateCoopSiegeBattle(MissionInitializerRecord rec, string mapEv
             CoopTroopSupplierRegistry.Register(defenderSupplier);
             CoopTroopSupplierRegistry.Register(attackerSupplier);
 
-            // BattleSizeType.Siege selects the siege battle-size config in the spawn logic.
+            var battleSizeType = isSallyOut
```
ShoT-UPfps, line 102: this doesnt preserve the sally-out spawn setup in coop. `BattleSizeProvider` still sends the field cap for `IsSiegeAmbush`, then `CoopBattleMissionSpawnHandler.AfterStart` clears these phases and reinitializes them with the normal battle wave settings, so the 25/75 sizing and custom reinforcement timer are lost

ShoT-UPfps (changes requested):
requesting changes for the disabled menu path and the ambush result, finalization, and spawn setup issues noted inline

source/GameInterface/Services/MapEvents/Messages/Start/NetworkStartSiegeMission.cs
garrettluskey: removed `DisableSiegeAmbushMenuPatch`, the vanilla condition is live again.

source/GameInterface/Services/MapEvents/Handlers/BattleMissionStartHandler.cs
garrettluskey: patched `OnMissionEnded` to clear the local ambush flag but skip `FinalizeBattle` during a coop mission, so the result-ready barrier owns teardown.

source/Missions/Battles/CoopSiegeBattleLauncher.cs
garrettluskey: `BattleSizeProvider` now sends the sally-out cap for ambushes, and the coop spawn handler reuses the vanilla sally-out sizing, reinforcement settings, and custom timer instead of the normal battle waves.

source/Missions/Battles/CoopSiegeBattleLauncher.cs
garrettluskey: `SallyOutEndLogic.MissionEnded` now normalizes an ambush return to `DefenderVictory` while preserving each client’s local victory/defeat flags.

source/GameInterface/Services/SiegeEvents/Patches/SiegeMissionEndPatches.cs
```diff
@@ -23,6 +26,41 @@ internal class SiegeMissionEndPatches
...
+        if (!__result || !BattleConclusionGate.IsInCoopBattleMission) return;
+        if (PlayerEncounter.Battle?.IsSiegeAmbush != true || __instance.Mission?.PlayerTeam == null) return;
+
+        missionResult = CreateSiegeAmbushResult(__instance.Mission.PlayerTeam.Side);
+    }
+
+    internal static MissionResult CreateSiegeAmbushResult(BattleSideEnum playerSide)
+    {
+        bool playerVictory = playerSide == BattleSideEnum.Defender;
+        return new MissionResult(BattleState.DefenderVictory, playerVictory, !playerVictory, enemyRetreated: false);
```
garrettluskey: this cant go through the normal victory result path. vanilla finalizes a siege ambush from `SiegeAmbushCampaignBehavior.OnMissionEnded` while `MapEvent.BattleState` is still `None`; `DefenderVictory` makes the server run `OnBattleWon` and `ControlAndUpdateDefeatedPartiesAfterBattle`, so the attacker-side parties are processed as a defeated battle side when the defenders merely return to the castle. this needs an ambush completion barrier that finalizes the event without applying a victory state.

source/Missions/Battles/CoopBattleMissionSpawnHandler.cs
```diff
@@ -416,14 +475,51 @@ private static int ReachableSpawnNumber(int sideNumber, CoopTroopSupplier suppli
...
         => Math.Min(sideNumber, ownedShareOfSideNumber);
 
+    internal static MissionSpawnSettings CreateSallyOutSpawnSettings()
+        => SallyOutMissionController.CreateSallyOutSpawnSettings(0.01f, 0.1f);
+
+    private SallyOutSizing CalculateSallyOutSizing(
+        int defenderTotal, int attackerTotal, int battleSize)
+    {
+        var controller = Mission.GetMissionBehavior<SallyOutMissionController>();
+        if (controller == null || controller.MissionAgentSpawnLogic.BattleSize != battleSize)
```
garrettluskey: the reserve `BattleSize` is the server's authoritative value, but `MissionAgentSpawnLogic.BattleSize` comes from this client's local `BannerlordConfig`. a client with a different sally-out battle-size setting returns the default targets here and aborts a valid mission. calculate the 25/75 cap from `battleSize` instead of requiring the local setting to match.

source/GameInterface/Configuration/ModConfigProvider.cs
```diff
@@ -24,8 +24,13 @@ public readonly struct ModOptions
     public readonly bool FastForwardEnabled { get; } = true;
     [ProtoMember(2)]
     public readonly bool AutoPauseEnabled { get; } = true;
+#if DEBUG
+    [ProtoMember(3)]
+    public readonly bool ClientsCanUseCheats { get; } = true;
```
garrettluskey: why is the global debug default for `ClientsCanUseCheats` changing in this PR? this enables client-side state-changing cheats for every DEBUG session whose config omits the option, and none of the siege tests use it. please keep the default false and remove this unrelated change.

garrettluskey (commented):
the mission routing and sally-out setup are covered well, and the targeted tests pass. there are still two blocking correctness issues in the ambush result and sizing paths, plus an unrelated client-cheat default that needs to come out.

source/Missions/Battles/CoopSiegeBattleLauncher.cs
```diff
@@ -152,7 +160,9 @@ private Mission CreateCoopSiegeBattle(MissionInitializerRecord rec, string mapEv
             behaviors.Add(new HighlightsController());
             behaviors.Add(new BattleHighlightsController());
             behaviors.Add(new EquipmentControllerLeaveLogic());
-            behaviors.Add(new MissionSiegeEnginesLogic(defenderWeapons, attackerWeapons));
+            behaviors.Add(isSallyOut
```
ShoT-UPfps, line 163: how is surviving ambush engine damage written back here? `SiegeEngineStateReporter` returns unless `mapEvent.IsSiegeAssault`, and the client write-back is blocked, so partially damaged engines keep their pre-mission HP

source/GameInterface/Services/SiegeEvents/Patches/SiegeMissionEndPatches.cs
```diff
@@ -23,6 +26,42 @@ internal class SiegeMissionEndPatches
...
+
+        missionResult = CreateSiegeAmbushResult(__instance.Mission.PlayerTeam.Side);
+    }
+
+    // DefenderPullBack is the completion-barrier signal; it is never applied to the campaign MapEvent.
+    internal static MissionResult CreateSiegeAmbushResult(BattleSideEnum playerSide)
+    {
+        bool playerVictory = playerSide == BattleSideEnum.Defender;
+        return new MissionResult(BattleState.DefenderPullBack, playerVictory, !playerVictory, enemyRetreated: false);
+    }
```
ShoT-UPfps, line 47: `DefenderPullBack` isnt only a barrier signal here. on an attacker-side client `CampaignMissionComponent.OnMissionResultReady` calls `SetPlayerSiegeContinueWithDefenderPullBack` before `BattleResultReadyLogic` reports completion, so the server applies pull-back first and then rejects the report because `IsSupportedBattleResult` requires `BattleState.None`

source/GameInterface/Services/MapEvents/Handlers/BattleFinalizeHandler.cs
```diff
@@ -205,6 +207,45 @@ private void Handle_MapEventConcluded(MessagePayload<MapEventConcluded> payload)
...
+            if (mapEvent.IsFinalized)
+            {
+                applied = true;
+                return;
+            }
+
+            if (!mapEvent.IsSiegeAmbush || mapEvent.BattleState != BattleState.None)
+                return;
+
+            var playerPartyIds = FinalizeAndCollectPlayers(mapEvent);
```
ShoT-UPfps, line 231: this retry cant recover if `FinalizeAndCollectPlayers` throws before `FinalizeEventAux` completes. `TryMarkFinalized` records the map event first, so every scheduled retry returns as a duplicate while `mapEvent.IsFinalized` is still false

source/Missions/Battles/CoopSiegeBattleLauncher.cs
ShoT-UPfps: the report is sent now, but does the server apply it with ambush semantics? `SetSiegeEngineStatesAfterSiegeMission` checks `PlayerEncounter.Battle?.IsSiegeAmbush`; the dedicated server has no player encounter, so surviving melee engines are removed instead of having their HP updated

garrettluskey: fixed in `2e186deaf`. ambush reports now use `mapEvent.IsSiegeAmbush` and apply surviving HP or broken engines directly on the server, so the write-back does not depend on `PlayerEncounter`.

source/GameInterface/Services/SiegeEvents/Patches/SiegeMissionEndPatches.cs
garrettluskey: fixed in `2e186deaf`. the server ignores the attacker client’s ambush `DefenderPullBack` campaign-state write, so the map event stays at `None` until the result-ready barrier finalizes it.

source/GameInterface/Services/MapEvents/Handlers/BattleFinalizeHandler.cs
garrettluskey: fixed in `f7078c5f8` and covered by `b2ad7b62d`. a failed finalize removes the marker while a completed finalize keeps it; the raid sibling now uses the same rule, and the retry E2E verifies the second attempt completes.

source/GameInterface/Services/MapEvents/Handlers/BattleMissionStartHandler.cs
```diff
@@ -212,7 +220,7 @@ private void Handle_NetworkBattleStartRequest(MessagePayload<NetworkBattleStartR
...
```
ShoT-UPfps, line 216: `IsSiegeAmbush` now enters the one-shot siege start path, but the start message still has no settlement id and `OpenSiegeMission` permanently returns when the replicated `MapEventSettlement` is null. that race is already documented in #3129 for walls assaults; this change extends it to ambushes after mission mode has been claimed, so the mission can never open and the temporary map event can be orphaned. carry the settlement identity and bind/retry the open before routing ambushes here.

source/Missions/Battles/CoopSiegeBattleLauncher.cs
```diff
@@ -131,8 +135,11 @@ private Mission CreateCoopSiegeBattle(MissionInitializerRecord rec, string mapEv
                 behaviors.Add(new WorkshopMissionHandler(currentTown));
             }
 
+            if (isSallyOut)
+                behaviors.Add(new SandBoxSallyOutMissionController(isSallyOutAmbush: true));
```
ShoT-UPfps, line 139: the native v1.4.8 sally-out branch does more than add `SandBoxSallyOutMissionController`: it builds `GetPriorityTroopsForSallyOutAmbush()` and passes that roster to the defender supplier, elevating defender heroes and mounted troops ahead of the ordinary high-level queue. the coop reserve path still prepares sides with no priority roster and only moves the receiver's own hero first, so the limited 10% initial / 25% lifetime defender allocation can field high-level infantry while intended cavalry/heroes remain in reserve. author and transmit a deterministic defender priority order for this branch.

ShoT-UPfps (commented):
rereviewed `637af851`. the prior ten review threads are resolved and CI is green, but two correctness issues remain: siege ambushes now inherit the one-shot missing-settlement mission-start race, and the coop supplier omits v1.4.8's sally-out defender priority roster.

source/GameInterface/Services/MapEvents/Handlers/SiegeEngineStateCommitHandler.cs
```diff
@@ -59,12 +63,44 @@ private void HandleReport(MessagePayload<NetworkSiegeEngineStatesReport> payload
...
+        {
+            var engine = side.SiegeEngines.DeployedSiegeEngines[engineIndex];
+            if (!engine.IsActive)
+                continue;
+
+            var missionWeapon = missionWeapons[missionWeaponIndex--];
+            if (missionWeapon.Health > 0f)
+                engine.SetHitpoints(missionWeapon.Health);
+            else
+                siegeEvent.BreakSiegeEngine(side, engine.SiegeEngine);
```
ShoT-UPfps: `BreakSiegeEngine` only identifies the engine by type. if two same-type engines are deployed and an earlier slot is destroyed while a later slot survives, this removes the surviving later slot and leaves the destroyed one deployed.

ShoT-UPfps (commented):
rereviewed `ed692c76`. the two open findings from `637af851` are addressed in code and CI is green. one correctness blocker remains: duplicate same-type siege engines can delete the surviving slot during ambush write-back.

source/GameInterface/Services/MapEvents/Handlers/BattleMissionStartHandler.cs
garrettluskey: fixed in `ed692c767`. the siege start carries `SettlementId`, resolves and binds it on the game thread, and retries the open until the settlement is registered.

source/Missions/Battles/CoopSiegeBattleLauncher.cs
garrettluskey: fixed in `ed692c767`. the server builds the vanilla hero/mounted defender priority roster before `MakeReadyForMission` and sends its side-wide order through `SupplyOrder`.

source/GameInterface/Services/MapEvents/Handlers/SiegeEngineStateCommitHandler.cs
garrettluskey: fixed in `6ed9b3989`. ambush write-back now uses the mission index to select the original campaign engine and removes that engine’s exact deployment slot; the duplicate-type case is covered by a regression test.

gamingmenacers-svg (dismissed):
reviewed exact head `6ed9b3989` against current `development` and rechecked the ambush routing, result barrier, start retry, reserve priority, and siege-engine write-back. the latest slot-based removal fixes the duplicate-type case without changing the compact mission index mapping, and all hosted suites are green. no remaining code finding, this just needs rebasing from 58 commits behind and the suites rerun on the rebased head.

ShoT-UPfps (commented):
rereviewed `6ed9b398`. the duplicate same-type removal finding is fixed and Build and Test #5203 is green. one correctness blocker remains: `MissionSiegeWeapon.Index` is a type-specific deployment slot, but the replacement uses it as an index into the compact deployed-engine list, so sparse or mixed ranged/melee layouts skip or target the wrong engine.

source/GameInterface/Services/MapEvents/Handlers/SiegeEngineStateCommitHandler.cs
```diff
@@ -59,12 +63,68 @@ private void HandleReport(MessagePayload<NetworkSiegeEngineStatesReport> payload
...
+    internal static void SetAmbushEngineStatesForSide(
+        SiegeEvent siegeEvent,
+        ISiegeEventSide side,
+        IEnumerable<IMissionSiegeWeapon> missionSiegeEngineData)
+    {
+        var missionWeapons = missionSiegeEngineData?.ToList();
+        if (missionWeapons == null || missionWeapons.Count == 0)
+            return;
+
+        var deployedEngines = side.SiegeEngines.DeployedSiegeEngines.ToList();
```
ShoT-UPfps, line 89: `MissionSiegeWeapon.Index` is the deployment slot within the ranged or melee slot array, but this indexes the compact `DeployedSiegeEngines` list. a lone ranged engine in slot 3 has list count 1 and is skipped, while ranged slot 0 and melee slot 0 both select the same list element, so one weapon's HP/removal is ignored. select the type-specific deployed slot array from `missionWeapon.Type.IsRanged`, then update or remove that exact slot; add sparse-slot and mixed ranged/melee regressions.

garrettluskey: this index isnt a type-specific deployment slot on this path. v1.4.8 `GetPreparedAndActiveSiegeEngines` creates it from the compact `DeployedSiegeEngines` loop index, and that value round-trips unchanged through the start/report messages. `ace1ee019` corrects the misleading comment and covers sparse ranged slot 3 plus melee slot 0; switching to type-specific indexing would break that case.

ShoT-UPfps (dismissed):
lgtm

ShoT-UPfps (approved):

PR #3175, opened 2026-08-19
Allow defenders to join active slow raids

garrettluskey:
## PR Checklist

## PR Type

Active slow village raids disable both encounter join options and reject defender join requests, so a player cannot attack an army raiding their village.

## What is the new behavior?

Defenders can join the existing slow raid through the server-authoritative battle join path. The defender membership is replicated before the battle mode/start messages, and the raid becomes contested while attacker behavior stays unchanged.

This addresses the raiding-army part of #2792 only. It does not address the separate lord-capture report.

## How to test

`dotnet test source/E2E.Tests/E2E.Tests.csproj -c Release --filter "FullyQualifiedName~E2E.Tests.Services.Villages.VillageHostileActionTests"`

## Bot Changelog Entry

- Players can now defend villages against armies performing an active raid.

### DIFF
--- a/source/E2E.Tests/Services/Villages/VillageHostileActionTests.cs
+++ b/source/E2E.Tests/Services/Villages/VillageHostileActionTests.cs
@@ -1995,20 +1995,32 @@ public class VillageHostileActionTests : MapEventTestBase
     }
 
     [Fact]
-    public void SlowRaidMapEvent_DefenderJoinRequest_IsRejected()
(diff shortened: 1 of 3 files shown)


ShoT-UPfps (approved):
verified the defender path through the encounter condition, authoritative join, replicated defender membership, mode replay, and mission entry. membership is applied before the mode on the game-thread queue, and the focused e2e coverage passes in ci.

PR #3176, opened 2026-08-19
Restore nearby AI parties joining player battles

### LINKED ISSUE #2670: [SIEGES] AI armies abandon sieges and disengage from combat
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

AI armies, parties, and warbands lose their target or stop entirely. During online testing, armies approaching a siege immediately turned toward the nearest friendly settlement when players entered the city.

## Steps to reproduce

Not provided

## Additional information

The same behavior affects lords and bandits: armies do not join when intercepting to help a lord, and enemy armies abandon friendly lords when combat begins. Client and server logs are attached.

## Discord source

https://discord.com/channels/709516043332354119/1534042266656702595/1534042266656702595

## Images and videos

[](<https://cdn.discordapp.com/attachments/1534042266656702595/1534042268011593840/image.png?ex=6aa4c8b1&is=6aa37731&hm=0d16af2cbdb42cd294233c7426737a5747a9b28eb18500459add0225b7f3f961&>) ([Discord message](https://discord.com/channels/709516043332354119/1534042266656702595/1534042266656702595))

## Logs

Client log: [0.1.1+87a4520ab6fd2a9e4e6aad964ccd395296a507fc - Download Coop_client_1.log](<https://cdn.discordapp.com/attachments/1534042266656702595/1534042268518842388/Coop_client_1.log?ex=6aa4c8b1&is=6aa37731&hm=340dfd6b8452fc7a9d015ffd70066c289f47acc944a186fda8017b6ef72ea1d5&>) · [0.1.1+87a4520ab6fd2a9e4e6aad964ccd395296a507fc - Download Coop_client.log](<https://cdn.discordapp.com/attachments/1534042266656702595/1534042269064237068/Coop_client.log?ex=6aa4c8b2&is=6aa37732&hm=2a0817383c109c752d0597efbbbb607c5aa2afc5f93a00cb5aea8ff744ee09cf&>)

Server log: [Download coop-server-20260803-190445.log](<https://cdn.discordapp.com/attachments/1534042266656702595/1534042270171402410/coop-server-20260803-190445.log?ex=6aa4c8b2&is=6aa37732&hm=b2fe3f017c9cc2c94f6f4862ecb74b8adc4bf7fdc113a2919227a6fa2c4bf253&>)

garrettluskey:
## PR Checklist

## PR Type

Co-op blocks the client-side nearby-party scan because it mutates the shared map event locally, but the server never replaces it. Nearby friendly lords and armies therefore stop instead of joining a player's battle.

## What is the new behavior?

Part of #2670

The server runs the native encounter-model selection when a player joins and on campaign ticks while the AI join window is open. The window is opened before the synchronous join trigger, and selected AI parties are added authoritatively on the game thread with sync patches active.

## How to test

`dotnet test source/GameInterface.Tests/GameInterface.Tests.csproj -c Release --filter FullyQualifiedName~NearbyPartyReinforcerTests`

## Bot Changelog Entry

- Nearby AI lords and armies can join player battles again.

## Scope within #2670

This PR fixes #2670's nearby-AI combat-join/disengagement symptom. PR #3181 owns the separate siege-target-retargeting symptom that occurs when a player enters the intended settlement. Keep #2670 open until both implementations merge and both reproductions pass.

### DIFF
--- a/source/GameInterface/Services/MapEvents/NearbyPartyReinforcer.cs
+++ b/source/GameInterface/Services/MapEvents/NearbyPartyReinforcer.cs
@@ -1,0 +1,164 @@
+﻿using GameInterface.Services.MapEvents.Patches;
+using GameInterface.Services.MobileParties.Extensions;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.Encounters;
+using TaleWorlds.CampaignSystem.MapEvents;
+using TaleWorlds.CampaignSystem.Party;
+using TaleWorlds.Core;
+
+namespace GameInterface.Services.MapEvents;
+
+/// <summary>Finds and adds AI parties selected by the native encounter model.</summary>
+internal interface INearbyPartyReinforcer
+{
+    void Reinforce(MapEvent mapEvent);
+    void ReinforceOpenPlayerBattles();
+}
+
+/// <summary>Applies nearby AI reinforcements to open player battles on the server.</summary>
+internal sealed class NearbyPartyReinforcer : INearbyPartyReinforcer
+{
+    public void Reinforce(MapEvent mapEvent)
+    {
+        var encounterModel = Campaign.Current?.Models?.EncounterModel;
+        if (encounterModel == null)
+            return;
+
+        Reinforce(
+            mapEvent,
+            encounterModel.FindNonAttachedNpcPartiesWhoWillJoinPlayerEncounter,
+            (side, party) => side.AddNearbyPartyToPlayerMapEvent(party));
+    }
+
+    public void ReinforceOpenPlayerBattles()
+    {
+        var mapEvents = Campaign.Current?.MapEventManager?.MapEvents;
+        if (mapEvents == null)
+            return;
+
+        foreach (var mapEvent in mapEvents.ToArray())
+            Reinforce(mapEvent);
+    }
+
+    internal void Reinforce(
+        MapEvent mapEvent,
+        Action<List<MobileParty>, List<MobileParty>> selectJoiners,
+        Action<MapEventSide, MobileParty> addJoiner)
+    {
+        if (!CanReinforce(mapEvent) || !TryGetPlayerParty(mapEvent, out var playerParty, out var playerSide))
+            return;
+
+        var enemySide = playerSide.GetOppositeSide();
+        var playerParties = CollectMobileParties(mapEvent, playerSide);
+        var enemyParties = CollectMobileParties(mapEvent, enemySide);
+        int playerPartyCount = playerParties.Count;
+        int enemyPartyCount = enemyParties.Count;
+
+        SelectJoiners(mapEvent, playerParty, playerSide, playerParties, enemyParties, selectJoiners);
+
+        AddJoiners(mapEvent.GetMapEventSide(playerSide), playerParties, playerPartyCount, addJoiner);
+        AddJoiners(mapEvent.GetMapEventSide(enemySide), enemyParties, enemyPartyCount, addJoiner);
+    }
+
+    internal static bool CanReinforce(MapEvent mapEvent)
+    {
+        if (mapEvent == null || mapEvent.IsFinalized || mapEvent.BattleState != BattleState.None)
+            return false;
+
+        if (mapEvent.IsRaid || mapEvent.IsSiegeAssault || mapEvent.IsForcingSupplies ||
+            mapEvent.IsForcingVolunteers || mapEvent.MapEventSettlement?.IsHideout == true)
+            return false;
+
+        return InteractionPatches.IsWithinAiJoinWindow(mapEvent) && mapEvent.ContainsPlayerParty();
+    }
+
+    private static void SelectJoiners(
+        MapEvent mapEvent,
+        MobileParty playerParty,
+        BattleSideEnum playerSide,
+        List<MobileParty> playerParties,
+        List<MobileParty> enemyParties,
+        Action<List<MobileParty>, List<MobileParty>> selectJoiners)
+    {
+        // Vanilla reads these encounter statics, so scope them to the actual client party without creating a host party.
+        var campaign = Campaign.Current;
+        var previousMainParty = campaign.MainParty;
+        var previousEncounter = campaign.PlayerEncounter;
+        var encounter = new PlayerEncounter
+        {
+            _mapEvent = mapEvent,
+            PlayerSide = playerSide,
+            OpponentSide = playerSide.GetOppositeSide()
+        };
+
+        try
+        {
+            campaign.MainParty = playerParty;
+            campaign.PlayerEncounter = encounter;
+            selectJoiners(playerParties, enemyParties);
+        }
+        finally
+        {
+            campaign.PlayerEncounter = previousEncounter;
+            campaign.MainParty = previousMainParty;
+        }
+    }
+
+    private static bool TryGetPlayerParty(
+        MapEvent mapEvent,
+        out MobileParty playerParty,
+        out BattleSideEnum playerSide)
+    {
+        foreach (var side in new[] { BattleSideEnum.Attacker, BattleSideEnum.Defender })
+        {
+            foreach (var mapEventParty in mapEvent.PartiesOnSide(side))
+            {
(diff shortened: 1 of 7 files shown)


source/GameInterface/Services/MapEvents/NearbyPartyReinforcer.cs
```diff
@@ -0,0 +1,164 @@
...
+        List<MobileParty> enemyParties,
+        Action<List<MobileParty>, List<MobileParty>> selectJoiners)
+    {
+        // Vanilla reads these encounter statics, so scope them to the actual client party without creating a host party.
+        var campaign = Campaign.Current;
+        var previousMainParty = campaign.MainParty;
+        var previousEncounter = campaign.PlayerEncounter;
+        var encounter = new PlayerEncounter
+        {
+            _mapEvent = mapEvent,
```
ShoT-UPfps, line 260: this synthetic encounter never sets `EncounterSettlementAux`. v1.4.8's sally-out path falls back to `PlayerEncounter.EncounterSettlement.SiegeEvent`; a player joining an existing AI sally-out from outside that siege has no `PlayerSiegeEvent`, so this scan throws and then retries every tick. please populate the settlement state from the map event and add a regression test.

source/GameInterface/Services/MapEvents/NearbyPartyReinforcer.cs
```diff
@@ -0,0 +1,164 @@
...
+            (side, party) => side.AddNearbyPartyToPlayerMapEvent(party));
+    }
+
+    public void ReinforceOpenPlayerBattles()
+    {
+        var mapEvents = Campaign.Current?.MapEventManager?.MapEvents;
+        if (mapEvents == null)
+            return;
+
+        foreach (var mapEvent in mapEvents.ToArray())
```
ShoT-UPfps: `CampaignTick` is emitted from `Campaign.Tick`, so this runs once per campaign frame. every open player battle then allocates party lists plus a `PlayerEncounter` and runs the native locator search for the whole configured window. when campaign time is paused, that window does not advance but frame ticks continue. please throttle the follow-up scan per map event instead of polling every frame.

ShoT-UPfps (changes requested):
requesting changes for one correctness issue and one avoidable server hot path.

the v1.4.8 selector validation is the blocker: the temporary encounter does not contain the settlement state used by the sally-out path. the campaign-frame polling also needs throttling before this runs on production servers.

ci was green at `ecc399c177dc6cf7a43e363170cfe48bede7e85a` (`Build and Test` #4893) when this review was prepared.

source/GameInterface/Services/MapEvents/NearbyPartyReinforcer.cs
garrettluskey: fixed in `482c56d55`: the synthetic encounter now copies `MapEventSettlement`, with a sally-out regression test.

source/GameInterface/Services/MapEvents/NearbyPartyReinforcer.cs
garrettluskey: fixed in `482c56d55`: follow-ups now track only opened player battles and run every 0.25 campaign hours, so paused campaign time stops the scans.

source/GameInterface/Services/MapEvents/NearbyPartyReinforcer.cs
```diff
@@ -0,0 +1,257 @@
...
+    public void Reinforce(MapEvent mapEvent)
+    {
+        var encounterModel = Campaign.Current?.Models?.EncounterModel;
+        if (encounterModel == null)
+            return;
+
+        Reinforce(
+            mapEvent,
+            encounterModel.FindNonAttachedNpcPartiesWhoWillJoinPlayerEncounter,
+            (side, party) => side.AddNearbyPartyToPlayerMapEvent(party));
```
ShoT-UPfps, line 53: what clears these when the last player joiner leaves an existing AI battle? v1.4.8 only calls `RemoveNearbyPartiesFromPlayerMapEvent` while removing `PartyBase.MainParty`, but the dedicated server has no main party, so the tracked AI need an authoritative replicated cleanup or they stay in the map event after every player leaves.

ShoT-UPfps (changes requested):
the two earlier findings are fixed. one authoritative leave cleanup is still missing.

source/GameInterface/Services/MapEvents/NearbyPartyReinforcer.cs
garrettluskey: fixed in `0b28c6437`: the server now routes map-event party removals through the reinforcer, waits until no player parties remain, then removes only the tracked nearby AI with patches live and broadcasts each `MapEventPartyRemoved`. added multiplayer leave, stale tracking, and rejoin tests.

source/GameInterface/Services/MapEvents/NearbyPartyReinforcer.cs
```diff
@@ -0,0 +1,337 @@
...
+            foreach (var mapEventParty in side.Parties)
+            {
+                if (mapEventParty.Party == nearbyParty.Party)
+                {
+                    removedParty = mapEventParty;
+                    break;
+                }
+            }
+
+            nearbyParty.MapEventSide = null;
```
ShoT-UPfps, line 158: `nearbyParty.MapEventSide = null` recursively removes this root and each attached army member in v1.4.8, but this loop snapshots and publishes only the root's `MapEventPartyRemoved`. the native selector rejects `AttachedTo != null`, not an unattached army leader with `AttachedParties`, so that shape is valid. clients receive only the leader's `NetworkRemoveMapEventParty`, and the client handler clears the direct field rather than recursing, leaving attached parties in the map event after the last player leaves. snapshot and publish every removed `MapEventParty`, then add an army cleanup regression test.

ShoT-UPfps (commented):
rereviewed `0b28c643`. the earlier settlement-state, follow-up-throttling, and no-player cleanup findings are addressed, but one replication blocker remains: clearing a selected nearby army leader removes its attached parties on the server while broadcasting only the leader's removal.

source/GameInterface/Services/MapEvents/NearbyPartyReinforcer.cs
garrettluskey: fixed in `bc5f9998d`: cleanup now snapshots the side before removing the tracked root and publishes every `MapEventParty` removed by the recursive setter. added the attached-army regression test.

ShoT-UPfps (commented):
rereviewed `bc5f9998`. the attached-army cleanup now publishes every recursively removed map-event party, but the documented zero-hour AI join window still admits the synchronous initial scan.

source/GameInterface/Services/MapEvents/NearbyPartyReinforcer.cs
```diff
@@ -0,0 +1,333 @@
...
+    internal static bool CanReinforce(MapEvent mapEvent)
+    {
+        if (mapEvent == null || mapEvent.IsFinalized || mapEvent.BattleState != BattleState.None)
+            return false;
+
+        if (mapEvent.IsRaid || mapEvent.IsSiegeAssault || mapEvent.IsForcingSupplies ||
+            mapEvent.IsForcingVolunteers || mapEvent.MapEventSettlement?.IsHideout == true)
+            return false;
+
+        return InteractionPatches.IsWithinAiJoinWindow(mapEvent) && mapEvent.ContainsPlayerParty();
```
ShoT-UPfps, line 243: `playerBattleAiJoinWindowHours: 0` still passes this check during the immediate join scan. `OpenAiJoinWindowAndPublish` creates the deadline at `CampaignTime.Now`, `AiJoinWindowExpired` uses `Now > deadline`, and the synchronous `PlayerJoinedBattle` handler reaches here before campaign time advances, so the documented zero value no longer disables AI joining. make the AI expiry inclusive or explicitly reject nonpositive windows, and add a zero-window regression test.

garrettluskey: fixed in `9858bfed0`: nonpositive AI join windows now expire immediately, and the zero-hour test verifies the synchronous scan does not call the selector.

ShoT-UPfps (approved):
lgtm

Issue #1863, opened 2026-07-11
[ISSUE] Disable surrendering with a map event that has multiple players


garrettluskey:
# For New Contributors
[How to contribute](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/wiki/Getting-Started-as-a-Contributor)

## Description

Clicking surrender while another player was on my team took both of us prisoner

## Location

Create a branch based from [development](https://github.com/Bannerlord-Coop-Team/BannerlordCoop)

## Definition of Done
- [ ] Class level comments exist for all new classes.
- [ ] XUnit tests exist for every method that does not require the game to be ran.

PR #3177, opened 2026-08-19
Fix Retreated Party outcomes

### LINKED ISSUE #2939: [BATTLES] Being taken prisoner after retreating from a battle
Two player parties, party A and party B, join a battle on the same side. Party A initiated the map encounter for the battle. 

If party A retreats their party from the battle, but party B decides to fight until the end, if party B ends up losing, party A is still taken prisoner even though they have already moved on from the battle.

### LINKED ISSUE #3194: [BATTLES] Retreating from a battle can grant the winner excessive renown and influence
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

When a player joins the opposing side of a friend's battle and then retreats, the friend may receive credit as though they defeated that player, granting substantial influence. In one example, the friend received 300 renown after winning a 200-versus-200 battle.

## Steps to reproduce

Join the opposing side of a friend's ongoing battle with a larger force, then immediately retreat. If the friend wins, the friend receives renown and influence as though they defeated you.

## Additional information

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1539919259096842370/1539919259096842370

## Logs

Client log: Not provided

Server log: Not provided

## Additional report details

- The player joined a 200-versus-200 battle with approximately 700 additional units, retreated via TAB > Retreat, and their party remained intact; the friend received the excessive renown and influence after winning the battle. ([Discord message](https://discord.com/channels/709516043332354119/1539919259096842370/1539956692211732521))

codebydean:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

When multiple allied player parties participate in the same battle, a player who retreats remains structurally attached to the map event so they can re-enter. If the remaining allied party later loses, result processing treats every party still attached to the defeated side as defeated. Consequently, the player who had already retreated is also taken prisoner.

## What is the new behavior?
Resolves #2939
Resolved #3194 

Player parties that gracefully retreat from a shared battle are tracked separately from their structural map-event membership. Retreated parties are excluded from the battle's later result processing, preventing them from being captured or receiving other victory/defeat consequences when an ally continues fighting.

If a retreated player **re-enters** the battle before it concludes, their retreat state is **cleared**, and they become eligible for the final result again. Retreat tracking is also cleared when the map event is finalized.

This preserves the existing ability to re-enter an ongoing battle and does not change the current party-specific surrender behavior related to #1863.

## Bot Changelog Entry
Players who retreat from a shared battle are no longer taken prisoner if an ally continues fighting and later loses.

## Other information
Added end-to-end regression coverage for:

- A retreated player remaining free when an allied player later loses.
- A retreated player re-entering the battle and becoming eligible for capture again.

Existing partial-surrender and retreat-casualty regression tests continue to pass.


codebydean:
Since this is related to #1863 

Should surrender be unavailable whenever another allied player party exists, or should individual partial surrender remain allowed?

source/Coop.Core/Server/Services/Instances/Handlers/ServerRetreatedMapEventPartyHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    private void Handle_MissionMemberEntered(MessagePayload<MissionMemberEntered> payload)
+    {
+        if (TryResolve(payload.What.ControllerId, payload.What.InstanceId, out var mapEvent, out var party))
+            tracker.MarkReentered(mapEvent, party);
+    }
+
+    private void Handle_MissionMemberDeparted(MessagePayload<MissionMemberDeparted> payload)
+    {
+        if (!payload.What.WasRetreat) return;
```
ShoT-UPfps: can this use a signal that only the retreat path sends? `CoopBattleController.OnLeaving` sends `NetworkMissionLeft` for normal battle completion too, and `ServerMissionMembershipHandler` publishes every one of those with `WasRetreat = true`, so a client that leaves normally while another result is still pending gets excluded from the final outcome.

source/GameInterface/Services/MapEvents/Interfaces/MapEventResultsInterface.cs
```diff
@@ -175,8 +179,10 @@ public void CalculateAndCommitMapEventResults(MapEvent mapEvent, out NetworkPlay
         {
             if (mapEvent.BattleState == BattleState.AttackerVictory || mapEvent.BattleState == BattleState.DefenderVictory)
             {
-                MBList<MapEventParty> defeatedParties = mapEvent.GetMapEventSide(mapEvent.DefeatedSide).Parties.ToMBList<MapEventParty>();
-                MBList<MapEventParty> winnerParties = mapEvent.GetMapEventSide(mapEvent.WinningSide).Parties.ToMBList<MapEventParty>();
+                MBList<MapEventParty> defeatedParties = mapEvent.GetMapEventSide(mapEvent.DefeatedSide).Parties
+                    .Where(party => !retreatedPartyTracker.IsRetreated(mapEvent, party.Party)).ToMBList<MapEventParty>();
```
ShoT-UPfps, line 184: should retreated parties also be excluded from `CommitCalculatedMapEventResults`? they stay in `side.Parties`, so this only skips loot and capture while the later commit still applies their xp, renown, influence, morale, and gold changes.

ShoT-UPfps (changes requested):
requesting changes for the normal-leave retreat misclassification; the second inline comment also needs clarification before this is merged.

garrettluskey:
> Since this is related to #1863
> 
> Should surrender be unavailable whenever another allied player party exists, or should individual partial surrender remain allowed?

I like this solution

source/Coop.Core/Server/Services/Instances/Handlers/ServerRetreatedMapEventPartyHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+/// <summary>Tracks graceful battle withdrawals separately from structural map-event membership.</summary>
+internal sealed class ServerRetreatedMapEventPartyHandler : IHandler
+{
+    private readonly IMessageBroker messageBroker;
+    private readonly IObjectManager objectManager;
+    private readonly IPlayerManager playerManager;
+    private readonly IRetreatedMapEventPartyTracker tracker;
+
+    public ServerRetreatedMapEventPartyHandler(IMessageBroker messageBroker, IObjectManager objectManager,
+        IPlayerManager playerManager, IRetreatedMapEventPartyTracker tracker)
```
garrettluskey: `GameInterfaceModule` isn't loaded by `Coop.Tests`, while the server test container auto-activates this handler. this dependency is unresolved and the `test` job currently fails 85 tests during container construction; register the tracker in the server test composition too.

source/Coop.Core/Server/Services/Instances/Handlers/ServerRetreatedMapEventPartyHandler.cs
garrettluskey: `WasRetreat` isn't retreat-specific here. `ServerMissionMembershipHandler` sets it to true for every `NetworkMissionLeft`, including `CoopBattleController.OnLeaving` after a normal resolved battle, so that player can be excluded while another result is still pending. drive this from a retreat-only signal.

source/GameInterface/Services/MapEvents/Interfaces/MapEventResultsInterface.cs
garrettluskey, line 184: this only removes retreated parties from loot and capture inputs. `mapEvent.CommitCalculatedMapEventResults()` still iterates every party in both sides and applies xp, renown, influence, morale, and gold to the retreated party; exclude it from commit too.

garrettluskey (changes requested):
tracking retreat separately from map-event membership is the right direction, but normal leaves are still misclassified, result commit still includes those parties, and the new handler currently breaks the server test container.

codebydean:
Re-review @ShoT-UPfps @garrettluskey

codebydean:
Issue #3194 was reported and appears to share the same underlying cause.

The reproduction based on the reporter's discord comment is this:
- Player A starts a roughly 200-vs-200 battle against an NPC party.
- Player B joins the NPC side with roughly 700 troops.
- Player B retreats through Tab → Retreat, and their party remains intact.
- Player A later defeats the remaining NPC party.
- After the battle ends, Player A receives excessive renown and influence, apparently including credit for Player B’s retreated force.

This PR currently tracks deliberate retreats and excludes retreated parties from later loot, captivity, and per-party result commits. However, #3194 may require excluding retreaters **earlier**, while native battle rewards are **calculated**, so they **do not inflate** the opposing winner’s calculated renown and influence.

Do you want me to possibly implement the solution to #3194 on this PR or open a new one once this is merged?

source/Missions/Battles/CoopBattleController.cs
```diff
@@ -538,6 +538,8 @@ protected override void OnLeaving()
         // Retry the result-ready report before tearing the instance down. Duplicate reports are idempotent.
         ResultCommitter.ReportResolvedResult(missionResult);
 
-        lifecycle.Leave();
+        bool wasRetreat = missionResult?.BattleResolved != true && !ResultCommitter.TryGetResolvedState(out _);
```
garrettluskey: this still treats every unresolved mission teardown as a deliberate retreat. `CoopBattleMissionSpawnHandler.AbortInvalidBattle` calls `Mission.EndMission()` with no result after the instance was joined, so this sends `NetworkBattleRetreated` for a party that never chose to retreat, drive this from the actual retreat action instead.

source/Coop.Core/Server/Services/Instances/Handlers/ServerRetreatedMapEventPartyHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            return;
+        }
+
+        messageBroker.Publish(this, new BattlePartyRetreated(player.ControllerId, payload.What.InstanceId));
+    }
+
+    private void Handle_BattlePartyRetreated(MessagePayload<BattlePartyRetreated> payload)
+    {
+        if (TryResolve(payload.What.ControllerId, payload.What.InstanceId, out var mapEvent, out var party))
```
garrettluskey: this runs synchronously from `NetworkBattleRetreated` on the Poller thread, then `TryResolve` reads campaign state and changes eligibility while results run on the game thread. put the lookup and `MarkRetreated` inside `GameThread.RunSafe`, otherwise loot and commit can observe different retreat states.

source/GameInterface/Services/MapEvents/Interfaces/MapEventResultsInterface.cs
```diff
@@ -175,8 +180,10 @@ public void CalculateAndCommitMapEventResults(MapEvent mapEvent, out NetworkPlay
...
```
garrettluskey, line 184: these filters run after `CalculateMapEventResults()` has used every party to calculate `GoldLost`, winner `PlunderedGold`, and renown/influence/morale shares. line 214 then skips the retreater's loss but pays the inflated winner values, so #3194 remains and can create gold, exclude retreaters during calculation too.

garrettluskey (changes requested):
separating retreat state from map-event membership is the right direction, and the old result-commit/container issues are fixed. changes are still needed for retreat classification, thread ordering, and reward calculation.

source/Coop.Core/Server/Services/Instances/Handlers/ServerRetreatedMapEventPartyHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        messageBroker.Unsubscribe<NetworkBattleRetreated>(Handle_NetworkBattleRetreated);
+        messageBroker.Unsubscribe<BattlePartyRetreated>(Handle_BattlePartyRetreated);
+        messageBroker.Unsubscribe<MissionMemberEntered>(Handle_MissionMemberEntered);
+        messageBroker.Unsubscribe<MapEventFinalized>(Handle_MapEventFinalized);
+    }
+
+    private void Handle_NetworkBattleRetreated(MessagePayload<NetworkBattleRetreated> payload)
+    {
+        if (payload.Who is not NetPeer peer || !playerManager.TryGetPlayer(peer, out var player))
```
ShoT-UPfps: should this verify that `peer` is still the controller's current peer, like `ServerMissionMembershipHandler.TryGetCurrentController` does? `SetPeer` leaves the old `peerToPlayer` mapping in place on reconnect, so a late retreat packet from the superseded connection can mark the re-entered current player's party as retreated.

source/Coop.Core/Server/Services/Instances/Handlers/ServerRetreatedMapEventPartyHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    private void Handle_MissionMemberEntered(MessagePayload<MissionMemberEntered> payload)
+    {
+        if (TryResolve(payload.What.ControllerId, payload.What.InstanceId, out var mapEvent, out var party))
+            tracker.MarkReentered(mapEvent, party);
+    }
+
+    private void Handle_MapEventFinalized(MessagePayload<MapEventFinalized> payload)
+    {
+        tracker.Clear(payload.What.MapEvent);
```
ShoT-UPfps, line 69: `MapEventFinalized` is published after `FinalizeEventAux` has already applied defeated-side cleanup. vanilla still applies `GetDefeatMoraleChange` to every party on `DefeatedSide`, so a tracked retreater receives a defeat penalty before this clear; should that path exclude retreaters too?

ShoT-UPfps (commented):
two additional correctness issues remain beyond the three current-head blocker threads already open.

codebydean:
Enabled auto-merge

codebydean:
The new fixes I believe satisfy your review comments and also Resolves #3194

ShoT-UPfps (commented):
rereviewed `127c6455` on current `development` `f5cab14b`. this head is only a merge of development into the previously reviewed `cd14f22f`; the incremental delta is five unrelated Issue-system files, so the retreat implementation is unchanged. the retreat-only signal, current-peer authentication, game-thread ordering, per-party commit filtering, and defeated-side morale exclusion remain fixed. one P1 reward blocker remains: the `CalculateMapEventResults` prefix removes an intact retreater from the defeated-party loop while v1.4.8 still reads cached full-side `StrengthOfSide`, `RenownValue`, and `InfluenceValue`; the retreater's current strength is therefore not subtracted, so the #3194 700-troop force is counted as destroyed. derive every reward input from one consistent eligible-party snapshot and add coverage that executes the real v1.4.8 calculation. Build and Test #5075 passes on this exact head, but the current outcome test disables both `CalculateMapEventResults` and `CommitCalculatedMapEventResults`, so it cannot catch the reward error.

source/Missions/Battles/BattleResultReadyLogic.cs
```diff
@@ -63,4 +67,11 @@ private void Handle_BattleHostAuthorityAcquired(MessagePayload<BattleHostAuthori
             resultCommitter.ReportAcceptedResult();
         }, context: nameof(Handle_BattleHostAuthorityAcquired));
     }
+
+    // Only report retreats triggered by the retreat action
+    public override void OnRetreatMission()
+    {
+        if (session.HasInstance)
+            relayNetwork.SendAll(new NetworkBattleRetreated(session.InstanceId));
```
garrettluskey, line 75: `OnRetreatMission` isn't retreat-only in v1.4.8. `Mission.OnEndMissionResult()` calls `RetreatMission()`, which invokes this callback after a normal battle result, so a player who finishes first can still be marked retreated while another result is pending. drive this from a callback or state that distinguishes an actual retreat and add a normal-result regression.

garrettluskey (changes requested):
the current-peer check, game-thread dispatch, commit filtering, and morale exclusion fix the earlier paths. changes are still needed because the new callback also fires for normal battle results, and the existing reward-calculation thread remains open: https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/3177#discussion_r3822871907

gamingmenacers-svg (changes requested):
reviewed exact head `127c64552` against current `development`. the open reward blocker is still present: removing the retreater from `side.Parties` immediately before `CalculateMapEventResults` does not rebuild the cached side strength/value inputs vanilla uses, so the large retreated force can still inflate the winner's renown and influence. the current E2E test disables the real calculate and commit methods, so it doesnt cover #3194's calculation path. this also needs rebasing from 94 commits behind before merge.

source/GameInterface/Services/MapEvents/Patches/MapEventPatches.cs
```diff
@@ -270,6 +272,95 @@ internal static int RemovePartiesWithoutParty(MapEvent mapEvent)
...
+    private static void Prefix_CalculateMapEventResults(MapEvent __instance, out List<RemovedMapEventParty> __state)
+    {
+        __state = null;
+        if (!ModInformation.IsServer || !ContainerProvider.TryResolve<IRetreatedMapEventPartyTracker>(out var tracker)) return;
+
+        __state = RemoveParties(__instance, party => tracker.IsRetreated(__instance, party.Party));
+
+        if (__state.Count == 0) return;
+        
+        __instance.RecalculateStrengthOfSides();
```
garrettluskey: `RecalculateStrengthOfSides()` rebuilds `StrengthOfSide` from the post-casualty rosters. v1.4.8 treats that field as the battle baseline and subtracts those same current party strengths in `CalculateWinnerPartiesRenownInfluenceAndMoraleShares`, so the multiplier becomes `0` and winners get no renown or influence whenever a retreater exists. preserve an eligible pre-battle baseline instead of replacing it with current strength.

source/E2E.Tests/Services/Missions/RetreatedPartyOutcomeTests.cs
```diff
@@ -0,0 +1,184 @@
...
+
+    private void CommitEnemyVictory(string mapEventId)
+    {
+        var disabledMethods = MapEventDisabledMethods
+            .Append(AccessTools.Method(typeof(DefaultBattleRewardModel), nameof(DefaultBattleRewardModel.GetCaptureMemberChancesForWinnerParties)))
+            .Append(AccessTools.Method(typeof(MapEvent), "LootDefeatedPartyCasualties"))
+            .Append(AccessTools.Method(typeof(MapEvent), "LootDefeatedPartyItems"))
+            .Append(AccessTools.Method(typeof(MapEvent), "LootDefeatedPartyPrisoners"))
+            .Append(AccessTools.Method(typeof(MapEvent), "LootDefeatedPartyShips"))
+            .Append(AccessTools.Method(typeof(MapEvent), "CalculateMapEventResults"))
```
garrettluskey: this disables both the real calculate and commit methods, so the #3194 path still isnt exercised and the zero-reward regression above passes. run the real calculation and assert that the remaining winner keeps the expected renown/influence while the retreater contributes none.

garrettluskey (changes requested):
the retreat signal is now authenticated to the current peer, ordered ahead of mission departure, and cleared on reentry and finalization. the result filters also cover commit and defeated-side morale, but the reward baseline is still wrong and needs another pass.

source/GameInterface/Services/MapEvents/Patches/MapEventPatches.cs
```diff
@@ -270,6 +272,86 @@ internal static int RemovePartiesWithoutParty(MapEvent mapEvent)
...
 
+    // Exclude retreated parties from reward calculations
+    [HarmonyPatch("CalculateMapEventResults")]
+    [HarmonyPrefix]
+    private static void Prefix_CalculateMapEventResults(MapEvent __instance, out List<RemovedMapEventParty> __state)
+    {
+        __state = null;
+        if (!ModInformation.IsServer || !ContainerProvider.TryResolve<IRetreatedMapEventPartyTracker>(out var tracker)) return;
+
+        __state = RemoveParties(__instance, party => tracker.IsRetreated(__instance, party.Party));
```
garrettluskey, line 283: this removes the retreater from `defeatedParties` but leaves its strength in `StrengthOfSide` and the cached `RenownValue`/`InfluenceValue`. `CalculateWinnerPartiesRenownInfluenceAndMoraleShares` subtracts only the remaining defeated parties' current strength, so the intact 700-troop retreater in #3194 is counted as destroyed again. exclude its contribution from the pre-battle reward inputs too, and cover a losing-side retreater with an expected reward assertion.

garrettluskey (changes requested):
the resolved-result guard, game-thread dispatch, and per-party outcome filters fix the earlier paths, and the reward test now runs the real calculation. removing the strength recalculation avoids zero rewards, but restores the inflated rewards from #3194 when the retreater is on the losing side.

ShoT-UPfps (approved):
Rereviewed exact head `5684126beb182dc6ba4591bb06ff3419f3a84b4f` against `development` at `4a84d5770bf3e6b3f1285f2e1f9fce207b82418f`, including the complete 20-file diff, current checks, and previous review discussions. I found no new actionable blocking regression in this head.

The latest reward fix addresses the outstanding intact losing-side retreater case from #3194: it subtracts the removed party's surviving strength from the cached battle baseline and recalculates the side reward values before calculating the eligible parties' shares, rather than replacing the entire baseline with post-casualty strength. The added losing-side regression executes the real reward calculation and asserts expected renown and influence; the winning-side regression also exercises result commits. This resolves the earlier inflated-reward/zero-reward concerns for those covered cases.

The earlier retreat-classification, current-peer authentication, game-thread retreat dispatch, commit filtering, and defeated-side morale concerns are also addressed in the current implementation. Normal resolved mission results are guarded before the retreat callback; ordinary teardown does not emit the dedicated retreat signal. Re-entry and map-event finalization clear the corresponding tracking state.

Validation: all 12 reported checks on this exact head are successful, including build, test, all eight E2E shards, the verification-plan check, and CodeFactor. I additionally checked the affected reward and mission/finalization semantics against the installed v1.4.8 managed implementation. I did not run a new local build or an interactive multiplayer gameplay reproduction. Approving on the reviewed code and available CI evidence.

This review applies only to head `5684126beb182dc6ba4591bb06ff3419f3a84b4f` compared from `4a84d5770bf3e6b3f1285f2e1f9fce207b82418f` with base tip `4a84d5770bf3e6b3f1285f2e1f9fce207b82418f` as validated at `2026-09-15T13:53:12.013643Z`. Later revisions and later development/base movement were not evaluated by this review.

- Pull request: https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/3177
- Original verdict: `APPROVE`
- Reviewed head: `5684126beb182dc6ba4591bb06ff3419f3a84b4f`
- Reviewed base tip: `4a84d5770bf3e6b3f1285f2e1f9fce207b82418f`
- Comparison base: `4a84d5770bf3e6b3f1285f2e1f9fce207b82418f`

This review applies only to head `5684126beb182dc6ba4591bb06ff3419f3a84b4f` compared from `4a84d5770bf3e6b3f1285f2e1f9fce207b82418f` with base tip `4a84d5770bf3e6b3f1285f2e1f9fce207b82418f` as validated at `2026-09-15T13:53:12.013643Z`. Later revisions and later development/base movement were not evaluated by this review.

## Exact review summary

```text
Rereviewed exact head `5684126beb182dc6ba4591bb06ff3419f3a84b4f` against `development` at `4a84d5770bf3e6b3f1285f2e1f9fce207b82418f`, including the complete 20-file diff, current checks, and previous review discussions. I found no new actionable blocking regression in this head.

The latest reward fix addresses the outstanding intact losing-side retreater case from #3194: it subtracts the removed party's surviving strength from the cached battle baseline and recalculates the side reward values before calculating the eligible parties' shares, rather than replacing the entire baseline with post-casualty strength. The added losing-side regression executes the real reward calculation and asserts expected renown and influence; the winning-side regression also exercises result commits. This resolves the earlier inflated-reward/zero-reward concerns for those covered cases.

The earlier retreat-classification, current-peer authentication, game-thread retreat dispatch, commit filtering, and defeated-side morale concerns are also addressed in the current implementation. Normal resolved mission results are guarded before the retreat callback; ordinary teardown does not emit the dedicated retreat signal. Re-entry and map-event finalization clear the corresponding tracking state.

Validation: all 12 reported checks on this exact head are successful, including build, test, all eight E2E shards, the verification-plan check, and CodeFactor. I additionally checked the affected reward and mission/finalization semantics against the installed v1.4.8 managed implementation. I did not run a new local build or an interactive multiplayer gameplay reproduction. Approving on the reviewed code and available CI evidence.
```

## Exact inline findings

None.
## Exact thread actions

None.

garrettluskey (approved):
the retreat-only signal, ordered tracking, and outcome filters address the earlier findings. the reward calculation now excludes the intact retreater without discarding the remaining battle losses, with regression coverage for both winning and losing sides. no actionable correctness or simplicity blockers found.

all 12 checks pass. reviewed surrounding callers and decompiled game methods; no new local build or live multiplayer test was run.

PR #3180, opened 2026-08-19
Keep released offline player parties parked

### LINKED ISSUE #2574: [ARMIES] Enemy army cannot capture an offline player's party
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

A large enemy army appears unable to capture an offline player's AI-controlled party and continues engaging it.

## Steps to reproduce

1. Leave a player's character active while that player is offline. 2. Have a large enemy army attack and defeat the AI-controlled party. 3. Observe the party reduced to one member while the army remains nearby and continues engaging it.

## Additional information

The server runs on Linux through Wine; all players involved so far have used Windows.

## Discord source

https://discord.com/channels/709516043332354119/1533199787988488374/1533199787988488374

## Logs

Client log: Not provided

Server log: [Download coop-server-20260801-183544.log](<https://cdn.discordapp.com/attachments/1533199787988488374/1533199788391137480/coop-server-20260801-183544.log?ex=6aa45b13&is=6aa30993&hm=bb5a43696899483d56b2ed274c76e1cf6123030475e2118c63a24af8c2a800ea&>)

garrettluskey:
## PR Checklist

## PR Type

Captivity release reactivates and recreates an offline or still-loading player's party, making the one-hero party attackable again.

## What is the new behavior?
Resolves #2574

Captivity state and release position are restored immediately, but party visibility is restored only for a connected player that completed campaign synchronization. Offline and loading parties stay inactive and hidden until the existing synchronization handler activates them.

Tested with the full `Coop.Tests` suite in Release: 700 passed, 1 skipped.

## Bot Changelog Entry
- Released offline player parties no longer reappear on the map and get captured repeatedly.

### DIFF
--- a/source/Coop.Core/Server/Connections/ConnectionCollection.cs
+++ b/source/Coop.Core/Server/Connections/ConnectionCollection.cs
@@ -1,6 +1,7 @@
-using Common.Messaging;
+﻿using Common.Messaging;
 using Common.Network.Messages;
 using Coop.Core.Server.Connections.Messages;
+using Coop.Core.Server.Connections.States;
 using LiteNetLib;
 using System;
 using System.Collections;
@@ -18,6 +19,7 @@ namespace Coop.Core.Server.Connections;
 public interface IConnectionCollection : IEnumerable<IConnectionLogic>, IDisposable
 {
     IEnumerable<IConnectionLogic> LoadingPeers { get; }
+    bool HasCompletedCampaignSynchronization(NetPeer peer);
 }
 
 /// <inheritdoc cref="IConnectionCollection"/>
@@ -28,6 +30,10 @@ public class ConnectionCollection : IConnectionCollection
     public IEnumerable<IConnectionLogic> LoadingPeers => ConnectionStates
         .Select(conn => conn.Value)
         .Where(conn => conn.IsLoading);
+
+    public bool HasCompletedCampaignSynchronization(NetPeer peer) =>
+        ConnectionStates.TryGetValue(peer, out var connection) &&
+        (connection.State is CampaignState || connection.State is MissionState);
 
     private readonly IMessageBroker messageBroker;
     private readonly ConnectionContext connectionContext;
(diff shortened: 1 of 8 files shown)


source/Coop.Core/Server/Connections/ConnectionCollection.cs
```diff
@@ -29,6 +31,10 @@ public class ConnectionCollection : IConnectionCollection
         .Select(conn => conn.Value)
         .Where(conn => conn.IsLoading);
 
+    public bool HasCompletedCampaignSynchronization(NetPeer peer) =>
+        ConnectionStates.TryGetValue(peer, out var connection) &&
+        (connection.State is CampaignState || connection.State is MissionState);
```
ShoT-UPfps: could we snapshot `connection.State` once here? each getter call locks independently, so a `MissionState` to `CampaignState` transition between the two reads makes both checks false even though both states are synchronized. the release handler then parks the connected party, and mission exit never publishes `PlayerCampaignSynchronized` to restore it.

ShoT-UPfps (changes requested):
one correctness issue found.

ShoT-UPfps (approved):
lgtm

source/Coop.Core/Server/Services/Players/Handlers/PlayerPartyVisibilityHandler.cs
```diff
@@ -186,13 +193,45 @@ private void Handle_PlayerCampaignSynchronized(MessagePayload<PlayerCampaignSync
...
+
+        ActivateParty(party, player.MobilePartyId);
+        Logger.Information("Restored released party {PartyId} for peer {Peer}", party.StringId, peer.Id);
+    }
+
+    private void ActivateParty(MobileParty party, string mobilePartyId)
+    {
+        party.IsActive = true;
+        CreateVisual(party, mobilePartyId);
+        party.Party.UpdateVisibilityAndInspected(party.Position);
```
gamingmenacers-svg, line 232: At-sea releases used to skip this call, and the client handler still does. Here it is unconditional. On a dedicated server `MobileParty.MainParty` is null, so the native visibility calculation can dereference it when true sight is off. The broker then logs/swallow the exception after the party has been activated and its visual recreated, leaving `IsVisible` false. Could we set `IsVisible = true` directly here since this is server-only, or restore the sea guard plus a null-main-party fallback and add an at-sea release test?

PR #3181, opened 2026-08-19
Weight player defenders in siege target scoring

### LINKED ISSUE #3053: [SIEGES] AI siege armies abandon intended targets and loop through other settlements when the player enters
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

AI siege stacks (about 2,000-3,000 troops) stop advancing to their intended target when the player enters the settlement, immediately switch to a different target, and do not start a siege at the original one.

## Steps to reproduce

In multiple ongoing wars (including 200 vs 3,000 and 50 vs 3,000), the AI abandons the intended siege, leaves immediately, and repeats switching settlements until it never commits to the original siege.

## Additional information

Observed in base game. If a siege is allowed to begin without the player present, the stack can set up with siege towers and catapults, enter siege mode, and remain stuck until starvation. Initial report had no logs.

## Discord source

https://discord.com/channels/709516043332354119/1538204234506567690/1538204234506567690

## Logs

Client log: Not provided

Server log: Not provided

garrettluskey:

Dedicated servers count controlled client parties inside a settlement at full strength during siege target scoring because vanilla only recognizes `MainHero` and `MainParty`. This can push the pre-siege score below its strength threshold and make an AI army retarget.

## What is the new behavior?

Controlled defenders and player-led army parties use vanilla's player weights, with the settlement-wide player-presence factor applied once after aggregation. NPC-only targets, nearby parties outside the settlement, hourly retargeting, and AI authority are unchanged.

Fixes #3053
Addresses the siege-target loss symptom in #2670

## How to test

`dotnet test source/GameInterface.Tests/GameInterface.Tests.csproj -c Release --filter "FullyQualifiedName~PlayerSiegeTargetScoringTests|FullyQualifiedName~PatchTest.HarmonyPatchesAll"`

## Bot Changelog Entry

- AI siege armies no longer overestimate settlement defenses when co-op players are inside.

## Scope note

This PR fixes #3053's player-entry siege-target-scoring and retargeting reproduction. The separate observation that an already-started siege can remain in preparation until the army starves is tracked by #2735 and PR #3182.


source/GameInterface/Services/Armies/Patches/PlayerSiegeTargetScoringPatches.cs
```diff
@@ -0,0 +1,117 @@
...
+        MethodInfo currentSettlementGetter = AccessTools.PropertyGetter(typeof(Hero), nameof(Hero.CurrentSettlement));
+        MethodInfo applyAdjustment = AccessTools.Method(
+            typeof(PlayerSiegeTargetScoringPatches),
+            nameof(ApplyPlayerSettlementDefense));
+
+        int playerPresenceIndex = FindPlayerPresenceCalculation(
+            codes,
+            mainHeroGetter,
+            currentSettlementGetter);
+        if (playerPresenceIndex < 4)
```
ShoT-UPfps, line 38: this throws on the second registered copy of the transpiler. `ContainerTest` installs the assembly under three Harmony IDs while earlier patches stay live, so the first copy removes this sequence and the next copy fails here. can this detect the already-injected `ApplyPlayerSettlementDefense` call and leave the method alone instead?

ShoT-UPfps (changes requested):
one blocker: applying the patch assembly under a second Harmony ID throws, which is already failing `ContainerTest` in CI.

source/GameInterface/Services/Armies/Patches/PlayerSiegeTargetScoringPatches.cs
```diff
@@ -0,0 +1,118 @@
...
+
+    private static void ApplyPlayerSettlementDefense(
+        Settlement targetSettlement,
+        ref float totalStrength,
+        ref float mobileLordStrength)
+    {
+        if (!ContainerProvider.TryResolve<IPlayerSiegeTargetScoring>(out var scoring))
+        {
+            Logger.Error("Unable to resolve {Scoring}", nameof(IPlayerSiegeTargetScoring));
+            return;
```
ShoT-UPfps, line 124: this return changes later singleplayer scoring after co-op teardown. the patches stay installed after `ContainerProvider` is cleared, but this transpiler replaced `MainHero.CurrentSettlement == targetSettlement` with false, so a player inside the target no longer gets vanilla's `0.8` presence factor.

ShoT-UPfps (changes requested):
the original Harmony reapplication issue is fixed. one blocker remains in the no-container fallback, which drops vanilla's player-presence factor after teardown.

source/GameInterface/Services/Armies/Patches/PlayerSiegeTargetScoringPatches.cs
garrettluskey: fixed in `d96c74788` and `7b55b7c11`. the no-container path reapplies vanilla’s `0.8` presence factor, and leaves the totals unchanged when the dedicated server has no player character.

ShoT-UPfps (approved):

PR #3182, opened 2026-08-19
Resolve starving AI siege stalls

### LINKED ISSUE #2735: [SIEGES] Seonon AI sieges stop progressing and never launch assaults
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

On a dedicated server hosted on another computer, Seonon’s siege stops progressing while the city is at war, and the AI never launches an assault. The city can remain under siege for weeks while its armies slowly starve; the siege stops when the city is not at war.

## Steps to reproduce

Not provided

## Additional information

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1534584063145279599/1534584063145279599

## Images and videos

[](<https://cdn.discordapp.com/attachments/1534584063145279599/1534584064030408865/image.png?ex=6aa4c708&is=6aa37588&hm=f42b2d7d36114d77e3ded79ac9efa62cd98e08924dd14c0b06ac40e33a23ed00&>) ([Discord message](https://discord.com/channels/709516043332354119/1534584063145279599/1534584063145279599))

[](<https://cdn.discordapp.com/attachments/1534584063145279599/1534584065959788604/image.png?ex=6aa4c708&is=6aa37588&hm=e76fefc7d866464e5e0c287462386c7ee3ecfa368c793e489a76c045df92ce1d&>) ([Discord message](https://discord.com/channels/709516043332354119/1534584063145279599/1534584063145279599))

## Logs

Client log: [0.1.1+324d59795711727d40bc1aa3eb56a652ae656306 - Download Coop_client.log](<https://cdn.discordapp.com/attachments/1534584063145279599/1534584065275986140/Coop_client.log?ex=6aa4c708&is=6aa37588&hm=7ee429a7cc16b5701b047c66d1f4ec7f0ee0f7c6f88ffe2b56ad88313e0f40ed&>)

Server log: [Download coop-server-20260805-160009.log](<https://cdn.discordapp.com/attachments/1534584063145279599/1534584064751960076/coop-server-20260805-160009.log?ex=6aa4c708&is=6aa37588&hm=10cc321f3d8d978701e4960e8c2119080c191e3fea017e8a46a057a182c9cc0b&>)

garrettluskey:
## PR Checklist

## PR Type

AI siege readiness combines power, equipment state, elapsed time, and a random roll. If the roll never produces an assault before the besieging army starves, vanilla disperses the army for its food problem but leaves the siege objective active, so the remaining leader can keep rebuilding indefinitely.

## What is the new behavior?

Resolves #2735

A server-side terminal policy runs after vanilla disperses an army for a food problem. It starts a prepared assault through `StartBattleAction` only when the current vanilla readiness calculation has a positive assault chance; otherwise it removes the siege parties through the vanilla camp authority so they can replan. Player-led sieges and active map-event transitions are ignored.

## Bot Changelog Entry

- Starving AI siege armies now assault when viable or abandon the siege instead of rebuilding forever.

## Other information

Automated tests:
- `dotnet test source/GameInterface.Tests/GameInterface.Tests.csproj -c Release --filter "FullyQualifiedName~AiSiege"` (10 passed)
- full `GameInterface.Tests` run (1044 passed, 11 skipped)

Manual Danustica check on the server:
1. Run `coop.debug.siege.start town_ES1`.
2. Run `coop.debug.siege.terminal_status town_ES1` to inspect preparation, starvation, power ratio, equipment-adjusted ratio, and the pending decision.
3. Run `coop.debug.siege.resolve_starvation town_ES1` to simulate the food terminal. A viable prepared siege creates the assault map event; an inviable or unprepared siege ends and its parties return to AI planning.

### DIFF
--- a/source/GameInterface/Services/BesiegerCamps/Patches/BesiegerCampAssaultPatches.cs
+++ b/source/GameInterface/Services/BesiegerCamps/Patches/BesiegerCampAssaultPatches.cs
@@ -1,10 +1,6 @@
-﻿using GameInterface.Services.Heroes.Extensions;
+﻿using GameInterface.Services.SiegeEvents;
 using HarmonyLib;
-using TaleWorlds.CampaignSystem;
-using TaleWorlds.CampaignSystem.Party;
 using TaleWorlds.CampaignSystem.Siege;
-using TaleWorlds.Core;
-using TaleWorlds.Library;
 
 namespace GameInterface.Services.BesiegerCamps.Patches;
 
@@ -20,71 +16,13 @@ internal class BesiegerCampAssaultPatches
     [HarmonyPrefix]
     private static bool StartingAssaultIsLogicalPrefix(BesiegerCamp __instance, ref bool __result)
     {
-        var siegeEvent = __instance.SiegeEvent;
-        var settlement = siegeEvent.BesiegedSettlement;
-
-        // The player-inside discount is a constant factor, so one pass collects both.
-        bool playerDefenderInside = false;
-        float defenderStrength = 0f;
-        foreach (var party in settlement.GetInvolvedPartiesForEventType())
+        if (!ContainerProvider.TryResolve<IAiSiegeAssaultReadiness>(out var readiness))
         {
-            if (!party.IsMobile || party.MobileParty.CurrentSettlement != settlement) continue;
-
-            if (party.LeaderHero != null && party.LeaderHero.IsPlayerHero())
-            {
-                playerDefenderInside = true;
-            }
-
-            if (party.MobileParty.Aggressiveness > 0.01f || party.MobileParty.IsMilitia || party.MobileParty.IsGarrison)
-            {
-                defenderStrength += party.CalculateCurrentStrength();
-            }
+            __result = false;
+            return false;
         }
 
-        defenderStrength *= playerDefenderInside ? 0.5f : 1f;
-
-        float attackerStrength = 0f;
-        foreach (var party in siegeEvent.BesiegerCamp.GetInvolvedPartiesForEventType())
-        {
-            attackerStrength += party.CalculateCurrentStrength();
-        }
-
-        bool hasRam = false;
-        bool hasSiegeTower = false;
-        foreach (var engine in siegeEvent.GetSiegeEventSide(BattleSideEnum.Attacker).SiegeEngines.AllSiegeEngines())
-        {
-            if (!engine.IsConstructed) continue;
-
-            if (engine.SiegeEngine == DefaultSiegeEngineTypes.Ram || engine.SiegeEngine == DefaultSiegeEngineTypes.ImprovedRam)
-            {
-                hasRam = true;
-            }
-            else if (engine.SiegeEngine == DefaultSiegeEngineTypes.SiegeTower)
-            {
-                hasSiegeTower = true;
-            }
-        }
-
-        float settlementAdvantage = Campaign.Current.Models.CombatSimulationModel.GetSettlementAdvantage(settlement);
-        float graceHours = (float)CampaignTime.HoursInDay * 4f;
-        float advantageExponent = 0.8f - ((siegeEvent.SiegeStartTime.ElapsedHoursUntilNow > graceHours)
-            ? ((siegeEvent.SiegeStartTime.ElapsedHoursUntilNow - graceHours) * 0.02f)
-            : 0f);
-        if (!hasRam) advantageExponent *= 1.25f;
-        if (!hasSiegeTower) advantageExponent *= 1.25f;
-
-        float powerRatio = attackerStrength / (defenderStrength * MathF.Pow(settlementAdvantage, advantageExponent));
-        __result = false;
-        if (powerRatio > 1f)
-        {
-            int equipmentsBuilt = Campaign.Current.Models.CombatSimulationModel.GetNumberOfEquipmentsBuilt(settlement);
-            powerRatio *= (float)MathF.Min(3, equipmentsBuilt) / 3f;
-            float equipmentProgress = Campaign.Current.Models.CombatSimulationModel.GetMaximumSiegeEquipmentProgress(settlement) + (0.25f * (float)(5 - equipmentsBuilt));
-            powerRatio *= 1f - (0.85f * (equipmentProgress * equipmentProgress));
-            float assaultChance = powerRatio * 0.1f;
-            __result = defenderStrength == 0f || MBRandom.RandomFloat < assaultChance;
-        }
-
+        __result = readiness.ShouldStartAssault(__instance);
         return false;
     }
 }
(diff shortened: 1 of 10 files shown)


source/GameInterface/Services/SiegeEvents/AiSiegeTerminalPolicy.cs
```diff
@@ -0,0 +1,143 @@
...
+        this.logger = logger;
+    }
+
+    public AiSiegeTerminalDecision GetDecision(AiSiegeTerminalContext context)
+    {
+        // World state is the idempotency marker, so a save cannot retain a stale pending transition.
+        if (!context.IsFoodProblem
+            || context.IsPlayerLed
+            || !context.IsCurrentSiege
+            || context.HasActiveTransition)
```
ShoT-UPfps: what retries this once the map event ends? the food-problem postfix runs only once after the army has already dispersed, and an assault without a winner keeps the siege alive. returning `None` here can put the surviving leader back into the same rebuild loop; please defer and rerun the terminal decision when the transition clears.

source/GameInterface/Services/SiegeEvents/AiSiegeAssaultReadiness.cs
```diff
@@ -0,0 +1,168 @@
...
+
+public readonly struct AiSiegeAssaultReadinessResult
+{
+    public float AttackerStrength { get; }
+    public float DefenderStrength { get; }
+    public float PowerRatioBeforeEquipment { get; }
+    public float PowerRatioAfterEquipment { get; }
+    public float AssaultChance { get; }
+
+    public bool IsViable => DefenderStrength == 0f ||
```
ShoT-UPfps: this changes the old zero-strength behavior. with both sides at `0f`, the ratio is `NaN`; `IsViable` returns true here and `ShouldStartAssault` also falls through its `<= 1f` guard, so an empty besieger starts an assault where the previous `powerRatio > 1f` check returned false. please preserve that gate and cover the zero/zero case.

ShoT-UPfps (changes requested):
the starvation terminal still has two correctness gaps.

source/GameInterface/Services/SiegeEvents/AiSiegeTerminalPolicy.cs
```diff
@@ -0,0 +1,196 @@
...
+    AiSiegeTerminalDecision ResolveFoodProblem(AiSiegeTerminalTransitionState state);
+}
+
+/// <summary>Resolves an AI siege after vanilla disperses its starving army.</summary>
+internal class AiSiegeTerminalPolicy : IAiSiegeTerminalPolicy, IDisposable
+{
+    private readonly IAiSiegeAssaultReadiness readiness;
+    private readonly ILogger logger;
+    private readonly IMessageBroker messageBroker;
+    private readonly List<AiSiegeTerminalTransitionState> deferredTransitions = new();
```
ShoT-UPfps, line 82: how is this recovered after a save/restart? the army has already dispersed before this is added, but `deferredTransitions` is only in memory and the host's `SaveHandler` can autosave the active map event. after load there is no callback left to recreate the terminal decision.

source/GameInterface/Services/SiegeEvents/AiSiegeTerminalPolicy.cs
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        deferredTransitions.Add(state);
+    }
+
+    private void Handle_MapEventFinalized(MessagePayload<MapEventFinalized> _)
+    {
+        if (ModInformation.IsClient) return;
+
+        RetryDeferredTransitions(state => { ResolveFoodProblem(state); });
```
ShoT-UPfps: can this retry on a later game-thread turn? `MapEventFinalized` is published synchronously before the old event's destroy and encounter-close messages, so this can create the replacement assault first. the later old-event close clears the player's current `MapEventSide` by party id, which detaches them from the new assault.

ShoT-UPfps (changes requested):
the deferred retry still has two correctness gaps.

source/GameInterface/Services/SiegeEvents/AiSiegeTerminalPolicy.cs
garrettluskey: fixed in `d474d8e2a`, deferred leader/siege pairs are saved through the siege campaign behavior and restored into the server policy on load. stale pairs are discarded by the current-siege checks.

source/GameInterface/Services/SiegeEvents/AiSiegeTerminalPolicy.cs
garrettluskey: fixed in `4e379b415`, `MapEventFinalized` now queues the retry with `GameThread.EnqueueSafe`, so old-event destroy and encounter-close processing finish before a replacement assault is created.

ShoT-UPfps (commented):
rereviewed `4e379b41`. the four earlier correctness comments are fixed, but one teardown regression remains: when the co-op container is absent, the still-installed assault prefix skips vanilla and returns `false`, suppressing all AI assault rolls.

source/GameInterface/Services/BesiegerCamps/Patches/BesiegerCampAssaultPatches.cs
```diff
@@ -20,71 +16,13 @@ internal class BesiegerCampAssaultPatches
...
-        if (powerRatio > 1f)
+        if (!ContainerProvider.TryResolve<IAiSiegeAssaultReadiness>(out var readiness))
         {
-            int equipmentsBuilt = Campaign.Current.Models.CombatSimulationModel.GetNumberOfEquipmentsBuilt(settlement);
-            powerRatio *= (float)MathF.Min(3, equipmentsBuilt) / 3f;
-            float equipmentProgress = Campaign.Current.Models.CombatSimulationModel.GetMaximumSiegeEquipmentProgress(settlement) + (0.25f * (float)(5 - equipmentsBuilt));
-            powerRatio *= 1f - (0.85f * (equipmentProgress * equipmentProgress));
-            float assaultChance = powerRatio * 0.1f;
-            __result = defenderStrength == 0f || MBRandom.RandomFloat < assaultChance;
+            __result = false;
```
ShoT-UPfps: this disables the assault roll whenever the co-op container is unavailable. `UnpatchAll()` is intentionally a no-op, and teardown clears `ContainerProvider` while these Harmony patches stay installed, so later campaign ticks take this branch, skip vanilla, and make every AI siege assault decision return `false`. please keep a container-independent fallback to the same dedicated-host-safe readiness calculation (calling vanilla is not safe on the headless host because v1.4.8 dereferences `MobileParty.MainParty`) and add a no-container regression test.

garrettluskey: fixed in `186bf91df`, the prefix now falls back to the dedicated-host-safe readiness evaluator when the container is unavailable, with a no-container regression test.

ShoT-UPfps (approved):

PR #3193, opened 2026-08-20
[3079] Converge weapon drops across clients

### LINKED ISSUE #3079: [ISSUE] Fix weapon-drop state divergence at its source
# For New Contributors

## Description
Issue #3071 proved that a client can receive `NetworkWeaponDropped` for an agent whose corresponding equipment slot is already empty. PR #3074 prevents the native empty-slot access violation, but it intentionally ignores that message and does not explain or repair the underlying weapon/equipment and world-item state divergence.

Find and fix why a natural siege weapon drop can reach another client after that client has already emptied, or never populated, the same slot. The result must converge the equipment slot and network world-item registration across every client instead of only preventing the crash.

## Intended Design
First reproduce and identify the actual source of the mismatch. Do not assume whether the remote copy independently ran `Agent.DropItem`, began with different equipment, or received stale state.

One logical weapon drop should have one authoritative transition. Every participating client should end with the same equipment state and the same registered world-item id. Receiving an already-applied transition must be idempotent, while a genuine state mismatch must be repaired or trigger a bounded resync instead of being silently discarded.

The protocol should carry enough identity or revision information to distinguish an already-applied drop from an unrelated empty slot. Keep the managed empty-slot crash guard from #3074 as defense in depth.

## Location
- `source/Missions/Agents/Patches/AgentDropPatch.cs`
- `source/Missions/Agents/Handlers/WeaponDropHandler.cs`
- `source/Missions/Agents/Messages/WeaponDropped.cs`
- `source/Missions/Agents/Messages/NetworkWeaponDropped.cs`
- `source/Missions/Services/Network/MissionContext.cs`
- `source/Missions/Services/Network/LiteNetP2PClient.cs`
- Mission synchronization tests under `source/E2E.Tests/Services/Missions`

## Related Issues
- Follow-up to #3071
- Crash containment currently lives in #3074

## Requirements
- Reproduce the mismatch through a natural weapon-drop action in an active siege with one server and two clients. Do not directly inject an invalid `NetworkWeaponDropped` message as the reproduction.
- Capture the controller id, agent id, equipment slot contents, weapon identity, drop origin, and world-item registration on the sender and both receivers before and after the drop.
- Document the exact event sequence that makes a receiver's slot empty before `WeaponDropReceive` applies the message.
- Make one logical drop converge to an empty source slot and one shared network world-item id on all clients.
- Make an already-applied drop idempotent without creating a second world item or losing the existing world-item registration.
- Reconcile or explicitly resync a genuine equipment mismatch instead of silently returning with divergent state.
- Add automated coverage for the natural sender/receiver sequence, the already-applied case, and a mismatched receiver state.
- Complete a one-server, two-client siege test that proves equipment and world-item registry convergence on both clients.
- Do not leave issue-specific `#if DEBUG` fixture messages or commands in production source.

## Additional information
The #3071 client dump shows `EquipmentIndex` value `4` reaching `WeaponDropHandler.WeaponDropReceive` while the receiver's extra weapon slot is empty. `Agent.GetWeaponEntityFromEquipmentSlot` then raises `0xc0000005` before it can return null.

`MissionContext.ControllersInMission` excludes the local controller, so `IBattleNetwork.SendAll` is not a simple self-echo. The mismatch is between different mission copies. The current message carries only agent id, equipment index, and world-item id, which is not enough to prove whether an empty receiver already applied this specific drop.

The original report has no sender or server log, and #3074's live test deliberately exercised the recovered invalid receive state rather than naturally producing it. That evidence validates crash containment but does not identify the upstream divergence.

## Definition of Done
- [ ] The natural siege reproduction and root-cause event sequence are documented.
- [ ] All clients converge on the same equipment slot and world-item registration after one logical drop.
- [ ] Duplicate/already-applied and mismatched-state paths have automated tests.
- [ ] A one-server, two-client siege test passes without direct invalid-message injection.
- [ ] No issue-specific DEBUG fixture protocol remains in production source.

## Cross-issue crash verification scope

This issue is the authoritative implementation owner for the confirmed weapon-drop divergence family.

### Supporting and candidate reports

- **#2478, August 15 evidence:** confirmed `0xC0000005` null read in `Agent.GetWeaponEntityFromEquipmentSlot` from `WeaponDropHandler`. Treat this as supporting evidence for #3079, not a separate code owner.
- **#3041:** moderate-confidence candidate only. Its pre-PR-#3074 build contains the vulnerable path, but its artifacts do not include an exception or stack. Keep #3041 distinct until verification.

### Required expanded live verification

In addition to the existing one-server/two-client natural-drop requirements:

1. Run a three-player castle siege approximating #3041 when practical.
2. Record equipment-slot contents, weapon identity/revision, drop origin, and network world-item registration on sender and every receiver.
3. Exercise rejoin while the same battle remains active.
4. Capture the complete Bannerlord hard-crash directory for every failed client.
5. Verify that one logical drop converges to one empty source slot and one shared world-item registration on every mission copy.
6. Verify duplicate/already-applied delivery remains idempotent after rejoin.
7. Compare any #3041 crash stack with the confirmed weapon-drop signature.

Merging #3079 does not by itself close #3041. Close or consolidate #3041 only after an exact signature match or a current-build reproduction demonstrates that the crash/rejoin sequence is resolved. If #3041 produces a different signature, retain it as a separate issue.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Natural weapon drops during active battles can leave clients disagreeing about whether the equipment slot is empty and which dropped world item represents the weapon. A later drop update or rejoin can then create a duplicate, lose the shared item identity, or hit the empty-slot crash path.

## What is the new behavior?

Resolves #3079

Natural weapon drops now converge each client on one empty source slot and one shared dropped world item, including duplicate delivery, equipment mismatches, pickup ordering, and active-battle rejoin. Existing matching drops are reused, missing state is repaired, and consumed drops are not replayed, preventing the confirmed empty-slot weapon-drop crash while preserving the guard.

## Bot Changelog Entry

- Fixed weapon drops sometimes duplicating, disappearing, or crashing a client during battles and rejoin.


source/Missions/NetworkAgentRegistry.cs
```diff
@@ -425,6 +425,7 @@ public bool TryTransferAuthority(string controllerId, Guid agentId)
             }
 
             agentInfo.CurrentAuthority = controllerId;
+            agentInfo.AuthorityRevision++;
```
garrettluskey: this is only a local transfer count, and the agent spawn/catch-up data doesnt carry it. a client that rejoins after earlier migrations registers the current authority at revision 0; if that client later becomes authority, its `NetworkWeaponPickupSlotState` has a lower revision and gets rejected by the existing peers in `WeaponPickupHandler.cs:806`. carry the revision in the synchronized spawn/migration state instead of incrementing an unsynchronized counter.

garrettluskey (dismissed):
the drop/pickup state machine covers a lot of ugly ordering cases and the new tests are thorough. the new slot-state revision is not stable across rejoin plus authority migration, so that needs to be synchronized before this merges.

source/Missions/Tournaments/TournamentManifestAuthority.cs
```diff
@@ -72,5 +73,9 @@ private static TournamentAgentSpawnData CopyWithOwner(TournamentAgentSpawnData d
             data.MountCharacterId,
             data.MountDescriptorSeed,
             data.MountEquipment,
-            data.MountHealth);
+            data.MountHealth,
+            data.AuthorityRevision + 1,
```
garrettluskey, line 77: this increments whichever manifest copy the receiver has, but the server keeps the originally submitted manifest. after two migrations a rejoin normalizes revision 0 to 1 while existing peers are at 2, so its slot-state response is rejected at `WeaponPickupHandler.cs:806` after it is promoted. persist the canonical authority generation and cover rejoin from the original manifest after multiple migrations.

source/Missions/Battles/CoopBattleController.cs
```diff
@@ -429,6 +431,8 @@ void ReplayJoinState()
             siegeEngineDeployment.CatchUpJoiner(controllerId);
             siegeMachineState.CatchUpJoiner(controllerId);
             Deployment.CatchUpJoiner(controllerId);
+            if (Session.IsLocalHost)
+                coopMissionComponent.WeaponDropHandler.CatchUpJoiner(controllerId);
```
garrettluskey, line 435: this one-shot catch-up isnt ordered against live drops. an owner can send a drop before learning about the joiner while the host handles `NetworkMissionPeerEntered` first, so both sends omit that item. reconcile after the join or have the host forward canonical drops after applying them.

source/Missions/Agents/Handlers/WeaponDropHandler.cs
```diff
@@ -49,57 +196,1778 @@ public WeaponDropHandler(
...
+        }
+
+        var observed = new ObservedDrop(
+            dropped.Agent,
+            dropped.EquipmentIndex,
+            dropped.DroppedWeapon,
+            dropped.DroppedItem);
+        queue.Enqueue(observed);
+        if (observed.Item != null)
+            messageBroker.Publish(this, new WorldItemIdentityPending(observed.Item));
```
garrettluskey, line 682: this pending marker is only consumed by `WeaponPickupHandler`; `TrySerializeRuntimeWorldItems` still gives the item another id through `GetOrCreateId`. a join in this window can replay that id and then materialize the owner's canonical drop as a second item. skip pending observations in tournament replay or wait for their canonical identity.

source/Missions/Taverns/CoopLocationsController.cs
```diff
@@ -75,6 +76,8 @@ public CoopLocationsController(
         // binding map are SHARED state, so the components are constructed here around single
         // instances instead of DI-resolving them (transient injection would give each its own).
         session = new LocationSession(controllerIdProvider, hostRegistry);
+        coopMissionComponent.WeaponDropHandler.ConfigureLocalHostProvider(
```
garrettluskey, line 80: the location host is now the resync source, but `SendJoinInfo` only replays NPCs and never calls `WeaponDropHandler.CatchUpJoiner`. a player entering after a location drop never receives that world item, so send drop catch-up after `npcReplicator.ReplicateCurrentAgentsTo`.

garrettluskey (dismissed):
the drop/pickup state machine covers the core ordering cases and the revision plumbing fixes battle and location migration. there are still gaps around join ordering and tournament identity/revision replay, so these need to be fixed before merge.

source/Missions/NetworkAgentRegistry.cs
ShoT-UPfps: fixed, catch-up manifests now preserve the current per-agent revision and battle migration advances from that revision.

source/Missions/Tournaments/TournamentManifestAuthority.cs
ShoT-UPfps: fixed, the session registry now persists the normalized manifest after each host change, so rejoin receives the current authority revisions after multiple migrations.

source/Missions/Battles/CoopBattleController.cs
ShoT-UPfps: fixed, the host now forwards every accepted canonical drop, so a join between the owner send and host catch-up still receives it.

source/Missions/Agents/Handlers/WeaponDropHandler.cs
ShoT-UPfps: fixed, tournament replay now skips a runtime item while its canonical identity is still pending in `WeaponDropHandler`.

source/Missions/Taverns/CoopLocationsController.cs
ShoT-UPfps: fixed, location join sends the live agent equipment and then calls `WeaponDropHandler.CatchUpJoiner` for the entering controller.

source/Missions/Agents/Handlers/WeaponPickupHandler.cs
```diff
@@ -19,177 +19,1150 @@ namespace Missions.Agents.Handlers
...
+                pendingWorldItemIdentities.Remove(resolved.WorldItem);
+                if (pendingIdentityPickups.TryGetValue(
+                        resolved.WorldItem,
+                        out Queue<PendingIdentityPickup> pending))
+                {
+                    pendingIdentityPickups.Remove(resolved.WorldItem);
+                    while (pending.Count > 0)
+                    {
+                        PendingIdentityPickup pendingPickup = pending.Dequeue();
+                        SendWeaponPickup(
```
garrettluskey, line 484: this sends the deferred pickup to the other mission peers but never records its `PickupId` in this peer's `WorldItemTransitionState`. if this peer is the host or becomes host later, `TrySendWorldItemStateResponse` cannot acknowledge that pickup during resync, so the requester keeps retrying; record the resolved pickup locally before broadcasting it.

garrettluskey (dismissed):
the catch-up forwarding and authority revision fixes cover the earlier ordering gaps. one deferred-pickup path still loses the pickup receipt on the originating peer.

source/Missions/Agents/Handlers/WeaponPickupHandler.cs
ShoT-UPfps: fixed, the delayed identity path now records the generated pickup id locally before broadcasting it.

garrettluskey (approved):
the deferred identity path now records the pickup receipt before broadcasting, and the rejoin and migration ordering cases are covered. looks good.

source/Missions/NetworkAgentRegistry.cs
ShoT-UPfps: still broken for host-only adoption: existing observers can remain D/rev0, the host advances to H/rev1, and a later joiner receives H/rev1. if the host leaves, the old observers transfer D to S at rev1 while the joiner transfers H to S at rev2, so that joiner rejects the successor rev1 slot repair.

source/Missions/Battles/CoopBattleController.cs
ShoT-UPfps: drops are forwarded now, but accepted pickups still have the same join race. the picker can omit the joiner and the host can send catch-up before receiving the pickup; the host applies it locally but never forwards it, so the joiner keeps the pre-pickup world item and misses the slot transition.

source/Missions/Agents/Handlers/WeaponDropHandler.cs
ShoT-UPfps: skipping the pending item makes the runtime snapshot incomplete. a peer that already has the canonical item can delete it during `ReconcileRuntimeWorldItems`; when the original drop is forwarded later, its duplicate `DropId` is ignored, so the item stays missing.

source/Missions/Agents/Handlers/WeaponDropHandler.cs
```diff
@@ -49,57 +199,1878 @@ public WeaponDropHandler(
...
+            message.ItemObjectId,
+            message.WorldItemId);
+        ForwardAcceptedDrop(message);
+    }
+
+    private void ForwardAcceptedDrop(NetworkWeaponDropped message)
+    {
+        if (!isLocalHost()) return;
+
+        network.SendAll(message);
```
ShoT-UPfps, line 861: `NetworkWeaponDropped` needs the agent authority revision and the receive path needs to reject an old authority generation before changing the slot. otherwise a delayed old-host drop can overwrite the successor equipment and get forwarded again here.

PR #3172, opened 2026-08-19
Allow stuck family members to be freed

jordanbrymora:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] All new classes have class-level documentation comments, if there are any at all
- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bug fix (non-breaking change that fixes an issue)

## What is the new behavior?

Resolves #3160

Not a root cause fix but should be sufficient enough to free stuck heroes on existing and new saves in a time frame that isn't noticable. This just uses the existing fix for stuck companions.

## Bot Changelog Entry

Fixed family members permanently becoming stuck as prisoners.

PR #3210, opened 2026-08-21
Repair all stuck prisoner heroes

gamingmenacers-svg:

The stale prisoner repair added in #3172 stops after the first match, so additional stuck heroes must wait another campaign day.

## What is the new behavior?

Every prisoner hero with no captor party is repaired in the same daily tick. Prisoners with a valid captor and non-prisoner heroes are left unchanged.

## How to test

`dotnet test source\GameInterface.Tests\GameInterface.Tests.csproj -c Release --no-restore`

1,060 passed, 11 skipped, 0 failed.


jordanbrymora (approved):
Looks good

PR #3211, opened 2026-08-21
More town NPCs

### LINKED ISSUE #2736: [ISSUE] Enable tavern employees and other town NPCs
# For New Contributors

## Location

## Definition of Done

jordanbrymora:
## PR Checklist

## PR Type

`TavernEmployeesCampaignBehavior` is disabled. This manages:
- Talking to tavern keepers for nearby issues and asking about nearby recruitable companions.
- Talking to tavern maids to buy a tun for a small recent events morale boost for your party.
- Talking to ransom brokers in person.

`GuardsCampaignBehavior` is disabled. This spawns guards around settlement locations as well as controlling dialogue options to bribe guards for access to certain locations.

`SettlementMusiciansCampaignBehavior` is disabled. This sets a playlist for musicians in towns when loading in.

## What is the new behavior?

Resolves #2736 

Enable `TavernEmployeesCampaignBehavior` and save relevant data in `CoopSession` to persist across sessions.

Enable `GuardsCampaignBehavior` and re-work `Settlement.BribePaid` to be client specific and saved in `CoopSession`. This also fixes an existing problem where roguery xp was not rewarded for bribing to get into a keep from the settlement menu.

Enable `SettlementMusiciansCampaignBehavior`. The played music isn't synced between clients but neither are the instruments musicians held.

## Bot Changelog Entry

More NPCs will now spawn in towns including tavern keepers, ransom brokers and guards.


source/GameInterface/Services/MobileParties/Interfaces/SessionInteractionsPlayerDataInterface.cs
```diff
@@ -130,6 +134,55 @@ public void AddSettlementSneakedIn(string playerHeroId, string settlementId)
...
+    }
+
+    public bool DailyTickDrinkThisDayInSettlement()
+    {
+        bool hasUpdated = false;
+        foreach (var playerData in InteractionsPlayerData.PlayerOrderedDrinkThisDayInSettlement)
+        {
+            if (playerData.Value != null)
+            {
+                InteractionsPlayerData.PlayerOrderedDrinkThisDayInSettlement[playerData.Key] = null;
```
gamingmenacers-svg: this mutates `PlayerOrderedDrinkThisDayInSettlement` while its enumerator is active. Bannerlord runs on CLR v4, so the next `MoveNext` throws `InvalidOperationException`; iterate a key snapshot here and in `WeeklyTickHasBoughtToTunToParty` otherwise the first non-default reset fails.

source/GameInterface/Services/Inventory/TradeSkills/Handlers/TradeDataInitializationHandler.cs
```diff
@@ -133,4 +135,17 @@ private Dictionary<Settlement, CampaignTime> GetEnteredSettlements(string player
 
         return enteredSettlements;
     }
+
+    private void LoadSettlementBribePaidData(string playerHeroId)
+    {
+        // Null and key check for players without existing bribe paid data
+        if (tradePlayerData?.PlayerSettlementBribePaid?.ContainsKey(playerHeroId) != true) return;
```
gamingmenacers-svg, line 148: `LoadSettlementBribePaidData` only writes settlements present for the new hero. every client first loads the same save and hero switches reuse these `Settlement` instances, so any value outside this dictionary carries into the player and can satisfy the keep/dungeon access checks; clear every `Settlement.BribePaid` before applying this player's dictionary.

source/GameInterface/Services/Actions/Handlers/BribeGuardsHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+        var data = obj.What;
+
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Hero>(data.MainHeroId, out var mainHero)) return;
+            if (!objectManager.TryGetObjectWithLogging<Settlement>(data.SettlementId, out var settlement)) return;
+
+            GiveGoldAction.ApplyBetweenCharacters(mainHero, null, data.Gold, false);
+
+            sessionTradePlayerDataInterface.UpdateSettlementBribePaid(data.MainHeroId, data.SettlementId, data.Gold);
```
gamingmenacers-svg, line 78: `GiveGoldAction` clamps the transfer to `mainHero.Gold`, but this records the requested amount and the XP roll below also uses it. if server gold changed after the client check, a partial payment gets full bribe credit and XP; validate the current balance before applying this request.

source/GameInterface/Services/Towns/Handlers/TavernEmployeesHandler.cs
```diff
@@ -0,0 +1,240 @@
...
+    private void Handle_NetworkPlayerBuysTun(MessagePayload<NetworkPlayerBuysTun> obj)
+    {
+        var data = obj.What;
+
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Hero>(data.MainHeroId, out var mainHero)) return;
+
+            GiveGoldAction.ApplyBetweenCharacters(mainHero, null, data.TunPrice, false);
+            if (mainHero.PartyBelongedTo != null) mainHero.PartyBelongedTo.RecentEventsMorale += 2f;
```
gamingmenacers-svg, line 229: `GiveGoldAction` clamps the charge to the hero's current gold, then the tun morale and session flag are always granted. validate the server balance first otherwise a stale client balance buys the tun for less.

gamingmenacers-svg (changes requested):
found 4 issues that need fixing before this goes in.

source/GameInterface/Services/MobileParties/Interfaces/SessionInteractionsPlayerDataInterface.cs
jordanbrymora: Changed to use key snapshot

source/GameInterface/Services/Inventory/TradeSkills/Handlers/TradeDataInitializationHandler.cs
jordanbrymora: Changed to clear existing `Settlement.BribePaid` values for all settlements first.

source/GameInterface/Services/Actions/Handlers/BribeGuardsHandler.cs
jordanbrymora: Added guard

source/GameInterface/Services/Towns/Handlers/TavernEmployeesHandler.cs
jordanbrymora: Added guard

source/GameInterface/Services/Actions/Handlers/BribeGuardsHandler.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Hero>(data.MainHeroId, out var mainHero)) return;
+            if (!objectManager.TryGetObjectWithLogging<Settlement>(data.SettlementId, out var settlement)) return;
+
+            if (mainHero.Gold - data.Gold < 0)
+            {
+                Logger.Error($"Rejecting player bribe due to a lack of gold. Requested change: {data.Gold} Player gold: {mainHero.Gold}");
+                return;
```
gamingmenacers-svg, line 73: [P1] this rejection happens after the client already applied `settlement.BribePaid += gold`. if an earlier spend has not synced yet, the client keeps the bribe access without paying, so apply it only after acceptance or send a correction on rejection.

source/GameInterface/Services/Towns/Handlers/TavernEmployeesHandler.cs
```diff
@@ -0,0 +1,246 @@
...
+        var data = obj.What;
+
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Hero>(data.MainHeroId, out var mainHero)) return;
+
+            if (mainHero.Gold - data.TunPrice < 0)
+            {
+                Logger.Error($"Rejecting player bought tun due to a lack of gold. Requested change: {data.TunPrice} Player gold: {mainHero.Gold}");
+                return;
```
gamingmenacers-svg, line 225: [P1] this rejection leaves `_hasBoughtTunToParty = true` from the client prefix. the player gets no morale and cannot retry, and the weekly reset may not broadcast because server state stayed false, so commit the flag only after acceptance or correct it on rejection.

source/GameInterface/Services/Towns/Handlers/TavernEmployeesHandler.cs
```diff
@@ -0,0 +1,246 @@
...
+
+    private void Handle_NetworkTavernKeeperFindCompanion(MessagePayload<NetworkTavernKeeperFindCompanion> obj)
+    {
+        var data = obj.What;
+
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Hero>(data.MainHeroId, out var mainHero)) return;
+
+            GiveGoldAction.ApplyBetweenCharacters(mainHero, null, 2, false);
```
gamingmenacers-svg, line 285: [P1] the client reveals and records each companion before this charge reaches the server, while its local gold is not deducted. choosing “someone else” again before gold sync lets later requests reveal information while `GiveGoldAction` only charges the remaining balance, so authorize the charge before running the client consequence.

gamingmenacers-svg (changes requested):
the dictionary iteration and load-time bribe reset are fixed. the payment paths still commit client effects before the server accepts them, and companion lookup has the same ordering problem.

source/GameInterface/Services/Towns/Handlers/TavernEmployeesHandler.cs
```diff
@@ -0,0 +1,246 @@
...
+
+    private void Handle_NetworkPlayerAcceptsClanInfoOffer(MessagePayload<NetworkPlayerAcceptsClanInfoOffer> obj)
+    {
+        var data = obj.What;
+
+        GameThread.RunSafe(() =>
+        {
+            if (!objectManager.TryGetObjectWithLogging<Hero>(data.MainHeroId, out var mainHero)) return;
+
+            GiveGoldAction.ApplyBetweenCharacters(mainHero, null, 500, false);
```
garrettluskey: [P1] this payment path has the same stale-balance problem as the other tavern purchases. vanilla already charged 500 on the client because the prefix returns `true`, while this server call clamps the charge to whatever gold remains; accept the purchase before committing the client charge, or correct the client when the server rejects it.

garrettluskey (changes requested):
the persistence changes and the first two review fixes look right. requesting changes because the three existing P1 payment-ordering threads are still current, and clan-info payment also needs server-side acceptance before its client consequence is committed.

source/GameInterface/Services/Towns/Handlers/TavernEmployeesHandler.cs
jordanbrymora: The results of `FindCompanionWithType` are needed immediately for the dialogue to continue properly and sending the authorised result from the server destroys this chain, leaving a tavern keeper never able to find companions.

This isn't really feasible afaik without forcing a dialogue refresh which could cause a variety of other problems.

source/GameInterface/Services/Towns/Handlers/TavernEmployeesHandler.cs
jordanbrymora: Fixed

source/GameInterface/Services/Actions/Patches/BribeGuardsActionPatches.cs
```diff
@@ -0,0 +1,29 @@
...
+    [HarmonyPrefix]
+    public static bool ApplyInternalPrefix(Settlement settlement, int gold)
+    {
+        // Block on server, only clients should be starting a player bribe
+        if (ModInformation.IsServer) return false;
+
+        // Don't update with no gold change like in vanilla
+        if (gold <= 0) return false;
+
+        var message = new PlayerBribesGuard(Hero.MainHero, settlement, gold);
```
garrettluskey, line 24: `conversation_prison_guard_visit_permission_bribe_on_consequence` opens the prison immediately after this returns, and `castle_guard_let_player_in` queues `OpenLordsHallMission` before the server response. a stale-balance rejection still grants this visit, so gate the location transition on the acceptance response too.

garrettluskey (changes requested):
the new response messages correctly defer the bribe, tun, and clan-info state until the server accepts them. guard location access still happens before that response, and the existing companion payment thread is still unresolved.

gamingmenacers-svg (dismissed):
the response messages fixed the replicated state ordering, but c174207 still grants the client-side guard/location consequence before the server accepts the bribe, and FindCompanionWithType still reveals the result before its asynchronous two-gold charge with no acceptance response. rejection or packet loss leaves the client past a purchase the server never committed. move both consequences behind server acknowledgements and cover rejected and delayed replies.

source/GameInterface/Services/Actions/Patches/BribeGuardsActionPatches.cs
jordanbrymora: Not needed

source/GameInterface/Services/Actions/Patches/BribeGuardsActionPatches.cs
```diff
@@ -0,0 +1,29 @@
...
+        MessageBroker.Instance.Publish(null, message);
```
garrettluskey, line 25: [P1] `BribeGuardsAction.Apply` returns to vanilla after this publish. `conversation_prison_guard_visit_permission_bribe_on_consequence` then opens the prison, and `castle_guard_let_player_in` queues `OpenLordsHallMission`, so a stale-balance rejection still grants access. move the location transition behind `NetworkPlayerBribesGuardClient`.

source/GameInterface/Services/Towns/Patches/TavernEmployeesCampaignBehaviorPatches.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    [HarmonyPatch(nameof(TavernEmployeesCampaignBehavior.FindCompanionWithType))]
+    [HarmonyPatch(new Type[] { typeof(TavernEmployeesCampaignBehavior.TavernInquiryCompanionType) })]
+    [HarmonyPostfix]
+    public static void FindCompanionWithTypePostfix(TavernEmployeesCampaignBehavior __instance)
+    {
+        var message = new TavernKeeperFindCompanion(Hero.MainHero);
+        MessageBroker.Instance.Publish(__instance, message);
```
garrettluskey, line 95: [P1] `FindCompanionWithType` has already selected and revealed the companion before this postfix runs, while client `GiveGoldAction` is blocked. the server handler has no balance check and `GiveGoldAction` clamps to the remaining gold, so stale gold or repeated "someone else" choices keep results that were not fully paid for. authorize the charge before committing the lookup result.

garrettluskey (changes requested):
the new responses fix the tun and clan-info ordering. guard access and companion lookup still commit their client consequence before the server has accepted the charge.

garrettluskey (approved):

PR #3213, opened 2026-08-21
Re-enable Tab in tournaments

### LINKED ISSUE #2058: [Tournaments] Reenable tab behavior
## Severity of issue/bug

- [ ] High
- [ ] Average
- [x] Low

Currently tournaments suppress tab for their initial implementation. This ticket is meant to renable it to work properly as it does in vanilla

This includes -
When a round is complete, press tab to immediately advance.
While spectating, pressing tab should open a menu to allow the spectator to leave the tournament.
I think thats it?

G07cha:
## PR Checklist

## PR Type

Tab button that brings up tournament overview is disabled, preventing spectators from leaving until the match is over.

## What is the new behavior?

Tab now casts the advance vote on a finished round (after the initial post-win delay) and toggles the spectator leave menu during the live match. Spectator can leave at any time while participant is only able to do so upon the match completion.

Resolves #2058 

## Bot Changelog Entry

- Re-enable Tab button in tournaments

### DIFF
--- a/source/GameInterface/Services/Tournaments/UI/CoopTournamentVM.cs
+++ b/source/GameInterface/Services/Tournaments/UI/CoopTournamentVM.cs
@@ -27,6 +27,7 @@ internal sealed class CoopTournamentVM : TournamentVM
         public readonly bool CanLeave;
         public readonly bool CanBet;
         public readonly bool IsMatchActive;
+        public readonly bool IsInCurrentMatch;
         public readonly int ReadyCount;
         public readonly int SkipCount;
         public readonly int VoterCount;
@@ -39,6 +40,7 @@ internal sealed class CoopTournamentVM : TournamentVM
             bool canLeave,
             bool canBet,
             bool isMatchActive,
+            bool isInCurrentMatch,
             int readyCount,
             int skipCount,
             int voterCount,
@@ -50,6 +52,7 @@ internal sealed class CoopTournamentVM : TournamentVM
             CanLeave = canLeave;
             CanBet = canBet;
             IsMatchActive = isMatchActive;
+            IsInCurrentMatch = isInCurrentMatch;
             ReadyCount = readyCount;
             SkipCount = skipCount;
             VoterCount = voterCount;
@@ -89,6 +92,7 @@ internal sealed class CoopTournamentVM : TournamentVM
     private bool canBet;
     private bool isJoinVisible;
     private bool isCoopMatchActive;
+    private bool leaveMenuOpen;
     private string readyCountText;
     private string skipCountText;
     private string selectedChoiceText;
@@ -96,6 +100,7 @@ internal sealed class CoopTournamentVM : TournamentVM
     private bool hasAcceptedBetResult;
     private long acceptedBetSequence;
     private BetSummary acceptedBetSummary;
+    private UIState lastUIState;
 
     public CoopTournamentVM(
         Action disableUI,
@@ -184,11 +189,24 @@ internal sealed class CoopTournamentVM : TournamentVM
         {
             if (isCoopMatchActive == value) return;
             isCoopMatchActive = value;
-            base.IsCurrentMatchActive = value;
+            base.IsCurrentMatchActive = value && !leaveMenuOpen;
             OnPropertyChangedWithValue(value, nameof(IsCoopMatchActive));
             OnPropertyChanged(nameof(ShouldShowUI));
         }
     }
+
+    internal bool IsLeaveMenuOpen
+    {
+        get => leaveMenuOpen;
+        set
+        {
+            if (leaveMenuOpen == value) return;
+            leaveMenuOpen = value;
+            base.IsCurrentMatchActive = isCoopMatchActive && !leaveMenuOpen;
+        }
+    }
+
+    public bool IsInCurrentMatch => lastUIState.IsInCurrentMatch;
 
     [DataSourceProperty]
     public bool ShouldShowUI => TournamentMissionPresentationState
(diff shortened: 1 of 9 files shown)


ShoT-UPfps:
just verifying- did you test this ingame? @G07cha 
changes lgtm tho

G07cha:
> just verifying- did you test this ingame? @G07cha changes lgtm tho

Yes, checked the following scenarios:
- spectator can open overview by pressing tab at any point of the match regardless if participating player is dead or alive
- spectator can close overview by pressing tab again at any point except for rounds end (had to add generic key events listeners to be able to do that)
- nothing happens for participating player on hitting tab when the round is still going 
- pressing tab in overview automatically advances participating player to the next round

Let me know if you can think of any other scenarios worth checking.

ShoT-UPfps (dismissed):
lgtm

zzzzzzzzott:
Some notes - 

1) **Marking this PR as changes requested for this note** - For spectators, holding tab in an ongoing round should behave the same as if the player was walking around the town. So it would bring up that progress bar to show they are about to leave the scene and go back to the campaign map. That would be the most vanilla way to handle it.

2) I initially disabled tab behavior bcs I found during testing that there would be some weird race conditions. Nothing that couldn't be solved, I just didn't feel like doing it lol. But I mention this because I want to make sure you tested various scenarios, I don't want to introduce any new bugs with this.

3) I noticed you said "Tab now casts the advance vote on a finished round (after the initial post-win delay) ". Does this mean that the tournament will **not** continue until all players press tab? Not sure im understanding this correctly.

zzzzzzzzott (dismissed):
cmt above

source/GameInterface/Services/Tournaments/UI/CoopTournamentVM.cs
```diff
@@ -622,12 +640,28 @@ internal static UIState CalculateUIState(
...
+    internal static TournamentPlayerChoice? GetAdvanceChoice(UIState state)
+    {
+        if (state.CanJoin) return TournamentPlayerChoice.Join;
+        if (state.CanSkip) return TournamentPlayerChoice.Skip;
+        if (state.CanWatch) return TournamentPlayerChoice.Watch;
+        return null;
+    }
+
+    internal static bool ShouldOpenLeaveMenu(UIState state)
+        => state.IsMatchActive && !state.IsInCurrentMatch && state.CanLeave;
```
garrettluskey: this also returns true for a human contestant who isnt in the current match. `TryLeaveActive` replaces that contestant immediately, so a resting participant can leave during another live match even though participants should wait for tournament completion. track whether the local controller is a pure spectator here instead of using `!IsInCurrentMatch`.

garrettluskey (changes requested):
the tab routing is much cleaner than the global input suppression, and the advance-choice paths are covered. one live-match leave predicate still includes resting contestants, so this needs one correction.

source/GameInterface/Services/Tournaments/UI/CoopTournamentVM.cs
G07cha: During my testing, a contestant player waiting for the next round was able to open an overview by pressing Tab button with "Leave" button enabled. They cannot skip the round as there are players participating in it of course.

G07cha:
@zzzzzzzzott thanks for taking the time to review my changes!

>  For spectators, holding tab in an ongoing round should behave the same as if the player was walking around the town. So it would bring up that progress bar to show they are about to leave the scene and go back to the campaign map. That would be the most vanilla way to handle it.

Checking vanilla Bannerlord I was presented with a confirmation window (screenshot below) upon pressing Tab, granted, I wasn't controlling the player's character but rather spectating AIs fighting directly but I wonder how tricky that would be with the existing vanilla implementation.

image

One more thing I've found during testing, you can actually forfeit the tournament by pressing Tab while being far away from other contestants and confirming your decision. Right now the player contestants cannot leave the tournament until the round is over, is it something we want to implement in the coop version as well?

image

> I initially disabled tab behavior bcs I found during testing that there would be some weird race conditions. Nothing that couldn't be solved, I just didn't feel like doing it lol. But I mention this because I want to make sure you tested various scenarios, I don't want to introduce any new bugs with this.

Realized that I haven't checked with two player participants yet, validated both pressing Tabs at various points of the match (in the match, at the end, disqualified), seems to be functioning as expected.

>I noticed you said "Tab now casts the advance vote on a finished round (after the initial post-win delay) ". Does this mean that the tournament will not continue until all players press tab? Not sure im understanding this correctly.

The Tab acts as a keybind for clicking the specific button on the overview screen based on the player's state - Join -> Skip -> Watch (watch is never triggered afaik). As round is over, all joined players will either have to press one of the respective buttons on the screen or hit the Tab keyboard button to proceed to the next round.

source/GameInterface/Services/Tournaments/UI/CoopTournamentVM.cs
garrettluskey: this still treats a resting or eliminated contestant as a spectator. those contestants have `CanLeave` but arent in `CurrentMatch`, so Tab opens `ExecuteLeave` and `TryLeaveActive` replaces them during another live fight. key this off `localIsSpectator` instead.

garrettluskey (changes requested):
the focused Tab routing is cleaner than the process-wide input patches, and the advance-choice paths are covered. changes requested because the live-match leave predicate still lets resting contestants leave before tournament completion.

source/GameInterface/Services/Tournaments/UI/CoopTournamentVM.cs
G07cha: Misunderstood your comment initially, updated to prevent resting contestant from opening leave UI ~~while still allowing eliminated ones to open it.~~ Eliminated contestants leaving seems to cause some issues, blocked leaving the arena for them as well.

garrettluskey (approved):
the role split now keeps resting and eliminated contestants out of the live-match leave path, and the fight controller blocks vanilla Tab handling until completion. looks good.

G07cha:
@zzzzzzzzott @ShoT-UPfps any objections to merge from your side?

gamingmenacers-svg (approved):
reviewed b7c4dcf through focused Leave-key routing, photo mode, advance choices, role calculation, spectator-only live-match leave, and native mission-exit blocking. the latest role split fixes the prior resting-contestant hole and the tests exercise the production view/VM seam. no actionable delta finding; rebase and rerun after the 185-commit development drift.

ShoT-UPfps (approved):
lgtm

PR #3216, opened 2026-08-21
Save details of the last connection

### LINKED ISSUE #2402: [LOBBY] Save Join co-op sandbox input so the IP is not required every time
# For New Contributors

Not provided

## Description

The Join co-op sandbox flow should persist user-entered connection input so players do not need to type the IP address each time they join.

## Intended Design

Not provided

## Location

Not provided

## Related Issues

Not provided

## Requirements

Not provided

## Additional information

Not provided

## Definition of Done

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1530933262720172052/1530933262720172052

## Consolidated duplicate evidence

This is the authoritative tracker for **persisting the last direct-connect server address between sessions**.

- #2892 independently requests that the previously entered server IP be retained and pre-filled on the next connection attempt.
- Expected behavior: store the last successfully used direct-connect address locally and restore it when the join screen is reopened, while leaving it editable.

Duinhil:
## PR Checklist

## PR Type

The join screen has no memory of previous sessions. Every time a player wants to reconnect they must re-enter the server address, port, and password manually. The Steam Lobbies tab also triggers a full lobby search immediately on selection, even when the player only wants to reconnect.

## What is the new behavior?
Resolves #2402
After joining a direct-connection server or a Steam lobby, the details are saved to the player's local coop options file. A new "Last Connection" tab appears in the join screen showing the last direct server and last Steam lobby with a one-click Reconnect button for each. Reconnecting this way never triggers a lobby search.

## Bot Changelog Entry
- Added a "Last Connection" tab to the join screen that saves and displays the last direct connection and Steam lobby you joined, allowing one-click reconnect without re-entering details.

## Other information
Saved connection data is stored in Documents\Mount and Blade II Bannerlord\Configs\BannerlordCoop\coop_options.json under the Connection.LastSession key. If the file is absent or corrupt the tab starts empty with no error. This means passwords are stored in plain text but only locally to the Client. Also, I have been unable to test the Steam connection as I could see no way to do so without a second Steam account setup but I believe it's logically sound. Happy to do more testing if there is a preferred method to do so.

### DIFF
--- a/UIMovies/CoopConnectionUIMovie.xml
+++ b/UIMovies/CoopConnectionUIMovie.xml
@@ -229,6 +229,7 @@
                       </Children>
                     </ListPanel>
                     <!-- End of Filter Buttons -->
+
                     <TextWidget WidthSizePolicy="StretchToParent" HeightSizePolicy="Fixed" SuggestedHeight="52"
                                 IsVisible="@IsSteamLobbyStatusVisible" Brush="Clan.TabControl.Text"
                                 Brush.FontSize="24" Brush.FontColor="#B8B8B8FF"
@@ -362,6 +363,50 @@
                                          IsDisabled="@IsNextSteamLobbyPageDisabled"
                                          Command.Click="ActionNextSteamLobbyPage"
                                          Parameter.Text="@NextPageButtonText"/>
+                      </Children>
+                    </ListPanel>
+                  </Children>
+                </ListPanel>
+
+                <!-- Last Connection Tab -->
+                <ListPanel WidthSizePolicy="StretchToParent" HeightSizePolicy="StretchToParent"
+                           IsVisible="@IsLastConnectionTabVisible"
+                           StackLayout.LayoutMethod="VerticalTopToBottom">
+                  <Children>
+                    <!-- Empty state -->
+                    <TextWidget IsVisible="@HasNoLastConnection"
+                                WidthSizePolicy="StretchToParent" HeightSizePolicy="Fixed" SuggestedHeight="60"
+                                Brush="Clan.TabControl.Text" Brush.FontSize="22" Brush.FontColor="#B8B8B8FF"
+                                Brush.TextHorizontalAlignment="Left" Brush.TextVerticalAlignment="Center"
+                                Text="No previous connections saved. Join a server to save it here."/>
+
+                    <!-- Last direct connection row -->
+                    <ListPanel IsVisible="@HasLastDirectConnection"
+                               WidthSizePolicy="StretchToParent" HeightSizePolicy="Fixed" SuggestedHeight="58"
+                               MarginBottom="12" StackLayout.LayoutMethod="HorizontalLeftToRight">
+                      <Children>
+                        <TextWidget WidthSizePolicy="StretchToParent" HeightSizePolicy="StretchToParent"
+                                    Brush="Clan.TabControl.Text" Brush.FontSize="22" Brush.FontColor="#B8B8B8FF"
+                                    Brush.TextHorizontalAlignment="Left" Brush.TextVerticalAlignment="Center"
+                                    Text="@LastDirectConnectionText"/>
+                        <Standard.Button WidthSizePolicy="Fixed" SuggestedWidth="160" HeightSizePolicy="Fixed" SuggestedHeight="58"
+                                         VerticalAlignment="Center"
+                                         Command.Click="ActionConnect" Parameter.Text="@ReconnectButtonText"/>
+                      </Children>
+                    </ListPanel>
+
+                    <!-- Last Steam lobby row -->
+                    <ListPanel IsVisible="@HasLastSteamLobby"
+                               WidthSizePolicy="StretchToParent" HeightSizePolicy="Fixed" SuggestedHeight="58"
+                               StackLayout.LayoutMethod="HorizontalLeftToRight">
+                      <Children>
+                        <TextWidget WidthSizePolicy="StretchToParent" HeightSizePolicy="StretchToParent"
+                                    Brush="Clan.TabControl.Text" Brush.FontSize="22" Brush.FontColor="#B8B8B8FF"
+                                    Brush.TextHorizontalAlignment="Left" Brush.TextVerticalAlignment="Center"
+                                    Text="@LastSteamLobbyText"/>
+                        <Standard.Button WidthSizePolicy="Fixed" SuggestedWidth="160" HeightSizePolicy="Fixed" SuggestedHeight="58"
+                                         VerticalAlignment="Center"
+                                         Command.Click="ActionReconnectSteamLobby" Parameter.Text="@ReconnectButtonText"/>
                       </Children>
                     </ListPanel>
                   </Children>
--- a/UIMovies/CoopConnectionUIMovie.xml
+++ b/UIMovies/CoopConnectionUIMovie.xml
@@ -391,7 +391,7 @@
                                     Text="@LastDirectConnectionText"/>
                         <Standard.Button WidthSizePolicy="Fixed" SuggestedWidth="160" HeightSizePolicy="Fixed" SuggestedHeight="58"
                                          VerticalAlignment="Center"
-                                         Command.Click="ActionConnect" Parameter.Text="@ReconnectButtonText"/>
+                                         Command.Click="ActionReconnectDirect" Parameter.Text="@ReconnectButtonText"/>
                       </Children>
                     </ListPanel>
 
--- a/source/Coop.Core/CoopartiveMultiplayerExperience.cs
+++ b/source/Coop.Core/CoopartiveMultiplayerExperience.cs
@@ -46,6 +46,7 @@ namespace Coop.Core
         private volatile bool hostedSession;
         private volatile bool clientConnectedOnce;
         private bool passwordInquiryPending;
+        private string pendingPreSuppliedPassword;
         private int coopStartGeneration;
         // Bumped when a new host attempt starts, so a prior attempt's deferred exit handling drops out.
         private volatile int hostSessionGeneration;
@@ -72,6 +73,7 @@ namespace Coop.Core
             this.coopLogFilePath = coopLogFilePath;
 
             messageBroker.Subscribe<AttemptJoin>(Handle);
+            messageBroker.Subscribe<JoinSteamLobby>(Handle);
             messageBroker.Subscribe<AttemptHost>(Handle);
             messageBroker.Subscribe<HostSaveGame>(Handle);
             messageBroker.Subscribe<EndCoopMode>(Handle);
@@ -268,6 +270,12 @@ namespace Coop.Core
         {
             clientConnectedOnce = true;
             setCrashPhase("connected");
+            messageBroker.Publish(this, new ClientNetworkConnected());
+        }
+
+        private void Handle(MessagePayload<JoinSteamLobby> payload)
+        {
+            pendingPreSuppliedPassword = payload.What.PreSuppliedPassword;
         }
 
         private void Handle(MessagePayload<SessionJoinInfoResolved> obj)
@@ -277,7 +285,16 @@ namespace Coop.Core
 
             if (joinInfo.PasswordRequired)
             {
-                PromptForSessionPassword(joinInfo);
+                if (!string.IsNullOrEmpty(pendingPreSuppliedPassword))
+                {
+                    joinInfo.Password = pendingPreSuppliedPassword;
+                    pendingPreSuppliedPassword = null;
+                    StartResolvedJoin(joinInfo);
+                }
+                else
+                {
+                    PromptForSessionPassword(joinInfo);
+                }
                 return;
             }
 
--- a/source/Coop.Core/CoopartiveMultiplayerExperience.cs
+++ b/source/Coop.Core/CoopartiveMultiplayerExperience.cs
@@ -241,6 +241,7 @@ namespace Coop.Core
             hostSessionGeneration++;
             hostedSession = false;
             clientConnectedOnce = false;
+            pendingPreSuppliedPassword = null;
         }
 
         private void Handle(MessagePayload<HostedServerExited> obj)
--- a/source/Coop.Core/CoopartiveMultiplayerExperience.cs
+++ b/source/Coop.Core/CoopartiveMultiplayerExperience.cs
@@ -47,6 +47,7 @@ namespace Coop.Core
         private volatile bool clientConnectedOnce;
         private bool passwordInquiryPending;
         private string pendingPreSuppliedPassword;
+        private string acceptedSessionPassword;
         private int coopStartGeneration;
         // Bumped when a new host attempt starts, so a prior attempt's deferred exit handling drops out.
         private volatile int hostSessionGeneration;
@@ -242,6 +243,7 @@ namespace Coop.Core
             hostedSession = false;
             clientConnectedOnce = false;
             pendingPreSuppliedPassword = null;
+            acceptedSessionPassword = null;
         }
 
         private void Handle(MessagePayload<HostedServerExited> obj)
@@ -271,7 +273,11 @@ namespace Coop.Core
         {
             clientConnectedOnce = true;
             setCrashPhase("connected");
-            messageBroker.Publish(this, new ClientNetworkConnected());
+            var accepted = acceptedSessionPassword;
+            acceptedSessionPassword = null;
+            GameThread.RunSafe(
+                () => messageBroker.Publish(this, new ClientNetworkConnected { AcceptedPassword = accepted }),
+                context: nameof(NetworkConnected));
         }
 
         private void Handle(MessagePayload<JoinSteamLobby> payload)
@@ -284,12 +290,14 @@ namespace Coop.Core
             var joinInfo = obj.What.JoinInfo;
             if (!CanStartResolvedJoin()) return;
 
+            var preSuppliedPassword = pendingPreSuppliedPassword;
+            pendingPreSuppliedPassword = null;
+
             if (joinInfo.PasswordRequired)
             {
-                if (!string.IsNullOrEmpty(pendingPreSuppliedPassword))
+                if (!string.IsNullOrEmpty(preSuppliedPassword))
                 {
-                    joinInfo.Password = pendingPreSuppliedPassword;
-                    pendingPreSuppliedPassword = null;
+                    joinInfo.Password = preSuppliedPassword;
                     StartResolvedJoin(joinInfo);
                 }
                 else
@@ -318,6 +326,7 @@ namespace Coop.Core
                 {
                     passwordInquiryPending = false;
                     joinInfo.Password = password ?? string.Empty;
+                    acceptedSessionPassword = joinInfo.Password;
                     StartResolvedJoin(joinInfo);
                 },
                 () =>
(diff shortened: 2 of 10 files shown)


garrettluskey (changes requested):
Can you add some pictures?

Duinhil:
Sure, default state prior to any connections being saved:
image
Direct connection saved with reconnect option:
image

garrettluskey (changes requested):
Are you able to move that to the default page (direct connect?) also we want to hide the IP for streamers

Duinhil:
Rationale for it being it's own tab is that it also remembers the details for the last Steam lobby joined (if any), and adding a reconnect button to the Steam Lobbies tab would still trigger the initial Steam lobby search when loading the tab.

Not sure how you'd prefer to proceed there? But can look at hiding the IP from that tab in the morning, though it would still be visible on the direct connect tab that opens by default.

Duinhil:
Have removed details of the connection and just to provide a more complete example screenshot with both a Steam Lobby and Direct connection saved,
image

UIMovies/CoopConnectionUIMovie.xml
```diff
@@ -366,6 +367,50 @@
...
+                               WidthSizePolicy="StretchToParent" HeightSizePolicy="Fixed" SuggestedHeight="58"
+                               MarginBottom="12" StackLayout.LayoutMethod="HorizontalLeftToRight">
+                      <Children>
+                        <TextWidget WidthSizePolicy="StretchToParent" HeightSizePolicy="StretchToParent"
+                                    Brush="Clan.TabControl.Text" Brush.FontSize="22" Brush.FontColor="#B8B8B8FF"
+                                    Brush.TextHorizontalAlignment="Left" Brush.TextVerticalAlignment="Center"
+                                    Text="@LastDirectConnectionText"/>
+                        <Standard.Button WidthSizePolicy="Fixed" SuggestedWidth="160" HeightSizePolicy="Fixed" SuggestedHeight="58"
+                                         VerticalAlignment="Center"
+                                         Command.Click="ActionConnect" Parameter.Text="@ReconnectButtonText"/>
```
garrettluskey: this calls `ActionConnect`, which reads the editable `Ip`, `Port`, and `Password` fields instead of the saved connection. edit the direct form without joining and this reconnects to those edits, then overwrites the saved server. use a dedicated direct reconnect command backed by the saved snapshot and test that path.

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
```diff
@@ -398,6 +443,7 @@ public void ActionConnect()
             }
 
             messageBroker.Publish(this, new AttemptJoin(ip, port, connectPassword, steamInvites));
+            SaveDirectConnection();
```
garrettluskey: this saves before `NetworkConnected`, so an unreachable or rejected attempt replaces the last successfully used address from #2402. keep the attempt pending and persist it only after the connection succeeds.

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
```diff
@@ -567,9 +620,110 @@ private void RequestSteamLobbyJoin(ulong lobbyId)
     {
         if (disposed || lobbyId == 0) return;
 
+        SaveSteamLobby(lobbyId);
```
garrettluskey: same problem for Steam, this runs before the lobby enter callback, join-info resolution, and the actual network connection. a failed lobby click becomes the reconnect target, so save it only after a successful connection.

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
```diff
@@ -297,6 +315,33 @@ public string Password
...
+    public bool HasLastDirectConnection
+    {
+        get => hasLastDirectConnection;
+        private set
+        {
+            if (hasLastDirectConnection == value) return;
+
+            hasLastDirectConnection = value;
+            OnPropertyChanged(nameof(HasLastDirectConnection));
+            OnPropertyChanged(nameof(LastDirectConnectionText));
```
garrettluskey, line 343: `HasNoLastConnection` is bound in the XML but neither this setter nor `SaveSteamLobby` notifies it. after the first join, the empty-state text stays visible beside the reconnect row until this VM is recreated. notify it on both transitions and cover the property notification.

source/GameInterface/Services/UI/LastConnectionData.cs
```diff
@@ -0,0 +1,16 @@
...
+
+/// <summary>Persisted data for the last direct-connection and Steam lobby the player joined.</summary>
+public class LastConnectionData
+{
+    public const string TabId = "Connection";
+    public const string SectionId = "LastSession";
+
+    [JsonPropertyName("directIp")] public string DirectIp { get; set; }
+    [JsonPropertyName("directPort")] public string DirectPort { get; set; }
+    [JsonPropertyName("directPassword")] public string DirectPassword { get; set; }
```
garrettluskey: dont persist the server password in `coop_options.json`. this puts a reusable credential in plaintext under Documents, where config files are commonly synced or shared. keep it out of `LastConnectionData` and prompt again or use protected credential storage.

source/GameInterface.Tests/Services/UI/CoopConnectMenuVMTests.cs
```diff
@@ -326,6 +327,147 @@ private static string[] VisibleHosts(CoopConnectMenuVM viewModel)
...
+        using var viewModel = new CoopConnectMenuVM(browser, messageBroker, store);
+
+        viewModel.Ip = "127.0.0.1";
+        viewModel.Port = "4200";
+        viewModel.Password = "secret";
+        viewModel.ActionConnect();
+
+        Assert.True(viewModel.HasLastDirectConnection);
+        Assert.False(viewModel.HasNoLastConnection);
+        Assert.Contains("127.0.0.1:4200", viewModel.LastDirectConnectionText);
```
garrettluskey: this and the three address/host assertions below still expect the details that `adb90af5f` removed from these labels. all four new persistence tests fail now, update the expectations and keep explicit checks that the labels dont expose the address or host.

garrettluskey (changes requested):
the persistence seam and tab wiring are straightforward, and corrupt options fall back without breaking the menu. changes requested because reconnect currently records or uses the wrong state, the bound empty state does not refresh, credentials are written in plaintext, and the new tests fail.

UIMovies/CoopConnectionUIMovie.xml
Duinhil: This now calls `ActionReconnectDirect` instead as a separate path.

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
Duinhil: Both paths now store their credentials in memory as pending and wait for the connection to be established before storing

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
Duinhil: Both paths now set this

source/GameInterface/Services/UI/LastConnectionData.cs
Duinhil: Fair, valid concern. I've added some logic to store this using Windows DPAPI which adds an extra package to the project. I'll admit I'm less familiar with this but research suggests this should be reasonably secure tying the decryption to both the Windows User and the physical machine so should be safe if shared. However, if this remains a concern or you don't want extra packages then it can be removed. Though speaking personally this was one of me key motivators in opening the PR

source/GameInterface.Tests/Services/UI/CoopConnectMenuVMTests.cs
Duinhil: Sorry about that, should have re-ran tests after the changes. Have updated appropriately now

source/Coop.Core/CoopartiveMultiplayerExperience.cs
```diff
@@ -265,6 +268,12 @@ private void Handle(MessagePayload<NetworkConnected> obj)
         {
             clientConnectedOnce = true;
             setCrashPhase("connected");
+            messageBroker.Publish(this, new ClientNetworkConnected());
```
garrettluskey: this relay runs from `CoopClient.OnPeerConnected` on the LiteNetLib poller thread, and `HandleNetworkConnected` immediately does file IO plus bound `OnPropertyChanged` calls. marshal that work through `GameThread.RunSafe` so the poller isnt blocked and Gauntlet isnt updated off-thread.

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
```diff
@@ -567,7 +664,135 @@ private void RequestSteamLobbyJoin(ulong lobbyId)
     {
         if (disposed || lobbyId == 0) return;
 
-        messageBroker.Publish(this, new JoinSteamLobby(lobbyId));
+        pendingSteamLobbyId = lobbyId;
+        pendingSteamHostName = discoveredSteamLobbies.FirstOrDefault(l => l.LobbyId == lobbyId)?.HostText
+            ?? string.Empty;
+        pendingSteamPassword = connectPassword;
+        messageBroker.Publish(this, new JoinSteamLobby(lobbyId, connectPassword));
```
garrettluskey: `connectPassword` belongs to the direct tab, and `LoadLastConnection` fills it from the saved direct server. passing it into every normal Steam join makes a protected lobby skip its password prompt with an unrelated credential. keep normal lobby joins prompt-based and only pre-supply the saved Steam password from `ActionReconnectSteamLobby`.

source/Coop.Core/CoopartiveMultiplayerExperience.cs
```diff
@@ -274,7 +283,16 @@ private void Handle(MessagePayload<SessionJoinInfoResolved> obj)
 
             if (joinInfo.PasswordRequired)
             {
-                PromptForSessionPassword(joinInfo);
+                if (!string.IsNullOrEmpty(pendingPreSuppliedPassword))
```
garrettluskey: `pendingPreSuppliedPassword` is only cleared when this branch consumes it. reconnecting to an open lobby or failing before resolution leaves the old value behind, so a later friends-list or invite join to a protected lobby can skip its prompt with that stale password. clear it on every resolved or failed Steam attempt.

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
```diff
@@ -567,7 +664,135 @@ private void RequestSteamLobbyJoin(ulong lobbyId)
...
+    }
+
+    private void HandleNetworkConnected(MessagePayload<ClientNetworkConnected> payload)
+    {
+        if (disposed) return;
+
+        if (!string.IsNullOrEmpty(pendingDirectIp))
```
garrettluskey, line 704: failed attempts never clear these pending fields. after a direct attempt fails, a Steam join on the same screen can succeed while `pendingDirectIp` still wins this branch, so the failed direct server is saved and the lobby isnt. clear failed state and replace the opposite pending kind when a new attempt starts.

garrettluskey (changes requested):
the reconnect UI, protected storage, and persistence tests are moving in the right direction. changes requested because the success relay touches UI and options on the poller thread, normal Steam joins reuse the direct password, and pending attempt state can be applied to later joins.

gamingmenacers-svg (commented):
the current head still has the three connection-order blockers: ClientNetworkConnected mutates options and VM properties on the network callback thread; a normal Steam lobby join copies the direct-connect password and can skip the real lobby prompt; and failed pending direct state is not cleared, so a later successful Steam join can persist the failed direct attempt. these need attempt IDs or explicit failure/reset handling plus game-thread marshalling before merge.

Duinhil:
Have addressed all the latest feedback and updated the tests accordingly

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
```diff
@@ -406,6 +470,45 @@ public void ActionConnect()
...
+                ip = parsedIp;
+            }
+            else
+            {
+                var addresses = Dns.GetHostAddresses(lastDirectIp);
+                ip = addresses.FirstOrDefault(x => x.AddressFamily == AddressFamily.InterNetwork);
+                if (ip == null) return;
+            }
+
+            messageBroker.Publish(this, new AttemptJoin(ip, port, lastDirectPassword ?? string.Empty, steamInvites));
```
garrettluskey, line 506: This reconnect path never replaces the pending snapshot. Save server A, make a form attempt to B that gets rejected, then reconnect A: `pendingDirectIp` is still B, so A's `ClientNetworkConnected` persists B and replaces the good snapshot with the failed server. Set the pending direct fields from the saved values and clear the Steam fields before publishing, or carry an attempt ID through the success event, and cover failed-attempt -> reconnect.

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
```diff
@@ -567,7 +671,139 @@ private void RequestSteamLobbyJoin(ulong lobbyId)
     {
         if (disposed || lobbyId == 0) return;
 
-        messageBroker.Publish(this, new JoinSteamLobby(lobbyId));
+        pendingSteamLobbyId = lobbyId;
```
garrettluskey, line 689: This advances the saved pending target even when the join request is not accepted. `SteamJoinListener` ignores lobby B while a join for lobby A is in flight, and these rows stay clickable; click A then B before A's callback and A can connect while `ClientNetworkConnected` persists B. Serialize/disable in-flight joins or propagate a lobby/attempt identity through resolution and connection, then test a delayed first callback.

source/Coop.Core/CoopartiveMultiplayerExperience.cs
```diff
@@ -268,16 +273,37 @@ private void Handle(MessagePayload<NetworkConnected> obj)
...
+            var accepted = acceptedSessionPassword;
+            acceptedSessionPassword = null;
+            GameThread.RunSafe(
+                () => messageBroker.Publish(this, new ClientNetworkConnected { AcceptedPassword = accepted }),
+                context: nameof(NetworkConnected));
+        }
+
+        private void Handle(MessagePayload<JoinSteamLobby> payload)
+        {
+            pendingPreSuppliedPassword = payload.What.PreSuppliedPassword;
```
garrettluskey, line 286: This password is still process-global and only cleared on the resolved path. If `JoinLobby` throws or `OnLobbyEntered` fails, `SessionJoinFailed` leaves the saved password here; the next protected friends-list/invite join has no `JoinSteamLobby` event and consumes the previous host's credential instead of prompting. `acceptedSessionPassword` has the same cross-attempt problem after a prompted connection is rejected. Scope both values to a lobby/attempt and clear them on every failure, abandonment, and session end, with cross-attempt tests.

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
```diff
@@ -567,7 +671,139 @@ private void RequestSteamLobbyJoin(ulong lobbyId)
...
+            {
+                return;
+            }
+
+            if (!string.IsNullOrEmpty(data.DirectIp))
+            {
+                lastDirectIp = data.DirectIp;
+                lastDirectPort = data.DirectPort ?? connectPort;
+                lastDirectPassword = LastConnectionData.UnprotectPassword(data.DirectPasswordProtected);
+                connectIP = lastDirectIp;
```
garrettluskey: This puts the persisted address into `Ip`, and the constructor immediately selects the Direct tab where the XML renders `Ip` as plain text. Merely opening the join screen now exposes the last server address on stream, despite the earlier request to hide it; hiding only the reconnect row does not help. Keep the saved snapshot separate from the default editable form, or open on a view that does not render the address.

garrettluskey (changes requested):
The protected storage, game-thread relay, and separation of normal Steam passwords from the direct form address the earlier security and threading problems. Changes are still needed because connection success is not correlated to the attempt that actually succeeded, password state can cross failed Steam attempts, and opening the menu re-exposes the saved IP.

Duinhil:
Have addressed as three new fixes:
- `pendingPreSuppliedPassword` and `acceptedSessionPassword` are now cleared in `Handle(SessionJoinFailed)` and a new `Handle(SessionJoinAbandoned)` handler in `CoopartiveMultiplayerExperience`.
- `ActionReconnectDirect` now sets `pendingDirectIp`/Port/Password from the saved snapshot (and clears Steam pending state) before publishing, mirroring `ActionReconnectSteamLobby`. Also added a `steamJoinInFlight` guard on `RequestSteamLobbyJoin` to prevent a potential second lobby-list click from overwriting the pending snapshot, guard is cleared on success `HandleNetworkConnected`, failure `HandleJoinFailed`, and cancellation `HandleJoinAbandoned`.
- `LoadLastConnection` no longer uses the saved details and goes back to it's original behaviour, localhost, 4200 and an empty password.

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
```diff
@@ -406,6 +474,52 @@ public void ActionConnect()
...
+            messageBroker.Publish(this, new AttemptJoin(ip, port, lastDirectPassword ?? string.Empty, steamInvites));
+        }
+        catch { }
+    }
+
+    public void ActionReconnectSteamLobby()
+    {
+        if (disposed || lastSteamLobbyId == 0) return;
+
+        steamJoinInFlight = true;
```
garrettluskey, line 516: this path doesnt check `steamJoinInFlight`: click lobby A, then this saved-lobby button before A enters and `SteamJoinListener` ignores B while these pending fields and `pendingPreSuppliedPassword` move to B. A can then use or save B's password and identity, so return while a join is in flight and cover the delayed callback.

source/GameInterface/Services/UI/CoopConnectMenuVM.cs
```diff
@@ -566,8 +683,156 @@ private void NotifySteamLobbyDisplayChanged()
...
+        }
+        catch
+        {
+            // IO failure — not critical
+        }
+    }
+
+    private void HandleJoinFailed(MessagePayload<SessionJoinFailed> _)
+    {
+        steamJoinInFlight = false;
```
garrettluskey, line 834: `SessionJoinFailed` only covers lobby discovery; an incorrect server password goes through `CoopClient.OnPeerDisconnected` and publishes `EndCoopMode`. clear these fields on session end too, otherwise `steamJoinInFlight` stays true and every later lobby-row click is ignored.

source/Coop.Core/CoopartiveMultiplayerExperience.cs
```diff
@@ -388,9 +416,17 @@ private bool CanStartResolvedJoin()
 
         private void Handle(MessagePayload<SessionJoinFailed> obj)
         {
+            pendingPreSuppliedPassword = null;
+            acceptedSessionPassword = null;
```
garrettluskey, line 420: an incorrect prompted password also ends through `EndCoopMode`, not either of these handlers, so `acceptedSessionPassword` survives. clear it on session end too, otherwise a later open Steam join can relay and save the rejected credential.

garrettluskey (changes requested):
the protected storage, game-thread relay, and saved-snapshot reconnect paths address the earlier issues. changes are still needed because a reconnect can overwrite an active Steam attempt, and connection rejection leaves both the menu guard and prompted password state attached to the next attempt.

garrettluskey (changes requested):
Instead of a whole new page, which I don't think is the best page flow. Can we simply save the last connected ip (if there is one) and auto fill the direct connect ip with that? For steam connections let's not save anything

PR #3217, opened 2026-08-21
[3206] Prevent policy decision menu from reopening

### LINKED ISSUE #3206: Policy decision menu reopens after voting timer expires
## Severity of issue/bug
- [x] High
- [ ] Average
- [ ] Low

## Describe the issue/bug that appears
When a client has a kingdom policy decision panel open and the authoritative voting timer expires, the active panel closes but the same decision immediately reopens without a timer. The reopened panel cannot submit or finish, leaving the player soft locked.

The server sends `NetworkKingdomDecisionResolved` before applying the chosen outcome and broadcasting `NetworkRemoveDecision`. On the client, applying the resolution clears the round state and closes the active `DecisionItemBaseVM`. The two messages cross onto the game thread separately, so `KingdomDecisionsVM.OnFrameTick` can run before removal applies. A decision opened directly through `RefreshWith` is not added to `_examinedDecisionsSinceInit`, so vanilla sees the still-unresolved decision and opens it again.

The later remove path closes registered decision items but cannot close the vanilla inquiry created during this gap. If that stale inquiry opens another decision panel, it has no round deadline because client state was already cleared. The co-op final-selection prefix still blocks vanilla resolution, while vote publication fails because the decision no longer has an unresolved-decision index.

## Steps to reproduce
1. Join the co-op campaign with two clients whose clans are eligible to vote in the same kingdom.
2. Open a kingdom policy decision directly from the kingdom management screen.
3. Leave one eligible clan without a final vote.
4. Keep the policy decision panel open until the 60-second voting deadline expires.
5. Observe that the timed panel closes, then the same decision reappears without a timer and cannot be completed.

## Additional information
The existing timeout coverage resolves a round without an active `KingdomDecisionsVM`. Existing active-panel coverage delivers resolution and removal without ticking `KingdomDecisionsVM` between those messages.

Add regression coverage that opens the decision through `RefreshWith`, applies `NetworkKingdomDecisionResolved`, runs one decision VM frame before `NetworkRemoveDecision`, and verifies that no inquiry or decision panel is reopened. The resolved decision must remain suppressed until authoritative removal applies.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Letting a kingdom policy decision voting timer expire closes the timed menu, then can reopen the same decision without a timer and leave the player unable to continue.

## What is the new behavior?

Resolves #3206

Letting the timer expire now closes the policy decision and keeps it closed while the resolved decision is removed, so the Kingdom screen remains usable.

## Bot Changelog Entry

- Fixed an issue where the same Kingdom Policy Decision could pop up twice, with the second menu being a soft lock

### DIFF
--- a/source/E2E.Tests/Services/Kingdoms/PlayerKingdomCreationFlowTests.cs
+++ b/source/E2E.Tests/Services/Kingdoms/PlayerKingdomCreationFlowTests.cs
@@ -1046,6 +1046,66 @@ public class PlayerKingdomCreationFlowTests : IDisposable
     }
 
     [Fact]
+    public void KingdomDecisionResolution_DoesNotReopenBeforeRemovalApplies()
+    {
+        var client = Clients.First();
+        client.Resolve<IControllerIdProvider>().SetControllerId(ControllerId);
+
+        var player = CreateSyncedPlayerContext(ControllerId, client);
+        var otherPlayer = CreateSyncedPlayerContext("PolicyTimeoutOtherClan", _ => false);
+        var kingdomId = TestEnvironment.CreateRegisteredObject<Kingdom>();
+
+        ConfigureClanInKingdom(player.ClanId, kingdomId);
+        ConfigureClanInKingdom(otherPlayer.ClanId, kingdomId);
+        EnsureKingdomRegisteredEverywhere(kingdomId);
+
+        client.Call(() =>
+        {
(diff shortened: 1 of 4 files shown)


garrettluskey (approved):
the resolved state now survives the resolution/removal gap, and the regression test exercises vanilla's reopen path before removal. the debug fixture keeps campaign mutations on the server. looks good.

PR #3220, opened 2026-08-21
[3133] Add log headers 

### LINKED ISSUE #3133: [MOD COMPATIBILITY] Log enabled client mods and game version during connection
# For New Contributors

Not provided

## Description

Current client logs already include startup and connection details, but they do not surface the enabled-mod list or their versions; adding these values would make debugging client setup issues easier.

## Intended Design

Not provided

## Location

Not provided

## Related Issues

Not provided

## Requirements

- Include all enabled mods for the client in logs, including versions for each mod (including base game addons like Sandbox and Campaign).
- Include the client game version in a user-readable format.
- Keep the new client log section ordered and organized for readability, similar to Dedicated Server logs.

## Additional information

Not provided

## Definition of Done

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1539086049504923729/1539086049504923729

casey-johnson:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Logs do not contain a header similar to the dedicated server log

## What is the new behavior?

Resolves [#3133](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/issues/3133)

Adds a header to the logs containing new information about current modules

## Bot Changelog Entry

- Added log headers to assist support tickets

### DIFF
--- a/source/Coop.Core/Diagnostics/StartupDiagnosticsSequence.cs
+++ b/source/Coop.Core/Diagnostics/StartupDiagnosticsSequence.cs
@@ -11,10 +11,13 @@ public static class StartupDiagnosticsSequence
 {
     public static void Run(
         Action<string> emitLogHeader,
-        Action<string> initializeCrashReporting)
+        Action<string> initializeCrashReporting,
+        Func<string> resolveInformationalVersion = null)
     {
-        string version = ModInformation.BuildVersion;
+        string version = (resolveInformationalVersion ?? DefaultResolveInformationalVersion)();
         emitLogHeader(version);
         initializeCrashReporting(version);
     }
+
+    private static string DefaultResolveInformationalVersion() => ModInformation.BuildVersion;
 }
--- a/source/Coop.Core/Diagnostics/StartupDiagnosticsSequence.cs
+++ b/source/Coop.Core/Diagnostics/StartupDiagnosticsSequence.cs
@@ -15,7 +15,16 @@ public static class StartupDiagnosticsSequence
         Func<string> resolveInformationalVersion = null)
     {
         string version = (resolveInformationalVersion ?? DefaultResolveInformationalVersion)();
-        emitLogHeader(version);
+        
+        try
+        {
+            emitLogHeader(version);
+        }
+        catch
+        {
+            // Best effort
+        }
+        
         initializeCrashReporting(version);
     }
 
--- a/source/Coop.Core/Diagnostics/StartupDiagnosticsSequence.cs
+++ b/source/Coop.Core/Diagnostics/StartupDiagnosticsSequence.cs
@@ -1,0 +1,20 @@
+using System;
+
+namespace Coop.Core.Diagnostics;
+
+/// <summary>
+/// Threads the resolved build version through crash-reporting startup so a caller cannot
+/// initialize crash reporting or emit the log header before the version is resolved.
+/// </summary>
+public static class StartupDiagnosticsSequence
+{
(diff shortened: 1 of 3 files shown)


ShoT-UPfps:
could you please fix the codefactor?

garrettluskey (dismissed):

ShoT-UPfps (dismissed):
cmt

source/Coop/CoopMod.cs
```diff
@@ -123,6 +123,13 @@ public override void NoHarmonyInit()
             SetupLogging();
             InitializeCrashReporting();
             
+            var moduleInfoProvider = new TaleWorldsModuleInfoProvider();
+            informationalVersion = typeof(ModInformation).Assembly
```
ShoT-UPfps: this assignment occurs after `InitializeCrashReporting()`, but that method immediately calls `CrashDiagnostics.Initialize(informationalVersion, role)` and launches `Coop.CrashReporter.exe` with `COOP_CRASH_BUILD` set from the same field. since the field is still its default "unknown", every startup loses the actual coop build in both crash-reporting paths. it also allows reporter-startup warnings to precede the supposed header. resolve the informational version and emit the header before calling `InitializeCrashReporting()`, and add a startup-order regression test.

gamingmenacers-svg (commented):
one startup hole remains on 85c8fa7: CoopLogHeader enumerates modules and dereferences the Native result before InitializeCrashReporting runs, with no fallback if module discovery fails or Native is absent. that turns the new diagnostic header into an unreported startup crash path. initialize crash capture first or make header collection best-effort with an unknown game version, then rebase this 160-commit-old draft.

source/Coop/CoopMod.cs
casey-johnson: regression tests in its current form will be difficult within Coopmod.NoHarmonyInit. I plan to extract the sequence into a regression-testable form.

source/Coop.Core/Diagnostics/StartupDiagnosticsSequence.cs
```diff
@@ -0,0 +1,23 @@
...
+/// </summary>
+public static class StartupDiagnosticsSequence
+{
+    public static void Run(
+        Action<string> emitLogHeader,
+        Action<string> initializeCrashReporting,
+        Func<string> resolveInformationalVersion = null)
+    {
+        string version = (resolveInformationalVersion ?? DefaultResolveInformationalVersion)();
+        emitLogHeader(version);
```
garrettluskey: If `emitLogHeader` throws, this line skips `initializeCrashReporting` immediately below it. The production delegate performs the new TaleWorlds module discovery and formatting, so a failure there now unwinds `OnSubModuleLoad` while `CrashDiagnostics` still has the unknown build/role and the external reporter has not started. Make header collection/emission best-effort while guaranteeing crash reporting still initializes, and add a regression test with a throwing header delegate.

garrettluskey (changes requested):
The build version is now resolved before both the header and crash-reporting startup, and the new tests cover that ordering. I’m requesting one remaining change because the diagnostic header can still disable the crash path it is meant to support.

garrettluskey (approved):
The prior fail-safe issue is resolved at the current head. Header collection is now best-effort, the resolved build is assigned before module enumeration can fail, and crash reporting still initializes afterward. The throwing-header regression test covers that path. I found no new defect in the affected startup sequence.

PR #3227, opened 2026-08-22
Block adding a hero to TroopRosters more than once

jordanbrymora:
## PR Checklist

## PR Type

There have been many reports of companions, spouses etc. being added two or more times to the same `TroopRoster`.

## What is the new behavior?

This aims to apply a band aid fix by blocking adding the same hero to a `TroopRoster` if they are already present in it.

Two tests are also removed that are no longer relevant with the changes to `PrefixAddToCountsAtIndex` as adding a hero to a `TroopRoster` two or more times is now blocked.

## Bot Changelog Entry

It should be no longer be possible for heroes/companions to take multiple slots within a party.


source/GameInterface/Services/TroopRosters/Patches/TroopRosterPatches.cs
```diff
@@ -47,31 +47,39 @@ internal static bool ShouldReportClientMutation(bool isClient, bool isRegistered
...
     [HarmonyPrefix]
-    private static void PrefixAddToCountsAtIndex(TroopRoster __instance, int index, int countChange,
+    private static bool PrefixAddToCountsAtIndex(TroopRoster __instance, int index, int countChange,
         int woundedCountChange, int xpChange, bool removeDepleted)
     {
-        if (CallOriginalPolicy.IsOriginalAllowed()) return;
+        // Resolve the element by identity now: the index is valid here (pre-mutation), but a subtract-to-zero
+        // with removeDepleted removes it before a postfix could read it back.
+        var character = __instance.GetElementCopyAtIndex(index).Character;
+        if (character.IsHero && __instance.GetTroopCount(character) + countChange > 1)
```
ShoT-UPfps: this blocks cleanup too. v1.4.8 `RemoveTroop` calls `AddToCountsAtIndex(index, -1, ...)`; from count 3, the result 2 still trips this guard. `TakePrisonerAction` then changes the hero to Prisoner and adds it to the captor while the source roster remains tripled. restrict the guard to positive additions or normalize legacy counts, and test capture/removal from count 3.

source/GameInterface/Services/TroopRosters/Patches/TroopRosterPatches.cs
```diff
@@ -47,31 +47,39 @@ internal static bool ShouldReportClientMutation(bool isClient, bool isRegistered
...
+        {
+            Logger.Error("Attempted to add a hero to a TroopRoster where they were already present.");
+            return false;
```
ShoT-UPfps, line 65: this is too late to reject outer `AddToCounts`. v1.4.8 has already updated the version and called `AddNewElement` before reaching this prefix, so an absent `AddToCounts(hero, 2)` leaves a zero-count slot; the allowed-thread outer postfix can also publish the rejected delta. guard before insertion and suppress publication, with a regression asserting no state or network change.

source/E2E.Tests/Services/Issues/AwaitingAlternativeSolutionTroopsTests.cs
```diff
@@ -656,64 +656,6 @@ public void OnHourlyTick_ListenServerHostOwnsAwaitingTroops_DrainsThemBackIntoMa
         });
     }
 
-    [Fact]
-    public void RequestAwaitingAlternativeSolutionTroopsDeposit_ClaimedRosterExceedsSentTroops_ClampedToWhatWasActuallySent()
```
ShoT-UPfps, line 660: this remains a server trust-boundary test. an old or modified client can submit raw `TroopRosterData`; server unpacking now turns the overclaim into zero, so `PrisonerSaleValidator` skips it instead of clamping to the one authoritative sent troop. keep the test, construct the packed element directly, and preserve the expected deposit of 1.

source/E2E.Tests/Services/MapEvents/MapEventEnvironmentTests.cs
```diff
@@ -553,65 +553,6 @@ public void DuplicateEscapeRequests_ReleaseOnlyOnce()
         }
     }
 
-    [Fact]
-    public void PartyScreenDiscard_DuplicatedPlayerPrisoner_ClearsEveryCopyAndRestoresPartyOnce()
```
ShoT-UPfps, line 557: the guard only blocks this test's seeding helper; it does not migrate old saves or eliminate a stale client that is one copy ahead. the release path still sends absolute-zero tombstones for that divergence. reseed malformed state with `SetElementNumber` or a raw helper and keep the 2-vs-3 cleanup regression.

ShoT-UPfps (commented):
reviewed `1fe749ae` against live development `45785371`; Build and Test #5133 passed all ten jobs. three P1 blockers remain: the predicate blocks a normal -1 removal from a legacy hero count of 3; an absent `AddToCounts(hero, 2)` has already versioned and inserted an element before the inner veto; and raw awaiting-troop overclaims now become zero instead of being clamped to the one authoritative troop. the deleted duplicate-prisoner test still covers existing or stale-peer corruption and should be reseeded, not removed.

source/GameInterface/Services/TroopRosters/Patches/TroopRosterPatches.cs
```diff
@@ -47,31 +47,39 @@ internal static bool ShouldReportClientMutation(bool isClient, bool isRegistered
...
     [HarmonyPatch(nameof(TroopRoster.AddToCountsAtIndex))]
     [HarmonyPrefix]
-    private static void PrefixAddToCountsAtIndex(TroopRoster __instance, int index, int countChange,
+    private static bool PrefixAddToCountsAtIndex(TroopRoster __instance, int index, int countChange,
         int woundedCountChange, int xpChange, bool removeDepleted)
     {
-        if (CallOriginalPolicy.IsOriginalAllowed()) return;
+        // Resolve the element by identity now: the index is valid here (pre-mutation), but a subtract-to-zero
+        // with removeDepleted removes it before a postfix could read it back.
+        var character = __instance.GetElementCopyAtIndex(index).Character;
```
gamingmenacers-svg, line 60: this dereference now runs before the original policy, client, registration, and bounds guards. `AddToCountsAtIndex` calls in the slot-padding or invalid-index cases that the old prefix deliberately ignored will now throw in the Harmony prefix instead of reaching vanilla behavior. move the existing `index < 0 || index >= Count` check before `GetElementCopyAtIndex`, then apply the hero predicate only to a valid non-null character; keep a padding-index regression.

gamingmenacers-svg (commented):
the existing duplicate-count regressions are still blocking, and there is an additional bounds regression in the same prefix:

source/GameInterface/Services/TroopRosters/Patches/TroopRosterPatches.cs
jordanbrymora: Changed guard to only consider positive additions and added test

source/GameInterface/Services/TroopRosters/Patches/TroopRosterPatches.cs
jordanbrymora: Added guard to prevent publishing in postfix

source/E2E.Tests/Services/Issues/AwaitingAlternativeSolutionTroopsTests.cs
jordanbrymora: Re-added

source/E2E.Tests/Services/MapEvents/MapEventEnvironmentTests.cs
jordanbrymora: Re-added

source/GameInterface/Services/TroopRosters/Patches/TroopRosterPatches.cs
```diff
@@ -208,10 +218,26 @@ private static void PrefixSetElementXp(TroopRoster __instance, int index, int nu
...
+    static bool Prefix(TroopRoster __instance, CharacterObject character, int count, out bool __state)
+    {
+        __state = true;
+        if (character == null ||
+            !TroopRosterPatches.ShouldRejectHeroAddition(character.IsHero, __instance.GetTroopCount(character), count))
+            return true;
+
+        __state = false;
+        Logger.Error("Attempted to add a hero to a TroopRoster where they were already present.");
+        return false;
```
ShoT-UPfps, line 235: `TryApplyTroopRosterDeltas` validates the full batch, commits every removal before additions, ignores `AddToCounts`' return value, and currently permits a hero's validated final count to exceed one. A `(source, -1)` plus `(destination, +2)` request therefore removes the source hero, hits this silent veto for the destination addition, then returns `true` and lets `PartyDoneLogicHandler` continue its remaining effects. Reject `character.IsHero && finalNumber > 1` during the prevalidation pass before either mutation pass, and add a regression asserting the invalid batch returns false with both rosters unchanged.

source/GameInterface/Services/TroopRosters/Patches/TroopRosterPatches.cs
```diff
@@ -45,33 +45,43 @@ internal static bool IsRegistered(TroopRoster roster)
...
+        if (index < 0 || index >= __instance.Count) return true;
+
+        // Resolve the element by identity now: the index is valid here (pre-mutation), but a subtract-to-zero
+        // with removeDepleted removes it before a postfix could read it back.
+        var character = __instance.GetElementCopyAtIndex(index).Character;
+        if (character != null &&
+            ShouldRejectHeroAddition(character.IsHero, __instance.GetTroopCount(character), countChange))
+        {
+            Logger.Error("Attempted to add a hero to a TroopRoster where they were already present.");
+            return false;
```
ShoT-UPfps, line 66: Both patched v1.4.8 methods return a roster index. When a Harmony prefix skips either original without writing `ref int __result`, the caller receives the CLR default `0`; a rejected duplicate at another index—or an absent `AddToCounts(hero, 2)`—can therefore be reported as slot zero. Set a deliberate result on both veto paths (the existing index where applicable, otherwise a failure sentinel such as `-1`) and assert the return values in the regressions.

ShoT-UPfps (changes requested):
Rereviewed `33105d68` against `c6daab28`. Build and Test #5241 succeeds, and the four earlier blockers are addressed. Two regressions remain in the new veto path: batched Party-screen deltas can partially commit while reporting success, and skipped vanilla `int` mutators report the default slot `0` unless the prefixes set `__result`.

source/GameInterface/Services/TroopRosters/Patches/TroopRosterPatches.cs
jordanbrymora: Changed to reject `character.IsHero && finalNumber > 1`

source/GameInterface/Services/TroopRosters/Patches/TroopRosterPatches.cs
jordanbrymora: Changed to override result in both paths

source/GameInterface/Services/Issues/Handlers/AwaitingAlternativeSolutionTroopsHandler.cs
```diff
@@ -193,8 +193,11 @@ private TroopRoster UnpackToRoster(TroopRosterData troops)
         var roster = TroopRoster.CreateDummyTroopRoster();
         foreach (var element in troopRosterInterface.UnpackTroopRosterData(troops))
         {
-            roster.AddToCounts(element.Character, element.Number, false, element.WoundedNumber, element.Xp, false);
+            int index = roster.AddNewElement(element.Character, -1);
```
ShoT-UPfps, line 196: `UnpackTroopRosterData` yields every array element and vanilla `AddNewElement` creates a distinct slot each time. A modified client can therefore send the same regular-troop ID twice; `PrisonerSaleValidator` clamps each slot independently against the same authoritative stack, then aggregates both into `validatedRoster` (for example, 5 available plus two requests of 5 becomes 10 deposited). The previous `AddToCounts` path coalesced duplicates before validation. Please reject or coalesce duplicate character IDs here, or track remaining availability per character in the validator, and add a duplicate-regular-entry E2E regression.

ShoT-UPfps (changes requested):
Rereviewed `40341091` against `c6daab28`.

The two blockers from the previous review are addressed: invalid hero delta batches are now rejected before mutation, and both skipped `int`-returning mutators set deliberate return values. All 11 checks on this head pass.

One P1 trust-boundary blocker remains. `AwaitingAlternativeSolutionTroopsHandler.UnpackToRoster` now creates one raw roster slot for every incoming element. A modified client can submit the same regular-troop character ID more than once. `PrisonerSaleValidator.Validate` then iterates those duplicate slots independently, clamps each one against the same full authoritative entry, and aggregates each accepted amount into the validated roster. For example, two requested entries of 5 against 5 available validate to 10, so the deposited/confirmed roster can exceed `AlternativeSolutionSentTroops`. The old `AddToCounts` unpack path coalesced duplicate regular entries before validation, so this is a regression introduced by the raw reconstruction.

Please coalesce or reject duplicate character IDs before raw insertion while preserving the intended raw hero-overclaim behavior, or make validation consume per-character remaining availability. Add an E2E trust-boundary regression with duplicate regular `TroopRosterElementData` entries asserting that the confirmed and deposited total never exceeds the authoritative sent count.

source/GameInterface/Services/Issues/Handlers/AwaitingAlternativeSolutionTroopsHandler.cs
jordanbrymora: Updated validator

source/GameInterface/Services/TroopRosters/Interfaces/TroopRosterInterface.cs
```diff
@@ -210,6 +210,7 @@ public bool TryApplyTroopRosterDeltas(
                 long finalXp = current.xp + elementData.Xp;
                 if (finalNumber < 0 ||
                     finalNumber > int.MaxValue ||
+                    (character.IsHero && finalNumber > 1) ||
```
ShoT-UPfps, line 213: This condition rejects removals as well as additions. For a legacy source stack containing 3 copies of a hero, a source -1 / empty-destination +1 batch now returns false because the source's finalNumber is 2. PartyDoneLogicHandler consequently aborts the whole Party-screen submission and asks the player to reopen it, which cannot repair the persisted count. The base accepted this reduction, and ShouldRejectHeroAddition deliberately permits it; ServerRemovesHeroFromMalformedCountThree_SyncsRemainingTwo only tests direct RemoveTroop, not this batch path. Restrict the duplicate check to positive number deltas while retaining prevalidation, and add a raw-seeded count-3 batch-transfer regression alongside the existing destination +2 rejection test.

ShoT-UPfps (changes requested):
Rereviewed `eecd57fae4dff9472b5110e203ec73fc8bb8d25c` against `c6daab28265c26028d5c778e517222966e2579d8`.

One P2 regression remains: the new hero-count check in `TroopRosterInterface.TryApplyTroopRosterDeltas` also rejects count-reducing changes to existing malformed hero stacks. With a legacy source count of 3 and an empty destination, a normal source -1 / destination +1 transfer is rejected because the source's final count is 2. `PartyDoneLogicHandler` then aborts the entire submission and tells the client to reopen the Party screen, although reopening cannot change that authoritative count. This differs from both the base implementation and this PR's positive-additions-only mutation guard; the new direct RemoveTroop regression does not exercise the batch path.

Implementation handoff: preserve prevalidation before either mutation pass, but apply the duplicate-hero rejection to positive number deltas rather than to every final count above one. Keep the other validation checks. Add a batch-level regression for a raw-seeded legacy source count of 3 and source -1 / destination +1, covering member and prisoner rosters and verifying success, final counts of 2 and 1, destination hero linkage, and client synchronization. Retain the existing invalid destination +2 test and its unchanged-roster/version/no-network assertions.

The earlier findings are addressed for their reported scenarios: invalid destination +2 batches are rejected before mutation, both skipped int-returning mutators now set deliberate results, and duplicate requested regular-troop entries share the remaining authoritative healthy/wounded availability while preserving XP only once. The legacy malformed-prisoner cleanup test and raw hero-overclaim test are retained.

Validation: read the complete 11-file diff, all PR context pages, relevant exact-head/base source, and the installed v1.4.8 vanilla roster mutators. All 11 reported checks on this head pass. This was source/vanilla analysis plus inspection of existing CI results; I did not execute a fresh local build, test suite, or live-game reproduction.

This review applies only to head `eecd57fae4dff9472b5110e203ec73fc8bb8d25c` compared from `c6daab28265c26028d5c778e517222966e2579d8` with base tip `c6daab28265c26028d5c778e517222966e2579d8` as validated at `2026-09-11T01:32:58.160863Z`. Later revisions and later development/base movement were not evaluated by this review.

- Pull request: https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/3227
- Original verdict: `REQUEST_CHANGES`
- Reviewed head: `eecd57fae4dff9472b5110e203ec73fc8bb8d25c`
- Reviewed base tip: `c6daab28265c26028d5c778e517222966e2579d8`
- Comparison base: `c6daab28265c26028d5c778e517222966e2579d8`

This review applies only to head `eecd57fae4dff9472b5110e203ec73fc8bb8d25c` compared from `c6daab28265c26028d5c778e517222966e2579d8` with base tip `c6daab28265c26028d5c778e517222966e2579d8` as validated at `2026-09-11T01:32:58.160863Z`. Later revisions and later development/base movement were not evaluated by this review.

## Exact review summary

```text
Rereviewed `eecd57fae4dff9472b5110e203ec73fc8bb8d25c` against `c6daab28265c26028d5c778e517222966e2579d8`.

One P2 regression remains: the new hero-count check in `TroopRosterInterface.TryApplyTroopRosterDeltas` also rejects count-reducing changes to existing malformed hero stacks. With a legacy source count of 3 and an empty destination, a normal source -1 / destination +1 transfer is rejected because the source's final count is 2. `PartyDoneLogicHandler` then aborts the entire submission and tells the client to reopen the Party screen, although reopening cannot change that authoritative count. This differs from both the base implementation and this PR's positive-additions-only mutation guard; the new direct RemoveTroop regression does not exercise the batch path.

Implementation handoff: preserve prevalidation before either mutation pass, but apply the duplicate-hero rejection to positive number deltas rather than to every final count above one. Keep the other validation checks. Add a batch-level regression for a raw-seeded legacy source count of 3 and source -1 / destination +1, covering member and prisoner rosters and verifying success, final counts of 2 and 1, destination hero linkage, and client synchronization. Retain the existing invalid destination +2 test and its unchanged-roster/version/no-network assertions.

The earlier findings are addressed for their reported scenarios: invalid destination +2 batches are rejected before mutation, both skipped int-returning mutators now set deliberate results, and duplicate requested regular-troop entries share the remaining authoritative healthy/wounded availability while preserving XP only once. The legacy malformed-prisoner cleanup test and raw hero-overclaim test are retained.

Validation: read the complete 11-file diff, all PR context pages, relevant exact-head/base source, and the installed v1.4.8 vanilla roster mutators. All 11 reported checks on this head pass. This was source/vanilla analysis plus inspection of existing CI results; I did not execute a fresh local build, test suite, or live-game reproduction.
```

## Exact inline findings

### 1. [P2] Preserve count-reducing deltas for legacy duplicate heroes

Original target: `source/GameInterface/Services/TroopRosters/Interfaces/TroopRosterInterface.cs:213 (RIGHT)` at `eecd57fae4dff9472b5110e203ec73fc8bb8d25c`

Context:

```text
Original reviewed diff location: source/GameInterface/Services/TroopRosters/Interfaces/TroopRosterInterface.cs:213 (RIGHT).
```

Finding:

```text
This condition rejects removals as well as additions. For a legacy source stack containing 3 copies of a hero, a source -1 / empty-destination +1 batch now returns false because the source's finalNumber is 2. PartyDoneLogicHandler consequently aborts the whole Party-screen submission and asks the player to reopen it, which cannot repair the persisted count. The base accepted this reduction, and ShouldRejectHeroAddition deliberately permits it; ServerRemovesHeroFromMalformedCountThree_SyncsRemainingTwo only tests direct RemoveTroop, not this batch path. Restrict the duplicate check to positive number deltas while retaining prevalidation, and add a raw-seeded count-3 batch-transfer regression alongside the existing destination +2 rejection test.
```

## Exact thread actions

### 1. resolve `PRRT_kwDOD37SzM6euKQ-`

Original latest-comment guard: `PRRC_kwDOD37SzM7puF7c`

```text
The reported source -1 / destination +2 partial-commit scenario is now rejected in the full prevalidation pass, and the new member/prisoner regression asserts unchanged rosters, versions, and network output. A separate inline finding covers the new overbroad rejection of legacy count-reducing operations.
```

### 2. resolve `PRRT_kwDOD37SzM6euKRA`

Original latest-comment guard: `PRRC_kwDOD37SzM7puGZN`

```text
Both skipping prefixes now set ref int __result deliberately: the existing roster index when present and -1 for an absent rejected AddToCounts. Tests assert an existing nonzero index for both entry points and -1 for the rejected absent hero.
```

### 3. resolve `PRRT_kwDOD37SzM6e6WC1`

Original latest-comment guard: `PRRC_kwDOD37SzM7p7QKZ`

```text
PrisonerSaleValidator now subtracts already-validated healthy and wounded counts from each character's availability and preserves authoritative XP only for the first accepted entry. The new duplicate-entry theory exercises raw separate roster slots and verifies bounded totals, single XP preservation, and unchanged authoritative input.
```

source/GameInterface/Services/TroopRosters/Interfaces/TroopRosterInterface.cs
jordanbrymora: Addressed in [92ff44e](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/3227/commits/92ff44e883081a86a64db66a11853d3747eb334a)

ShoT-UPfps (approved):
Rereviewed `92ff44e883081a86a64db66a11853d3747eb334a` against `c6daab28265c26028d5c778e517222966e2579d8`.

No remaining blocking findings in the reviewed changes. The last reported regression is addressed: `TryApplyTroopRosterDeltas` now rejects a hero count above one only for a positive number delta. A legacy source count of 3 with source -1 / empty destination +1 is therefore no longer rejected by this condition. The complete prevalidation pass and removals-before-additions ordering remain intact, so the invalid destination +2 scenario is still rejected before either roster is mutated.

The earlier fixes also remain in place: both skipping prefixes set deliberate return values; the outer AddToCounts guard rejects before vanilla can version or insert a slot and suppresses its publication postfix; valid-index checks precede dereferencing; and duplicate requested regular-troop entries share remaining authoritative healthy/wounded availability while preserving XP only once. The raw hero-overclaim and malformed-prisoner cleanup regressions are retained. I also inspected the awaiting-troop withdrawal consumer, which clamps each removal against the then-current roster, and the client AddCounts replay path.

Nonblocking handoff: add an explicit successful batch-transfer regression for raw-seeded source hero count 3, source -1, and empty destination +1, parameterized for member and prisoner rosters. Assert success, final counts 2 and 1, destination hero linkage, and client synchronization. Keep the existing invalid destination +2 atomic-rejection assertions. The added direct RemoveTroop count-3 regression does not exercise this batch entry point.

Validation: reviewed the complete 11-file diff, every PR context page, relevant exact-head source and the exact-base batch implementation, plus current installed v1.4.8 vanilla roster mutators. All 12 reported checks on this head passed, including build, unit-test job, all eight E2E shards, verification profiles, and CodeFactor. This is source/vanilla analysis and inspection of existing CI results; I did not execute a fresh build, test suite, or live-game reproduction.

This review applies only to head `92ff44e883081a86a64db66a11853d3747eb334a` compared from `c6daab28265c26028d5c778e517222966e2579d8` with base tip `c6daab28265c26028d5c778e517222966e2579d8` as validated at `2026-09-24T16:52:27.965305Z`. Later revisions and later development/base movement were not evaluated by this review.

- Pull request: https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/3227
- Original verdict: `APPROVE`
- Reviewed head: `92ff44e883081a86a64db66a11853d3747eb334a`
- Reviewed base tip: `c6daab28265c26028d5c778e517222966e2579d8`
- Comparison base: `c6daab28265c26028d5c778e517222966e2579d8`

This review applies only to head `92ff44e883081a86a64db66a11853d3747eb334a` compared from `c6daab28265c26028d5c778e517222966e2579d8` with base tip `c6daab28265c26028d5c778e517222966e2579d8` as validated at `2026-09-24T16:52:27.965305Z`. Later revisions and later development/base movement were not evaluated by this review.

## Exact review summary

```text
Rereviewed `92ff44e883081a86a64db66a11853d3747eb334a` against `c6daab28265c26028d5c778e517222966e2579d8`.

No remaining blocking findings in the reviewed changes. The last reported regression is addressed: `TryApplyTroopRosterDeltas` now rejects a hero count above one only for a positive number delta. A legacy source count of 3 with source -1 / empty destination +1 is therefore no longer rejected by this condition. The complete prevalidation pass and removals-before-additions ordering remain intact, so the invalid destination +2 scenario is still rejected before either roster is mutated.

The earlier fixes also remain in place: both skipping prefixes set deliberate return values; the outer AddToCounts guard rejects before vanilla can version or insert a slot and suppresses its publication postfix; valid-index checks precede dereferencing; and duplicate requested regular-troop entries share remaining authoritative healthy/wounded availability while preserving XP only once. The raw hero-overclaim and malformed-prisoner cleanup regressions are retained. I also inspected the awaiting-troop withdrawal consumer, which clamps each removal against the then-current roster, and the client AddCounts replay path.

Nonblocking handoff: add an explicit successful batch-transfer regression for raw-seeded source hero count 3, source -1, and empty destination +1, parameterized for member and prisoner rosters. Assert success, final counts 2 and 1, destination hero linkage, and client synchronization. Keep the existing invalid destination +2 atomic-rejection assertions. The added direct RemoveTroop count-3 regression does not exercise this batch entry point.

Validation: reviewed the complete 11-file diff, every PR context page, relevant exact-head source and the exact-base batch implementation, plus current installed v1.4.8 vanilla roster mutators. All 12 reported checks on this head passed, including build, unit-test job, all eight E2E shards, verification profiles, and CodeFactor. This is source/vanilla analysis and inspection of existing CI results; I did not execute a fresh build, test suite, or live-game reproduction.
```

## Exact inline findings

None.
## Exact thread actions

### 1. resolve `PRRT_kwDOD37SzM6hTzzu`

Original latest-comment guard: `PRRC_kwDOD37SzM7uXxRa`

```text
The exact-head prevalidation condition now restricts duplicate rejection to positive number deltas, addressing the reported legacy count-3 reduction regression while retaining prevalidation and removals-before-additions ordering.
```

PR #3228, opened 2026-08-22
Config for smithed item value

### LINKED ISSUE #3093: [SMITHING] Add a config to multiply the value of crafted items
# For New Contributors

Not provided

## Description

Add a config to multiply the value of crafted items so they aren't insanely profitable if players don't want that on their server.

## Intended Design

Not provided

## Location

Not provided

## Related Issues

Not provided

## Requirements

Crafted weapon value being set to something like 0.3 of what it is, if possible would let you still train skill as usual without it taking forever.

## Additional information

Not provided

## Definition of Done

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1537968304235225089/1537968304235225089

jordanbrymora:
## PR Checklist

## PR Type

## What is the new behavior?

Resolves #3093

## Bot Changelog Entry

Added a new config option to multiply value of items crafted/smithed by players.


source/GameInterface/Configuration/ModConfigProvider.cs
```diff
@@ -47,16 +47,18 @@ public readonly struct ModOptions
     [ProtoMember(13)]
     public readonly float SmithingStaminaRecoveryMultiplier { get; } = 0.1f;
     [ProtoMember(14)]
```
garrettluskey, line 49: dont renumber existing protobuf members. `development` now has `PlayerWoundedBattleEntry` at 20, so this auto-merges into two `[ProtoMember(20)]` fields and breaks model construction. keep 14-20 unchanged and append this as 21.

source/GameInterface/Services/ItemObjects/Patches/CraftedItemValuePatch.cs
```diff
@@ -0,0 +1,22 @@
...
+    [HarmonyPatch(nameof(ItemObject.DetermineValue))]
+    [HarmonyPostfix]
+    public static void DetermineValuePostfix(ItemObject __instance)
+    {
+        if (__instance.IsCraftedByPlayer)
+        {
+            var multiplier = ModConfigProvider.ModOptions.SmithingCraftedItemsValueMultiplier;
+            if (multiplier < 0) multiplier = 1;
+
+            __instance.Value *= (int)multiplier;
```
garrettluskey: `(int)multiplier` truncates before multiplication, so `0.5` makes every crafted item worth 0 and `1.5` leaves values unchanged. multiply as a float first, then convert or round the resulting value to int.

garrettluskey (changes requested):
The config is wired through the default file and network serialization, but the multiplier arithmetic and current-base protobuf numbering both need fixing.

source/GameInterface/Configuration/ModConfigProvider.cs
jordanbrymora: No already explained why previously. It doesn't make sense to keep smithing config options separated and the order should match the config file to ensure we can track all config options correctly. The field numbers changing won't affect players with different versions because different versions are already blocked from joining each other. This is a non-issue.

source/GameInterface/Services/ItemObjects/Patches/CraftedItemValuePatch.cs
jordanbrymora: Changed

source/GameInterface/Configuration/ModConfigProvider.cs
```diff
@@ -47,18 +47,20 @@ public readonly struct ModOptions
...
```
garrettluskey, line 49: dont renumber existing protobuf members. these tags are the wire contract, so matching the config-file order isnt a reason to change fields 14-20. keep those tags and append `SmithingCraftedItemsValueMultiplier` as 21.

garrettluskey (changes requested):
the multiplier arithmetic is fixed and the option is wired through the config and network message. the protobuf tags still need to stay stable before this can merge.

gamingmenacers-svg (commented):
the protobuf blocker remains on 60e7d60: inserting the new option at tag 14 and renumbering every existing field through 21 changes the established ModOptions wire contract. append the new field at the next unused tag and leave every existing tag unchanged, then regenerate the serialization assertions against current development. the multiplier patch itself now applies the configured value correctly.

source/GameInterface/Configuration/ModConfigProvider.cs
garrettluskey: Not an issue

source/Coop.IntegrationTests/Serialization/ModConfigNetworkMessageSerializationTest.cs
```diff
@@ -72,6 +72,7 @@ public void NetworkLoadModConfig_RoundTrips_ConfiguredValues()
                 WandererLimitScalesWithPlayers = true,
                 PlayerKingdomClanTierRequired = 2,
                 SmithingStaminaRecoveryMultiplier = 2.5f,
+                SmithingCraftedItemsValueMultiplier = 3.5f,
```
garrettluskey, line 75: this same-model round trip still passes after fields 14-20 are renumbered, so it doesnt protect the established wire format. add a development-format payload assertion, keep tags 14-20 unchanged, and append this option as 21.

source/GameInterface/Services/ItemObjects/Patches/CraftedItemValuePatch.cs
```diff
@@ -0,0 +1,22 @@
...
+internal class CraftedItemValuePatch
+{
+    [HarmonyPatch(nameof(ItemObject.DetermineValue))]
+    [HarmonyPostfix]
+    public static void DetermineValuePostfix(ItemObject __instance)
+    {
+        if (__instance.IsCraftedByPlayer)
+        {
+            var multiplier = ModConfigProvider.ModOptions.SmithingCraftedItemsValueMultiplier;
+            if (multiplier < 0) multiplier = 1;
```
garrettluskey: this accepts Newtonsoft's `NaN` and positive `Infinity`, and a large finite value can overflow the product. the cast can then set `Value` to `int.MinValue`, so reject non-finite values and cap the product before casting.

garrettluskey (changes requested):
the missing-config default and the real protobuf round trip are wired correctly, and the value patch runs after `ItemValueModel.CalculateValue`. the existing tag-renumbering issue still blocks this, and the new multiplier needs finite/range validation before the cast.

source/Coop.IntegrationTests/Serialization/ModConfigNetworkMessageSerializationTest.cs
jordanbrymora: Not needed

source/GameInterface/Configuration/ModConfigProvider.cs
```diff
@@ -47,18 +47,20 @@ public readonly struct ModOptions
...
-    public readonly float MaximumLootersMultiplier { get; } = 1f;
+    public readonly float SmithingCraftedItemsValueMultiplier { get; } = 1f;
```
garrettluskey, line 50: these tags are still renumbered. `MaximumLootersMultiplier` already owns tag 14 and the existing fields continue through tag 20, so keep those unchanged and append this as tag 21.

source/GameInterface/Services/ItemObjects/Patches/CraftedItemValuePatch.cs
```diff
@@ -0,0 +1,22 @@
...
+internal class CraftedItemValuePatch
+{
+    [HarmonyPatch(nameof(ItemObject.DetermineValue))]
+    [HarmonyPostfix]
+    public static void DetermineValuePostfix(ItemObject __instance)
+    {
+        if (__instance.IsCraftedByPlayer)
+        {
+            var multiplier = ModConfigProvider.ModOptions.SmithingCraftedItemsValueMultiplier;
+            if (multiplier < 0 || multiplier == float.PositiveInfinity || multiplier == float.NaN) multiplier = 1;
```
garrettluskey: `float.NaN` never equals itself, so this check still accepts it. use `float.IsNaN(multiplier)`.

source/GameInterface/Services/ItemObjects/Patches/CraftedItemValuePatch.cs
```diff
@@ -0,0 +1,22 @@
...
+
+            __instance.Value = (int)(__instance.Value * multiplier);
```
garrettluskey: a finite multiplier can still make this result exceed `int.MaxValue`, so the cast is outside the valid range. reject or clamp the scaled value before converting.

source/Coop.IntegrationTests/Serialization/ModConfigNetworkMessageSerializationTest.cs
```diff
@@ -88,6 +89,7 @@ public void NetworkLoadModConfig_RoundTrips_ConfiguredValues()
             Assert.True(copy.WandererLimitScalesWithPlayers);
             Assert.Equal(2, copy.PlayerKingdomClanTierRequired);
             Assert.Equal(2.5f, copy.SmithingStaminaRecoveryMultiplier);
+            Assert.Equal(3.5f, copy.SmithingCraftedItemsValueMultiplier);
```
garrettluskey, line 92: this only round trips the current schema, so it stays green when tags are renumbered on both sides. pin a development payload or the established tag numbers here.

garrettluskey (changes requested):
the previous blockers remain: existing `ModOptions` tags 14-20 are still shifted, `NaN` still passes validation, and large finite values still reach an out-of-range cast.

source/GameInterface/Configuration/ModConfigProvider.cs
jordanbrymora: Not needed

source/GameInterface/Services/ItemObjects/Patches/CraftedItemValuePatch.cs
jordanbrymora: Fixed

source/GameInterface/Services/ItemObjects/Patches/CraftedItemValuePatch.cs
jordanbrymora: Fixed

source/Coop.IntegrationTests/Serialization/ModConfigNetworkMessageSerializationTest.cs
jordanbrymora: Not needed

source/GameInterface/Services/ItemObjects/Patches/CraftedItemValuePatch.cs
```diff
@@ -0,0 +1,31 @@
...
+internal class CraftedItemValuePatch
+{
+    [HarmonyPatch(nameof(ItemObject.DetermineValue))]
+    [HarmonyPostfix]
+    public static void DetermineValuePostfix(ItemObject __instance)
+    {
+        if (__instance.IsCraftedByPlayer)
+        {
+            var multiplier = ModConfigProvider.ModOptions.SmithingCraftedItemsValueMultiplier;
+            __instance.Value = ApplyMultiplier(__instance.Value, multiplier);
```
ShoT-UPfps, line 17: This only runs when the player-crafted flag is already set, but `CraftingCampaignBehavior.InitializeCraftedItemData` reconstructs a saved weapon through `InitializePreCraftedWeaponOnLoad` / `GenerateItem` (which calls `DetermineValue`) before marking it player-crafted. There is also no authoritative multiplier during cold-start reconstruction: `ModConfigProvider` begins at 1.0, and [LoadModConfigHandler](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/blob/ea235069c7bf93db0e8f0c931186d8015740b890/source/GameInterface/Services/CampaignService/Handlers/LoadModConfigHandler.cs#L45-L60) applies the server configuration only at `CampaignReady`; receiving it on a client only replaces the options and publishes `ModConfigApplied`. No existing-item repricing follows.

A non-default discount can therefore disappear from saved weapons after restarting/loading, while a late-joining client can reconstruct an unscaled value for an item that remains scaled on the server. Recalculate already initialized player-crafted items from their base model value after the authoritative options are applied on each peer, and add a save/restart/rejoin regression with a non-default multiplier. Keep this idempotent rather than multiplying an already scaled `Value`.

ShoT-UPfps (changes requested):
Reviewed head `ea235069c7bf93db0e8f0c931186d8015740b890` against base `c3d9ef041dbddd4020ab17cbec62d81369e5ec8f`.

## Finding

**[P2] Reapply crafted-item valuation after save reconstruction and authoritative config loading.** The new postfix only scales an item while `IsCraftedByPlayer` is already true. The installed game's saved-item reconstruction regenerates each crafted weapon, including `DetermineValue`, before calling `InitAsPlayerCraftedItem`. On a cold process start the configured multiplier is also unavailable at that point: `ModConfigProvider` initially supplies 1.0, and `LoadModConfigHandler` does not load the server's options until `CampaignReady`, published after `MapScreen.OnInitialize`. A joining client similarly receives the options after loading. Neither the config-load nor config-receive path recalculates existing crafted items. Consequently, a weapon created with a non-default multiplier can revert to its unscaled value after a server restart/load, and joining clients can reconstruct a different value from the running server.

Recompute existing player-crafted items from their unmodified model value once their initialization and authoritative configuration are complete, on both server and client; do not multiply an already adjusted value. Add regression coverage for crafting with 0.1, saving, restarting the host, and joining/rejoining a client. The existing weapon and a newly crafted equivalent should retain the same configured valuation, with no repeated scaling and no changes to non-player-crafted items.

## Earlier feedback

The fractional-multiplier arithmetic, NaN/infinity rejection, and integer conversion bounds are corrected in `ApplyMultiplier`. I am not re-raising the previously rejected protobuf-tag compatibility comments: the current tags are unique, and the exact-head client validation explicitly rejects a mismatched co-op build before advancing the join. The normal authoritative crafting path correctly sets the player-crafted flag before generating the item; the remaining issue is the reconstruction/configuration lifecycle rather than that creation path.

## Verification

Read the complete diff, all five changed-file entries, all six commits, all review/comment/thread pages, and relevant exact-head repository code. All 12 reported checks for the reviewed head passed. Inspected and validated matching installed v1.4.8 managed evidence for valuation and saved-item reconstruction. No local build, live-game reproduction, or restart/rejoin test was executed; the finding follows from the inspected code paths. The overview also reported merge conflicts, separate from this correctness finding.

This review applies only to head `ea235069c7bf93db0e8f0c931186d8015740b890` compared from `c3d9ef041dbddd4020ab17cbec62d81369e5ec8f` with base tip `c3d9ef041dbddd4020ab17cbec62d81369e5ec8f` as validated at `2026-09-11T01:34:59.432806Z`. Later revisions and later development/base movement were not evaluated by this review.

- Pull request: https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/3228
- Original verdict: `REQUEST_CHANGES`
- Reviewed head: `ea235069c7bf93db0e8f0c931186d8015740b890`
- Reviewed base tip: `c3d9ef041dbddd4020ab17cbec62d81369e5ec8f`
- Comparison base: `c3d9ef041dbddd4020ab17cbec62d81369e5ec8f`

This review applies only to head `ea235069c7bf93db0e8f0c931186d8015740b890` compared from `c3d9ef041dbddd4020ab17cbec62d81369e5ec8f` with base tip `c3d9ef041dbddd4020ab17cbec62d81369e5ec8f` as validated at `2026-09-11T01:34:59.432806Z`. Later revisions and later development/base movement were not evaluated by this review.

## Exact review summary

```text
Reviewed head `ea235069c7bf93db0e8f0c931186d8015740b890` against base `c3d9ef041dbddd4020ab17cbec62d81369e5ec8f`.

## Finding

**[P2] Reapply crafted-item valuation after save reconstruction and authoritative config loading.** The new postfix only scales an item while `IsCraftedByPlayer` is already true. The installed game's saved-item reconstruction regenerates each crafted weapon, including `DetermineValue`, before calling `InitAsPlayerCraftedItem`. On a cold process start the configured multiplier is also unavailable at that point: `ModConfigProvider` initially supplies 1.0, and `LoadModConfigHandler` does not load the server's options until `CampaignReady`, published after `MapScreen.OnInitialize`. A joining client similarly receives the options after loading. Neither the config-load nor config-receive path recalculates existing crafted items. Consequently, a weapon created with a non-default multiplier can revert to its unscaled value after a server restart/load, and joining clients can reconstruct a different value from the running server.

Recompute existing player-crafted items from their unmodified model value once their initialization and authoritative configuration are complete, on both server and client; do not multiply an already adjusted value. Add regression coverage for crafting with 0.1, saving, restarting the host, and joining/rejoining a client. The existing weapon and a newly crafted equivalent should retain the same configured valuation, with no repeated scaling and no changes to non-player-crafted items.

## Earlier feedback

The fractional-multiplier arithmetic, NaN/infinity rejection, and integer conversion bounds are corrected in `ApplyMultiplier`. I am not re-raising the previously rejected protobuf-tag compatibility comments: the current tags are unique, and the exact-head client validation explicitly rejects a mismatched co-op build before advancing the join. The normal authoritative crafting path correctly sets the player-crafted flag before generating the item; the remaining issue is the reconstruction/configuration lifecycle rather than that creation path.

## Verification

Read the complete diff, all five changed-file entries, all six commits, all review/comment/thread pages, and relevant exact-head repository code. All 12 reported checks for the reviewed head passed. Inspected and validated matching installed v1.4.8 managed evidence for valuation and saved-item reconstruction. No local build, live-game reproduction, or restart/rejoin test was executed; the finding follows from the inspected code paths. The overview also reported merge conflicts, separate from this correctness finding.
```

## Exact inline findings

### 1. [P2] Reprice saved crafted items after initialization and config loading

Original target: `source/GameInterface/Services/ItemObjects/Patches/CraftedItemValuePatch.cs:14-17 (RIGHT)` at `ea235069c7bf93db0e8f0c931186d8015740b890`

Context:

```text
if (__instance.IsCraftedByPlayer)
{
    var multiplier = ModConfigProvider.ModOptions.SmithingCraftedItemsValueMultiplier;
    __instance.Value = ApplyMultiplier(__instance.Value, multiplier);
}
```

Finding:

```text
This only runs when the player-crafted flag is already set, but `CraftingCampaignBehavior.InitializeCraftedItemData` reconstructs a saved weapon through `InitializePreCraftedWeaponOnLoad` / `GenerateItem` (which calls `DetermineValue`) before marking it player-crafted. There is also no authoritative multiplier during cold-start reconstruction: `ModConfigProvider` begins at 1.0, and [LoadModConfigHandler](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/blob/ea235069c7bf93db0e8f0c931186d8015740b890/source/GameInterface/Services/CampaignService/Handlers/LoadModConfigHandler.cs#L45-L60) applies the server configuration only at `CampaignReady`; receiving it on a client only replaces the options and publishes `ModConfigApplied`. No existing-item repricing follows.

A non-default discount can therefore disappear from saved weapons after restarting/loading, while a late-joining client can reconstruct an unscaled value for an item that remains scaled on the server. Recalculate already initialized player-crafted items from their base model value after the authoritative options are applied on each peer, and add a save/restart/rejoin regression with a non-default multiplier. Keep this idempotent rather than multiplying an already scaled `Value`.
```

## Exact thread actions

None.

jordanbrymora:
Not worth it

PR #3239, opened 2026-08-22
Log client and server FPS

### LINKED ISSUE #1847: [ISSUE] Log fps on server and client
~~add to packet profiler~~

G07cha:
## PR Checklist

## PR Type

FPS is not recorded in the logs

## What is the new behavior?

~~- New `coop.debug.fps.log` command~~
- Works on client and server, logs average, min, and max FPS every 30 seconds

Resolves #1847

## Bot Changelog Entry

- Log client and server FPS at regular intervals

## Other information

I've considered two other options:
- Always on FPS logging at the regular intervals - ~~pollutes logs and likely isn't helpful on it's own in identifying performance issues without additional debugging enabled~~ In the end we went for this one
- Logging if FPS drops - similar to the above but also much more complex than a plain logging with user's hardware having an additional influence


ShoT-UPfps:
I think we'll want to do this:

"Always on FPS logging at the regular intervals - pollutes logs and likely isn't helpful on it's own in identifying performance issues without additional debugging enabled"

It can be irregular like once every 30 seconds.

This is not really for helping identify performance issues as much as validating what the user's are claiming. Users will claim they have low fps and then have like 60 fps. That's what this is for

ShoT-UPfps (changes requested):
cmt

G07cha:
@ShoT-UPfps makes sense, updated the implementation, let me know what you think.

garrettluskey (changes requested):
Instead make this a low FPS threshold of 10-20 FPS and log once when that threshold is crossed

G07cha:
@garrettluskey this contradicts initial feedback from @ShoT-UPfps in https://github.com/Bannerlord-Coop-Team/BannerlordCoop/pull/3239#issuecomment-5381787682. I can also incorporate both, just want to make sure we're not going back and forth on it.

ShoT-UPfps:
> Instead make this a low FPS threshold of 10-20 FPS and log once when that threshold is crossed

I don't think this measurably impacts performance? Not sure why we need the low FPS threshold. Personally I'd want to always be able to see the client FPS because users tend to smoke crack and act like their FPS is much lower than it is. I'd rather be able to say "your log shows you had 60 fps" than "your log is missing our low-fps log so you must have had more than 20fps"

source/Common/FpsLogger.cs
```diff
@@ -0,0 +1,93 @@
...
+        windowFrames = 0;
+        windowSeconds = 0d;
+        minimumFps = double.MaxValue;
+        maximumFps = double.MinValue;
+    }
+
+    private static void LogReport(FpsWindowStats stats)
+    {
+        Logger.Information(
+            "[Fps] frames={Frames} seconds={Seconds:0.00} avg={Average:0.0} min={Minimum:0.0} max={Maximum:0.0}",
```
ShoT-UPfps, line 68: the PR's documented behavior says each 30-second report contains current, minimum, and average FPS, but this payload emits average, minimum, and maximum. `FpsWindowStats` never retains the last/current sample, so `max` cannot serve as the advertised current value and readers get a different signal than promised. either retain and log the current/last-frame FPS with regression coverage, or change the stated behavior to explicitly be min/max/average.

ShoT-UPfps (commented):
rereviewed `e32e36ae` against live development `6b7a8215`. the earlier always-on 30-second request is implemented and Build and Test #5084 is green, but product direction is now unresolved because the later threshold-only review contradicts that request. independently, the advertised current/min/average report actually emits min/max/average.

source/Common/FpsLogger.cs
garrettluskey, line 68: this logs maximum FPS, but the PR says each report includes current FPS. the last sample isnt retained, so this cant provide that value. track and log the last-frame FPS with a test, or change the PR contract to min/max/average.

garrettluskey (changes requested):
`FpsLogger` is initialized after the shared logger, runs on the application thread for both roles, and the focused tests/netstandard build pass. requesting changes because the emitted statistic still doesnt match the PR's current/min/average contract.

source/Common/FpsLogger.cs
G07cha: Updated PR description to accurately reflect the changes. Capturing the current FPS seems pointless since there is too much time distance between logging points.

G07cha: Updated PR description to accurately reflect the changes. Capturing the current FPS seems pointless since there is too much time distance between logging points.

ShoT-UPfps (approved):
lgtm

garrettluskey (approved):
the description now matches the min/max/average report, and the logger is initialized after logging setup for both roles. common tests pass.

gamingmenacers-svg (approved):
the 30-second window reports frame-weighted average plus min and max instantaneous FPS, initializes after logging, and has focused reset and invalid-frame coverage. no code finding on e32e36a; the branch is 82 development commits behind and must be rebased and rerun before merge.

PR #3290, opened 2026-08-23
Armytracker

NatteTosti69:
## PR Checklist

- [ x ] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [ x ] New feature (non-breaking change that adds functionality)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Players couldn't see the maptracker of an army in their kingdom

## What is the new behavior?

Resolves #

Players see a maptracker on the armies in their kingdom
Also works for late joining clients.

## Other information
1: `MapTrackerProvider.ResetTrackers()` gets called before the mainhero is properly assigned, which causes late joining clients to always evaluating `army.Kingdom == Hero.MainHero.MapFaction` false.
2: `OnArmyCreated` was never sent to clients, thus it would never be added.

### DIFF
--- a/source/GameInterface/Services/Armies/Messages/NetworkArmyFullyCreated.cs
+++ b/source/GameInterface/Services/Armies/Messages/NetworkArmyFullyCreated.cs
@@ -4,7 +4,7 @@ using ProtoBuf;
 namespace GameInterface.Services.Armies.Messages;
 
 [ProtoContract(SkipConstructor = true)]
-public readonly struct NetworkArmyFullyCreated : ICommand
+public readonly struct NetworkArmyFullyCreated : IServerToClientCommand
 {
     [ProtoMember(1)]
     public readonly string ArmyId;
--- a/source/GameInterface/Services/Armies/Messages/NetworkArmyFullyCreated.cs
+++ b/source/GameInterface/Services/Armies/Messages/NetworkArmyFullyCreated.cs
@@ -1,0 +1,16 @@
+﻿using Common.Messaging;
+using ProtoBuf;
+
+namespace GameInterface.Services.Armies.Messages;
+
+[ProtoContract(SkipConstructor = true)]
+public readonly struct NetworkArmyFullyCreated : ICommand
+{
+    [ProtoMember(1)]
+    public readonly string ArmyId;
+
+    public NetworkArmyFullyCreated(string armyId)
+    {
+        ArmyId = armyId;
+    }
+}
--- a/source/GameInterface/Services/UI/Notifications/Patches/DefaultNotificationsCampaignBehaviorPatches.cs
+++ b/source/GameInterface/Services/UI/Notifications/Patches/DefaultNotificationsCampaignBehaviorPatches.cs
@@ -255,13 +255,13 @@ internal class DefaultNotificationsCampaignBehaviorPatches
     }
 
     [HarmonyPatch(nameof(DefaultNotificationsCampaignBehavior.OnArmyCreated))]
-    [HarmonyPostfix]
-    public static void OnArmyCreatedPostfix(ref DefaultNotificationsCampaignBehavior __instance, Army army)
+    [HarmonyPrefix]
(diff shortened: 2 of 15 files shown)


source/GameInterface/Services/UI/Notifications/Patches/DefaultNotificationsCampaignBehaviorPatches.cs
```diff
@@ -255,13 +255,13 @@ public static void OnClanChangedFactionPostfix(ref DefaultNotificationsCampaignB
...
 
     [HarmonyPatch(nameof(DefaultNotificationsCampaignBehavior.OnArmyCreated))]
-    [HarmonyPostfix]
-    public static void OnArmyCreatedPostfix(ref DefaultNotificationsCampaignBehavior __instance, Army army)
+    [HarmonyPrefix]
+    public static bool OnArmyCreatedPrefix(ref DefaultNotificationsCampaignBehavior __instance, Army army)
     {
-        if (ModInformation.IsClient) return;
-
+        if (ModInformation.IsClient) return false;
```
ShoT-UPfps, line 262: this suppresses vanilla army-created notifications after co-op teardown. `GameInterface.UnpatchAll()` is intentionally a no-op, while the container-scoped `DefaultNotificationsHandler` unsubscribes during teardown; once a client has joined co-op, this still-installed prefix keeps returning `false` for every later singleplayer `OnArmyCreated` call, so v1.4.8's vanilla quick notification has no remaining replacement. please only skip vanilla while the active co-op replacement path exists (for example, while a co-op container/session is present), and add coverage for both active co-op and post-teardown/no-container behavior.

ShoT-UPfps (commented):
reviewed `4fbc674d` against live development `843d3efd`. the live and late-join tracker paths match v1.4.8's event/reset behavior, and Build and Test #5112 is green, but one P2 lifecycle regression remains: the new unconditional `OnArmyCreated` prefix outlives the co-op notification handler and suppresses vanilla army-created notifications in a later singleplayer campaign. the 17-commit development drift does not touch the PR's 13 changed files.

source/GameInterface/Services/Armies/Messages/NetworkArmyFullyCreated.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Common.Messaging;
+using ProtoBuf;
+
+namespace GameInterface.Services.Armies.Messages;
+
+[ProtoContract(SkipConstructor = true)]
+public readonly struct NetworkArmyFullyCreated : ICommand
```
ShoT-UPfps: this packet is only emitted by the authoritative server after `Kingdom.CreateArmy` finishes, and its receive handler replays `CampaignEventDispatcher.Instance.OnArmyCreated(army)`. declaring it as plain `ICommand` leaves the reverse direction open: `MessagePacketHandler` rejects inbound traffic on the server only for `IServerToClientCommand`, so a modified client can send any registered army id, reach `ArmyHandler` on the server, and replay the full `ArmyCreated` event without an army being created. that immediately republishes the army-created notification to clients through this PR's server prefix and also reruns every other registered server listener. make this implement `IServerToClientCommand` and add a forged-client server-receipt regression proving it is dropped before broker dispatch.

ShoT-UPfps (commented):
rereviewed `10c483c` against live development `45785371`. the lifecycle fix is correct: the army-created prefix now allows vanilla whenever no active co-op `IGameInterface` can be resolved, and Build and Test #5128 passes with active-container, no-container, and post-clear coverage, so the prior thread can be resolved. one P1 trust-boundary blocker remains: `NetworkArmyFullyCreated` is emitted only by the server but implements plain `ICommand`, so a modified client can send it back with any registered army id and make the server replay `CampaignEventDispatcher.OnArmyCreated`, including another army-notification broadcast. mark it `IServerToClientCommand` and cover forged server receipt. the 25-commit development drift does not overlap the reviewed tracker or message paths.

ShoT-UPfps (approved):

PR #3291, opened 2026-08-23
Potential fix to the hero meeting syncronization

codebydean:
## PR Checklist

- [x] All new classes have class-level documentation comments, if there are any at all
- [x] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update
- [ ] Other... Please describe:

## What is the current behavior?

Hero meeting sync can lose access to the local player hero after `Hero.SetHasMet` executes, preventing the correct `PlayerMetHero` message from being published.

## What is the new behavior?

Resolves #3231

The local player hero is captured before `Hero.SetHasMet` executes and reused afterward when publishing the `PlayerMetHero` message. The server continues to skip this client-side synchronization.

## Bot Changelog Entry

- Fixed some hero meetings not synchronizing correctly.

## Other information

Added tests covering the hero meeting patch behavior.

PR #3312, opened 2026-08-24
Add hero meeting persistence regression coverage

gamingmenacers-svg:
## PR Checklist

## PR Type

- [x] Other: regression test coverage

Follow-up to #3291: the patch-level test invokes the postfix directly, so it does not verify that a real client meeting reaches the server persistence layer. The test file also introduced nullable warnings and did not follow the repository's source encoding/line-ending convention.

## What is the new behavior?

Adds a focused end-to-end regression that calls the real patched `Hero.SetHasMet()` method on a client and verifies the exact meeting timestamp is recorded in the server session. It also removes the nullable warnings and normalizes the existing test file.

## Other information

Tested with:

- `dotnet test source\GameInterface.Tests\GameInterface.Tests.csproj -c Release --no-restore --filter FullyQualifiedName~HeroMetPatchesTests`
- `dotnet test source\E2E.Tests\E2E.Tests.csproj -c Release --no-restore --filter FullyQualifiedName~HeroMeetingPersistenceTests`


ShoT-UPfps (approved):
lgtm

PR #3316, opened 2026-08-24
Add opt-in diagnostic bug reports

garrettluskey:
## PR Checklist

## PR Type

The server has no component for collecting client diagnostics and submitting a bug report.

## What is the new behavior?

Server features can request a diagnostic report through `IBugReportService` and identify the trigger. The service collects redacted logs from the server and every connected client that opted in, then posts the report to the configured endpoint as `application/json`.

The JSON includes a schema version, report id, reporting client network id, summary, description, trigger, timestamps, semantic mod version, source commit, full build version, collection counts, the redacted server log, and pseudonymous client logs. Logs use gzip-compressed bytes represented as base64 with their encoding and uncompressed length. Failed uploads retain a quota-limited local ZIP for seven days; cleanup runs at startup and hourly.

Clients get a `Report Coop Bug` button at the top right during map and mission gameplay. The server-side unstuck handler uses the same component as a side effect of `coop.debug.mobileparty.unstuck`.

Log consent is disabled by default, saved per client, and can be changed with `coop.bug_report_log_sharing enable|disable`. The prompt discloses that network ids are published in a public GitHub issue and included logs use publicly accessible links with no guaranteed remote expiry; existing consent is invalidated so clients see the new disclosure. Submitting the in-game form while sharing is disabled opens the privacy prompt; `No thanks` cancels the report. Remote upload is disabled by default and requires `BANNERLORDCOOP_BUG_REPORT_ENDPOINT`, `BANNERLORDCOOP_BUG_REPORT_PUBLISHABLE_KEY`, and a separately provisioned `BANNERLORDCOOP_BUG_REPORT_TOKEN`. The publishable key is used only as `apikey`; the server token is used for bearer authorization.

image

image

Creates these bug reports https://github.com/Bannerlord-Coop-Team/Bannerlord-Coop-Issues/issues/61

## Bot Changelog Entry

- Added an in-game bug-report button with opt-in diagnostic log collection.

## How to test

1. Start a server with the three bug-report endpoint environment variables provisioned, then connect two clients with diagnostic log sharing disabled on one client.
2. Click `Report Coop Bug` at the top right on that client during campaign-map gameplay.
3. Enter summary `Cannot leave Danustica` and description `Leaving Danustica keeps reopening the town menu.`, then submit.
4. Choose `No thanks` in the privacy prompt and confirm the report is discarded.
5. Repeat, choose `Allow`, and confirm the client reports submission and the Supabase function creates the GitHub issue. A failed upload retains the ZIP under `Documents/Mount and Blade II Bannerlord/Coop Bug Reports/Pending`.
6. Run `coop.debug.mobileparty.unstuck` and confirm it starts the same diagnostic collection flow.

Automated coverage:

- `dotnet test source/GameInterface.Tests/GameInterface.Tests.csproj --filter "FullyQualifiedName~CoopLogBugReportTests|FullyQualifiedName~BugReportConsentTests|FullyQualifiedName~BugReportVMTests"`
- `dotnet test source/Coop.IntegrationTests/Coop.IntegrationTests.csproj --filter "FullyQualifiedName~BugReportNetworkMessageSerializationTest|FullyQualifiedName~UnstuckNetworkMessageSerializationTest"`
- `dotnet test source/E2E.Tests/E2E.Tests.csproj --filter "FullyQualifiedName~UnstuckCommandTests"`


source/GameInterface/Services/BugReporting/Handlers/UnstuckBugReportHandler.cs
```diff
@@ -0,0 +1,566 @@
...
+    {
+        if (!ModInformation.IsServer || !(payload.Who is NetPeer requester)) return;
+        if (!playerManager.TryGetPlayer(requester, out var player) ||
+            player.MobilePartyId != payload.What.PartyId)
+        {
+            Logger.Warning("Ignoring unstuck log collection request from an unregistered or mismatched peer");
+            return;
+        }
+
+        StartCollection(requester);
```
ShoT-UPfps: `StartCollection` has no requester/session cooldown, and `ClaimFinalization` clears the sole active slot before `PackageAndUploadAsync` finishes. a regular registered player can therefore loop the unstuck command: every completed round makes all opted-in clients reread/compress their current logs and creates another pending ZIP. the placeholder uploader leaves those ZIPs on disk, and the only cleanup is age-based—there is no pending count/byte quota—so this is a remote CPU/network/disk-exhaustion path. please add server-side throttling/backpressure, keep packaging inside the in-flight bound, enforce a storage quota, and cover repeated requests.

source/GameInterface/Services/BugReporting/BugReportService.cs
```diff
@@ -0,0 +1,566 @@
...
+        if (client.NextChunkIndex != client.ChunkCount) return true;
+        if (client.Data.Length != client.CompressedLength) return false;
+
+        var combinedLength = collection.Clients.Values
+            .Where(item => item.Status == ClientCollectionStatus.Collected)
+            .Sum(item => (long)item.CompressedLength);
+        if (combinedLength + client.CompressedLength > MaximumCombinedCompressedBytes) return false;
+
+        client.Responded = true;
+        client.Status = ClientCollectionStatus.Collected;
```
ShoT-UPfps, line 484: `client.Data` is still peer-controlled compressed bytes here; the checks above validate only framing and declared lengths. a client can send matching-count random bytes and be marked `Collected`, after which `BugReportArchiveBuilder.AddLog` throws while reading the `GZipStream`; `Create` deletes the whole ZIP on that exception. one malformed or hostile peer can therefore discard every valid client's diagnostics. validate/decompress each payload before marking it collected, or skip only that client and count it failed during packaging, and add a mixed valid/invalid-client regression test.

source/GameInterface/Services/BugReporting/BugReportService.cs
```diff
@@ -0,0 +1,566 @@
...
+
+    public UnstuckBugReportHandler(
+        IMessageBroker messageBroker,
+        INetwork network,
+        IPlayerManager playerManager,
+        IUnstuckLogSharingPreference logSharingPreference,
+        ICoopLogSnapshotProvider logSnapshotProvider,
+        IBugReportArchiveBuilder archiveBuilder,
+        IBugReportUploader uploader,
+        CancellationTokenSource sessionCancellation)
```
ShoT-UPfps, line 66: `ServiceModule` auto-activates every `IHandler`, so this constructor makes every `GameInterfaceModule`-only container require a session `CancellationTokenSource`; resolving `ICoopLogSnapshotProvider` also requires the `ICoopLogFile` registration that currently lives only in `CommonModule`/the session composition. Build and Test #5144 already fails `GameInterface.Tests.ContainerTest` on the missing cancellation token, with unrelated tests cascading afterward. move this handler to a client/server composition root or provide the complete shared dependencies wherever it is auto-activated, and keep the container test green.

ShoT-UPfps (commented):
reviewed `a800f8a9` against live development `45785371`. three P1 blockers remain: repeated unstuck requests are unthrottled and can accumulate pending archives while forcing every opted-in client to recapture its log; peer-controlled bytes are marked collected before gzip validation, so one malformed client aborts the entire archive; and the auto-activated handler makes GameInterface-only containers unresolvable. Build and Test #5144 is red in the unit-test job on the missing `CancellationTokenSource` dependency.

source/GameInterface/Services/BugReporting/Handlers/UnstuckBugReportHandler.cs
garrettluskey: fixed, collection stays in flight through packaging/upload, new reports are throttled for five minutes, and pending archives are capped at 20 files / 256 MiB.

source/GameInterface/Services/BugReporting/BugReportService.cs
garrettluskey: fixed, client gzip is decompressed and length-checked off the poll thread before packaging. invalid logs are counted as failed and valid logs still get archived/uploaded.

source/GameInterface/Services/BugReporting/BugReportService.cs
garrettluskey: fixed, `GameInterfaceModule` now supplies preserved fallback registrations for the log path and session cancellation token. `ContainerTest` passes.

source/GameInterface/Services/UI/BugReporting/BugReportConsentCoordinator.cs
```diff
@@ -0,0 +1,98 @@
...
+/// <summary>Displays and persists the client's diagnostic bug-report log-sharing choice.</summary>
+public class BugReportConsentCoordinator
+{
+    public const string TabId = "BugReporting";
+    public const string SectionId = "BugReportLogSharingConsent";
+    public const string PromptTitle = "Share Co-op Diagnostic Logs?";
+    public const string PromptText =
+        "When the dedicated server creates a diagnostic bug report, allow this client's current " +
+        "BannerlordCoop log to be sent to the dedicated server, packaged with logs from other consenting " +
+        "clients, and submitted to BannerlordCoop's bug-report service? Reports may be triggered " +
```
ShoT-UPfps: `Bannerlord-Coop-Issues#61` demonstrates that the configured production endpoint creates a public GitHub issue containing the reporter's network ID and public `r2.dev` links to the server and client logs. this consent says only that data is sent to a service; it does not disclose public publication or remote retention, even though it warns that logs may contain Steam IDs, IP addresses, and file paths and the server log can contain other peers' data. users have not consented to that public exposure. disclose the public issue/log destination and retention before opt-in (or make the logs private and expiring), then cover the exact disclosure in both startup and submit-time consent tests.

source/GameInterface/Services/BugReporting/BugReportUploader.cs
```diff
@@ -0,0 +1,278 @@
...
+        BugReportArchiveContents report,
+        CancellationToken cancellationToken);
+}
+
+/// <inheritdoc />
+public class BugReportUploader : IBugReportUploader, IDisposable
+{
+    public const string Endpoint =
+        "https://wfvqnijwuyqjibhlcrhz.supabase.co/functions/v1/create-github-issue-bug-report";
+    public const string SupabasePublishableKey = "sb_publishable_tseZMeJ-RYeSHI0KwU2p0g_PE4GVtN6";
```
ShoT-UPfps: `SupabasePublishableKey` is intentionally public and is the only credential this request sends (the same value is copied into `apikey` and `Authorization`). the request contains no user JWT, server identity, signature, nonce, or challenge, so from the request itself the Edge Function cannot distinguish an official BannerlordCoop server from any caller who copies this source. the in-game five-minute throttle therefore does not protect the live GitHub/R2 endpoint from direct issue and storage abuse. require an authenticated, rate-limited broker or short-lived server-bound authorization; do not treat a publishable key as origin authentication.

source/GameInterface/Services/BugReporting/BugReportService.cs
```diff
@@ -0,0 +1,761 @@
...
+            {
+                if (activeCollection?.RequestId == finalized.Contents.RequestId)
+                    activeCollection = null;
+            }
+        }
+    }
+
+    private async Task PackageAndUploadCoreAsync(FinalizedCollection finalized)
+    {
+        if (finalized.Contents.Logs.Count == 0 && finalized.Contents.Submissions.Count == 0)
```
ShoT-UPfps: `unstuck` calls `RequestReport` without a submission, and client log sharing defaults disabled. when every client declines, `Logs` and `Submissions` are both empty, so this return runs before `CaptureServerLog`; the default automatic recovery path produces no archive, upload, or server diagnostics at all. capture and package the server log regardless of client opt-in (or explicitly disable this trigger), and add an all-clients-declined unstuck regression.

source/GameInterface/Services/BugReporting/Messages/NetworkBugReportLogChunk.cs
```diff
@@ -0,0 +1,47 @@
+﻿using Common.Messaging;
+using ProtoBuf;
+
+namespace GameInterface.Services.BugReporting.Messages;
+
+/// <summary>
+/// One bounded chunk of a gzip-compressed client co-op log.
+/// </summary>
+[ProtoContract(SkipConstructor = true)]
+internal readonly struct NetworkBugReportLogChunk : ICommand
```
ShoT-UPfps, line 10: this 64 KiB chunk is an ordinary `ICommand`, so `MessagePacket` sends it on reliable-ordered channel 0—the same ordered command/world stream—while `SendSnapshot` synchronously enqueues up to 512 chunks from the game thread. a max-sized consented log can fragment into tens of thousands of reliable packets and hold later client commands behind the upload, precisely when unstuck recovery is running. move diagnostics to the existing bulk lane (or another isolated stream) and pace/backpressure the send; cover that world traffic remains responsive during a large upload.

source/GameInterface/Services/BugReporting/BugReportService.cs
```diff
@@ -0,0 +1,761 @@
...
+
+        if (client.Data.Length + chunk.Data.Length > client.CompressedLength) return false;
+
+        client.Data.Write(chunk.Data, 0, chunk.Data.Length);
+        client.NextChunkIndex++;
+        if (client.NextChunkIndex != client.ChunkCount) return true;
+        if (client.Data.Length != client.CompressedLength) return false;
+
+        var combinedLength = collection.Clients.Values
+            .Where(item => item.Status == ClientCollectionStatus.Collected)
```
ShoT-UPfps: `MaximumCombinedCompressedBytes` does not bound the collection's actual memory: only clients already marked `Collected` are summed, the check happens after the final chunk was written, pending and failed streams remain allocated, and `ClaimFinalization` duplicates every accepted stream with `ToArray` while the active collection stays alive through upload. enough peers can therefore retain far more than 512 MiB before JSON encoding starts. reserve the aggregate budget before every write across every status, clear/dispose rejected streams, and transfer rather than duplicate completed buffers; add a concurrent near-limit regression.

source/GameInterface/Services/BugReporting/BugReportUploader.cs
```diff
@@ -0,0 +1,278 @@
...
+        if (report == null) throw new ArgumentNullException(nameof(report));
+        if (!IsConfigured)
+        {
+            return new BugReportUploadResult(
+                false,
+                false,
+                "The bug-report upload endpoint is not configured.");
+        }
+
+        var json = JsonSerializer.Serialize(CreateRequest(report), JsonOptions);
```
ShoT-UPfps, line 96: `JsonSerializer.Serialize` base64-encodes every allowed `byte[]` into one UTF-16 string, then `StringContent` allocates another UTF-8 copy. at the 512 MiB client allowance (plus the server log and retained stream/array copies), encoding alone is roughly a 1.3 GiB string plus 0.7 GiB request buffer; this can exhaust a normal dedicated server, and the configured hosted Edge Function has a 256 MB worker-memory limit. stream logs directly to object storage/multipart and send only metadata/URLs, or enforce a small end-to-end body limit before collection; test the maximum accepted report.

source/GameInterface/Services/BugReporting/BugReportService.cs
```diff
@@ -0,0 +1,761 @@
...
+        }
+
+        string archivePath = null;
+        string clientMessage;
+        try
+        {
+            var contents = await Task.Run(
+                () => CaptureServerLog(ValidateClientLogs(finalized.Contents)),
+                cancellationToken).ConfigureAwait(false);
+            archivePath = await Task.Run(
```
ShoT-UPfps, line 666: `archiveBuilder.Create` is a mandatory precondition for the configured upload, even though archive retention is described as the fallback for failed uploads. its 256 MiB pending quota is lower than the service's 512 MiB accepted client budget, so a valid large report—or an unwritable/full Documents folder—throws here and never reaches the HTTP endpoint. creating this temporary ZIP can also evict an older failed report before a successful upload immediately deletes the new one. upload independently first (or make archive failure non-fatal) and retain/enforce quota only on the failure copy.

source/Coop/CoopMod.cs
```diff
@@ -634,7 +643,40 @@ private void TryShowCrashReportingConsent(bool isAtMainMenu)
...
+            }
+
+            bugReportConsent.TryShowPrompt(
+                isAtMainMenu && !InformationManager.IsAnyInquiryActive(),
+                inquiry => InformationManager.ShowInquiry(inquiry));
+        }
+
+        private void TryInitializeBugReportOverlay()
+        {
+            if (isServer || Game.Current == null || bugReportOverlay != null ||
```
ShoT-UPfps: `bugReportOverlay` is a scope-owned `IDisposable`: `DestroyContainerCore` disposes it and `BugReportOverlay.Dispose` removes the global layer, but this module field is never cleared. after leaving the first co-op session, the stale non-null reference makes this guard skip resolving and initializing the overlay from every later session, so the Report Coop Bug button disappears until process restart. clear the field on scope teardown or `CoopModeEnded`, and add a leave/rejoin UI lifecycle test.

source/GameInterface/Services/BugReporting/BugReportArchiveBuilder.cs
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    public string Create(BugReportArchiveContents contents)
+    {
+        if (contents == null) throw new ArgumentNullException(nameof(contents));
+        if (!Guid.TryParseExact(contents.RequestId, "N", out _))
+            throw new ArgumentException("Request id must be a compact GUID.", nameof(contents));
+
+        Directory.CreateDirectory(outputDirectory);
+        PruneExpiredArchives();
```
ShoT-UPfps, line 214: the advertised seven-day retention is enforced only when another report later calls `Create`. if the last upload fails—or the server is retired before another report—its ZIP is never revisited and can retain network IDs and diagnostic logs indefinitely; one locked stale file can also abort the whole best-effort sweep. run expiry cleanup at startup and on a periodic timer (handling files independently), and test that an archive disappears after the retention window without requiring a subsequent report.

ShoT-UPfps (commented):
rereviewed `b155ea0b` against live development `1e6b891d`. the three prior blockers are fixed and Build and Test #5162 passes, but seven P1 blockers plus two P2 defects remain: consent omits that network IDs and logs are published through a public issue and public R2 links; the live endpoint is authenticated only by an exposed publishable key; all-opt-out unstuck reports skip even the server log; 64 KiB log chunks flood reliable channel 0 from the game thread; aggregate buffering is not actually bounded and whole-report JSON/base64 materialization exceeds normal server and Edge Function memory; mandatory ZIP creation can prevent the upload; the overlay is not recreated after leave/rejoin; and the claimed seven-day local retention runs only when another report is created.

source/GameInterface/Services/UI/BugReporting/BugReportConsentCoordinator.cs
garrettluskey: confirmed against issue 61. both consent prompts and the form now say the network id goes to a public GitHub issue, logs use publicly accessible links, and remote expiry is not guaranteed. the disclosure version was bumped so existing opt-ins are prompted again, with tests for both consent paths.

source/GameInterface/Services/BugReporting/BugReportUploader.cs
garrettluskey: confirmed. the live endpoint and key are no longer shipped as defaults, and the publishable key cannot be used as bearer auth. upload now requires endpoint, publishable key, and a separately provisioned server bearer token through environment configuration; otherwise it stays local. the broker still needs to validate and rate-limit that token.

source/GameInterface/Services/BugReporting/BugReportService.cs
garrettluskey: fixed, preparation now captures the server log before deciding the report is empty. an all-client-empty report continues when the server log exists, covered by `PrepareContents_WithNoClientLogs_StillCapturesServerLog`.

source/GameInterface/Services/BugReporting/BugReportService.cs
garrettluskey: fixed, the collection reserves each client exact-sized array before its first write against an 8 MiB aggregate cap. rejected/timed-out buffers are released, and accepted arrays are transferred into the report without `ToArray` copies. the exact-limit and one-byte-over cases are covered.

source/GameInterface/Services/BugReporting/BugReportUploader.cs
garrettluskey: fixed by bounding before serialization: 1 MiB compressed per log, 8 MiB combined client logs, and 10 MiB total compressed server+client data. `UploadAsync` refuses one byte over the total before creating the JSON string, with boundary coverage.

source/GameInterface/Services/BugReporting/BugReportArchiveBuilder.cs
garrettluskey: fixed, archive cleanup now runs when the builder starts and every hour, and deletion failures are isolated per file. tests cover timer-driven expiry without another `Create` and a locked stale file not blocking deletion of another.

source/GameInterface/Services/BugReporting/BugReportService.cs
garrettluskey: fixed, upload now runs before any ZIP work. a successful upload creates no archive; the quota-managed ZIP is attempted only when authorization is unavailable or upload fails, and archive failure no longer prevents the already-attempted HTTP request.

source/Coop/CoopMod.cs
garrettluskey: fixed, initialization now resolves the current scope instance first and skips only when it is the same object as the cached overlay. a replacement scope returns a different instance, so leave/rejoin initializes the new overlay.

source/GameInterface/Services/BugReporting/Messages/NetworkBugReportLogChunk.cs
garrettluskey: confirmed. diagnostics stay on the ordinary message stream, but the per-log cap is now 1 MiB compressed, chunks are 4 KiB, the aggregate cap is 8 MiB, and the sender blocks on one game-thread send per frame instead of enqueueing the whole log at once. this bounds and interleaves the channel-0 traffic without adding special message routing.

gamingmenacers-svg (approved):
reviewed 86cf4eb through the consent, collection, chunk validation, redaction, archive quota, authenticated upload, fallback, UI lifecycle, and automatic-report paths. the earlier blockers are addressed, the full hosted matrix is green, and I found no remaining actionable delta against current development.

ShoT-UPfps (approved):

Issue #1827, opened 2026-07-11
[ISSUE] Crash when joining another player's kingdom


jordanbrymora:
# For New Contributors
[How to contribute](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/wiki/Getting-Started-as-a-Contributor)

## Description

When talking to another player to join their kingdom and after taking the oath, the joining client hard crashes.

## Location

Create a branch based from [development](https://github.com/Bannerlord-Coop-Team/BannerlordCoop)

## Definition of Done
- [ ] Class level comments exist for all new classes.
- [ ] XUnit tests exist for every method that does not require the game to be ran.

EgardA:
I am not seeing this behaviour when testing it out locally

jordanbrymora:
> I am not seeing this behaviour when testing it out locally

Sorry I forgot to mention this happens specifically when talking to another player in a settlement. Seems to be fine when just interacting with another player's party.

casey-johnson:
I tested this a bit and was unable to reproduce. I tried:
Joining Player1's kingdom from Player2 through dialogue while sitting in Syronea on both clients, but through the keep
The same as above, except while taking a walk through town twice (different clients being the host)

I can only assume this was fixed already. I'll have a regression test generated, and then close the issue once that is merged.

PR #3317, opened 2026-08-24
[1827] Add regression test for player->player vassal oath in settlements

casey-johnson:
issue #1827

## PR Checklist

## PR Type

- [x] Other... Please describe:
Added a regression test for issue https://github.com/Bannerlord-Coop-Team/BannerlordCoop/issues/1827 as I was not able to reproduce the issue. It seems to have been fixed.

### DIFF
--- a/source/E2E.Tests/Services/Kingdoms/PlayerKingdomCreationFlowTests.cs
+++ b/source/E2E.Tests/Services/Kingdoms/PlayerKingdomCreationFlowTests.cs
@@ -1,4 +1,5 @@
 ﻿using Common;
+using Common.Network;
 using Common.Messaging;
 using Common.Util;
 using Coop.Core.Client.Services.Kingdoms.Handlers;
@@ -20,6 +21,8 @@ using GameInterface.Services.Kingdoms.Data;
 using GameInterface.Services.Kingdoms.Extentions;
 using GameInterface.Services.Kingdoms.Messages;
 using GameInterface.Services.Kingdoms.Patches;
+using GameInterface.Services.Locations.Messages.Conversation;
+using GameInterface.Services.MapEvents.Messages.Conversation;
 using GameInterface.Services.MobileParties.Extensions;
 using GameInterface.Services.MobileParties.Handlers;
 using GameInterface.Services.MobileParties.Messages.Behavior;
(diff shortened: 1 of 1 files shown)


source/E2E.Tests/Services/Kingdoms/PlayerKingdomCreationFlowTests.cs
```diff
@@ -287,6 +290,55 @@ public void VassalServiceAccepted_JoinsPlayerClanAuthoritativelyAndRejectsReplay
...
+        var kingdomId = TestEnvironment.CreateRegisteredObject<Kingdom>();
+        ConfigureClanInKingdom(ruler.ClanId, kingdomId);
+        SetClanTierEverywhere(joiner.ClanId, 2);
+
+        Server.Call(() => Server.Resolve<IPlayerManager>().SetPeer(ControllerId, client.NetPeer));
+
+        // The joiner walks up to the ruler's hero inside a settlement and opens a location conversation,
+        // exactly as LocationConversationPatches gates it for a synced hero met in a settlement.
+        Server.NetworkSentMessages.Clear();
+        client.Call(() => client.Resolve<INetwork>().SendAll(
```
ShoT-UPfps: this never enters the settlement-dialogue path reported in #1827. `LocationConversationPatches.OnAgentInteractionPrefix` resolves the real mission location and target, stores `pending`, and publishes this request; sending the packet directly leaves `pending` null, so the approval callback returns before `MissionConversationLogic.StartConversation`. the later direct `VassalServiceAccepted` also bypasses `VassalServiceConversationPatch.ConversationPlayerIsAcceptedAsVassalPrefix`, the oath consequence that suppresses vanilla and routes the join through the server. this is therefore the existing vassal-service happy path plus a server tracker reservation and would stay green if the actual settlement oath crashed. drive those production entry points (or a mission fixture), end through the real conversation callback, and prove the same target can be acquired again; `NetworkPlayerInteractionEnded` comes from a separate waiting-party dictionary and does not by itself prove the lock was released.

ShoT-UPfps (commented):
reviewed `995c1e7a` against live development `45785371`; Build and Test #5146 passed all ten jobs. one P2 validation gap remains: the new case sends the location request and vassal-acceptance message directly, so it never starts a settlement conversation or executes the patched oath consequence. it only composes two existing message-level happy paths and would pass if #1827's actual settlement dialogue crash returned.

garrettluskey (changes requested):
the added E2E setup is readable and the existing vassalage assertions are useful. requesting changes for the unresolved coverage gap in `discussion_r3845712201`; this still needs to exercise the settlement conversation and oath entry points before it can protect #1827.

garrettluskey (dismissed):
the updated test now drives both production prefixes, verifies the settlement lock is held through the oath, and proves the same target can be acquired after the conversation ends. the focused E2E passes locally.

gamingmenacers-svg (dismissed):
the revised test now exercises both production prefixes, holds the real settlement conversation lock through the oath, proves kingdom replication, and reacquires the same target after the conversation ends. no test-logic finding on 34ca06d; the branch still needs its current development conflict resolved without weakening those entry-point assertions.

ShoT-UPfps (approved):
lgtm

PR #3323, opened 2026-08-24
Add settlement location E2E test environment

garrettluskey:
## PR Checklist

## PR Type

Mission E2E networking does not model mesh membership, packet delivery, or latency, and location tests cannot compose isolated missions for multiple clients.

## What is the new behavior?

Adds deterministic virtual-time mission mesh routing and a settlement test environment with one isolated mission per client. Location NPC authority and conversation state are scoped per client, and settlement hero population reconciliation includes the co-op player roster.

How to test:

- `dotnet build source/E2E.Tests/E2E.Tests.csproj -c Debug -p:PostBuildEvent=`
- Run the focused location, mission mesh, conversation, and NPC gate tests.


source/E2E.Tests/Environment/Mock/MockBattleNetwork.cs
```diff
@@ -86,29 +105,50 @@ public void SendAllBut(string controllerId, IMessage message)
...
+        if (RoutePackets)
+            router.SendAll(this, packet);
+    }
+
     public void SendAll(IPacket packet, byte[] serializedPacket)
     {
         NetworkSentPackets.Add(packet);
         SerializedPacketSends.Add(new SerializedPacketSend(null, packet, serializedPacket));
+        if (RoutePackets)
+            router.SendAll(this, packet);
```
gamingmenacers-svg: can these serialized overloads route `serializedPacket` through the receiver's real deserialize/restore path instead of reserializing `packet`? the real mesh sends these exact bytes, but this mock discards them; the new test even passes `{ 1, 2, 3 }` and still delivers a valid `MovementPacket`, so malformed/stale compression payloads and wire-size behavior can pass E2E.

gamingmenacers-svg (commented):
the production/state changes look consistent with the v1.4.8 location behavior. i found one mesh-fidelity hole in the new packet router:

source/GameInterface/Services/Locations/SettlementPopulationTracker.cs
```diff
@@ -144,12 +146,11 @@ private void Handle_SettlementRosterHeroesChanged(MessagePayload<SettlementRoste
...
             if (settlement.LocationComplex == null) return;
 
+            var ambientCandidates = new HashSet<Hero>(heroSpawnPool.GetAmbientCandidates(settlement));
             var refreshed = false;
             foreach (var hero in heroes)
             {
-                if (hero == null) continue;
-                // Player heroes are the player agents themselves and are never placed as roster NPCs.
-                if (PlayerManager.TryGetControlledObjectInfo(hero, out _)) continue;
+                if (hero == null || !ambientCandidates.Contains(hero)) continue;
```
ShoT-UPfps: `ambientCandidates` cannot gate a change-event cleanup: it deliberately removes heroes once they belong to a player-controlled party. if a companion already has a dungeon/location roster entry and is released back into a player's party, this skips the reconciliation triggered by the prisoner-change event and leaves the stale roster entry in server state. `BroadcastRosterSnapshot` filters only `IsPlayerHero`, so a later snapshot can still send that companion while the mission mesh also spawns it. process affected heroes that may need removal, or explicitly remove an existing roster entry when a hero ceases to be ambient, and add a prisoner-to-player-party regression.

ShoT-UPfps (commented):
reviewed `41b3fd1b`. the existing serialized-payload thread already covers the mesh wire-fidelity defect. one additional production blocker remains: roster-change reconciliation drops heroes as soon as they enter a player-controlled party, so an existing settlement roster entry can outlive the transition and later duplicate the companion spawned by the mission mesh.

source/E2E.Tests/Environment/Mock/MockBattleNetwork.cs
garrettluskey: fixed in `d89b5c85b`. the serialized overloads now copy and deliver the supplied bytes through receiver deserialization and compressed-packet restore. the regression sends an invalid compressed envelope and confirms it is discarded.

source/GameInterface/Services/Locations/SettlementPopulationTracker.cs
garrettluskey: fixed in `f5b2fd32c`. affected heroes that are no longer ambient now have any existing roster entry removed before the snapshot, with a prisoner-to-player-party regression.

source/E2E.Tests/Environment/MeshNetworkRouter.cs
```diff
@@ -7,47 +8,172 @@
...
-/// client-to-client with no server, so this knows only about clients and addresses them by their
-/// controller id. Delivery serializes the message first (matching the real wire) and publishes it on the
-/// receiving client's broker, exactly as a received mesh packet would.
+/// Routes mission-mesh traffic between active members of the same in-process mission instance. Delivery uses
+/// a virtual-time queue and serialized copies so it follows the same process and wire boundaries as the real mesh.
 /// </summary>
 public class MeshNetworkRouter
 {
-    private readonly List<(ClientInstance instance, MockBattleNetwork mesh)> clients = new();
+    private const string MessageChannel = "message";
```
ShoT-UPfps: messages and reliable packets are not separate ordered streams on the real mesh. `LiteNetP2PClient.Send(IMessage)` wraps the message in `MessagePacket` (`ReliableOrdered`), and its production `netPeer.Send(data, method)` call uses LiteNetLib channel 0; `AgentActionPacket` is also `ReliableOrdered` on that same channel. this `"message"` key is therefore distinct from `"packet:ReliableOrdered"` only in the mock. because `VirtualNetworkScheduler` clamps due times per key, lowering latency after an earlier message lets a later reliable packet arrive first, an ordering production cannot produce. use the shared reliable-ordered stream key and add a mixed message/packet latency regression.

ShoT-UPfps (commented):
rereviewed `f5b2fd32`. both prior findings are fixed: serialized overloads now deliver the supplied bytes through receiver deserialization/restore, and non-ambient player-party companions remove stale roster entries with regression coverage. one mesh-ordering blocker remains: `IMessage` traffic is scheduled outside the `ReliableOrdered` packet stream, so a latency reduction can reorder traffic that production sends on LiteNetLib channel 0. the current-head workflow build passed; run #5220 is red on the `CallOriginalPolicy` thread-scope unit test outside this PR's diff.

source/E2E.Tests/Environment/MeshNetworkRouter.cs
garrettluskey: confirmed against `LiteNetP2PClient` and LiteNetLib 1.3.1: both sends use `ReliableOrdered` channel 0. fixed in `433891980` by sharing the scheduler stream key, with a mixed message/action-packet latency regression.

ShoT-UPfps (commented):
rereviewed `f4717e6b`. the prior reliable-ordering finding is fixed and its mixed message/packet latency regression is correct. Build and Test #5256 passed all ten jobs on the last synthetic merge. development advanced afterward; GitHub now reports this head dirty/unmergeable, with the new base overlapping six E2E harness/test files. no new production-code correctness finding, but resolve the conflicts deliberately and rerun the full workflow before merge.

ShoT-UPfps (approved):
lgtm

PR #3324, opened 2026-08-24
Cover deterministic mesh transport edges

garrettluskey:
## PR Checklist

## PR Type

- [x] Other: test coverage

The deterministic mission mesh test infrastructure does not cover cancellation, channel ordering, latency overrides, unknown recipients, or non-empty wire payloads at its edges.

## What is the new behavior?

Adds focused scheduler and mesh routing coverage. This stacks on #3323.

How to test:

- `dotnet test source/E2E.Tests/E2E.Tests.csproj -c Debug -p:PostBuildEvent= -p:NuGetAudit=false --filter "FullyQualifiedName~VirtualNetworkSchedulerTests|FullyQualifiedName~BattleMeshNetworkTests"`
- `dotnet build source/E2E.Tests/E2E.Tests.csproj -c Debug -p:PostBuildEvent= --no-restore`


source/E2E.Tests/Services/Missions/BattleMeshNetworkTests.cs
```diff
@@ -84,6 +84,182 @@ public void Mesh_LatencyDefersDelivery_AndStopCancelsPendingTraffic()
...
+        packetManager.RegisterPacketHandler(receiver);
+
+        try
+        {
+            byte[] packetPayload = { 2, 4, 6, 8 };
+            byte[] serializedPayload = { 9, 7, 5, 3 };
+            var packet = new CompressedMovementPacket(12, packetPayload);
+            MockBattleNetwork sender = Mesh(clients[0]);
+
+            clients[0].Call(() => sender.SendAll(packet, serializedPayload));
```
gamingmenacers-svg, line 250: this test currently proves the opposite of the name: `serializedPayload` is deliberately unrelated to `packet`, but the receiver still gets a clone of `packet` because the parent mock discards the supplied bytes. please route/deserialise the exact payload and make this assert the receiver's wire result; as written it locks in the gap called out on #3323 and would let broken pre-serialization pass.

gamingmenacers-svg (commented):
one actionable test-fidelity issue on this stack delta:

ShoT-UPfps (approved):
lgtm

source/E2E.Tests/Services/Missions/BattleMeshNetworkTests.cs
garrettluskey: fixed in `b25ed50c8`; the test now sends distinct logical and wire packets and asserts the receiver gets the exact deserialized wire length and payload.

PR #3331, opened 2026-08-25
Preserve settlement agent replication fidelity

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Stacks #3323. Settlement agent replication only has count-level coverage, and non-default action speed falls back to 1x on puppets.

## What is the new behavior?

Settlement NPC and companion tests now cover identity, authority, transforms, movement, actions, virtual timing, deduplication, and despawn. Action channel speed is appended to the wire data and applied for both transitions and unchanged action indices.

## Other information

Tests:
- `dotnet restore source/E2E.Tests/E2E.Tests.csproj -p:NuGetAudit=false`
- `dotnet build source/E2E.Tests/E2E.Tests.csproj -c Debug --no-restore --nologo -v:minimal`
- `dotnet test source/E2E.Tests/E2E.Tests.csproj -c Debug --no-build --no-restore --filter FullyQualifiedName~SettlementAgentReplicationTests --logger "console;verbosity=minimal"`
- `dotnet test source/E2E.Tests/E2E.Tests.csproj -c Debug --no-build --no-restore --filter "FullyQualifiedName~AgentActionPacketSerializationTests|FullyQualifiedName~BattleBlockingSyncTests|FullyQualifiedName~RapidGuardTransitionSyncTests" --logger "console;verbosity=minimal"`
- `dotnet test source/E2E.Tests/E2E.Tests.csproj -c Debug --no-build --no-restore --filter "FullyQualifiedName~SettlementTestEnvironmentTests|FullyQualifiedName~SettlementAgentReplicationTests" --logger "console;verbosity=minimal"`
- `git diff --check`


source/Missions/Agents/Handlers/AgentActionHandler.cs
```diff
@@ -217,6 +220,8 @@ private void PollAgentAction(
 
         int action0 = agent.GetCurrentAction(0).Index;
         int action1 = agent.GetCurrentAction(1).Index;
+        float action0Speed = AgentActionData.GetCurrentActionSpeed(agent, 0);
+        float action1Speed = AgentActionData.GetCurrentActionSpeed(agent, 1);
```
gamingmenacers-svg: can these skeleton reads be gated to action classes whose speed we may actually publish, and can ambient speed updates be debounced/quantized? `PollActionsAfterNativeTick` visits every locally authoritative agent every rendered frame, so this adds two `GetSkeleton`/native speed calls per battle agent; for location NPCs, any >0.001 speed jitter now also emits ReliableOrdered packets because `locationAmbientAgent` accepts every non-point action, including locomotion. a many-NPC test with small speed variation should assert bounded packet output.

gamingmenacers-svg (commented):
the speed propagation works in the focused case, but it broadens the hot path enough to need one guardrail:

ShoT-UPfps (commented):
reviewed `06431833` head-only. the nullable action-speed fields, 1x fallback for absent fields, transition/unchanged-index application, and settlement replication coverage are otherwise consistent with the installed v1.4.8 `Agent.SetActionChannel` and skeleton APIs. one P2 hot-path issue remains: ambient packet jitter is now bounded, but every active non-point settlement action still traverses visuals/skeleton and performs a native speed read every display tick; the new 24-NPC test codifies 2,880 reads over 120 frames while sending zero packets. sample ambient speed on a bounded cadence and assert a bounded read count. Build and Test #5306 built successfully and passed all eight E2E shards; the exact-head run is red only on the untouched `GameInterface.Tests.Policies.CallOriginalPolicyTests.CurrentOperationScope_IsLimitedToInvokingThread`.

source/Missions/Agents/Handlers/AgentActionHandler.cs
```diff
@@ -623,6 +677,81 @@ public void HandlePacket(NetPeer peer, IPacket packet)
...
+        float previousSpeed,
+        out bool speedChanged)
+    {
+        if (!publishActionSpeed)
+        {
+            speedChanged = false;
+            return 1f;
+        }
+
+        float speed = AgentActionData.GetCurrentActionSpeed(
```
ShoT-UPfps, line 741: The new 0.05 hysteresis bounds packet emission, but this call still traverses `AgentVisuals -> Skeleton -> GetAnimationSpeedAtChannel` on every display tick for every publishable ambient channel. `locationAmbientAgent` makes every active non-point settlement action publishable, including generic `Other`/`Idle` classes, and the added stress test explicitly expects `24 * 120 = 2,880` getter calls while zero packets are emitted. That leaves the original CPU/native-crossing cost linear in NPC count and render FPS even after the wire flood is fixed. Poll ambient speed on a bounded cadence (or exclude locomotion/idle where speed is movement-owned), and change the stress test to cap getter calls as well as packets.

garrettluskey: fixed in `71ff5ebeb`; unchanged ambient speeds sample every 0.25 seconds, while transitions, missing baselines, and discrete actions still read immediately. the stress test now asserts 192 reads instead of 2,880 and zero jitter packets.

source/Missions/Agents/Handlers/AgentActionHandler.cs
garrettluskey: fixed across `064318336` and `71ff5ebeb`; speed reads are gated, ambient jitter uses a separate threshold, and unchanged ambient actions now sample on a bounded cadence with many-NPC packet and read-count coverage.

ShoT-UPfps (commented):
rereviewed `71ff5ebe` head-only. the prior P2 is fixed: unchanged settlement ambient actions now sample rendered speed every 0.25 seconds, while action transitions, first observations, and discrete actions still read immediately. the 24-NPC/120-frame regression now bounds native reads at 192, emits no jitter packets, and batches a material speed change into three eight-agent packets. the branch's `f3b1f116` development merge preserves the action-speed wire/apply changes and the mock-engine coverage. Build and Test #5336 passed build, unit tests, and all eight E2E shards. no actionable correctness findings on this head.

ShoT-UPfps (approved):

PR #3339, opened 2026-08-25
Suppress unnecessary "failed to get id for item" errors

jordanbrymora:
## PR Checklist

## PR Type

`TradeHandler.TryResolveItemRosterId` outputs an error when it fails to resolve the id of an item. The problem is that an `ItemRoster` can have empty `ItemRosterElement`s with null item values.

## What is the new behavior?

Add an earlier check to return if the item in the `ItemRosterElement` is null to avoid logging unnecessary errors.


source/GameInterface/Services/Inventory/Handlers/TradeHandler.cs
```diff
@@ -305,6 +305,11 @@ private bool TryResolveItemRosterId(ItemRosterElement itemRosterElement, out Ite
     {
         result = default;
 
+        if (itemRosterElement.EquipmentElement.Item == null)
```
gamingmenacers-svg: can this stay in `ResolveLeftLootIds` by iterating the roster/`Count` instead? the valid empty elements are unused slots in `FromRoster._data`; putting this in the shared resolver also silently drops a null item from `BoughtItems`/`SoldItems`, where the existing error is useful.

garrettluskey (changes requested):
the empty roster-slot diagnosis is right, but the unresolved scope issue in the existing thread still needs to be fixed before this can go in.

source/GameInterface/Services/Inventory/Handlers/TradeHandler.cs
garrettluskey, line 240:
```suggestion
if (item.EquipmentElement.Item == null)
    continue;
    
if (TryResolveItemRosterId(item, out var resolvedItem))
{
resolvedItems.Add((resolvedItem, count));
}
```

source/GameInterface/Services/Inventory/Handlers/TradeHandler.cs
```diff
@@ -305,6 +305,11 @@ private bool TryResolveItemRosterId(ItemRosterElement itemRosterElement, out Ite
...
+        {
+            return false;
+        }
+
```
garrettluskey:
```suggestion
```

garrettluskey (changes requested):

source/GameInterface/Services/Inventory/Handlers/TradeHandler.cs
```diff
@@ -231,6 +231,9 @@ private void Handle_UpdateEquipmentClients(MessagePayload<UpdateEquipmentClients
 
         foreach (var (item, count) in items)
         {
+            if (item.EquipmentElement.Item == null)
```
garrettluskey, line 234: this check is in the wrong resolver. `ResolveLeftLootIds` still walks `what.FromRoster._data` by capacity and calls `TryResolveItemRosterId` for every empty tail slot, so discarding/loot still logs the same error. pass the `ItemRoster` into `ResolveLeftLootIds` and iterate `roster.Count` with `GetElementCopyAtIndex` instead.

garrettluskey (changes requested):
the empty roster-slot diagnosis is right. this still leaves the noisy path unchanged, so changes are needed before this fixes the reported error.

source/GameInterface/Services/Inventory/Handlers/TradeHandler.cs
jordanbrymora: Changed `ResolveLeftLootIds` to use `ItemRoster` and guard against null items.

garrettluskey, line 234: `GetBoughtItems()` and `GetSoldItems()` return transaction-history entries with positive counts, not unused roster-array slots, and `ResolveLeftLootIds` now limits itself to `discardedItems.Count`. A null item in either active input is malformed data, so these guards silently drop it instead of preserving `TryResolveItemRosterId`'s failure log. Remove both null-item guards; iterating `Count` is what excludes genuine default `ItemRosterElement`s while leaving the helper to report malformed or unregistered entries.

garrettluskey (changes requested):
Switching the discard path from the backing array to the roster's active range fixes the noisy default tail-slot case. Changes are still needed because the new caller-side guards also suppress real resolution failures.

source/GameInterface/Services/Inventory/Handlers/TradeHandler.cs
jordanbrymora: The entire point of this PR is to suppress errors for failed id lookups for null items in item roster elements. Null items can be safely ignored because item rosters can legitimately contain empty elements with null items. This change would defeat the point of the PR and leave unneeded error logging.

The original error is for handling cases where the id lookup failed for an actual `ItemObject` instead of when the item was allowed to be null.

garrettluskey, line 234: `ItemRoster` can have empty elements in `_data` beyond `Count`, but those arent roster entries. `GetBoughtItems()` and `GetSoldItems()` build entries from transaction history, and `ResolveLeftLootIds` now stops at `discardedItems.Count`, so neither loop sees that empty capacity. remove both null guards, otherwise a null in an active input is silently dropped instead of reported.

garrettluskey (changes requested):
using `ItemRoster.Count` fixes the empty backing-array tail that caused the report. the two null guards still suppress failures from active inputs, so they need to be removed.

PR #3344, opened 2026-08-25
Fix contested village raid hanging on the battle loading screen

radinaldhs:
## PR Checklist

## PR Type

Raiding a village that resists (militia and villagers fight back, so the raid becomes a defense battle) can hang the client on the battle loading screen while the campaign keeps ticking underneath. The client log floods with `Failed to get TroopRoster using MapEventParty_Created_<id>_(DiedInBattle|WoundedInBattle|RoutedInBattle)`. This is the raid counterpart of the siege battle-launch races (sibling of #3253).

`BattleMissionStartHandler` opens the attack mission as soon as `NetworkStartAttackMission` arrives, gated only by `TryGetValidBattle`, which checks the local party side membership but not that the opponent (defender) side graph has committed on the client. `CoopTroopSupplier` then has no defender parties to deploy, deployment never completes, and the loading window never clears.

## What is the new behavior?

`ShowLoadingScreenAndQueueAttackMission` checks `initializationBarrier.IsPending(battle)`. While the event is still pending it defers the open through `IMapEventInitializationBarrier.RunAfterCommit`, so the mission opens only after the defender `MapEventParty` graph commits and `CoopTroopSupplier` has real parties to deploy. Already-committed events open immediately as before, so field battles and other already-synced paths are unchanged.

A `CampaignTick` safety net abandons the deferred open and disables the loading window if the graph is aborted and destroyed (the `RunAfterCommit` callback is dropped in that case), so a never-committing event returns to the map instead of hanging. The decision is the pure `DecideDeferredOpenAction`, covered by a theory matrix in `BattleMissionStartHandlerTests`.

## Bot Changelog Entry

- Fixed the loading screen hanging when raiding a village that fights back.

## Other information

Scoped to the attack-mission open. It does not touch the encounter-menu "Attack!" gating (that belongs with #3253) or the tally-roster client registration gap behind the `Failed to get TroopRoster` spam.

Manual in-game validation is still pending: raid a resisting village as a client and confirm `Coop_client.log` shows `Attack mission queued` then a delayed `executing queued open` after the commit, no `Failed to get TroopRoster ... MapEventParty_Created_*` flood, and the defense mission loads. Unit tests pass: `dotnet test GameInterface.Tests --filter BattleMissionStartHandlerTests` (5 of 5).

### DIFF
--- a/source/E2E.Tests/Services/MapEvents/BattleMissionStartHandlerTests.cs
+++ b/source/E2E.Tests/Services/MapEvents/BattleMissionStartHandlerTests.cs
@@ -5,6 +5,7 @@ using Common.Network;
 using GameInterface;
 using GameInterface.Services.MapEvents;
 using GameInterface.Services.MapEvents.Handlers;
+using GameInterface.Services.MapEvents.Initialization;
 using GameInterface.Services.MapEvents.Logging;
 using GameInterface.Services.MapEvents.Messages.Start;
 using GameInterface.Services.Players;
@@ -76,7 +77,8 @@ public class BattleMissionStartHandlerTests : MapEventTestBase
                 client.Resolve<IPlayerManager>(),
                 client.Resolve<INetwork>(),
                 client.Resolve<IMapEventLogger>(),
-                missionInitializerResolver);
+                missionInitializerResolver,
+                new Mock<IMapEventInitializationBarrier>().Object);
 
             try
             {
@@ -189,7 +191,8 @@ public class BattleMissionStartHandlerTests : MapEventTestBase
                     client.Resolve<IPlayerManager>(),
                     client.Resolve<INetwork>(),
                     client.Resolve<IMapEventLogger>(),
-                    missionInitializerResolver);
+                    missionInitializerResolver,
+                    new Mock<IMapEventInitializationBarrier>().Object);
 
                 messageBroker.Publish(this, new NetworkStartAttackMission(
                     mapEvent.MapEventId,
--- a/source/GameInterface.Tests/Services/MapEvents/BattleMissionStartHandlerTests.cs
+++ b/source/GameInterface.Tests/Services/MapEvents/BattleMissionStartHandlerTests.cs
@@ -8,11 +8,27 @@ namespace GameInterface.Tests.Services.MapEvents;
 public class BattleMissionStartHandlerTests
 {
     [Fact]
+    public void DecideDeferredOpenAction_ResolvesByMissionAndBattleState()
+    {
+        // Keep waiting while the battle is still valid, abandon the loading window once its graph is gone
+        // (the aborted contested-raid case), and clear once a mission has opened.
+        Assert.Equal(BattleMissionStartHandler.DeferredOpenAction.Keep,
+            BattleMissionStartHandler.DecideDeferredOpenAction(missionOpen: false, battleStillValid: true));
+        Assert.Equal(BattleMissionStartHandler.DeferredOpenAction.Abandon,
+            BattleMissionStartHandler.DecideDeferredOpenAction(missionOpen: false, battleStillValid: false));
+        Assert.Equal(BattleMissionStartHandler.DeferredOpenAction.Clear,
+            BattleMissionStartHandler.DecideDeferredOpenAction(missionOpen: true, battleStillValid: false));
+        Assert.Equal(BattleMissionStartHandler.DeferredOpenAction.Clear,
+            BattleMissionStartHandler.DecideDeferredOpenAction(missionOpen: true, battleStillValid: true));
+    }
+
+    [Fact]
     public void GetOrCreateMissionInitializerSnapshot_ReusesFirstBattleInitializer()
     {
         using var messageBroker = new MessageBroker();
         using var handler = new BattleMissionStartHandler(
             messageBroker,
+            null!,
             null!,
             null!,
             null!,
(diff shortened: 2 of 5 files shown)


source/GameInterface/Services/MapEvents/Handlers/BattleMissionStartHandler.cs
```diff
@@ -90,6 +100,7 @@ public void Dispose()
         messageBroker.Unsubscribe<NetworkStartAttackMission>(Handle_NetworkStartAttackMission);
         messageBroker.Unsubscribe<NetworkStartSiegeMission>(Handle_NetworkStartSiegeMission);
         messageBroker.Unsubscribe<MapEventFinalized>(Handle_MapEventFinalized);
+        messageBroker.Unsubscribe<CampaignTick>(Handle_CampaignTick);
```
gamingmenacers-svg, line 119: if this scope is disposed while an attack open is pending, this removes the only `CampaignTick` safety net but leaves the global loading window enabled. `MapEventInitializationBarrier` may later drop the callback when it disposes, or may invoke the captured handler first depending on lifetime disposal order, so the client can remain covered by the loading window or open a mission after this handler is dead. clear/abandon the deferred open here, disable the window, and make `OpenQueuedAttackMission` ignore a callback after disposal; cover the pending -> dispose path.

gamingmenacers-svg (changes requested):
reviewed this against current `development`, including the map-event barrier's register/commit/abort paths and the mission-start arbitration. the pending/commit decision fixes the reported ordering race, but the new deferred lifetime has one cleanup hole before it is safe to merge. the focused theory only tests the pure three-state decision; please add a handler-level pending -> commit test and a pending -> dispose/abort test with loading-window cleanup. manual contested-raid validation from the PR description is also still required.

source/GameInterface/Services/MapEvents/Handlers/BattleMissionStartHandler.cs
radinaldhs: fixed in c7a64b011. dispose now clears the deferred open and disables the loading window before unsubscribing, and OpenQueuedAttackMission returns early once disposed so a late RunAfterCommit callback does nothing. window toggles go through a raw IGlobalLoadingWindow seam that doesnt touch ForceLoadingWindow, otherwise the mission load completion disable in MissionPatches would be suppressed. added the pending to dispose and pending to commit handler tests asserting the window cleanup, 7 of 7 passing. manual contested raid validation is still pending and will be posted here when done.

source/E2E.Tests/Services/MapEvents/BattleMissionStartHandlerTests.cs
```diff
@@ -76,7 +77,8 @@ public void AttackMissionStart_NonInitiatingClientInitializesJoinedPlayerEncount
                 client.Resolve<IPlayerManager>(),
                 client.Resolve<INetwork>(),
                 client.Resolve<IMapEventLogger>(),
-                missionInitializerResolver);
+                missionInitializerResolver,
+                new Mock<IMapEventInitializationBarrier>().Object);
```
ShoT-UPfps, line 81: `BattleMissionStartHandler` now requires an `IGlobalLoadingWindow` after the barrier, so this seven-argument call no longer compiles. the second constructor call in this file and both direct calls in `SiegeAssaultLeaveTests` have the same omission, leaving four E2E call sites broken. pass a mock or resolved `IGlobalLoadingWindow` to all four and build the E2E project/solution.

source/GameInterface.Tests/Services/MapEvents/BattleMissionStartHandlerTests.cs
```diff
@@ -37,4 +97,155 @@ public void GetOrCreateMissionInitializerSnapshot_ReusesFirstBattleInitializer()
...
+
+        Assert.Equal(MapEventId, handler.DeferredAttackMissionMapEventId);
+        window.Verify(w => w.Enable(), Times.Once);
+        barrier.Verify(b => b.RunAfterCommit(It.IsAny<MapEvent>(), It.IsAny<Action>()), Times.Once);
+        // The open was registered on the barrier, not opened immediately, so the window is still held.
+        window.Verify(w => w.Disable(), Times.Never);
+
+        // The battle ends before the queued open runs (the race TryGetValidBattle guards); the commit
+        // callback still fires and drives the open, which re-validates and drops the window.
+        Campaign.Current.MainParty = null;
```
ShoT-UPfps, line 149: this invalidates the battle before the captured commit callback runs, so the test never exercises the success path named here: `OpenAttackMission`/the field-battle launcher is not reached. keep the battle valid, observe the launcher, invoke the callback, and assert exactly one mission open; keep the invalid-before-callback cleanup as a separate test.

ShoT-UPfps (commented):
rereviewed `2e5604c6`. the barrier gate itself matches the map-event commit ordering, but three issues remain: the new `IGlobalLoadingWindow` constructor parameter leaves four E2E call sites uncompilable; the pending-to-commit test invalidates the battle and never proves a mission opens; and disposal still strands the loading window when an already-committed battle is in the normal one-frame queue because only barrier-deferred opens are tracked. the current workflow is awaiting approval and has run no jobs, and manual contested-raid validation remains pending.

source/GameInterface/Services/MapEvents/Handlers/BattleMissionStartHandler.cs
ShoT-UPfps: the pending-barrier disposal case is covered now, but the same guard still leaks the window on the already-committed path. that path enables the window and uses `GameThread.EnqueueSafe` without setting `deferredAttackMissionMapEventId`; if the scope disposes before the next update, `Dispose()` sees no deferred id and does not disable, then `OpenQueuedAttackMission` returns on `disposed`. track window ownership for both queue types and add an immediate-queue -> dispose regression.

PR #3356, opened 2026-08-25
Fix bug report consent, saves, and archive eviction

garrettluskey:
## PR Checklist

## PR Type

`coop.bug_report_log_sharing enable` can save consent without displaying the current privacy disclosure. Pending archive quota enforcement stops when the oldest ZIP cannot be deleted. Bug reports also omit the server campaign save.

## What is the new behavior?

The enable command displays the current disclosure before saving consent, and quota enforcement skips undeletable archives and tries later candidates.

Each accepted bug report snapshots the server's current campaign as native `.sav` data in memory without writing to the normal Game Saves directory or creating a co-op session sidecar. The save is included in fallback ZIPs and uploaded before the JSON report. Existing consent is invalidated so the new public server-save disclosure is shown.

The ingestion endpoint receives:

1. An authenticated raw `PUT` with `X-Bug-Report-Artifact: server-save`, the report/file headers, and the save body.
2. The authenticated schema-v2 JSON `POST`, whose `serverSave` field identifies the uploaded artifact.

The endpoint must store the PUT by report id before accepting the matching JSON report.

Tested with:

- `dotnet test source/CoopUnitTests.slnf -c Release --no-build --no-restore`
- `dotnet test source/E2E.Tests/E2E.Tests.csproj -c Release --filter "FullyQualifiedName~UnstuckCommandTests"`

## Bot Changelog Entry

- Bug reports now save and upload the current server campaign, and log sharing shows the updated privacy disclosure before being enabled.

### DIFF
--- a/source/GameInterface/Services/BugReporting/BugReportServerSaveProvider.cs
+++ b/source/GameInterface/Services/BugReporting/BugReportServerSaveProvider.cs
@@ -4,7 +4,7 @@ using System;
 
 namespace GameInterface.Services.BugReporting;
 
-/// <summary>Creates the persistent server campaign save attached to a diagnostic report.</summary>
+/// <summary>Creates the native-format server campaign save attached to a diagnostic report.</summary>
 public interface IBugReportServerSaveProvider
 {
     bool TryCapture(out CollectedBugReportServerSave save);
@@ -31,7 +31,7 @@ public class BugReportServerSaveProvider : IBugReportServerSaveProvider
         save = null;
         try
         {
-            var result = saveInterface.SaveCurrentGameToFile(SaveName);
+            var result = saveInterface.SaveCurrentGameAsFileData(SaveName);
             if (!result.Success || result.Data == null || result.Data.Length == 0)
             {
                 logger.Warning("The server campaign save for the diagnostic bug report could not be created");
--- a/source/GameInterface/Services/BugReporting/BugReportServerSaveProvider.cs
+++ b/source/GameInterface/Services/BugReporting/BugReportServerSaveProvider.cs
@@ -1,0 +1,50 @@
+﻿using GameInterface.Services.Heroes.Interfaces;
+using Serilog;
+using System;
+
+namespace GameInterface.Services.BugReporting;
+
+/// <summary>Creates the persistent server campaign save attached to a diagnostic report.</summary>
+public interface IBugReportServerSaveProvider
+{
+    bool TryCapture(out CollectedBugReportServerSave save);
+}
+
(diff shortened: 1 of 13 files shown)


source/GameInterface/Services/BugReporting/BugReportServerSaveProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+/// <summary>Creates the persistent server campaign save attached to a diagnostic report.</summary>
+public interface IBugReportServerSaveProvider
+{
+    bool TryCapture(out CollectedBugReportServerSave save);
+}
+
+/// <inheritdoc />
+public class BugReportServerSaveProvider : IBugReportServerSaveProvider
+{
+    internal const string SaveName = "coop_bug_report";
```
ShoT-UPfps, line 16: this fixed name is in Bannerlord's ordinary user-save namespace. v1.4.8's `FileDriver` resolves it under `Game Saves`, and `PlatformFileHelperPC.SaveFile` reaches `File.WriteAllBytes`, so an existing `coop_bug_report.sav` is silently replaced. the server's existing `Game.Save` prefix also publishes `GameSaved("coop_bug_report")`, and `SaveGameHandler` writes the matching `coop_bug_report.json`, replacing the co-op session half of the save pair too. capture the diagnostic bytes without writing a fixed normal save (for example via an isolated driver/report directory), and add a regression that starts with an existing `.sav` / `.json` pair.

ShoT-UPfps (commented):
reviewed `3d139318` against live development `3e7fb745`. the consent and archive-eviction changes look good, and Build and Test #5247 passed all ten jobs. one P1 data-loss blocker remains: installed v1.4.8 writes the fixed `coop_bug_report` name into the normal Game Saves directory, and the existing `GameSaved` path writes the matching co-op session json, so an automatic report can overwrite an existing save pair. development's intervening kingdom-decision changes do not overlap the reviewed bug-report/save paths.

source/GameInterface/Services/BugReporting/BugReportServerSaveProvider.cs
garrettluskey: confirmed against v1.4.8, `FileDriver.Save` writes through `PlatformFileHelperPC` and the `Game.Save` prefix also creates the session sidecar. bug report saves now use a FileDriver-format in-memory driver, and `GameSaved` only publishes for real `FileDriver` saves. added a regression that keeps an existing `.sav` / `.json` pair unchanged.

ShoT-UPfps (approved):
lgtm

PR #3364, opened 2026-08-26
Update MBBodyPropertyRegistry

jordanbrymora:
## PR Checklist

## PR Type

`MBBodyPropertyRegistry` uses `CharacterObject` string ids instead of `MBBodyProperty.StringId`. This may be causing several failed object resolutions such as the one below obtained by entering a tavern:

image

## What is the new behavior?

Updated `MBBodyPropertyRegistry` to use `MBBodyProperty.StringId` as the base id instead.


garrettluskey (approved):
This now keys each body property from the same StringId that vanilla serializes and uses in MBObjectManager, rather than borrowing an incidental CharacterObject id. Enumerating MBObjectManager also covers shared and non-campaign body properties, and both vanilla constructors are accounted for. The focused lifetime tests pass; no blocking issues found.

PR #3365, opened 2026-08-26
Reduce reliable battle message backlog

garrettluskey:
## PR Checklist

## PR Type

Large battles put every combat message in its own reliable packet, while routed damage also serializes the full engine structs into payloads over 9 KB. The 64-packet reliable window backs up, delaying deaths and leaving stale presentation messages able to reach removed agents.

## What is the new behavior?

Related to #3363

Small reliable messages are batched through one shared transient component on both campaign and mission networking, including direct and relay mission routes. Routed damage uses a versioned owned wire format that stays below the batching budget, and blood presentation ignores inactive agents.

## How to test

- `dotnet test source/Common.Tests/Common.Tests.csproj -c Debug`
- run the `BattleMessageBatchingTests`, `BattleDamageCodecTests`, and existing routed-damage E2E tests
- join the same large field battle with two clients and confirm deaths do not fall minutes behind ongoing combat

## Bot Changelog Entry

- Reduced delayed deaths and ghost hit effects in large co-op battles.

### DIFF
--- a/source/Missions/Services/Network/LiteNetP2PClient.cs
+++ b/source/Missions/Services/Network/LiteNetP2PClient.cs
@@ -382,6 +382,7 @@ public class LiteNetP2PClient : INatPunchListener, INetEventListener, IUpdateabl
     public void OnPeerConnected(NetPeer peer)
     {
         bool rejectPeer = false;
+        string controllerIdToPromote = null;
         lock (peerGate)
         {
             if (pendingPeerControllers.TryGetValue(peer, out var controllerId))
@@ -400,11 +401,12 @@ public class LiteNetP2PClient : INatPunchListener, INetEventListener, IUpdateabl
                 }
                 else
                 {
-                    PromotePeer(peer, controllerId);
+                    controllerIdToPromote = controllerId;
                 }
             }
         }
 
+        if (controllerIdToPromote != null) PromotePeer(peer, controllerIdToPromote);
         if (rejectPeer) netManager.DisconnectPeer(peer);
 
         // Proof-of-P2P diagnostic: the remote endpoint here is the OTHER CLIENT's socket, reached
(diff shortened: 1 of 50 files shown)


source/Missions/Services/Network/LiteNetP2PClient.cs
```diff
@@ -546,6 +574,8 @@ private NetPeer RemoveTrackedPeer(string controllerId)
 
     private void PromotePeer(NetPeer peer, string controllerId)
     {
+        // Submit earlier relay traffic before switching this controller to the direct route.
+        reliableMessageBatcher.Flush(controllerId, SendReliableMessagePayload);
```
ShoT-UPfps: this is the `PromotePeer` flush, not either disconnect-path flush. `PromotePeer` is called while `peerGate` is held; this call takes the controller's destination-buffer lock and can invoke `SendReliableMessagePayload`, which re-enters `peerGate` while that buffer lock is still held. a concurrent reliable send can hold the buffer lock first and wait on `peerGate`, while promotion holds `peerGate` and waits on the buffer lock. move this flush outside the `peerGate` section or establish one lock order, and add a concurrent promotion/send regression.

ShoT-UPfps (commented):
rereviewed `b1f51a2`. the development merge and Coop Options movement-limit change do not fix the transport P1: `PromotePeer` still flushes the per-controller batch while its callers hold `peerGate`, and the flush callback can re-enter `peerGate` while holding the destination-buffer lock, opposite a concurrent reliable send. that can deadlock promotion and reliable sending. the publisher resolves only the added flush inside `PromotePeer`; identical disconnect-path flushes are excluded.

ShoT-UPfps (changes requested):

ShoT-UPfps (approved):
lgtm

PR #3366, opened 2026-08-26
Handle mount-affector death replays

garrettluskey:
## PR Checklist

## PR Type

Replicated deaths rebuild a `Blow` without weapon data, leaving its default weapon slot at zero. When the affector is a mount, vanilla tries to read slot zero from the mount's missing equipment and throws before the puppet dies or deregisters.

## What is the new behavior?

Related to #3363

Death replays mark the weapon slot as absent. Vanilla keeps the mount as the affector for attribution but uses an invalid weapon instead of reading mount equipment.

## How to test

- run `BattleDeathMirrorTests`
- run `BattleMountIdentityTests` and `BattleDamageMirrorTests`

## Bot Changelog Entry

- Fixed some mount-charge deaths leaving troops alive on other clients.

### DIFF
--- a/source/E2E.Tests/Environment/MockEngine/MissionEngineFixture.cs
+++ b/source/E2E.Tests/Environment/MockEngine/MissionEngineFixture.cs
@@ -511,6 +511,18 @@ public sealed class MissionEngineFixture : IDisposable
             throw new KeyNotFoundException(
                 $"Missile index {blow.WeaponRecord.AffectorWeaponSlotOrMissileIndex} not in the mock mission's missile set (models Mission.OnAgentHit)");
 
+        int affectorWeaponSlot = blow.WeaponRecord.AffectorWeaponSlotOrMissileIndex;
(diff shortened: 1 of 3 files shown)


ShoT-UPfps (commented):
reviewed `17a5130b` head-only. no actionable correctness findings. installed v1.4.8 confirms non-missile `Mission.OnAgentHit` indexes the affector's equipment only when `AffectorWeaponSlotOrMissileIndex >= 0`; setting the replayed weapon slot to `-1` therefore keeps the mount as affector while passing `MissionWeapon.Invalid`. the new E2E regression models the previous mount-equipment failure and verifies the replay deactivates and deregisters the victim without removing the affector. Build and Test #5269 passed build, unit tests, and all eight E2E shards. this validates the mount-affector replay crash only; the broader #3363 desynchronization report was not independently reproduced.

ShoT-UPfps (approved):

PR #3368, opened 2026-08-26
StanceLink duplicate id register overflow

NatteTosti69:
## PR Checklist

- [ x ] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Currently, `StanceLinks` that contain one or more eliminated factions, still send a message to the server,  where the server creates the new `StanceLink`, because `_stances` is empty.

## What is the new behavior?

Resolves #

`StanceLinks` do not get sent to the server or to other clients when one of the factions are eliminated.

### DIFF
--- a/source/E2E.Tests/Services/StanceLinks/StanceLinkRegistrationGuardTests.cs
+++ b/source/E2E.Tests/Services/StanceLinks/StanceLinkRegistrationGuardTests.cs
@@ -54,7 +54,6 @@ public class StanceLinkRegistrationGuardTests : IDisposable
 
         FirstClient.Call(() =>
         {
-
             Assert.True(FirstClient.ObjectManager.TryGetObject<IFaction>(faction1Id, out var clientFaction1));
             Assert.True(FirstClient.ObjectManager.TryGetObject<IFaction>(faction2Id, out var clientFaction2));
 
@@ -66,7 +65,6 @@ public class StanceLinkRegistrationGuardTests : IDisposable
         Assert.Empty(FirstClient.NetworkSentMessages.GetMessages<StanceLinkConstructed>());
         Assert.Empty(Server.NetworkSentMessages.GetMessages<StanceLinkConstructed>());
     }
-
 
     [Fact]
     public void BanditMismatchedFaction_Server_NeverPublishesRequestStanceLinkConstructed()
--- a/source/E2E.Tests/Services/StanceLinks/StanceLinkRegistrationGuardTests.cs
+++ b/source/E2E.Tests/Services/StanceLinks/StanceLinkRegistrationGuardTests.cs
@@ -1,4 +1,5 @@
-﻿using E2E.Tests.Environment;
+﻿using Common.Util;
+using E2E.Tests.Environment;
 using E2E.Tests.Environment.Instance;
 using GameInterface.Services.StanceLinks.Messages;
 using TaleWorlds.CampaignSystem;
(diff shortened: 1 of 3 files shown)


source/GameInterface/Services/StanceLinks/StanceLinkPatches.cs
```diff
@@ -23,6 +23,7 @@ private static bool Prefix(FactionManager __instance, IFaction faction1, IFactio
     {
         if (CallOriginalPolicy.IsOriginalAllowed()) return true;
         if (faction1.IsBanditFaction != faction2.IsBanditFaction) return true;
+        if (faction1.IsEliminated || faction2.IsEliminated) return true;
```
garrettluskey, line 26: This only stops lookups after the local faction has been marked eliminated. A `StanceLinkConstructed` already in flight—or sent by a client that has not received the elimination yet—is handled later by `HandleStanceLinkConstructed`; on the server it creates a link, vanilla `AddStance` rejects it because the faction is eliminated, then `AddExisting` registers the orphan and the server broadcasts it. That leaves the same duplicate-ID/counter path open. Reject eliminated pairs in `HandleStanceLinkConstructed` after resolving both factions (on both roles), and add a test that delivers the command after `DeactivateKingdom()`.

garrettluskey (changes requested):
The local eliminated-faction guard matches vanilla behavior, and the repeated-lookup coverage is useful. I'm requesting one change because a stale or in-flight client command can still recreate the orphaned registration this fix is meant to prevent.

ShoT-UPfps (commented):
reviewed `e22a6c54` head-only. the origin-side eliminated-faction guard correctly prevents new network construction once the caller already observes elimination, and removing the outer registration log preserves `ObjectManager`'s specific diagnostics. Garrett's existing line-26 thread already covers the one P1 blocker, so I am not duplicating it: an in-flight or stale `StanceLinkConstructed` can reach `HandleStanceLinkConstructed` after destruction; vanilla `AddStance` rejects the eliminated pair, but `AddExisting` still registers the orphan and the server rebroadcasts it. add the authoritative eliminated-pair guard before creation/registration and a delayed-delivery regression. Build and Test #5287 passed all ten jobs.

source/E2E.Tests/Services/StanceLinks/StanceLinkRegistrationGuardTests.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            Assert.True(FirstClient.ObjectManager.TryGetObject<IFaction>(faction1Id, out var clientFaction1));
+            Assert.True(FirstClient.ObjectManager.TryGetObject<IFaction>(faction2Id, out var clientFaction2));
+            Assert.False(clientFaction2.IsEliminated);
+            for (int i = 0; i < 10; i++)
+            {
+                Assert.NotNull(FactionManager.Instance.GetStanceLinkInternal(clientFaction1, clientFaction2));
+            }
+        });
+        Assert.Empty(FactionManager.Instance._stances.GetStanceLinks());
```
garrettluskey, line 98: This only checks `_stances`, which also stayed empty in the old failure because vanilla `AddStance` rejects eliminated factions. Check `Server.ObjectManager` for the computed stance ID and `Server.NetworkSentMessages` for `StanceLinkConstructed` here; otherwise removing the new handler guard still leaves this test green.

garrettluskey (changes requested):
The authoritative guard in `HandleStanceLinkConstructed` fixes the stale-client path: both factions are resolved on the game thread, and an eliminated pair returns before construction, registration, or rebroadcast. The delayed-delivery test still needs to assert the failure mode it is meant to cover.

garrettluskey (approved):
The eliminated-faction checks now cover both local lookup and the authoritative receive path. The delayed-delivery regression checks the object-manager registration and server rebroadcast failure mode, and the focused six-test suite passes.

PR #3387, opened 2026-08-28
Add bug report button visibility option

garrettluskey:
## PR Checklist

## PR Type

The `Report Coop Bug` button is always shown during gameplay.

## What is the new behavior?

Coop Options has a Bug Report tab with a `Show Coop Bug Report Button` checkbox. The choice is saved per client and applies immediately. Hiding the button does not change diagnostic log-sharing consent or automatic reports triggered by recovery actions.

Tested with:

`dotnet test source/CoopUnitTests.slnf -c Release --no-restore`

## Bot Changelog Entry

- The in-game bug-report button can now be hidden from Coop Options.

### DIFF
--- a/source/GameInterface/Services/UI/CoopOptions/Providers/BugReportTab/Sections/BugReportSectionOptions.cs
+++ b/source/GameInterface/Services/UI/CoopOptions/Providers/BugReportTab/Sections/BugReportSectionOptions.cs
@@ -1,0 +1,14 @@
+﻿namespace GameInterface.Services.UI.CoopOptions.Providers.BugReportTab.Sections;
+
+/// <summary>Persists the client's bug-report button visibility choice.</summary>
+public class BugReportSectionOptions
+{
+    public const bool DefaultShowBugReportButton = true;
+
+    public bool? ShowBugReportButton { get; set; }
+
+    public bool GetShowBugReportButtonOrDefault()
+    {
+        return ShowBugReportButton ?? DefaultShowBugReportButton;
+    }
+}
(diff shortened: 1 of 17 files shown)


source/GameInterface/Services/UI/CoopOptions/Providers/BugReportTab/Sections/BugReportSectionOptions.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace GameInterface.Services.UI.CoopOptions.Providers.BugReportTab.Sections;
+
+/// <summary>Persists the client's bug-report button visibility choice.</summary>
+public class BugReportSectionOptions
+{
+    public const bool DefaultShowBugReportButton = true;
+
+    public bool? ShowBugReportButton { get; set; }
```
ShoT-UPfps, line 11: this is the only Co-op Options DTO without an explicit JSON field name. `CoopOptionsData` serializes and deserializes section DTOs with no naming policy or case-insensitive matching, so this is persisted as `ShowBugReportButton`; every neighboring option and the documented config shape use lower camel case. a `showBugReportButton` entry is therefore ignored and falls back to `true`, and adding the convention later would strand the PascalCase values already written. add `[JsonPropertyName("showBugReportButton")]` before release and assert the raw saved key.

ShoT-UPfps (commented):
reviewed `d800f953` head-only. the new Bug Report tab, option save/apply flow, consent-section preservation, message-broker update, overlay visibility gate, generated movie, and focused tests are otherwise consistent. Build and Test #5325 passed build, unit tests, and all eight E2E shards. one P2 persisted-schema issue remains: `BugReportSectionOptions` omits the explicit lower-camel JSON name used by every neighboring option, so this release writes `ShowBugReportButton`; a conventional `showBugReportButton` value is ignored and defaults true, and correcting the name later would require migration. add the property name now and assert the raw saved key.

ShoT-UPfps (changes requested):

source/GameInterface/Services/UI/CoopOptions/Providers/BugReportTab/Sections/BugReportSectionOptions.cs
garrettluskey: added `[JsonPropertyName("showBugReportButton")]` and the persistence test now checks the raw lower-camel key.

ShoT-UPfps (approved):

PR #3390, opened 2026-08-28
Reuse stable tournament portraits during bracket refresh

gamingmenacers-svg:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Every canonical tournament bracket refresh resets and initializes the participant VMs again. Even when a bracket slot still contains the same character and descriptor, that rebuilds its portrait bindings and character texture on every tournament update.

## What is the new behavior?

Canonical state is rebound in place. A participant portrait is reused only when the same character reference and descriptor seed remain in the same valid slot; character changes, descriptor changes, slot changes, and empty intervals still force a fresh portrait. Scores, winners, team colors, match state, and cleared slots continue to refresh from the authoritative bracket.

## Bot Changelog Entry

- Tournament bracket updates no longer rebuild unchanged participant portraits every refresh.

## Other information

Validation:
- `dotnet build source\CoopTests.slnf -c Release` (0 errors)
- 14 focused tournament bracket tests passed
- full `CoopUnitTests.slnf` passed: 2,206 passed, 14 expected skips


source/GameInterface/Services/Tournaments/UI/CoopTournamentVM.cs
```diff
@@ -437,24 +455,160 @@ internal static TournamentMatchVM RebindCanonicalBracket(
...
+        participantViewModel.TeamColor = teamColor;
+        bool isPlayerCharacter = ReferenceEquals(
+            canonicalParticipant.Character,
+            Game.Current?.PlayerTroop);
+        participantViewModel.State = isPlayerCharacter ? 2 : 1;
+        participantViewModel.Name = canonicalParticipant.Character.Name.ToString();
+        participantViewModel.IsMainHero = isPlayerCharacter;
+        participantViewModel.IsInitialized = true;
+        participantViewModel.IsValid = true;
+        participantViewModel.IsDead = false;
```
garrettluskey: `ResetRoundViewModel` and native `TournamentParticipantVM.Refresh` both preserve `IsDead`, so this clears the knockout marker from a finished match when its result refreshes the bracket. only reset this for a changed or new identity, a stable slot should keep it.

source/GameInterface/Services/Tournaments/UI/CoopTournamentVM.cs
```diff
@@ -406,10 +406,25 @@ internal static TournamentMatchVM RebindCanonicalBracket(
...
+            getRoundTitle,
+            (participantViewModel, participant, teamColor) =>
+                participantViewModel.Refresh(participant, teamColor));
+
+    internal static TournamentMatchVM RebindCanonicalBracket(
+        IReadOnlyList<TournamentRoundVM> roundViewModels,
+        IReadOnlyList<TournamentRound> canonicalRounds,
+        TournamentMatch canonicalCurrentMatch,
+        Func<int, TextObject> getRoundTitle,
+        Action<TournamentParticipantVM, TournamentParticipant, Color> refreshParticipantPortrait)
```
garrettluskey: this callback only exists for the tests and gets passed through the round, match, and team binders, while production always calls `TournamentParticipantVM.Refresh`. call `Refresh` directly in `BindCanonicalParticipant` and test `CanReuseParticipantPortrait` plus the binding notifications instead.

garrettluskey (dismissed):
the repeated portrait rebuild is real, and keeping stable `Character`/`Visual` bindings is the right direction. changes requested because the stable rebind changes knockout state and the implementation carries a test-only refresh callback through the whole binder. the tests show that binding reconstruction is avoided, but they dont show a stall or freeze, so please say this avoids repeated portrait rebuilding unless runtime timing shows the stall.

source/GameInterface/Services/Tournaments/UI/CoopTournamentVM.cs
gamingmenacers-svg: fixed in `21441eea3`. stable slots now keep `IsDead`; it is only cleared after `Refresh` replaces a changed or new identity. the stable-slot test covers the knockout state.

source/GameInterface/Services/Tournaments/UI/CoopTournamentVM.cs
gamingmenacers-svg: fixed in `21441eea3`. `BindCanonicalParticipant` calls `Refresh` directly, and the callback overload and binder plumbing are removed. the tests cover `CanReuseParticipantPortrait` and the `Character`/`Visual` binding notifications.

ShoT-UPfps (approved):

PR #3391, opened 2026-08-28
Skip relation changes with unresolved heroes

gamingmenacers-svg:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

The relation-action patch only decides whether native `ApplyInternal` runs on this process. On the server, Bannerlord's diplomacy model can resolve one effective hero to null when a clan has no leader, after which the native relation action throws during callbacks such as the daily party tick.

## What is the new behavior?

The exact native overload is patched and the server validates the effective hero pair before allowing a non-zero relation change through. Valid pairs and Bannerlord's both-leaders-missing fallback still run natively, zero changes remain no-ops without resolution, and clients still skip the native action. Invalid pairs are logged once per identity tuple with a bounded 256-entry diagnostic set.

## Bot Changelog Entry

- Leaderless clans no longer make relation updates throw during campaign ticks.

## Other information

Validation:
- `dotnet build source\CoopTests.slnf -c Release` (0 errors)
- 8 focused relation-action tests passed
- full `CoopUnitTests.slnf` passed: 2,209 passed, 14 expected skips

### DIFF
--- a/source/GameInterface.Tests/Services/Actions/ChangeRelationActionPatchesTests.cs
+++ b/source/GameInterface.Tests/Services/Actions/ChangeRelationActionPatchesTests.cs
@@ -1,0 +1,136 @@
+﻿using Common;
+using GameInterface.Services.Actions.Patches;
+using GameInterface.Tests.Bootstrap;
+using HarmonyLib;
+using System;
+using System.Reflection;
+using TaleWorlds.CampaignSystem;
+using TaleWorlds.CampaignSystem.Actions;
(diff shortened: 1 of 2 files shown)


garrettluskey (approved):
this keeps client suppression intact, skips only unresolved non-zero server changes, and preserves vanilla's both-leaders-missing fallback. the focused tests cover the important branches and the Harmony target is valid.

PR #3400, opened 2026-08-28
Enter character creation before intro playback

garrettluskey:
## PR Checklist

## PR Type

- [ ] Breaking change (fix or feature that causes existing functionality to stop working as expected)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update
- [ ] Other

A new player remains in the client validation state while the character creation intro plays, so the forced `Validating modules...` window covers the video until the real character creation state activates.

## What is the new behavior?

The explicit new-player response now enters the co-op character creation state on the game thread before starting the vanilla new game. This releases the validation window before intro playback without allowing game-state activation events to bypass server approval.

## How to test

- Connect with a controller id that has no saved player and confirm the intro is visible without the validation window over it.
- Finish character creation and confirm the client continues into the host campaign.
- `dotnet test source\Coop.Tests\Coop.Tests.csproj -c Release --filter FullyQualifiedName~Coop.Tests.Client.States.ValidateModuleStateTests`
- `dotnet test source\GameInterface.Tests\GameInterface.Tests.csproj -c Release --filter FullyQualifiedName~GameInterface.Tests.Services.GameDebug.Patches.CharacterCreationIntroPatchTests`

## Bot Changelog Entry

- Fixed the `Validating modules...` screen covering the new-character intro movie.

### DIFF
--- a/source/Coop.Core/Client/States/ValidateModuleState.cs
+++ b/source/Coop.Core/Client/States/ValidateModuleState.cs
@@ -74,7 +74,6 @@ public class ValidateModuleState : ClientStateBase
         this.gameStateInterface = gameStateInterface;
         messageBroker.Subscribe<NetworkModuleVersionsValidated>(Handle_NetworkModuleVersionsValidated);
         messageBroker.Subscribe<NetworkClientValidated>(Handle_NetworkClientValidated);
-        messageBroker.Subscribe<CharacterCreationStarted>(Handle_CharacterCreationStarted);
 
 #if DEBUG
         controllerIdProvider.SetControllerFromProgramArgs();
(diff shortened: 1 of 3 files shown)


ShoT-UPfps (commented):
reviewed `eb9a6a3a` head-only.

actionable: `Handle_NetworkClientValidated` claims `completionClaimed` before it merely queues `BeginCharacterCreation`. if `clientLogic.Disconnect()` runs before the game-thread queue drains, `ValidateModuleState.Disconnect()` loses that already-claimed latch and skips teardown while `Logic.State` is still this validation instance; the queued callback then passes its guard, enters `CharacterCreationState`, and calls `StartNewGame()` after the disconnect. move the terminal claim into the same game-thread action as the state transition (or otherwise let disconnect cancel the pending transition), and add a regression that disconnects after the validation response but before draining `GameThread`.

Build and Test #5344 failed in the unit-test job at `Run Unit Tests`, so this head has no green exact-head CI.

ShoT-UPfps (approved):

PR #3401, opened 2026-08-28
Kingdom force ally and force trade agreement debug commands

casey-johnson:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?

Adds 2 debug commands:
1. forcing two kingdoms to ally
2. forcing a trade agreement between 2 kingdoms.

Previously there were debug commands for forcing a vote, but this adds cheats to bypass the vote.

## Bot Changelog Entry

- Added a force_ally cheat command
- Added a force_trade_agreement cheat command


source/GameInterface/Services/Kingdoms/Commands/KingdomDebugCommand.cs
```diff
@@ -1131,6 +1132,131 @@ public static string MakePeace(List<string> args)
...
+        }
+
+        if (objectManager.TryGetObject(args[1], out Kingdom kingdom2) == false)
+        {
+            return $"Kingdom not found with id: {args[1]}";
+        }
+
+        if (kingdom1 == kingdom2)
+        {
+            return "A kingdom cannot ally with itself.";
```
ShoT-UPfps: can this validation reject either `IsEliminated` kingdom and be shared with `ForceTradeAgreement`? `KingdomRegistry` has no destroy/unregister path, so destroyed kingdom ids still resolve; vanilla rejects eliminated realms, and the alliance/trade behaviors only clean their records in `OnKingdomDestroyed`. invoking either command after that event can re-add a relation for a dead realm with no later cleanup. return an error before making peace or calling the behavior, and cover both commands with a destroyed-kingdom regression.

ShoT-UPfps (commented):
reviewed `38d69071` head-only. Build and Test #5345 passed all ten jobs on this head. one correctness issue remains: both force commands accept eliminated kingdoms. those kingdoms remain registered after destruction, while vanilla rejects them and the alliance/trade behaviors perform their cleanup only when `OnKingdomDestroyed` fires. running either command afterward can therefore recreate a persistent relation for a dead realm. reject either `IsEliminated` input before making peace or adding the relation.

ShoT-UPfps (changes requested):

source/GameInterface/Services/Kingdoms/Commands/KingdomDebugCommand.cs
```diff
@@ -1131,6 +1132,149 @@ public static string MakePeace(List<string> args)
...
+            return pairError;
+        }
+
+        var behavior = Campaign.Current.GetCampaignBehavior<TradeAgreementsCampaignBehavior>();
+        if (behavior == null)
+        {
+            return "TradeAgreementsCampaignBehavior is not available.";
+        }
+
+        if (behavior.TryGetTradeAgreement(kingdom1, kingdom2, out _))
```
ShoT-UPfps: `TryGetTradeAgreement` only checks whether this pair exists in `_tradeAgreements`; it never inspects `EndTime`. vanilla's public `HasTradeAgreement` removes an expired entry and returns `false`. after an agreement expires but before another caller performs that lazy cleanup, this command reports "already have" and cannot create the requested new agreement. call `HasTradeAgreement(kingdom1, kingdom2, out _)` here and add an expired-agreement renewal regression.

ShoT-UPfps (commented):
rereviewed `7228857f` head-only after the merge of `development` at `8617247b`. the merge preserves the eliminated-kingdom fix: both force commands now share `TryGetKingdomPair`, reject either eliminated realm, and the two regressions pass, so the prior thread can be resolved. one P2 correctness edge remains in `force_trade_agreement`: it calls the raw-list `TryGetTradeAgreement`, so an expired agreement that has not yet been lazily removed is treated as active and prevents the force command from renewing it; use `HasTradeAgreement(..., out _)`, which removes expired entries. Build and Test #5396 built successfully and all eight E2E shards passed; the unit job is red only on the unrelated `CallOriginalPolicyTests.CurrentOperationScope_IsLimitedToInvokingThread`.

ShoT-UPfps (approved):
lgtm

PR #3403, opened 2026-08-28
Add injectable command framework

garrettluskey:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Console commands use static attributed methods and handle argument parsing, validation, and usage text individually. The command metadata and execution path cannot be reused cleanly by other transports.

## What is the new behavior?

Adds an Autofac-backed command registry with structured results, generated usage, centralized argument validation, trimming, and quoted argument parsing. Bannerlord console and live testing use adapters over the same registry, while existing attributed commands continue to work during migration.

Migrates `coop.debug.hero.id` as the first injected command.

## Other information

Tests:

- `dotnet test source/Common.Tests/Common.Tests.csproj -c Release --no-restore`
- `dotnet test source/GameInterface.Tests/GameInterface.Tests.csproj -c Release --no-restore --filter "FullyQualifiedName~CoopCommandLineRegistrarTests|FullyQualifiedName~LiveTestCommandDispatcherTests"`
- `dotnet test source/Coop.Tests/Coop.Tests.csproj -c Release --no-restore --filter "FullyQualifiedName~ContainerBuildTests"`

### DIFF
--- a/source/GameInterface/Services/Heroes/Commands/HeroIdCommand.cs
+++ b/source/GameInterface/Services/Heroes/Commands/HeroIdCommand.cs
@@ -1,0 +1,70 @@
+﻿using Common.Commands;
+using GameInterface.Services.ObjectManager;
+using GameInterface.Services.ObjectManager.Extensions;
+using System;
+using System.Linq;
+using System.Text;
+using TaleWorlds.CampaignSystem;
+
+namespace GameInterface.Services.Heroes.Commands;
+
+public interface IHeroIdCommand : ICoopCommand
+{
+}
+
+public sealed class HeroIdCommand : IHeroIdCommand
+{
+    private readonly IObjectManager objectManager;
+
+    public HeroIdCommand(IObjectManager objectManager)
+    {
+        if (objectManager == null) throw new ArgumentNullException(nameof(objectManager));
+
+        this.objectManager = objectManager;
+    }
+
+    public string Prefix => "coop.debug.hero";
+
+    public string Name => "id";
+
+    public string Description => "Finds registered ids for heroes with an exact display name.";
+
+    public IExpectedArgs[] ExpectedArgs { get; } = new IExpectedArgs[]
+    {
+        new ExpectedArgs(
+            "heroName",
+            "The exact display name of the hero to find."),
+    };
+
+    public CoopCommandResult ProcessCommand(ICoopCommandArgs args)
+    {
+        Campaign campaign = Campaign.Current;
+        if (campaign == null)
+            return new CoopCommandResult(false, "Campaign is not loaded.", "campaign_unavailable");
+
+        string heroName = args[0];
+        var heroes = campaign.CampaignObjectManager.GetAllHeroes()
+            .Where(hero => string.Equals(
+                hero.Name?.ToString(),
+                heroName,
+                StringComparison.OrdinalIgnoreCase))
+            .ToList();
+        if (heroes.Count == 0)
+            return new CoopCommandResult(false, $"No hero named '{heroName}' was found.", "hero_not_found");
+
+        var output = new StringBuilder();
+        foreach (Hero hero in heroes)
+        {
+            if (objectManager.TryGetId(hero, out string id))
+            {
+                output.AppendLine($"ID: '{id}', Name: '{hero.Name}', Game StringId: {hero.StringId}");
+            }
+            else
+            {
+                output.AppendLine($"Name: '{hero.Name}' was not registered with object manager");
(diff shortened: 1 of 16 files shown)


source/GameInterface/Utils/Commands/CoopCommandLineRegistrar.cs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    private void RemoveRegistrations()
+    {
+        foreach (KeyValuePair<string, Registration> registration in registrations)
+        {
+            if (!gameCommands.Contains(registration.Key) ||
+                !ReferenceEquals(gameCommands[registration.Key], registration.Value.GameCommand))
+            {
+                continue;
```
ShoT-UPfps, line 132: when a superseded scope is disposed after its replacement has installed this command, this branch skips the only unlinking path. the replacement still holds this game command through `PreviousRegistration`; its delegate target holds the disposed registrar, which holds the scoped registry and `HeroIdCommand`/`IObjectManager`. normal campaign teardown does not clear that object manager, so repeating A -> B -> C overlap retains every prior campaign object map from static `AllFunctions` until the current scope finally exits. splice disposed predecessors out of the active chain and add a three-generation overlap regression.

ShoT-UPfps (commented):
reviewed `12780276` head-only. the shared command model, argument parsing/validation, live-test adapter, hero-id migration, and reflected Bannerlord v1.4.8 registry contract are otherwise consistent. Build and Test #5346 succeeded on the exact head. one P1 reconnect-lifecycle issue remains: disposing a registrar after its replacement is installed leaves the disposed scope linked through `PreviousRegistration`; repeated overlapping reconnects therefore retain every prior command registry and campaign `IObjectManager` from static `AllFunctions` until the latest scope exits. unlink disposed predecessors and cover three overlapping generations.

ShoT-UPfps (changes requested):
c

source/GameInterface/Utils/Commands/CoopCommandLineRegistrar.cs
garrettluskey: fixed, disposing an older registrar now prunes disposed registrations from the active predecessor chain. added a three-generation weak-reference regression.

source/GameInterface/Services/Heroes/Commands/HeroIdCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+        this.objectManager = objectManager;
+    }
+
+    public string Prefix => "coop.debug.hero";
+
+    public string Name => "id";
+
+    public string Description => "Finds registered ids for heroes with an exact display name.";
+
+    public IExpectedArgs[] ExpectedArgs { get; } = new IExpectedArgs[]
```
ShoT-UPfps, line 32: this changes an existing invocation contract. before this migration, `coop.debug.hero.id John Doe` worked because `FindIds` joined all console tokens into `John Doe`; the new command declares exactly one expected argument, so the registry rejects those two tokens before `HeroIdCommand.ProcessCommand` runs and only `coop.debug.hero.id "John Doe"` works. preserve the legacy unquoted multi-word-name behavior during migration and add a regression for that form.

ShoT-UPfps (commented):
rereviewed `17174d22` head-only. the earlier reconnect-lifecycle P1 is fixed by pruning disposed predecessor registrations. one compatibility regression remains in the first migrated command: `coop.debug.hero.id` previously accepted an unquoted multi-word hero name by joining console tokens, while the new one-argument registry rejects the same invocation unless the user adds quotes. preserve the existing unquoted form and cover it with a regression. Build and Test #5354 is red on this exact head (`test` / `Run Unit Tests`).

ShoT-UPfps (changes requested):

source/GameInterface/Services/Heroes/Commands/HeroIdCommand.cs
garrettluskey: No, new way better

garrettluskey: We are sunsetting the underscore method. That only was for these commands and the quote way is industry standard. Nothing relies on the old method

ShoT-UPfps (approved):
rereviewed `807bb498` head-only. prior reconnect unlink still looks right, quoted hero names are the intended contract, and nothing new on parse, registration lifetime, or the hero.id sample. ci is green on this head.

PR #3445, opened 2026-08-28
Network optimizations

Gerseras:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

## What is the new behavior?
Reduces Bandwith usage due to protobuf tags and inefficient bool transmission
- Arrays save 1 byte per entry if there are more then 2 entries
- Bools save 2 bytes per bool if there are more then 2 bools without any bandwith downside. So Potential save on AgentActionData is up to 8 Bytes per instance This could happen up to 30 times a second per Agent
->So worst case right now this would be 8 bytes x 1000 Agents x 30 per second = 240kb/s per Player (real impact will be way lower as usually there are only up to 3 flags active at a time if its a mounted unit blocking)


ShoT-UPfps (dismissed):
lgtm after solving merge cofns

ShoT-UPfps (approved):

PR #3456, opened 2026-08-29
[3345] Reduce Vec3 wire size without allocations

### LINKED ISSUE #3345: Reduce Vec3Surrogate wire size without per-serialization allocations
## Description

`Vec3Surrogate` currently serializes `X`, `Y`, and `Z` as three separate protobuf `float` fields. For a normal vector with all three components nonzero, the surrogate body is 15 bytes: three one-byte field tags plus three four-byte float values.

The `network_optimizations` experiment in commit `9767d6c913ccfcd7d5bab0ad60ae23cfea0b46ba` attempted to reduce that normal representation to 14 bytes by replacing the three scalar fields with a packed `float[3]`:

```csharp
[ProtoMember(1, IsPacked = true)]
public float[] V;
```

That implementation was intentionally removed from the branch because every conversion allocated a new managed array:

```csharp
new[] { value.X, value.Y, value.Z }
```

`Vec3Surrogate` is registered globally and is used heavily by movement snapshots. Movement batching can serialize multiple candidate packet sizes for multiple recipients while finding the largest packet that fits. A per-`Vec3` array therefore adds avoidable Gen0 pressure in a hot serialization path.

Track a replacement that retains the normal-case byte reduction without allocating an array.

## Goal

Represent the raw bits of `X` and `Y` in one `ulong`, while keeping `Z` as a normal `float`:

```csharp
[ProtoMember(1, DataFormat = DataFormat.FixedSize)]
public ulong XY;

[ProtoMember(2)]
public float Z;
```

For the normal all-nonzero case, this produces the same 14-byte surrogate body as the removed packed-array experiment:

| Representation | Normal all-nonzero `Vec3` body | Surrogate conversion allocation |
| --- | ---: | --- |
| Current three scalar floats | 15 bytes | None |
| Removed packed `float[3]` | 14 bytes | One managed array |
| Target `ulong XY` plus `float Z` | 14 bytes | None |

When `Z` is exactly zero, the target representation is nine bytes because protobuf can omit the `Z` field. When all components are positive zero, both fields remain their defaults and may be omitted.

## Intended design

Use allocation-free bit reinterpretation compatible with the repository's `netstandard2.0` target. One possible implementation is:

```csharp
using ProtoBuf;
using System.Runtime.InteropServices;
using TaleWorlds.Library;

namespace GameInterface.Surrogates;

[ProtoContract]
internal struct Vec3Surrogate
{
    [ProtoMember(1, DataFormat = DataFormat.FixedSize)]
    public ulong XY;

    [ProtoMember(2)]
    public float Z;

    public static implicit operator Vec3Surrogate(Vec3 value) =>
        new Vec3Surrogate
        {
            XY = ((ulong)FloatToBits(value.Y) << 32) |
                FloatToBits(value.X),
            Z = value.Z,
        };

    public static implicit operator Vec3(Vec3Surrogate value) =>
        new Vec3(
            BitsToFloat((uint)value.XY),
            BitsToFloat((uint)(value.XY >> 32)),
            value.Z);

    private static uint FloatToBits(float value)
    {
        FloatUInt32 bits = default;
        bits.Float = value;
        return bits.UInt32;
    }

    private static float BitsToFloat(uint value)
    {
        FloatUInt32 bits = default;
        bits.UInt32 = value;
        return bits.Float;
    }

    [StructLayout(LayoutKind.Explicit)]
    private struct FloatUInt32
    {
        [FieldOffset(0)]
        public float Float;

        [FieldOffset(0)]
        public uint UInt32;
    }
}
```

The exact helper implementation may differ, but it must remain allocation-free and compatible with the current target framework. `DataFormat.FixedSize` is required on the `ulong`; serializing arbitrary float bits as a normal variable-length integer would lose the intended size benefit.

## Deliberate zero-value tradeoff

Do **not** add scalar fallback fields, `ShouldSerialize...` methods, or branches for the case where exactly one of `X` or `Y` is zero.

The optimization is aimed at ordinary movement positions and look directions, where exact component zeroes are not expected to be common enough to justify extra schema and implementation complexity. A rare exact-zero `X` or `Y` may serialize a few bytes larger than the current three-scalar representation. That is an accepted tradeoff unless production measurements demonstrate a material whole-packet regression.

Because this is a global surrogate, benchmarks should still include a representative `Mat3` or `MatrixFrame` payload. That verification is intended to expose a material real-world regression, not to require special-case zero fallbacks.

## Requirements

- [ ] Replace the current three-scalar `Vec3Surrogate` with one `ulong` containing the raw `X` and `Y` bits plus one normal `float` for `Z`.
- [ ] Encode the `ulong` using `DataFormat.FixedSize`.
- [ ] Allocate no array, list, memory stream, boxed value, or other managed object while converting between `Vec3` and `Vec3Surrogate`.
- [ ] Do not use `BitConverter.GetBytes`, since that allocates on the current target framework.
- [ ] Do not require unsafe compilation or add a runtime dependency solely for bit reinterpretation.
- [ ] Preserve `X` and `Y` bit patterns through a serialize/deserialize round trip.
- [ ] Preserve ordinary `Z` float round-trip behavior.
- [ ] Do not add exact-zero fallback fields or conditional serialization branches.
- [ ] Keep the separate `IsPacked = true` changes on repeated `ushort[]` and `long[]` packet fields; those changes are independent and are not part of this issue.
- [ ] Treat this as a network-schema change. Confirm that mixed mod builds are rejected before these packets can be exchanged; do not add a dual old/new wire format unless a concrete compatibility requirement is identified.
- [ ] Deliver this work in its own PR based on current `development`.

## Tests

Add focused tests covering:

- Representative positive and negative `X`, `Y`, and `Z` values.
- Fractional values and large finite values.
- Positive zero and negative zero.
- Representative `NaN` and infinity inputs where supported by existing serialization behavior.
- Bitwise preservation of `X` and `Y` after round trip.
- Semantic round-trip equality for normal finite vectors.
- An all-nonzero vector serializing to a 14-byte surrogate body.
- A vector with nonzero `X`/`Y` and zero `Z` serializing to a nine-byte surrogate body.
- An all-positive-zero vector remaining default/empty where protobuf-net normally omits default members.
- No managed allocation attributable to the implicit `Vec3` to surrogate conversion after warmup.
- Representative `AgentData`, mounted `AgentData`, `Mat3`, and `MatrixFrame` round trips.

Do not rely only on semantic packet round trips. Include direct size assertions for the surrogate and final packet-size comparisons for realistic workloads.

## Performance verification

Compare current `development` against the proposed implementation using the same machine and runtime configuration.

Report at minimum:

- Raw serialized size for representative `Vec3` values.
- Allocated bytes per 1,000 surrogate conversions.
- Serialization time per operation.
- Gen0 collections over a fixed high-volume workload where practical.
- Final `MovementPacketCompressor.Serialize` payload size for one-agent, three-agent, and realistic larger batches.
- Both on-foot and mounted movement snapshots.
- A representative non-movement payload containing `Mat3` or `MatrixFrame`.

The important movement comparison is the final post-LZ4 payload, because raw protobuf savings may not translate one-for-one after compression.

## Definition of done

- [ ] Normal all-nonzero `Vec3` representation is 14 bytes rather than 15 bytes.
- [ ] The surrogate conversion creates no managed allocation.
- [ ] No `float[]` or other collection is introduced.
- [ ] Round-trip and size tests pass.
- [ ] Final movement payloads improve or do not materially regress in representative batches.
- [ ] Representative global-surrogate payloads do not show a material whole-packet regression.
- [ ] Relevant `GameInterface`, `Missions`, integration, and E2E tests pass.
- [ ] The implementation PR includes benchmark results.

## Related work

- Related to #3076, which tracks allocations and copies inside the general `ProtoBufSerializer` buffering pipeline.
- This issue is distinct: it removes an allocation that would otherwise occur inside every compact `Vec3` surrogate conversion.
- Historical experiment: commit `9767d6c913ccfcd7d5bab0ad60ae23cfea0b46ba` on `network_optimizations`.

## Non-goals

- Reintroducing the packed `float[3]` implementation.
- Adding fallback fields for exact-zero `X` or `Y` values.
- Changing `Vec2Surrogate`.
- Rewriting protobuf-net or introducing a custom serializer.
- Changing movement batching, candidate sizing, cadence, prioritization, or LZ4 policy.
- Removing or reverting the valid `IsPacked = true` changes on repeated packet ID and sequence arrays.

ShoT-UPfps:
## PR Checklist

## PR Type

- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation update

Movement and other network messages encode ordinary three-dimensional vectors as three separate values, producing a 15-byte body for vectors whose components are all nonzero.

## What is the new behavior?

Resolves #3345

Movement and other network messages now use a smaller allocation-free vector representation while preserving the exact component values, including signed zero and non-finite values.

| Measurement | Current | New |
| --- | ---: | ---: |
| Ordinary nonzero vector body | 15 bytes | 14 bytes |
| Vector body with zero Z | 10 bytes | 9 bytes |
| Managed conversion allocation after warmup | 0 bytes | 0 bytes |
| Final 32-agent on-foot movement payload | 688 bytes | 662 bytes |
| Final 32-agent mounted-rider movement payload | 6,409 bytes | 6,276 bytes |
| Final 32-agent standalone-mount movement payload | 4,233 bytes | 4,164 bytes |
| Representative MatrixFrame body | 72 bytes | 68 bytes |
| Paired serialization median | 217.092 ns/op | 204.617 ns/op |

## Bot Changelog Entry

- Reduced network payload size for movement and other three-dimensional position updates without adding runtime allocations.

### DIFF
--- a/source/E2E.Tests/Services/Missions/MovementTrafficTests.cs
+++ b/source/E2E.Tests/Services/Missions/MovementTrafficTests.cs
@@ -1347,6 +1347,11 @@ public class MovementTrafficTests : MissionTestEnvironment
             Assert.Equal(movement.IdentityScopeId, roundTripped.IdentityScopeId);
             Assert.Equal(movement.AgentIds, roundTripped.AgentIds);
             Assert.Equal(movement.Agents.Length, roundTripped.Agents.Length);
+            for (int i = 0; i < agents.Length; i++)
+            {
+                Assert.Equal(agents[i].Position, roundTripped.Agents[i].Position);
+                Assert.Equal(agents[i].LookDirection, roundTripped.Agents[i].LookDirection);
+            }
             Assert.All(
                 roundTripped.Agents,
                 agent => Assert.Equal(
@@ -1398,10 +1403,20 @@ public class MovementTrafficTests : MissionTestEnvironment
 
             var serializer = new ProtoBufSerializer(new SerializableTypeMapper());
             var compressor = new MovementPacketCompressor(serializer);
-            AssertFitsAndDispatchesThroughRelay(serializer, compressor, packetManager, peer.NetPeer,
-                new MovementPacket("76561198000000042", ids, riders));
-            AssertFitsAndDispatchesThroughRelay(serializer, compressor, packetManager, peer.NetPeer,
-                new MountMovementPacket("76561198000000042", ids, mounts));
+            var restoredRiders = Assert.IsType<MovementPacket>(
+                AssertFitsAndDispatchesThroughRelay(serializer, compressor, packetManager, peer.NetPeer,
+                    new MovementPacket("76561198000000042", ids, riders)));
+            var restoredMounts = Assert.IsType<MountMovementPacket>(
+                AssertFitsAndDispatchesThroughRelay(serializer, compressor, packetManager, peer.NetPeer,
+                    new MountMovementPacket("76561198000000042", ids, mounts)));
+
+            for (int i = 0; i < ids.Length; i++)
+            {
+                Assert.Equal(riders[i].Position, restoredRiders.Agents[i].Position);
+                Assert.Equal(riders[i].LookDirection, restoredRiders.Agents[i].LookDirection);
+                Assert.Equal(mounts[i].MountPosition, restoredMounts.Mounts[i].MountPosition);
+                Assert.Equal(mounts[i].MountLookDirection, restoredMounts.Mounts[i].MountLookDirection);
+            }
         });
     }
 
@@ -1428,7 +1443,7 @@ public class MovementTrafficTests : MissionTestEnvironment
             LiteNetP2PClient.SelectDeliveryMethod(ordinaryUnreliable, datagramCeiling - 1, 0));
     }
 
-    private static void AssertFitsAndDispatchesThroughRelay(
+    private static IPacket AssertFitsAndDispatchesThroughRelay(
         ProtoBufSerializer serializer,
         MovementPacketCompressor compressor,
         IPacketManager packetManager,
@@ -1463,6 +1478,7 @@ public class MovementTrafficTests : MissionTestEnvironment
             Assert.Equal(1, restoredHandler.HandleCount);
             Assert.Same(sourcePeer, restoredHandler.SourcePeer);
             Assert.Equal(packet.GetType(), restoredHandler.Received.GetType());
+            return restoredHandler.Received;
         }
         finally
         {
--- a/source/GameInterface.Tests/Serialization/Vec3SurrogateTests.cs
+++ b/source/GameInterface.Tests/Serialization/Vec3SurrogateTests.cs
@@ -55,6 +55,17 @@ public class Vec3SurrogateTests
     }
 
     [Fact]
+    public void Serialize_UsesFieldTwoForZ()
+    {
+        Vec3Surrogate surrogate = new Vec3(1f, 2f, 3f);
+
+        using var stream = new MemoryStream();
+        Serializer.Serialize(stream, surrogate);
+
+        Assert.Equal((byte)0x15, stream.ToArray()[9]);
+    }
+
+    [Fact]
     public void ImplicitConversions_AllocateNoMemoryAfterWarmup()
     {
         var value = new Vec3(1024.25f, -2048.5f, 512.75f);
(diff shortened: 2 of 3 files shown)


Gerseras (approved):
Lgtm thanks for the bit shift idea 😁

PR #3473, opened 2026-08-30
[3388] Fix player barter in army softlocks

casey-johnson:
## PR Checklist

## PR Type

A player talking to a player army leader opens NPC bartering.
The player army leader trying to barter with a player army member softlocks the player army leader until the player army member leaves the army & has their menu update to see the barter
A player in an army attempting to trade with a player who is the army leader opens NPC bartering.

## What is the new behavior?

Resolves # 3388

Players are correctly routed through synced bartering

## Bot Changelog Entry

- Fixed players bartering with a player army leader from outside the army
- Fixed player army members bartering with army leader
- Fixed army leader soft-locking when trying to barter with a player army member


source/GameInterface/Services/MapEvents/PlayerPartyInteractions/PlayerPartyInteractionHandler.cs
```diff
@@ -47,6 +47,17 @@ internal class PlayerPartyInteractionHandler : IHandler
...
         typeof(TransferPrisonerBarterable).GetField("_prisonerCharacter", BindingFlags.Instance | BindingFlags.NonPublic);
 
+    // Parked, non-interactive campaign-map menus a player-party map conversation may safely open over: vanilla's
+    // "Talk to the army leader" opens a map conversation straight from army_encounter. army_dispersed and
+    // settlement menus (town/village) are deliberately excluded - they are handled elsewhere and out of scope here.
+    private static readonly HashSet<string> BenignConversationMenuIds = new HashSet<string>
+    {
+        "army_wait",
+        "army_wait_at_settlement",
+        "army_encounter",
```
ShoT-UPfps, line 57: this makes every `army_encounter` eligible without checking that it belongs to this session's `OtherPartyId`. the army-join path is approved without creating a P2P tracker entry, and custom P2P admission only checks that tracker, so player B can start an interaction while player A is already in an army encounter with C. A then opens B's dialog over C's menu; when B's session ends, `CloseLocalPlayerPartyEncounter` sees the existing `PlayerEncounter.Current` and finishes C's encounter. restrict `army_encounter` / `game_menu_army_talk_to_other_members` to the initiating request or verify that the current encountered/selected party matches the session counterpart, and add a three-player regression. the parked `army_wait*` carve-out can remain global.

ShoT-UPfps (commented):
reviewed `3521e746` head-only. the new client redirect correctly targets the installed v1.4.8 two-argument `ConversationManager.OpenMapConversation`, preserves attacker/defender ordering, and covers the parked `army_wait*` path. Build and Test #5397 passed build, unit tests, and all eight E2E shards.

one P2 remains: `army_encounter` and `game_menu_army_talk_to_other_members` are admitted by menu ID alone. an incoming player interaction can therefore open over an unrelated existing army encounter, and ending the incoming session finishes that pre-existing `PlayerEncounter`. scope those active encounter menus to the current session counterpart/initiating request and add a three-player regression.

source/GameInterface/Services/MapEvents/PlayerPartyInteractions/PlayerPartyInteractionHandler.cs
```diff
@@ -1148,18 +1159,71 @@ private void TryOpenPendingMapConversation()
...
+
+    // True when the local player is in a PlayerEncounter whose other side is not the party this interaction
+    // session is against.
+    internal static bool HasUnrelatedLiveEncounter(PartyBase sessionOtherParty)
+    {
+        if (PlayerEncounter.Current == null)
+        {
+            return false;
+        }
+
```
garrettluskey, line 1223: P2: `game_menu_army_talk_to_other_members` keeps the party originally encountered in `PlayerEncounter.EncounteredParty`; the selected attached party is separate and is the party this session targets. this comparison therefore rejects the normal "Talk to other members" route, `TryOpenMapConversation` retries forever, and the initiator never gets the dialog. scope this submenu to its selected/session party and add a regression that drives the selected-member route through session open and teardown.

garrettluskey (changes requested):
the redirect targets the current two-argument `OpenMapConversation`, and the encounter guard prevents teardown of a third party's encounter. one army-member route is still rejected by that guard, so this can still leave the interaction stuck.

source/GameInterface/Services/MapEvents/PlayerPartyInteractions/PlayerPartyInteractionHandler.cs
casey-johnson: Ahh the one thing I did not manually test was person outside army -> talk to not army leader, but another member. Good catch, I'll write up a fix.

source/GameInterface/Services/MapEvents/PlayerPartyInteractions/PlayerPartyInteractionHandler.cs
```diff
@@ -1148,18 +1159,76 @@ private void TryOpenPendingMapConversation()
...
+    }
+
+    // True when the local player is idle in a PlayerEncounter whose other side is not the party this
+    // interaction session is against.
+    internal static bool HasUnrelatedLiveEncounter(PartyBase sessionOtherParty, bool localPlayerInitiated)
+    {
+        if (localPlayerInitiated)
+        {
+            return false;
+        }
```
gamingmenacers-svg, line 1217: P2: scope this exemption to the original encounter, because `IsInitiator` stays true even if the player leaves that encounter and opens another army encounter before the session state arrives; `Handle_NetworkPlayerPartyInteractionState` does not consume the captured restart context, so this gate opens the old dialog and `CloseLocalPlayerPartyEncounter` later finishes the replacement encounter. retain the selected-member case, but reject a changed encounter and add a regression that replaces `PlayerEncounter.Current` between the request and session presentation.

gamingmenacers-svg (changes requested):
reviewed `6384d8b`. the selected army-member route now gets past the earlier party mismatch, but the initiator exemption still needs to be tied to the encounter that produced the request, as noted inline.

CodeFactor passes. the build fails with CS0246 for `CommandLineArgumentFunction` at `KingdomDebugCommand.cs:1086` and `:1136`, both already present in the base, so unit and E2E tests were skipped; please update the branch with the build fix and get those checks running before merge. this review checked the source, native army menu calls and CI logs; no local test run or live multiplayer reproduction was performed.

PR #3481, opened 2026-08-31
Fix smithing order completion popup

gamingmenacers-svg:
## PR Checklist

## PR Type

Completing a smithing order can remove the order and switch the screen back to free-build mode before the result popup is closed. Finalization then refreshes a stale order index and can leave the popup stuck.

## What is the new behavior?

Finalization uses the result popup's captured order mode, so completed orders clear their result without refreshing a removed order.

## Bot Changelog Entry

- Fixed the smithing order completion popup getting stuck after an order is completed.

## Other information

- Tests: `CraftingResultPatchesTests` (2 passed)


jordanbrymora (approved):

garrettluskey (approved):
the popup's captured order mode is the right source here, and the regression tests cover the screen mode changing before finalization. no issues found.

PR #3483, opened 2026-08-31
Accept settlement center for land entry

gamingmenacers-svg:
## PR Checklist

## PR Type

Land settlement entry validates only against the gate coordinate. A server-synced party at the settlement center can be rejected when the center and gate are farther apart than the interaction range.

## What is the new behavior?

Land entry accepts either the settlement center or gate coordinate. Port entry remains restricted to the port coordinate.

## Bot Changelog Entry

- Fixed valid settlement and siege entry being rejected when a party is at the settlement center.

## Other information

- Tests: `SettlementEncounterDistanceValidatorTests` (3 passed)


garrettluskey (approved):
the validator keeps port entry pinned to `PortPosition` and accepts both valid land convergence points, with coverage for the rejection cases. no issues found.

PR #3490, opened 2026-09-01
Keep empty battle host successor lists non-null

gamingmenacers-svg:
## PR Checklist

## PR Type

An empty battle host successor array can be omitted by protobuf. Because the message skips its constructor during deserialization, receivers then see a null list and host migration can fail while processing the assignment.

## What is the new behavior?

The serialized backing field is normalized by a non-null public accessor, including when protobuf omits an empty array.

## Bot Changelog Entry

- Fixed battle host migration when no backup host is available.

## Other information

Tested with a protobuf round-trip regression for an empty successor list.

### DIFF
--- a/source/E2E.Tests/Services/Missions/NetworkBattleHostAssignedSerializationTests.cs
+++ b/source/E2E.Tests/Services/Missions/NetworkBattleHostAssignedSerializationTests.cs
@@ -1,0 +1,28 @@
+﻿using Missions.Messages;
+using ProtoBuf;
+using System;
+using System.IO;
+using Xunit;
+
+namespace E2E.Tests.Services.Missions;
+
+/// <summary>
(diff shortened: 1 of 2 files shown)


garrettluskey (approved):
the backing accessor covers both constructor nulls and protobuf’s skipped-constructor empty case without changing tag 3. the zero-entry round trip passes in E2E shard 1, and consumers receive a non-null collection.

ShoT-UPfps (approved):
lgtm. `Handle_NetworkBattleHostAssigned` reads the accessor when it stores the assignment, so an omitted empty successor list stays usable on the client receive path.

garrettluskey (approved):
the accessor fixes the skipped-constructor empty-array case while keeping protobuf tags and wire bytes unchanged. the receive path uses the accessor, and empty and populated successors both survive the production serializer.

PR #3491, opened 2026-09-01
Disengage parties after leaving field battles

gamingmenacers-svg:
## PR Checklist

## PR Type

A server-approved field-battle leave closes the local encounter inside the network receive bypass. That suppresses the normal movement update, so the authoritative EngageParty order can remain and immediately reopen the encounter.

## What is the new behavior?

After the field-battle leave is applied, the leaving main party clears its engage order and publishes the hold through the authoritative party-behavior channel. Siege leaves and non-local parties keep their existing behavior.

## Bot Changelog Entry

- Fixed field-battle encounters reopening after a player leaves the fight.

## Other information

Tested end to end through the client request, server removal, broadcast, encounter close, and synchronized movement update.


source/GameInterface/Services/MapEvents/Handlers/BattleJoinLeaveHandler.cs
```diff
@@ -496,11 +498,23 @@ private void ApplyNetworkLeave(PartyBase party, bool leaveSiege, bool finishLoca
...
                 mobileParty?.SetMoveModeHold();
         }
+
+        // PlayerEncounter.Finish runs inside AllowedThread while applying the server's leave, so its normal
+        // postfix intentionally skips movement publication. Clear the stale engage order after leaving the
+        // receive-path bypass and send that change through the authoritative party-behavior channel.
+        if (fieldBattlePartyToDisengage?.Ai == null)
+            return;
+
+        fieldBattlePartyToDisengage.SetMoveModeHold();
```
garrettluskey: this makes the authoritative disengage a second client round trip: the server sends `NetworkPartyLeftBattle` while it is still `EngageParty`, and `fieldBattlePartyToDisengage` is never set if the owner already has no `PlayerEncounter`. clear the party on the server in the same `RemovePartyFromBattleAndBroadcast` action before sending the leave, otherwise that valid late/no-menu path remains engaged.

source/E2E.Tests/Services/MapEvents/BattleSurrenderTests.cs
```diff
@@ -430,6 +430,53 @@ public void RecipientSurrender_WithHealthyAllyOnSide_CapturesOnlyTheSurrenderer(
...
+        SetMockPlayerEncounter(leavingClient, mapEventId: setup.ctx.MapEventId);
+        foreach (var instance in Clients.Prepend(Server))
+        {
+            instance.Call(() =>
+            {
+                Assert.True(instance.ObjectManager.TryGetObject<MobileParty>(setup.initiatorPartyId, out var target));
+                Assert.True(instance.ObjectManager.TryGetObject<MobileParty>(setup.recipientPartyId, out var leavingParty));
+                using (new AllowedThread())
+                    leavingParty.SetMoveEngageParty(target, MobileParty.NavigationType.Default);
+
```
garrettluskey, line 466: this does not stage the state that reopens an encounter. vanilla `SetMoveEngageParty` resets `ShortTermBehavior` to `None` and only sets the default behavior and target, while `EncounterManager` checks the short-term engage state. stage a real short-term engage and run the encounter tick.

source/E2E.Tests/Services/MapEvents/BattleSurrenderTests.cs
```diff
@@ -430,6 +430,53 @@ public void RecipientSurrender_WithHealthyAllyOnSide_CapturesOnlyTheSurrenderer(
...
+                Assert.Equal(AiBehavior.EngageParty, leavingParty.DefaultBehavior);
+            }, MapEventDisabledMethods);
+        }
+
+        leavingClient.Call(() =>
+        {
+            Assert.True(leavingClient.ObjectManager.TryGetObject<MobileParty>(setup.recipientPartyId, out var leavingParty));
+            leavingClient.Resolve<IMessageBroker>().Publish(
+                this,
+                new PlayerLeaveBattleAttempted(leavingParty.Party));
```
garrettluskey: this bypasses `EncounterLeavePrefix`, but the real Leave option calls `ClearEngageOrder` immediately after publishing this event. invoke the patched menu consequence instead, otherwise the test only proves a helper-only path that players do not take.

garrettluskey (changes requested):
the receive-side cleanup is pointed at the right symptom, but the authoritative party is still left engaged for another server frame and the regression test does not exercise the real leave path. requesting changes so the hold is part of the server leave transaction and the reopen race is covered.

ShoT-UPfps (dismissed):
lgtm. vanilla `PlayerEncounter.Finish` doesnt hold, and `FinishPostfix` returns under AllowedThread, so leftover EngageParty would reopen on the next encounter poll. holding after the bypass and publishing `PartyBehaviorChangeAttempted` matches `FinishPostfix` / `ClearEngageOrder`.

source/E2E.Tests/Services/MapEvents/BattleSurrenderTests.cs
```diff
@@ -430,6 +430,53 @@ public void RecipientSurrender_WithHealthyAllyOnSide_CapturesOnlyTheSurrenderer(
...
 
+    /// <summary>
+    /// A server-approved field-battle leave must also clear the leaving party's authoritative engage order,
+    /// otherwise the encounter immediately reopens after its local menu closes.
+    /// </summary>
+    [Fact]
+    public void FieldBattleLeave_WithHealthyAlly_DisengagesLeavingPartyOnAllInstances()
+    {
+        var setup = SetupTwoOpposingPlayersInBattle();
+        JoinNewServerPartyToSide(setup.ctx.MapEventId, BattleSideEnum.Defender);
```
garrettluskey: this leaves `setup.recipientPartyId` as `DefenderSide.LeaderParty`; appending an ally doesnt make the recipient a joiner. `EncounterLeavePrefix` sends `MapEventFinalizeAttempted` for that party, so the direct `PlayerLeaveBattleAttempted` publish below skips the real menu path. make the client control a party joined after the leader and invoke `game_menu_encounter_leave_on_consequence` so this covers the reported leave.

garrettluskey (dismissed):
the receive-side hold is sent after leaving `AllowedThread` and that path looks sound. changes requested because the new E2E skips the real leave-menu path and sets up the leaving party as the side leader, so it doesn't cover the reported joiner leave.

ShoT-UPfps (approved):
Re-reviewed the latest head. The previous blockers are addressed:

- The server now clears both AI behavior and navigation state before broadcasting the leave, then publishes one authoritative behavior snapshot. The narrow `AllowedThread` scope suppresses duplicate movement publication while preserving the explicit post-scope replication.
- The E2E regression now uses a joined non-leader party, stages both default and short-term `EngageParty`, invokes the actual encounter-menu consequence, ticks `EncounterManager`, and verifies Hold plus cleared targets on the server and every client while the healthy side leader remains in battle.

Installed vanilla v1.4.8 confirms that `SetMoveModeHold` clears default and short-term behavior, navigation hold clears the movement target, and `HandleEncounterForMobileParty` can reopen an encounter when short-term `EngageParty` survives after `MapEventSide` removal. All required checks are green. Approved.

PR #3492, opened 2026-09-01
Protect player-visible wanderers from daily culling

gamingmenacers-svg:
## PR Checklist

## PR Type

The dedicated server runs Bannerlord's daily wanderer removal against Hero.MainHero. That leftover host hero does not represent the players, so an unhired wanderer can be removed while a player is visiting the same settlement.

## What is the new behavior?

The server mirrors the native selection and removal rules but protects a candidate when any player hero shares its current settlement. Settlement equality, including null equality, remains consistent with native behavior.

## Bot Changelog Entry

- Prevented tavern wanderers from disappearing while a player is visiting them.

## Other information

Tested the protected, unwatched, missing, hired, and null-settlement decision boundaries.


source/GameInterface/Services/Companions/Patches/Disable/DisableCompanionsCampaignBehavior.cs
```diff
@@ -71,6 +72,61 @@ public static bool DesiredTotalCompanionCountGetterPrefix(ref float __result)
...
+        bool sharesSettlementWithPlayer) =>
+        candidateExists && !isHired && !sharesSettlementWithPlayer;
+
+    internal static bool IsAnyPlayerAtSettlement(
+        Settlement settlement,
+        IEnumerable<Hero> playerHeroes)
+    {
+        foreach (Hero playerHero in playerHeroes)
+        {
+            if (playerHero.CurrentSettlement == settlement) return true;
```
garrettluskey, line 132: null isnt a shared settlement here. vanilla uses `candidate.CurrentSettlement == null || ...`, so a null candidate stays eligible regardless of the player's location. fired wanderers become fugitives with no settlement, and this protects them whenever any player is traveling; keep the candidate-null branch cullable and update the null test.

source/GameInterface.Tests/Services/Companions/Patches/Disable/CompanionsCampaignBehaviorPatchesTests.cs
```diff
@@ -3,13 +3,66 @@
...
 
 namespace GameInterface.Tests.Services.Companions.Patches.Disable;
 
-/// <summary>Verifies stale prisoner repair behavior.</summary>
+/// <summary>Verifies companion lifecycle replacement behavior.</summary>
 public class CompanionsCampaignBehaviorPatchesTests
 {
+    [Fact]
+    public void ShouldCullWanderer_PlayerSharesSettlement_ReturnsFalse()
+    {
```
garrettluskey, line 19: these only test helpers with precomputed booleans, so they never exercise `TryKillCompanionPrefix` or its role/random/template branches. add coverage around the prefix so the client suppression and server kill paths are tested.

garrettluskey (changes requested):
the server-only execution and the mirrored random/template selection look right. changes are needed because null is being treated as a shared settlement, and the tests never exercise the patched method.

ShoT-UPfps (changes requested):
null settlement still gets treated as shared, so a traveling player blocks the cull vanilla would still run (`CurrentSettlement == null` is always eligible). the prefix also still isnt covered by tests.

source/GameInterface/Services/Companions/Patches/Disable/DisableCompanionsCampaignBehavior.cs
garrettluskey, line 132: null isnt a shared settlement here. vanilla explicitly keeps the candidate eligible when `candidate.CurrentSettlement == null`, but this returns true whenever any player also has a null `CurrentSettlement`; require a non-null candidate settlement before comparing so settlement-less wanderers remain eligible for cleanup.

source/GameInterface.Tests/Services/Companions/Patches/Disable/CompanionsCampaignBehaviorPatchesTests.cs
```diff
@@ -3,13 +3,66 @@
...
+        Assert.False(CompanionsCampaignBehaviorPatches.ShouldCullWanderer(
```
garrettluskey: these tests only call detached helpers, so they can pass while `TryKillCompanionPrefix` has broken role, random/template, or kill wiring. add coverage that invokes the prefix and asserts the client suppression and server removal paths.

garrettluskey (dismissed):
the server-only prefix and the native random/template selection are kept intact. changes are still needed because null isnt a shared settlement, and the new tests dont execute the patched method.

source/GameInterface.Tests/Services/Companions/Patches/Disable/CompanionsCampaignBehaviorPatchesTests.cs
```diff
@@ -1,15 +1,100 @@
...
+/// <summary>Verifies companion lifecycle replacement behavior.</summary>
 public class CompanionsCampaignBehaviorPatchesTests
 {
+    [Fact]
+    public void TryKillCompanionPrefix_Client_SuppressesVanillaBeforeReadingServerState()
+    {
+        bool wasServer = ModInformation.IsServer;
+        try
+        {
+            ModInformation.IsServer = false;
```
ShoT-UPfps, line 23: This assignment changes the process-wide server/client role while `GameInterface.Tests` still runs test classes in parallel. The repository provides `ModInformationRoleCollection` specifically to serialize tests that mutate this flag; without it, another role-sensitive test can observe `IsServer == false` or overwrite the saved value, producing nondeterministic failures. Add `[Collection(ModInformationRoleCollection.Name)]` to this class.

ShoT-UPfps (changes requested):
The gameplay fix now matches the installed Bannerlord v1.4.8 implementation, including the corrected null-settlement behavior, and the earlier functional concerns are addressed. One deterministic test-safety issue remains: the new client-role test mutates the process-wide `ModInformation.IsServer` flag without joining the repository's `ModInformationRoleCollection`, so xUnit may run it concurrently with other role-sensitive tests and let those tests observe or overwrite the temporary client role. Add `[Collection(ModInformationRoleCollection.Name)]` to this test class (or otherwise avoid mutating the shared role) before merging.

ShoT-UPfps (approved):
Re-reviewed head `a1c81d0afffd4b805aa1b7d500b77039748c3a67` against base `6624f20db0c7af8d45ddc3cce814907b2bcb2846`. No blocking findings remain.

The latest commit addresses the outstanding test-isolation finding: `CompanionsCampaignBehaviorPatchesTests` now joins `ModInformationRoleCollection`, whose definition sets `DisableParallelization = true`, and the client-role test restores the original process-wide role in `finally`.

The gameplay replacement preserves the installed Bannerlord v1.4.8 random gate, template selection, first matching wanderer selection, hired-companion exclusion, null-settlement eligibility, and removal action. It substitutes the repository's player-hero collection for the single `Hero.MainHero` settlement comparison and suppresses vanilla execution on clients. In particular, a candidate with no settlement remains eligible even when a player is traveling.

The tests now exercise the actual client prefix and the shared server decision path for the random gate, empty template pool, null-settlement removal, and occupied-settlement protection. I also checked the prior reviews and discussion threads; the earlier functional findings are addressed.

Validation: exact-head source/diff review, surrounding repository code, direct installed-game managed-code comparison, and the successful current-head GitHub `Build and Test` workflow (run 33897925672). I did not run the test suite locally or perform an in-game reproduction.

PR #3494, opened 2026-09-01
Fix infinite synchronisation after player deletion

### LINKED ISSUE #3399: [CONNECTIONS] Infinite synchronization when a client connects after a player is deleted
## Severity of issue/bug

- [ ] High
- [ ] Average
- [ ] Low

An infinite synchronization occurs when a client connects after a player has been deleted.

## Steps to reproduce

Not provided

## Additional information

Not provided

## Discord source

https://discord.com/channels/709516043332354119/1542920517714055349/1542920517714055349

## Logs

Client log: [0.1.3+de68069945e8fd0593b098ffba2f891ead56b56b - Download Coop_client_9304.log](<https://cdn.discordapp.com/attachments/1542920517714055349/1542920603835564033/Coop_client_9304.log?ex=6aa4c88a&is=6aa3770a&hm=9da0cea75c4e1876c6e5b6d901e603e24b0b2836738bc03453f44c82be766bdc&>)

Server log: Not provided

## Images and videos

[](<https://cdn.discordapp.com/attachments/1542920517714055349/1543052786638258236/image.png?ex=6aa49ae4&is=6aa34964&hm=6a722628fd0424ce6b835b5eddee2c5a86085b0b5925bd321fc587e16c8be935&>) ([Discord message](https://discord.com/channels/709516043332354119/1542920517714055349/1543052786961485834))

jordanbrymora:
## PR Checklist

## PR Type

## What is the new behavior?

Resolves #3399

## Bot Changelog Entry

Fixed an issue where a deleted player could cause infinite synchronisation when trying to re-connect.


source/GameInterface/Services/MobileParties/MobilePartyRegistry.cs
```diff
@@ -138,8 +139,10 @@ public override void OnServerDestroyed(MobileParty obj, string id)
 
             messageBroker.Publish(this, new MobilePartyDestroyed(obj));
 
-            // Sole publisher of the PartyBase teardown: PartyBaseRegistry.DestroyMethods is empty, so
-            // a PartyBase's lifetime ends with its party's, while everything is still registered.
+            // These attachments have no destroy hooks, so their lifetime ends with their party.
+            messageBroker.Publish(this, new InstanceDestroyed<ItemRoster>(obj.ItemRoster));
```
garrettluskey, line 151: please add a regression test that registers all three rosters, destroys the party, and checks their ids are removed on the server and client. `DeletePlayerCommandTests` only checks the `MobileParty`, so this cleanup can regress without failing.

garrettluskey (changes requested):
the cleanup now follows the complete registered party graph, and the child destroy messages stay ordered ahead of the mobile-party destroy. requesting changes because the regression itself is still untested.

source/GameInterface/Services/MobileParties/MobilePartyRegistry.cs
garrettluskey, line 151: drop pending coalesced updates for all three roster ids before publishing these destroys. otherwise same-tick item updates and the deltas queued by `MemberRoster.Clear()` / `PrisonRoster.Clear()` flush after the destroy and clients look up ids that were already removed.

garrettluskey (changes requested):
the registry cleanup is covered now, but roster updates can still be queued behind the new destroy messages.

source/GameInterface/Services/MobileParties/MobilePartyRegistry.cs
jordanbrymora: Changed to drop roster ids from coalescer before instance destruction.

source/GameInterface.Tests/Services/MobileParties/MobilePartyRegistryCoalescingTests.cs
```diff
@@ -76,6 +79,78 @@ public void Destroy_DropsPendingBehaviorBeforeNetworkDestroy_AndFlushHasNoLateUp
...
+        var serverBroker = new MessageBroker();
+        var clientBroker = new MessageBroker();
+        using var serverNetwork = new TestNetwork();
+        var clientPeer = serverNetwork.CreatePeer();
+        var clientNetwork = Mock.Of<INetwork>();
+        var logger = Mock.Of<ILogger>();
+        var factory = Mock.Of<IAutoRegistryFactory>();
+
+        using var serverMobilePartyHandler = new AutoRegistryHandler<MobileParty>(
+            new MobilePartyRegistry(Mock.Of<IControllerIdProvider>(), serverBroker, logger, factory, serverObjectManager),
```
garrettluskey: this test passes no coalescer, and the other test only queues a party behavior update, so removing all three roster `DropInstance` calls still leaves both tests green. please inject a real coalescer, queue updates under all three compact roster ids, destroy the party, then flush and assert no roster update follows its destroy.

garrettluskey (changes requested):
the compact roster ids match the queued updates, and dropping them before the destroy messages addresses the previous ordering finding. requesting changes because that ordering fix still needs regression coverage.

garrettluskey (approved):
the new test covers the previous finding, all three compact roster ids are queued before destruction, neither side retains the roster registrations, and flushing sends no late roster updates. no further blocking findings.

build, unit tests and all e2e shards passed in ci. the verification-plan job failed because `.github/scripts/run-verification-plan.sh` is missing. local and live-game tests were not run.

PR #3496, opened 2026-09-01
Accept direct connection ports in server address

garrettluskey:
## PR Checklist

## PR Type

Direct connection asks for the server address and port in separate fields.

## What is the new behavior?

Direct connection uses one server address field. An address without a port uses 4200, and a custom port can be entered as `localhost:4300`. The last address used is saved in `coop_options.json` and filled in when the page is opened again.

## How to test

1. Open Join Co-op and select Direct.
2. Join a server on port 4200 with `localhost`.
3. Join a server on port 4300 with `localhost:4300`.
4. Reopen Join Co-op and confirm `localhost:4300` is filled in.

## Bot Changelog Entry

- Direct connection now accepts the port as part of the server address, uses port 4200 when omitted, and remembers the last address used.

### DIFF
--- a/UIMovies/CoopConnectionUIMovie.xml
+++ b/UIMovies/CoopConnectionUIMovie.xml
@@ -70,29 +70,6 @@
                         <TextWidget WidthSizePolicy="Fixed" SuggestedWidth="270" HeightSizePolicy="StretchToParent"
                                     Brush="Clan.TabControl.Text" Brush.FontSize="28"
                                     Brush.TextHorizontalAlignment="Right" Brush.TextVerticalAlignment="Center"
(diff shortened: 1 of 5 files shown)


ShoT-UPfps (approved):
Reviewed the complete diff and surrounding exact-head/base source. The combined server-address parser correctly handles the default port, explicit ports, raw IPv6, bracketed IPv6 with a port, and invalid port ranges. The connection path otherwise preserves the existing host-resolution and `AttemptJoin` behavior. Persistence uses the shared co-op options structure without discarding unrelated settings, and the screen lifecycle recreates the view model so the saved address is reloaded when reopened. All reported checks pass. No blocking findings.

PR #3515, opened 2026-09-03
Include co-op session data in bug reports

garrettluskey:
## PR Checklist

## PR Type

Diagnostic bug reports upload `coop_bug_report.sav` but omit its paired `coop_bug_report.json`, so the player-to-hero and co-op session state needed to reproduce the save is missing.

## What is the new behavior?

The server reads the paired co-op session file after creating the report save and includes both files in multipart uploads and fallback archives. A missing sidecar is logged but does not discard the campaign save. The consent text now discloses the paired session data, and the disclosure version is bumped.

The bug-report edge function was updated to version 27 to validate, store, and link the optional sidecar while remaining compatible with existing reports.

## How to test

- Submit a Coop Bug Report and confirm its Server Save section links both `coop_bug_report.sav` and `coop_bug_report.json`.
- Confirm a fallback archive contains both files under `server/`.
- `dotnet test source/GameInterface.Tests/GameInterface.Tests.csproj -c Release -p:NuGetAudit=false`

## Bot Changelog Entry

- Bug reports now include the co-op session data paired with the server save.

### DIFF
--- a/source/GameInterface.Tests/Services/BugReporting/BugReportServerSaveProviderTests.cs
+++ b/source/GameInterface.Tests/Services/BugReporting/BugReportServerSaveProviderTests.cs
@@ -1,0 +1,116 @@
+﻿using Common.Messaging;
+using GameInterface.Services.BugReporting;
+using GameInterface.Services.Heroes.Interfaces;
+using GameInterface.Services.Heroes.Messages;
+using Moq;
+using Serilog;
+using System;
+using System.IO;
+using System.Text;
+using Xunit;
+
+namespace GameInterface.Tests.Services.BugReporting;
+
+public class BugReportServerSaveProviderTests : IDisposable
+{
+    private readonly string tempRoot = Path.Combine(
+        Path.GetTempPath(), "BugReportServerSaveProviderTests_" + Guid.NewGuid().ToString("N"));
+    private readonly byte[] campaignData = Encoding.UTF8.GetBytes("current campaign save");
+    private readonly byte[] currentSidecarData = Encoding.UTF8.GetBytes("{\"players\":[\"current\"]}");
+    private string SidecarPath => Path.Combine(tempRoot, "coop_bug_report.json");
+
+    public BugReportServerSaveProviderTests()
+    {
+        Directory.CreateDirectory(tempRoot);
+        File.WriteAllText(SidecarPath, "{\"players\":[\"previous\"]}");
+    }
+
+    [Fact]
+    public void TryCapture_PreviousSidecarAndSuppressedCurrentWriteFailure_OmitsSidecar()
+    {
+        using var broker = new MessageBroker();
+        using var logger = new LoggerConfiguration().CreateLogger();
+        var writeFailed = false;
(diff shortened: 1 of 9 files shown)


ShoT-UPfps (changes requested):
Blocking — a stale sidecar can be paired with a fresh campaign save (`source/GameInterface/Services/BugReporting/BugReportServerSaveProvider.cs:42`). `Game.Save` publishes `GameSaved`, but `MessageBroker.Publish` catches and suppresses exceptions from `SaveGameHandler`. Because `coop_bug_report.json` is a fixed persistent filename, once a prior report has created it, a later `SaveCoopSession` serialization or write failure can leave that old valid JSON in place while `SaveCurrentGameToFile` still succeeds. `ReadSidecar` then uploads the previous report's player/session data alongside the newly created `.sav`, producing a mismatched diagnostic pair and potentially disclosing stale player data.

Please remove or invalidate the existing sidecar before starting the save, use a unique per-capture name, or otherwise require proof that the JSON was produced by this exact save operation. Add a regression test covering an existing sidecar plus a failure to write the current sidecar. Build, unit tests, and all E2E shards otherwise pass on the reviewed head.

ShoT-UPfps (approved):
Re-reviewed c2a0fe8a7a8ed4d4617d0a99a703088f27cf6b49 against d9a8e88bb92a902dcc4a158001299ea2293bd8f3. The previous stale-sidecar blocker is addressed; I found no remaining blocking code findings in the reviewed changes.

The provider now deletes the previous coop_bug_report.json before starting the campaign save. SaveInterface.DeleteSaveFile checks that the file is actually gone rather than trusting FileHelper's void deletion wrapper. If deletion fails, capture stops before saving or reading the stale sidecar. If the subsequent session write fails and leaves no readable nonempty sidecar, the current campaign save is still returned without the JSON. The new regression tests cover the suppressed GameSaved-subscriber write failure with a preexisting sidecar, successful replacement, and deletion failure preventing both save and read.

I also checked the exact-head Game.Save prefix, SaveGameHandler and CoopSaveManager path; matching save/sidecar filenames; multipart attachment and JSON metadata; fallback ZIP entry lifetimes and contents; the sidecar upload-size limit; and disclosure-version migration from 3 to 4. Both upload and archive tests verify the added attachment's bytes, and reports without a sidecar remain supported.

Validation: reviewed the complete diff, all nine changed files, all PR context pages, and the relevant save-path implementation. Installed v1.4.8 managed FileHelper/FileDriver behavior was inspected and its private evidence validated. On this head, build, unit tests, all eight E2E shards, and CodeFactor pass.

Outstanding verification caveat: the 'Verification plan and local harness profiles' job fails. Its job details show failures in 'Generate plan and run repository-owned profiles' and 'Upload verification plan and evidence'; I could not establish the underlying error from the available job details. This approval is not a claim that CI is fully green or that the PR should bypass its merge gates.

I did not rerun tests locally or exercise a live in-game capture/upload. The separately deployed edge-function v27 described in the PR is outside this diff and was not independently verified. No inline review threads exist to resolve.

PR #3520, opened 2026-09-05
Test loading-screen dismissal before character creation intro

garrettluskey:
#3475 reports intro audio behind the joining screen. #3400 already changed the lifecycle to hide that screen before starting character creation, but the tests didnt check the order.

## What is the new behavior?
checks that the loading screen has been hidden when `StartNewGame` runs, and that an unapproved character-creation event neither hides it nor starts a game. production code is unchanged.

## How to test
`ValidateModuleStateTests`: 22 passed in Release and 22 in Debug. `CharacterCreationIntroPatchTests`: 3 passed in Release.

live playback has not been tested. this protects the existing overlay fix, not a confirmed fix for the full video stall.

### DIFF
--- a/source/Coop.Tests/Client/States/ValidateModuleStateTests.cs
+++ b/source/Coop.Tests/Client/States/ValidateModuleStateTests.cs
@@ -139,14 +139,20 @@ namespace Coop.Tests.Client.States
             var validateState = clientLogic.SetState<ValidateModuleState>();
             var gameStateInterface = clientComponent.Container.Resolve<Mock<IGameStateInterface>>();
             IClientState? stateWhenGameStarted = null;
+            bool loadingScreenHidden = false;
+            bool loadingScreenHiddenWhenGameStarted = false;
             int startThreadId = 0;
             int gameThreadId = 0;
 
+            clientComponent.Container.Resolve<Mock<ILoadingInterface>>()
+                .Setup(x => x.HideLoadingScreen())
+                .Callback(() => loadingScreenHidden = true);
             gameStateInterface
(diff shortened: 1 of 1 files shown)


ShoT-UPfps (approved):
No actionable findings. Approved as test-only regression coverage.

Reviewed head `3495ea79ed7b060377a75ec127037e50652892ec` against base `f3a853bb35813d72529adb3aa8e16885807633dd`.

The new positive assertion snapshots the loading-screen flag inside the `StartNewGame` callback, rather than inspecting the flag only after execution. It therefore catches hiding the screen too late as well as not hiding it. The existing exactly-once checks for both calls, character-creation state checks, and game-thread check remain intact. This matches the current implementation: `BeginCharacterCreation` enters `CharacterCreationState`, whose constructor calls `HideLoadingScreen`, before invoking `StartNewGame`.

The unapproved-event test now checks both prohibited side effects in addition to retaining the validation state. The test broker dispatches subscribers synchronously, and the mock registrations expose the same instances used by the state machine.

Scope inspected: the complete diff and changed test file; all PR discussion/review/check pages; the relevant validation and character-creation states, client state machine, test component registrations, and test message broker.

Verification: all 12 reported checks on the reviewed head completed successfully, including build, unit tests, E2E shards 0–7, verification-plan/local-harness profiles, and CodeFactor. I did not independently rerun the tests or perform live gameplay testing.

Limitation: this protects the existing mocked loading-screen/startup ordering. It does not demonstrate that the full intro-video stall in #3475 is resolved; the PR description appropriately leaves live playback unverified.

PR #3524, opened 2026-09-05
Keep initialization party cache resets local

gamingmenacers-svg:
party cache resets publish synchronization events on the server even when its sync policy allows local initialization. this can send unregistered parties through the reset handler during setup.

## What is the new behavior?
respect the existing original-call policy before publishing the reset. native cache invalidation and normal running-server updates are unchanged.

## How to test
- six regression cases cover server/client roles, initialization policy and local-only scopes. four fail before the fix and all pass afterward.
- GameInterface.Tests: 1,315 passed, 11 existing skips. focused reset and policy tests: 12 passed.

## Bot Changelog Entry
- Avoid unnecessary party-cache synchronization during server initialization.

### DIFF
--- a/source/GameInterface.Tests/Services/MobileParties/ResetCachedPatchTests.cs
+++ b/source/GameInterface.Tests/Services/MobileParties/ResetCachedPatchTests.cs
@@ -1,0 +1,84 @@
+﻿using Autofac;
+using Common;
+using Common.Messaging;
+using Common.Util;
+using GameInterface.Policies;
+using GameInterface.Services.MobileParties.Messages;
+using GameInterface.Services.MobileParties.Patches;
+using System;
+using System.Collections.Generic;
+using TaleWorlds.CampaignSystem.Party;
+using Xunit;
+
+namespace GameInterface.Tests.Services.MobileParties;
+
+/// <summary>
+/// Checks that local cache initialization does not become a replicated party change.
+/// </summary>
+[Collection(ModInformationRoleCollection.Name)]
+public class ResetCachedPatchTests
+{
+    [Theory]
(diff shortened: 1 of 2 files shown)


ShoT-UPfps (approved):
Reviewed head `5dbd7d2fb52fd18b74261a57c3b524fe9aca6c92` against base `f3a853bb35813d72529adb3aa8e16885807633dd`.

No actionable findings. Approve.

The additional `CallOriginalPolicy.IsOriginalAllowed()` guard matches the existing lifecycle contract: `ServerSyncPolicy` permits local-only calls outside `ServerRunningState`, while normal running-server calls still publish. Client suppression is unchanged. Allowed-thread, all-thread save/load, and current-operation scopes now suppress this synchronization event consistently with that policy.

Traced the reset handler: publication otherwise reaches a registered-party ID lookup before a network broadcast, so avoiding local-initialization events prevents that unnecessary path. The change is solely in the postfix and does not skip the original cache invalidation. Installed v1.4.8 managed analysis corroborated the vanilla initialization/reset relationship; supporting evidence is retained privately.

The six regression cases cover ordinary server publication, client suppression, initialization-policy suppression, and all three local-only scopes. The scope cases also verify publication resumes after disposal. The tests use the nonparallel role-mutation collection and clean up their subscription and role/container state.

Verification: read the complete diff, both changed files, relevant policy/handler/test-isolation code, and all PR context sections. All 12 reported checks passed, including build, test, eight E2E shards, verification-plan/local-harness profiles, and CodeFactor. I did not independently execute tests or run a live server/client session; the test totals in the PR description remain author-reported.

PR #3525, opened 2026-09-05
Defer caravan trade item resolution until campaign apply

gamingmenacers-svg:
caravan trade messages resolve native items while decoding the packet. a crafted item can still be waiting in the game-thread queue, so the trade record loses its item before that creation runs.

## What is the new behavior?
keep item and modifier ids in the message and resolve them in the existing queued caravan apply. native StringIds and protobuf field numbers are unchanged. records with unavailable references are skipped.

## How to test
- six tests cover historical wire compatibility, unresolved ids, native registration and missing item/modifier references. both wire regressions fail before the fix.
- GameInterface.Tests: 1,315 passed, 11 existing skips on the unchanged rerun. the initial full run had two command-discovery failures; those six dispatcher tests pass in isolation. the intermittent failure cause is unconfirmed.

## Bot Changelog Entry
- Preserve caravan trade records when their crafted items are still being synchronized.

### DIFF
--- a/source/GameInterface.Tests/Serialization/CaravanTradeItemSerializationTests.cs
+++ b/source/GameInterface.Tests/Serialization/CaravanTradeItemSerializationTests.cs
@@ -1,0 +1,91 @@
+﻿using GameInterface.Services.Caravans.Messages;
+using GameInterface.Surrogates;
+using ProtoBuf;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using TaleWorlds.CampaignSystem;
+using Xunit;
+
+namespace GameInterface.Tests.Serialization;
+
+/// <summary>Checks that caravan trade messages retain unresolved item identities.</summary>
+public class CaravanTradeItemSerializationTests
+{
+    [Theory]
+    [InlineData(null)]
+    [InlineData("unregistered-caravan-modifier")]
+    public async Task Receive_PreservesUnknownItemsAndHistoricalWireShape(string? modifierId)
+    {
+        _ = new SurrogateCollection();
+        var original = new HistoricalMessage
+        {
+            PartyId = "caravan-party",
+            Rows = new List<HistoricalRow>
+            {
+                new HistoricalRow
+                {
+                    BoughtSettlementId = "town_ES1",
+                    BuyPrice = 14,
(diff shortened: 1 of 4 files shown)


ShoT-UPfps (approved):
## Review result

Approve — no blocking correctness findings in the reviewed revision.

Reviewed head `5c1d798c16638b4d8d81e86fb7700c31b2dec23c` against base `f3a853bb35813d72529adb3aa8e16885807633dd`. Read the complete diff, all four changed files, all PR context sections, and the relevant message, surrogate, crafting-handler, and game-thread implementations.

### Checks performed

- **Wire compatibility and identity:** `TradeActionLogData` retains its existing field numbers, and the field-4 payload retains the old surrogate's fields 1/2/3 and types. Outgoing item/modifier identities still come from native `StringId`, not compact co-op IDs. The historical test contract matches the actual message and base row layout.
- **Resolution timing:** the network message now contains only item identity data rather than an eagerly reconstructed roster element. The production receive handler calls `UnpackTradeActionLogData` inside its existing `GameThread.RunSafe` action. The client crafted-item creation handler also invokes registration inside `GameThread.RunSafe`, and the game-thread queue drains off-thread submissions in FIFO order. This addresses the stated case where item creation was queued first but had not executed when the trade packet was decoded.
- **Apply behavior:** successful lookup uses the same `ItemRosterElement(item, amount, modifier)` construction as the old surrogate. Missing items and explicitly referenced but unavailable modifiers are skipped; settlement lookup, prices, amount, and campaign time remain intact. Installed managed-game evidence confirms that the caravan rumor consumer assumes a valid trade item, supporting the missing-item guard.
- **Tests and CI:** reviewed both wire-compatibility cases and all four apply cases. All 12 reported checks on this exact head completed successfully, including build, test, eight E2E shards, verification-plan checks, and CodeFactor.

### Validation limits

I did not independently run the test suite or a live server/client session. The added apply tests call the unpack helper directly, so they are not an end-to-end queue-ordering test. A regression that queues item registration, decodes the trade message before draining, then applies the caravan update would strengthen coverage. This is a non-blocking coverage opportunity, not a demonstrated defect in this revision.

No inline findings or existing review threads require action. Current installed-game evidence is attached privately to the review handoff.

PR #3526, opened 2026-09-05
Synchronize militia reserves without counting troops twice

gamingmenacers-svg:
the militia setter broadcasts the requested total, but clients apply it as the ready militia reserve. a settlement with 100 militia troops and a requested total of 100.5 therefore receives a reserve of 100.5 instead of 0.5.

## What is the new behavior?
publish the remaining reserve after the native setter finishes. troop roster changes still synchronize separately, and client and initialization policy behavior is unchanged.

## How to test
- seven tests exercise the real patched native setter and reserve apply: active, inactive and absent militia parties, negative/zero/fractional reserves, and client/policy suppression. three regression cases fail before the fix.
- GameInterface.Tests: 1,316 passed, 11 existing skips. the fixture isolates and restores patch owners left by the patch-all tests.

## Bot Changelog Entry
- Fix clients counting settlement militia troops twice.

### DIFF
--- a/source/GameInterface.Tests/Services/Settlements/MilitiaSettlementPatchTests.cs
+++ b/source/GameInterface.Tests/Services/Settlements/MilitiaSettlementPatchTests.cs
@@ -1,0 +1,157 @@
+﻿using Autofac;
+using Common;
+using Common.Messaging;
+using Common.Util;
+using GameInterface.Policies;
+using GameInterface.Services.Settlements.Messages;
+using GameInterface.Services.Settlements.Patches;
+using HarmonyLib;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using TaleWorlds.CampaignSystem.Party;
+using TaleWorlds.CampaignSystem.Party.PartyComponents;
+using TaleWorlds.CampaignSystem.Roster;
+using TaleWorlds.CampaignSystem.Settlements;
+using Xunit;
+
+namespace GameInterface.Tests.Services.Settlements;
+
+/// <summary>Exercises the native militia setter and its replicated reserve.</summary>
+[Collection(ModInformationRoleCollection.Name)]
+public class MilitiaSettlementPatchTests
+{
+    static MilitiaSettlementPatchTests()
+    {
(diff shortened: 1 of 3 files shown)


ShoT-UPfps (commented):
## Review outcome

No blocking correctness or regression findings at head `599d1db3b255f377c432da3488b559c3f6fec623`, reviewed against base `f3a853bb35813d72529adb3aa8e16885807633dd`.

## What I verified

- The complete diff and all three changed files, the original base patch, all PR context sections, and the relevant server/client settlement handlers, network message, synchronization policy, existing militia integration test, and test collection configuration.
- The installed v1.4.8 managed implementation confirms that the public militia value includes active party members plus the reserve. Publishing `_readyMilitia` in the postfix fixes the total-versus-reserve mismatch and observes the value after the setter's possible troop conversion. The existing server and client handlers forward the float unchanged, and `RunMiltiaChange` assigns that float directly to `_readyMilitia` without rerunning conversion.
- The prefix initializes the per-call state and retains the existing original-call-policy and client suppression decisions. Both paths still execute the original setter and suppress publication. The new tests exercise the real patched setter for these paths and the fractional, negative, zero, inactive-party, and absent-party reserve cases.
- The fixture uses the nonparallel role collection and has cleanup for its subscription, process role, container, and captured Harmony patches.

## Non-blocking test improvement

`source/GameInterface.Tests/Services/Settlements/MilitiaSettlementPatchTests.cs:30-35`: all added numerical cases stay in the no-conversion portion of the native setter. They cover the double-counting regression, but do not directly prove the post-conversion timing that motivates the postfix. Add a case that actually crosses a whole-troop conversion threshold, then assert that the emitted value equals the final reserve and that client/server militia totals agree after applying the corresponding roster state. A removal/conversion case would strengthen this further. This is a coverage suggestion, not an identified production defect or merge blocker.

## Validation and limits

All 12 checks reported success on this exact head: build, test, eight E2E shards, verification-plan/local-harness profiles, and CodeFactor. There were no existing reviews, inline comments, or review threads to address.

I performed source review and installed-game managed analysis, not an independent test execution. The PR author's reported 1,316 GameInterface.Tests passes and 11 existing skips were not independently rerun. The existing militia integration test verifies message routing/counts, not the final converted roster/reserve invariant. Live server/client gameplay verification was not performed.

ShoT-UPfps (approved):

Issue #1318, opened 2026-06-11
Client attack-mission handler drifts from vanilla EncounterAttackConsequence


ShoT-UPfps:
## Description

`BattleHandler.OpenAttackMission` (the client field-battle path, invoked from `Handle_NetworkStartAttackMission`) is a hand-rolled copy of the field-battle branch of vanilla `Helpers.MenuHelper.EncounterAttackConsequence` and has diverged from it since the feature was first written — it was an incomplete port from inception, never a complete copy that was later changed. After investigation: most of the divergence is compensated for by the server-authoritative architecture; one drift point is a minor cosmetic downgrade, and one is a genuine (if narrow) correctness gap.

**Correction to the original framing:** the only live code path is the client's `OpenAttackMission`. The dedicated host has no party and never opens the encounter menu, so vanilla `EncounterAttackConsequence` never runs on the server — the patch's `if (IsServer) return true` branch is effectively dead. Client battle resolution arrives via the `NetworkMapEventFinalized` flow, not the client's own encounter state machine.

## Drift points (assessed)

**1. Missing tail — `PlayerEncounter.StartAttackMission()` + `MapEvent.PlayerMapEvent.BeginWait()` → harmless, by design.**
- `StartAttackMission()` only does `_campaignBattleResult = new CampaignBattleResult()`. Clients don't resolve battles locally (`MapEvent.Update` is a client no-op, `OnBattleWon` is skipped, the client exits via `NetworkMapEventFinalized → PlayerEncounter.Finish`), so the null result is inert and `DoWait` safely skips `ContinueBattle`.
- `BeginWait()` only does `State = MapEventState.Wait`. `MapEvent.State` is already set to `Wait` at map-event creation server-side and is a replicated DynamicSync property, so the client receives `Wait` via sync regardless.
- No behavior change — can be left as-is.

**2. Caravan/villager mission type → minor / cosmetic downgrade.**
- Vanilla dispatches to `OpenCaravanBattleMission` when defenders are caravans/villagers; the client always calls `OpenBattleMission`. The only difference is two non-mechanical mission behaviors: `CaravanBattleMissionHandler` (decorative scattered-goods prop + non-interactable pack animals) and `MissionCaravanOrVillagerTacticsHandler` (a defensive-line AI tactic). Combat spawns, win/lose conditions, casualties, and loot are identical.
- Reachable only from a loaded save that already contains caravan/villager parties — fresh co-op games never spawn them (their spawn behaviors are disabled).

[...]

PR #3529, opened 2026-09-05
Add mission receive-path diagnostics

garrettluskey:

mission packets from unmapped peers and Steam-to-UDP forwarding failures can disappear without useful logs. #1318 shows one client losing allied spawns after its Steam mission connection takes over, but the logs dont establish where delivery stops.

related: https://github.com/Bannerlord-Coop-Team/Bannerlord-Coop-Issues/issues/1318 and https://github.com/Bannerlord-Coop-Team/Bannerlord-Coop-Issues/issues/1332

## What is the new behavior?

adds `[ReceivePath]` logs for peer mapping, selected send routes, mapped/unmapped receives, and Steam UDP forwarding results. first events are logged immediately, repeats are aggregated every ten seconds while traffic continues, and teardown records totals. includes UTC and connection/mission identifiers, and removes the misleading direct-P2P readiness claim.

diagnostics only, routing and identity checks are unchanged. this does not claim to fix either report.

## How to test

42 focused tests passed across counter, Steam tunnel, mission routing/batching and disconnect tests. Release and Debug compile-only builds passed.

live test not run: use an in-game `/server` and two Steam clients, enter the same field battle near Danustica and complete deployment. run `coop.debug.battle.state` and `coop.debug.movement.state` on both clients, keep traffic running for 30 seconds, then leave and join another battle. preserve both client logs and the server log. compare `[ReceivePath]` mapping and route records, `SteamReceive`, `UdpForwarded`, `UdpForwardFailed`, and `UnmappedDrop`, alongside spawn transfer ids. check that rejoining starts fresh diagnostic lifetimes. forwarded UDP and mapped receives are not gameplay-delivery acknowledgements.

### DIFF
--- a/source/Common.Tests/Logging/ReceivePathDiagnosticsTests.cs
+++ b/source/Common.Tests/Logging/ReceivePathDiagnosticsTests.cs
@@ -1,0 +1,80 @@
+﻿using Common.Logging;
+using Serilog;
+using Serilog.Core;
+using Serilog.Events;
+using System.Collections.Concurrent;
+using System.Net.Sockets;
+using Xunit;
+
+namespace Common.Tests.Logging;
+
+public class ReceivePathDiagnosticsTests
+{
+    [Fact]
+    public void Categories_EmitFirstEvidenceThenAggregateUntilIntervalOrTeardown()
+    {
+        long now = 0;
+        var sink = new CaptureSink();
+        using var logger = new LoggerConfiguration().WriteTo.Sink(sink).CreateLogger();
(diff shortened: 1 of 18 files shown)


gamingmenacers-svg (commented):
reviewed `3881a0fde`, no repair findings in the current diff. the three `ReceivePathDiagnosticsTests` passed locally and all current CI checks passed, including all eight E2E shards. the live Steam battle test is still needed to verify the logs across an actual disconnect and rejoin.

gamingmenacers-svg (approved):

PR #3538, opened 2026-09-06
Add server/client availability to co-op commands

garrettluskey:
## Summary
- Add required `ICoopCommand.Side` metadata (`Server`, `Client`, or `Both`) and expose it through command descriptors.
- Reject wrong-side commands centrally before argument validation or execution, returning `command_wrong_side` with exactly `This command is only available on the server` / `This command is only available on the client`.
- Update every command and test implementation. Preserve dual-role diagnostics and conditional read/write commands; retain helper guards for direct callers.
- Add role-matrix, role-change, metadata-validation, production classification, and wrong-side regression tests.

## Verification
- Full solution builds in Release and Debug (`dotnet build source/Coop.sln -c <configuration> -p:Platform=x64 -p:ModName=`); deployment disabled.
- Common.Tests Release: **164 passed**.
- GameInterface.Tests filtered to `FullyQualifiedName~Commands|FullyQualifiedName~LiveTestCommandDispatcherTests`: **78 passed Release**, **90 passed Debug**.
- `git diff --check` passed.
- Existing dependency/framework/analyzer warnings remain. No in-game smoke test or deployment performed.

### DIFF
--- a/source/Common.Tests/Commands/CoopCommandRegistryTests.cs
+++ b/source/Common.Tests/Commands/CoopCommandRegistryTests.cs
@@ -4,8 +4,135 @@ using Serilog;
 
 namespace Common.Tests.Commands;
 
-public class CoopCommandRegistryTests
+[CollectionDefinition("ModInformation", DisableParallelization = true)]
+public class ModInformationCollection
 {
+}
+
+[Collection("ModInformation")]
+public class CoopCommandRegistryTests : IDisposable
+{
+    private readonly bool originalIsServer = ModInformation.IsServer;
+
+    public void Dispose()
+    {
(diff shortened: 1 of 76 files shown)


ShoT-UPfps (approved):
Reviewed the full PR diff and repository context. The centralized side check runs before argument validation/execution, descriptor metadata is snapshotted and validated, and the production command classifications preserve existing dual-role commands while marking commands with explicit client/server restrictions appropriately. I also checked ambiguous dual-role town mutation/debug commands against their implementations; these had no pre-existing side guard, so `Both` preserves their existing availability rather than introducing new authority behavior. All reported checks on the reviewed head are green. No blocking findings.

PR #3540, opened 2026-09-06
Improve staged MCP runs and menu automation

garrettluskey:
## What is the current behavior?

MCP starts every requested process together, screenshots need separate polling/conversion, and an exited game root can leave its watchdog alive. This builds on merged #3536.

## What is the new behavior?

- add clients to an owned run as needed, with read-only commit-headroom and matching DEBUG bridge checks before launch/join.
- return completed PNG images with artifact metadata, preserve operation errors, and confirm owned process-tree cleanup before reporting stopped.
- list local saves read-only and optionally select a server startup save without silent fallback.
- add bounded Gauntlet menu inspection/actions, Debug/Release `/autoconnect` endpoints, and repo-local Pi/Codex setup. no VOIP changes.

Existing `start_run` and async screenshot calls remain available. The new tools require a matching DEBUG bridge; truncated UI trees remain inspection-only.

## How to test

- MCP Release tests: 165/165, including official-SDK output schemas, PNG content, preflight, saves and owned-job cleanup. UI Debug: 63/63. Launch arguments: 43/43 each in Debug and Release. Guarded DEBUG build passed without deployment.
- Direct Pi MCP: select `MP`, start server only, add client1, join once, inspect options, click the discovered Map Time tab, capture the rendered PNG, close without Apply, then stop. Both reached the same campaign; actual server load logs name `MP`. Readiness alone does not confirm the filename.

```text
campaign: xTyMXZwWIn2S (server and client1)
options snapshot: 422 widgets, not truncated
capture: image/png, 1440x1080, 1539760 bytes, frames 2825 to 2845
stop: both processTreeAlive=false, cleanupComplete=true
independent process scan: []; module restoration differences: {}
```

`MP.sav` and Configs were unchanged. Ordinary generated `TransferSave.json` changed and was preserved. Historical text/slider/toggle/scroll checks were not repeated on the combined candidate. Release `/autoconnect` runtime, actual Codex startup and clean-package runtime remain unrun.

PR #3541, opened 2026-09-06
Add proximity voice chat

garrettluskey:

nearby players have no in-game voice chat over the co-op connection.

## What is the new behavior?

adds proximity voice for campaign, settlement, battle and tournament instances, with local voice settings, native push-to-talk binding, microphone test, talking-player list and server voice control. Voice settings save immediately; closing or Cancel does not undo them. other options tabs remain staged until Apply.

depends on #3540. this PR includes its MCP ancestry until that PR merges; the incremental VOIP change is 3306bfd6aca380a8da7db2a42b0d32555b295f87.

## How to test

536 reviewed tests passed before the final docs-only correction: 167 MCP, 230 voice/UI, 73 routing/args/container, 64 wire/registry and 2 scene-source tests. the final correction removes stale Refresh/retry instructions and clarifies immediate saving; no production code changed after those checks.

synthetic-pipeline screenshots cover the latest Voice menu and one client's Danustica tavern HUD, not a physical microphone test. physical-mic, dedicated-server and ten-player audio checks remain unrun; the reverse-view second-client screenshot is missing. see [Proximity voice](doc/ProximityVoice.md) for controls and test steps.


source/GameInterface/Services/Voice/VoiceAudio.cs
```diff
@@ -0,0 +1,519 @@
...
+        foreach (var pair in speakers)
+        {
+            if (now - pair.Value.LastArrival > VoiceJitterBuffer.MaximumAgeMilliseconds)
+            {
+                expired.Add(pair.Key);
+                continue;
+            }
+            if (!current.Position.CanHear || options.Deafened || testing || now - current.SampledAt > 100) continue;
+            if (!pair.Value.Buffer.TryRead(now, out var audio)) continue;
+            var samples = pair.Value.Decoder.Decode(audio);
```
ShoT-UPfps: When a structurally valid relayed packet causes `Decoder.Decode(audio)` to throw, this exception escapes `Mix` to the worker-wide catch and `SetFailure`. On the next iteration the failed state disposes both microphone capture and playback and clears every speaker, so one bad stream takes down all voice on the receiving client. The relay checks packet shape/size and permissions, not Opus contents, and `OpusVoiceCodecFactory` explicitly throws when the decoded frame duration is not 960 samples. Catch decode failures at this per-speaker boundary and drop/reset only the offending stream rather than entering the device-failure path. Add a bad-frame/throwing-decoder test alongside a healthy second stream and assert that healthy playback and local capture continue without global device disposal.

source/GameInterface/Services/UI/CoopOptions/Providers/VoiceTab/Sections/VoiceSection.xml
```diff
@@ -0,0 +1,147 @@
...
+                        </Widget>
+                        <Widget WidthSizePolicy="StretchToParent" HeightSizePolicy="Fixed" SuggestedHeight="12" MarginBottom="2" Sprite="BlankWhiteSquare_9" Color="#262626FF">
+                            <Children>
+                                <Widget WidthSizePolicy="StretchToParent" HeightSizePolicy="StretchToParent" Sprite="BlankWhiteSquare_9" Color="#3FBE64FF" IsVisible="@InputLevelLow" />
+                            </Children>
+                        </Widget>
+                    </Children>
+                </ListPanel>
+                <TextWidget WidthSizePolicy="Fixed" SuggestedWidth="60" HeightSizePolicy="Fixed" SuggestedHeight="24" VerticalAlignment="Top" Brush="SPOptions.OptionName.Text" Brush.FontSize="16" Text="@InputPeakText" />
+                <ButtonWidget WidthSizePolicy="Fixed" SuggestedWidth="245" HeightSizePolicy="Fixed" SuggestedHeight="42" VerticalAlignment="Center" Brush="ButtonBrush1" Command.Click="ExecuteTest" MarginRight="12">
```
ShoT-UPfps, line 138: This layout exposes `ExecuteTest` but never binds a control to `ExecuteRetry`, although voice errors instruct the player to select Retry. `VoiceSection.ExecuteRetry` is the implemented device-refresh/recovery path, and `VoiceClient.Retry` is the path that clears the client-level error which otherwise makes `Tick` return immediately. Applying settings, starting microphone testing, or reopening this menu does not clear that error, so those controls are not equivalent recovery. Add a visible Retry/Refresh button bound to `ExecuteRetry`, regenerate the movie, and add a binding/recovery regression test that reaches the action through the UI, refreshes microphone choices, and preserves healthy reception.

ShoT-UPfps (changes requested):
## Decision: request changes

Reviewed head `235eb7909354c30899b65ffff57b6a0f15a32bb2` against base `30350115b5f5c2229570212b1d592f8e2dabfacc`.

Completed the full 143-file, 12,704-line raw diff, including the MCP ancestry from #3540, all commit/check/discussion context, and the relevant installed-game integration checks. Two functional issues need correction before approval:

1. **P1 — isolate received-frame decoder failures.** An exception from one speaker's decoder escapes `Mix` into the worker-wide `SetFailure` path. That disables capture and playback for the recipient instead of discarding the offending frame/stream. The relay's structural packet validation does not validate Opus contents, and the codec wrapper also explicitly throws for a decoded frame with the wrong duration. Handle this at the per-speaker decode boundary, preserving healthy streams and local microphone capture.
2. **P2 — expose the implemented Retry action.** The Voice layout has a microphone-test control but no binding to `ExecuteRetry`, even though failures tell the player to select Retry. This is more than missing convenience: `VoiceClient.Retry` clears the client error that otherwise makes `Tick` return immediately, while ordinary Apply/test/menu reopening paths do not clear it. The same action refreshes the cached device choices.

## Implementation handoff

Address both inline findings in focused changes. Add a regression test with a decoder that throws for one stream while a second stream remains healthy: playback and capture must continue, and the global audio-failure/device-disposal path must not run. Also cover a structurally valid received packet rejected by the codec, rather than relying exclusively on valid encoder output.

Add a visible Retry/Refresh control bound to `ExecuteRetry` in the source Voice layout and regenerate the corresponding movie. Add a layout-binding test and a recovery test that exercises the user-accessible action after a latched client error or microphone reconnect. Verify device choices refresh and microphone retry preserves healthy reception.

## Validation and limits

The 12 checks reported for this exact head are successful, including build, tests, verification, E2E jobs, and CodeFactor; these were inspected, not rerun by this review. Current installed v1.4.8 managed evidence validates the camera and popup-cleanup assumptions. The decoder failure is a code-path finding, not a reproduced live-network exploit.

Physical microphone use, dedicated-server audio, and ten-player live audio remain unverified, as acknowledged in the PR. Synthetic codec/device tests and the reported screenshots do not establish those outcomes. No game deployment, source edits, merges, or test execution were performed during this review.

source/GameInterface/Services/Voice/VoiceAudio.cs
garrettluskey: fixed in 3e188823a. decode failures now discard only the affected stream. regression tests cover a throwing decoder and a valid opus packet with the wrong duration while healthy playback and microphone capture continue.

source/GameInterface/Services/UI/CoopOptions/Providers/VoiceTab/Sections/VoiceSection.xml
garrettluskey: fixed in 3e188823a. added Retry / Refresh bound to `ExecuteRetry` and regenerated the movie. tests invoke the bound action to clear a latched error, refresh microphone choices, and restore capture without interrupting reception.

gamingmenacers-svg (commented):
reviewed the voice-specific changes through 3e188823a055615512563b102ba4261ba4a77fcd, including server routing/context fences, reconnect generations, capture/playback lifecycle, settings and keybinding integration. the decoder exception now evicts only the offending stream, and both source/generated settings layouts expose ExecuteRetry. the regression tests cover continued healthy playback/capture and recovery through that button. i found no additional blocking issue in those paths.

local Release results: 39 Common voice tests, 120 GameInterface voice/keybinding tests and 31 Coop voice/container/client-logic tests passed, with no skips. the normal GameInterface test command stopped during restore because Scriban 7.2.0 audit warnings (NU1902/NU1903) are errors in Coop.Core; i reran with -p:TreatWarningsAsErrors=false for diagnostic validation, then ran the Coop tests against those built assemblies. this is not a clean default-build result. the existing GitHub checks are green.

this is a code and automated-test review of the voice feature, not live-audio sign-off or a separate review of the inherited #3540 MCP changes. physical microphone/device recovery, two-way multiplayer audio, dedicated-server and ten-player audio validation remain outstanding as documented; those should be demonstrated before calling this fully end-to-end validated.

ShoT-UPfps (dismissed):

ShoT-UPfps (dismissed):

ShoT-UPfps (dismissed):

ShoT-UPfps (approved):

PR #3544, opened 2026-09-07
Prevent duplicate poller loops after a restart

### LINKED ISSUE #3114: [ISSUE] Make Poller restart and disposal deterministic to prevent duplicate loops
# For New Contributors

## Description

`Poller.Start()` currently restarts an existing poller by:

1. Calling `Stop()`, which only cancels the current `CancellationTokenSource`.
2. Immediately replacing the mutable `cts` field.
3. Starting a new `Task.Run(PollAsync)`.

`PollAsync` does not capture the token/source that belongs to that invocation. It repeatedly reads the mutable `cts` field in both the loop condition and `Task.Delay`.

This creates a restart race:

- The old loop is cancelled.
- `Start()` replaces `cts` with a new uncancelled source before the old loop observes cancellation.
- The old loop reads the new source and can continue running alongside the newly started loop.

Once that happens, only the newest `pollingTask` is tracked. `StopAndWait` can wait for the newest task while an older orphan loop continues invoking the polling callback.

Additional lifetime problems:

- Replaced `CancellationTokenSource` instances are never disposed.
- `Stop()` does not clear or await task ownership.
- `IsRunning` reports the current `cts` state, not whether exactly one task is alive.
- Repeated start/stop cycles have no serialized ownership state.

This is conditional on restart timing, but it can create duplicate network/event polling, repeated callbacks, stale mission access, and retained task/closure state.

## Intended Design

Each polling-loop invocation must own immutable loop state:

- Its own `CancellationTokenSource`.
- Its own captured `CancellationToken`.
- Its own tracked `Task`.

Serialize `Start`, `Stop`, `StopAndWait`, and disposal under a lifecycle gate.

A restart must not launch the replacement until the prior loop has been cancelled and ownership has been safely transferred. The prior source must be disposed after its task has finished.

Define and document exact semantics:

- `Start()` when stopped.
- `Start()` when already running.
- `Stop()` as non-blocking cancellation.
- `StopAndWait()` as bounded deterministic shutdown.
- `Dispose()` as final cancellation/wait/resource cleanup.
- Calls made from inside the active polling callback, which must not self-deadlock.

Do not retain an old task/source after restart. Do not make every existing `Stop()` call synchronously block the game thread without auditing callers.

## Location

Primary location:

- `source/Common/Util/Poller.cs`

Audit all poller owners, especially:

- Network polling in `CoopNetworkBase`
- Mission movement polling
- Packet profiling or diagnostics polling
- Disposal paths that currently call `Stop()` versus `StopAndWait()`

## Related Issues

- `#1645` / PR `#1649` demonstrate why deterministic poller shutdown matters when callbacks access mission/native state.
- This issue is about the reusable `Poller` lifecycle itself, not that already-fixed battle teardown call site.

## Requirements

- Every loop captures the token belonging to that loop.
- At most one callback loop can be active for a `Poller` instance.
- Restart cannot orphan the previous loop.
- Replaced cancellation sources are disposed.
- Final disposal releases the task, source, callback reference where appropriate, and other lifecycle state.
- `StopAndWait` waits for the correct loop.
- `IsRunning` reflects actual lifecycle state.
- Repeated `Start` / `Stop` / `Start` cycles are deterministic.
- Calling shutdown from the active callback does not deadlock.
- Exceptions continue to be logged and do not silently kill the loop unless disposal/cancellation is underway.
- Existing polling interval behavior remains materially equivalent.

## Verification

Create deterministic tests using barriers/manual events rather than timing guesses:

- Start a loop and block inside its callback.
- Call `Start()` again while the old callback is in flight.
- Prove a second loop does not run concurrently.
- Release the old callback and prove restart completes with exactly one active loop.
- Repeatedly restart hundreds or thousands of times and assert maximum concurrent callbacks is one.
- Verify `Stop()` prevents future ticks.
- Verify `StopAndWait()` waits for an in-flight tick.
- Verify callback-initiated stop does not self-deadlock.
- Verify disposal is idempotent.
- Verify old `CancellationTokenSource` instances/tasks are no longer retained.
- Verify an exception is logged and the intended loop continues.
- Add tests for callers that require non-blocking stop semantics.

## Definition of Done

- [ ] A Poller instance can never own two live polling loops.
- [ ] Restart waits/transfers ownership safely.
- [ ] Cancellation sources and tasks are deterministically cleaned up.
- [ ] Existing poller users have been audited for correct stop semantics.
- [ ] Common, Missions, Coop.Core, integration, and E2E tests remain green.

Thomas-at-sb:
## PR Checklist

## PR Type

`Poller.Start()` cancels the running loop and then replaces the `cts` field, and
`PollAsync` re-reads that field in its loop condition (`Poller.cs:82`) and in
`Task.Delay` (`Poller.cs:118`). A loop that sits in its callback while the swap
happens comes back to the new, uncancelled source and keeps polling next to the
replacement, because it never owned the token it was cancelled with. Only the
newest `pollingTask` is tracked, so `StopAndWait` reports success while the older
loop still calls the polling function.

`AGENTS.md:23` states that the single `Poller` task drives
`netManager.PollEvents()` so receive callbacks fire one at a time, never
concurrently. The mission client sends and receives on that path
(`LiteNetP2PClient.cs:724` wraps an `IMessage` in a `MessagePacket`, `:1020`
publishes what arrives), and it is the one that restarts: its `Poller` is built per
client instance (`LiteNetP2PClient.cs:127`) and the client is registered per
lifetime scope (`MissionModule.cs:58`), the `Start`/`Stop` pair runs on every
mission change (`BattleInstanceLifecycle.cs:102` and `:121`,
`CoopLocationsController.cs:424` and `:507`, `CoopTournamentController.Initialize`
and `OnLeaving`), and `LiteNetP2PClient.cs:164` only cancels. Two loops then call
`netManager.PollEvents()` (`LiteNetP2PClient.cs:237`) at the same time.

That does not deliver a message twice. `PollEvents` detaches the whole pending
chain under `_eventLock` and processes it outside the lock, so the two loops split
the queue between them. What breaks is the ordering and the
one-at-a-time property `AGENTS.md:23` describes: a loop held up inside a receive
callback lets the other loop take everything that arrived after it and publish that
to the `MessageBroker` first, and receive callbacks of both loops run at the same
time.

The four other pollers do not restart, so they are unaffected.
`CoopNetworkBase.StartNetworkPoller()` (`CoopNetworkBase.cs:92`) allows a single
start through a `CompareExchange`, `PacketProfiler` starts once in its constructor
(`PacketProfiler.cs:45`), and `SteamTunnelClient.cs:62` and `SteamTunnelHost.cs:99`
build a fresh `Poller` on every start. `LiteNetP2PClient.cs:164` is also the only
teardown that stops on cancellation alone, because `CoopNetworkBase.cs:112`,
`PacketProfiler.cs:147`, `SteamTunnelClient.cs:152` and `SteamTunnelHost.cs:138`
already use `StopAndWait`.

## What is the new behavior?

Resolves #3114

`Start` cancels the current source and schedules `RunAsync(previousTask, source)`,
which waits for its predecessor, polls with `source.Token` instead of re-reading
the field, and in its `finally` clears `cts` only when no restart has already put
a newer source there, then disposes its own source under the lock. The completed
task stays tracked until the next start. `Stop` and `StopAndWait` share
`CancelAndCaptureTask`, so the cancel and the capture happen under the lock while
the wait and the self-wait check stay outside it. The api is unchanged, and `Stop`
still only cancels, so the teardown at `LiteNetP2PClient.cs:164` keeps the
behavior it has today.

The trade is that a restart waits for the tick in flight, so a polling function
that hangs now stops the replacement from starting while `IsRunning` still
reports true. It is the same callback that would have hung before, without a
second loop next to it. Note that the `StopAndWait` timeout now also covers a
queued successor, not just the loop being stopped.

Two requirements from the issue are not covered. There is no disposal api, so the
task, the source and the callback reference are not released on a final teardown,
and `IsRunning` still reports the state of the current source rather than of the
loop. Neither is added here, because this change keeps the existing api.

`PollerTests` keeps the restart cases and adds a restart started from inside the
polling function, which cancels the very loop it is called from.

## Bot Changelog Entry

- Fixed a second background loop polling mission network events after leaving and
  re-entering a battle, tavern or tournament, which delivered messages out of send
  order and ran receive callbacks at the same time.

### DIFF
--- a/source/Common.Tests/Utils/PollerTests.cs
+++ b/source/Common.Tests/Utils/PollerTests.cs
@@ -169,6 +169,10 @@ public sealed class PollerTests
     [Fact]
     public async Task RepeatedRestarts_KeepASingleTickInFlightAndStopCompletely()
     {
+        const int restarts = 20;
+        using var tickEntered = new SemaphoreSlim(0);
+        using var releaseTick = new SemaphoreSlim(0);
+        using var stopStarted = new ManualResetEventSlim(false);
         var sync = new object();
         int ticksEntered = 0;
         int ticksInFlight = 0;
@@ -182,33 +186,57 @@ public sealed class PollerTests
                 if (ticksInFlight > mostTicksInFlight) mostTicksInFlight = ticksInFlight;
             }
 
-            Thread.Sleep(1);
+            tickEntered.Release();
+            releaseTick.Wait();
 
             lock (sync) ticksInFlight--;
         }, TimeSpan.FromMilliseconds(1));
 
         try
         {
-            for (int restart = 0; restart < 200; restart++)
+            poller.Start();
+            Assert.True(tickEntered.Wait(TimeSpan.FromSeconds(5)));
+
+            for (int restart = 1; restart <= restarts; restart++)
             {
                 poller.Start();
+
+                // The replacement loop has to wait out the tick that is still running, so nothing enters here.
+                Assert.False(tickEntered.Wait(TimeSpan.FromMilliseconds(100)), $"restart {restart} started a second tick next to the one in flight");
+
+                releaseTick.Release();
+
+                // A restart only counts as a handoff once the replacement loop has run a tick of its own.
+                Assert.True(tickEntered.Wait(TimeSpan.FromSeconds(5)), $"restart {restart} never reached its replacement tick");
             }
 
-            Assert.True(poller.StopAndWait(TimeSpan.FromSeconds(10)));
+            Task<bool> stopTask = Task.Run(() =>
+            {
+                stopStarted.Set();
+                return poller.StopAndWait(TimeSpan.FromSeconds(10));
+            });
+            Assert.True(stopStarted.Wait(TimeSpan.FromSeconds(5)));
+            Task firstCompleted = await Task.WhenAny(stopTask, Task.Delay(TimeSpan.FromMilliseconds(100)));
+            Assert.NotSame(stopTask, firstCompleted);
 
-            int ticksWhenStopped;
-            lock (sync) ticksWhenStopped = ticksEntered;
+            releaseTick.Release();
+
+            Assert.True(await stopTask);
 
             await Task.Delay(TimeSpan.FromMilliseconds(200));
 
             lock (sync)
             {
-                Assert.Equal(ticksWhenStopped, ticksEntered);
-                Assert.True(mostTicksInFlight <= 1, $"{mostTicksInFlight} ticks ran at the same time");
+                // Every restart handed over to exactly one replacement tick, and nothing ticked after the stop.
+                Assert.Equal(restarts + 1, ticksEntered);
+                Assert.Equal(1, mostTicksInFlight);
             }
         }
         finally
         {
+            // Cancel before releasing, so the freed ticks end their loops instead of starting another round.
+            poller.Stop();
+            releaseTick.Release(restarts + 2);
             poller.StopAndWait(TimeSpan.FromSeconds(5));
         }
     }
--- a/source/Common.Tests/Utils/PollerTests.cs
+++ b/source/Common.Tests/Utils/PollerTests.cs
@@ -265,4 +265,58 @@ public sealed class PollerTests
             poller.StopAndWait(TimeSpan.FromSeconds(5));
         }
     }
+
+    [Fact]
+    public void Start_CalledFromThePollingFunction_RunsTheReplacementLoopOnItsOwn()
+    {
+        using var restartRequested = new ManualResetEventSlim(false);
+        using var releaseFirstTick = new ManualResetEventSlim(false);
+        using var laterTickEntered = new ManualResetEventSlim(false);
+        var sync = new object();
+        int ticksEntered = 0;
+        int ticksInFlight = 0;
+        int mostTicksInFlight = 0;
+        Poller poller = null!;
+        poller = new Poller(_ =>
+        {
+            bool isFirstTick;
+            lock (sync)
+            {
+                isFirstTick = ++ticksEntered == 1;
+                ticksInFlight++;
+                if (ticksInFlight > mostTicksInFlight) mostTicksInFlight = ticksInFlight;
+            }
+
+            if (isFirstTick)
+            {
+                poller.Start();
+                restartRequested.Set();
+                releaseFirstTick.Wait();
+            }
+            else
+            {
+                laterTickEntered.Set();
+            }
+
+            lock (sync) ticksInFlight--;
+        }, TimeSpan.FromSeconds(1));
+
+        try
+        {
+            poller.Start();
+            Assert.True(restartRequested.Wait(TimeSpan.FromSeconds(5)));
+
+            Assert.False(laterTickEntered.Wait(TimeSpan.FromSeconds(1)));
+
+            releaseFirstTick.Set();
+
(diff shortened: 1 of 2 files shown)


garrettluskey:
the restart race needs fixing, but please keep the existing API and simplify this to one lifecycle lock, one tracked task and one cancellation source instead of `PollLoop`. `Start` can cancel the current source and schedule `RunAsync(previousTask, source)`, which waits for its predecessor before calling `PollAsync(source.Token)` and clears its source field only if it still owns it, then disposes its source under the lock in `finally`.

keep the completed task tracked until the next start. `Stop` and `StopAndWait` can share a method that cancels and captures the task under the lock, with the wait and self-wait protection outside it. please retain the restart tests and add callback-initiated restart coverage, without expanding this into a disposal API or changing polling intervals.

source/Common.Tests/Utils/PollerTests.cs
```diff
@@ -69,4 +69,254 @@ public void StopAndWait_CalledFromThePollingFunction_DoesNotWaitOnItself()
...
+            Thread.Sleep(1);
+
+            lock (sync) ticksInFlight--;
+        }, TimeSpan.FromMilliseconds(1));
+
+        try
+        {
+            for (int restart = 0; restart < 200; restart++)
+            {
+                poller.Start();
```
gamingmenacers-svg, line 202: non-blocking test coverage note: this can cancel all 200 starts before any callback runs, and the final assertions still pass with `ticksEntered == 0` and `mostTicksInFlight == 0`. please synchronize the repeated restarts with callback-entry/release barriers and require a replacement tick, so this test proves repeated handoffs between active loops.

gamingmenacers-svg (commented):
reviewed `8ecd017a0b53f220e86cc0c8fa4edac38fdbe253`. the per-loop token and predecessor wait address the overlapping-loop race while keeping `Stop()` non-blocking and the callback self-wait check intact. no blocking findings in the lifecycle change; one test coverage note inline.

all 8 `PollerTests` passed with `dotnet test source/Common.Tests/Common.Tests.csproj -c Release --filter FullyQualifiedName~PollerTests`. this was focused local validation, not a full regression suite or live mission test.

garrettluskey (changes requested):
the restart race needs fixing, but please keep the existing API and simplify this to one lifecycle lock, one tracked task and one cancellation source instead of PollLoop. Start can cancel the current source and schedule RunAsync(previousTask, source), which waits for its predecessor before calling PollAsync(source.Token) and clears its source field only if it still owns it, then disposes its source under the lock in finally.

keep the completed task tracked until the next start. Stop and StopAndWait can share a method that cancels and captures the task under the lock, with the wait and self-wait protection outside it. please retain the restart tests and add callback-initiated restart coverage, without expanding this into a disposal API or changing polling intervals.

Thomas-at-sb:
Hi, this is up to date with development now and all checks are green. The run had been waiting for approval since 08.09, so there was no test result to look at before.

Your points are all in since 09.09. The one thing still open is the question from that comment: keep the old rapid-fire restart loop next to the synchronised one, or leave it out?

garrettluskey (approved):
the lifecycle simplification and callback restart coverage are in place. the synchronized restart test is enough, no need to keep the old rapid-fire loop. all 8 poller tests passed locally.

Issue #3113, opened 2026-08-16
[ISSUE] Clear battle agent registry during mission teardown


ShoT-UPfps:
# For New Contributors
[How to contribute](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/wiki/Getting-Started-as-a-Contributor)

## Description

`NetworkAgentRegistry` is registered `InstancePerLifetimeScope`, which currently gives the client one registry for the co-op session.

The registry strongly references mission agents through multiple indexes:

- `Dictionary<Agent, CoopAgentInfo>`
- `Dictionary<Guid, CoopAgentInfo>`
- Movement-ID mappings
- Controller-to-agent lists

Each `CoopAgentInfo` retains the Bannerlord `Agent`, ownership strings/IDs, and potentially authoritative equipment state.

`BattleInstanceLifecycle.Handle_PlayerEnteredBattle` clears the agent registry when a battle starts. `BattleInstanceLifecycle.Leave` stops the mission network, clears the world-item registry, and ends the mission context, but it does not clear the agent registry.

Surviving agents from the completed battle can therefore remain reachable while the client is back on the campaign map. They are normally released only when another mission starts and the entry path clears the registry, or when the whole co-op session ends.

This can make post-battle managed memory remain elevated and retains wrappers around native mission objects longer than their valid lifetime.

There is also a smaller container leak inside `NetworkAgentRegistry.RemoveAgentInternal`: when the final agent is removed from one controller's list, the empty controller key/list is not removed. `TryTransferAuthority` already prunes an emptied source list, so the individual remove path should be consistent.

## Intended Design

Clear every mission-owned agent index as part of mission teardown, not as deferred cleanup at the beginning of a later mission.

For battle teardown:

- Stop and dispose background pollers/handlers that read the registry.
- Prevent new registrations.
- Clear the registry before stale native agent wrappers can be used.
- Ensure the clear runs from a `finally`/guaranteed teardown path even if another cleanup step throws.

Audit location and tournament teardown as well because the registry is shared by the mission stack. Do not add duplicate clears blindly; establish one clear ownership point per mission lifecycle and make it idempotent.

When an individual agent removal empties a controller list, remove the controller entry.

This issue must preserve the teardown ordering established by `#1645` / PR `#1649`: a poller must not race native agent destruction or registry cleanup.

## Location

Create a branch based from [development](https://github.com/Bannerlord-Coop-Team/BannerlordCoop)

[...]

PR #3545, opened 2026-09-07
Clear the agent registry on mission teardown

Thomas-at-sb:
## PR Checklist

## PR Type

`Handle_PlayerEnteredBattle` clears the agent registry on entry
(`BattleInstanceLifecycle.cs:104`, line numbers in this section are on
`development`). `Leave` stops the socket and clears the world-item registry
(`BattleInstanceLifecycle.cs:121` and `:122`) but never clears the agent registry.
`NetworkAgentRegistry` is registered `InstancePerLifetimeScope`
(`MissionModule.cs:242`), so it outlives the mission and the finished battle's
agents stay reachable through `AgentToInfo`, `IdToInfo`, `MovementIdToInfo` and
`ControllerAgentMap` while the client is back on the campaign map. They are
released only when the next mission's entry path clears the registry, or when the
session ends.

`RemoveAgentInternal` (`NetworkAgentRegistry.cs:262`) takes the agent out of its
controller's list but leaves the emptied list in `ControllerAgentMap`, so
`GetControllerIds` returns a controller that holds no agent. Its documentation at
`NetworkAgentRegistry.cs:51` says it returns every controller that currently holds
at least one registered agent, and `TryTransferAuthority` prunes the same map for
that reason.

## What is the new behavior?

Part of #3113

`CoopMissionController.OnEndMissionInternal` now detaches the mission's services
and clears the agent registry from a `finally`, so every exit path empties it,
including the one where one of the twelve teardown steps before the clear throws:
the nine handler disposals at `CoopMissionController.cs:159-168` (line numbers for
that file are on this branch) and, in `CoopBattleController.OnLeaving`,
`damageRouter.FlushForMissionEnd`, the siege-engine report and
`ResultCommitter.ReportResolvedResult` (`CoopBattleController.cs:593-602`).
`Dispose` moves from the end of the method into that `finally` and runs before the
clear, because the services that register battle agents (`PuppetSpawner.cs:304`,
`OwnedAgentReplicator.cs:281`) are detached there and clearing first would let a
spawn still in flight refill the registry. `RemoveAgentInternal` drops the
controller key with its last agent.

Battles, locations and tournaments share `CoopMissionController`, so all three now
clear on the way out instead of leaving it to the next mission's entry path. The
entry clears stay (`BattleInstanceLifecycle.cs:104`,
`CoopLocationsController.cs:439`, `CoopTournamentController.cs:182`), because a
mission whose teardown never ran at all still needs them.

`network.Stop()` stays in `BattleInstanceLifecycle.Leave`
(`BattleInstanceLifecycle.cs:121`), so on every path that reaches `Leave` the
socket is stopped before the clear and the ordering from `#1649` is unchanged. A
throw before `network.Stop()` now empties the registry while the socket is still
up, so a late packet finds no agent instead of resolving one, and nothing refills
it because the registering services are detached first. Locations and tournaments
have the same shape, their `network.Stop()` sits inside `OnLeaving`
(`CoopLocationsController.cs:507`, `CoopTournamentController.cs:2027`).

Covered by six tests in `CoopMissionControllerTeardownTests` (every index cleared,
a throwing handler disposal, a throwing `OnLeaving`, detach before clear, teardown
twice, a second mission not seeing the first mission's agents) and
`NetworkAgentRegistryTests.RemoveAgent_DropsTheControllerEntryWithItsLastAgent`.
All seven fail on `development` and pass here. Manual check with two clients:
fight a field battle to the end, return to the campaign map, then fight a second
one and confirm both sides still see each other's troops.

Two limits worth naming. A throw in the first of the nine handler disposals still
skips the remaining eight, which predates this change and is left alone. `Prevent
new registrations` from #3113 is not implemented here: nothing stops a
`GameThread` action queued before the clear from registering an agent after it,
which needs a mission generation stamp on the queued action and is a separate
change.


source/Missions/Battles/BattleInstanceLifecycle.cs
```diff
@@ -121,6 +121,11 @@ public void Leave()
         network.Stop();
         worldItemRegistry.Clear();
 
+        // Release this battle's agents here rather than leaving them to the next mission's entry path, because
+        // the registry outlives the mission and would otherwise keep wrappers around destroyed native agents
+        // reachable while the client is back on the campaign map.
+        coopMissionComponent.AgentRegistry.Clear();
+
```
garrettluskey: this clear is skipped if the leave announcement or an earlier teardown step throws, so the agent references are still retained on exception teardown. #3113 explicitly requires guaranteed cleanup; put the registry clear in a `finally` path while preserving network shutdown ordering, and add a test with a throwing teardown collaborator.

garrettluskey (changes requested):
the normal battle-exit clear and empty-controller pruning are small and straightforward. the guaranteed cleanup required by #3113 is still missing. source review only; the new tests and live battle exit were not run.

source/Missions/CoopMissionController.cs
```diff
@@ -151,23 +151,39 @@ protected virtual void OnLeaving() { }
...
+            // Detach and clear on every exit path, because the registry outlives the mission and a throw in a
+            // step above would otherwise keep wrappers around destroyed native agents reachable from the
+            // campaign map. Detach first: the services that register agents are torn down in Dispose.
+            try
+            {
+                Dispose();
+            }
+            finally
+            {
+                coopMissionComponent.AgentRegistry.Clear();
```
gamingmenacers-svg, line 185: please keep #3113 open for the queued-registration follow-up instead of marking this as resolving the whole issue. `PuppetSpawner.Dispose()` only unsubscribes; an action already queued by `Handle_NetworkSpawnBattleAgents` can still run afterward, and `SpawnPuppetBatch` only checks that `Mission.Current` is non-null. the clear handles retained entries, but it does not prevent that old action from registering agents again.

gamingmenacers-svg (commented):
reviewed `832bf9bfda91e85a37a92d2e613ae5550495b2f6`. the nested `finally` addresses the earlier skipped-clear finding, including a throw from `Dispose()`, and removing the last agent now prunes its controller entry under the existing registry lock. no additional blocking findings in those changes; the acknowledged queued-registration gap still needs to remain tracked.

all 19 tests selected by `dotnet test source/E2E.Tests/E2E.Tests.csproj -c Release --no-restore --filter "FullyQualifiedName~CoopMissionControllerTeardownTests|FullyQualifiedName~NetworkAgentRegistryTests" -p:PostBuildEvent=` passed. restore required `dotnet restore source/E2E.Tests/E2E.Tests.csproj -p:NuGetAudit=false --force` because existing Scriban advisories were treated as errors. the full E2E suite and live two-client battle-exit test were not run.

garrettluskey (approved):
the nested finally fixes the earlier skipped-clear finding, and empty controller entries are removed under the existing registry lock. no blocking findings for this scoped cleanup; the queued-registration follow-up remains separate under #3113. all 19 teardown and agent-registry tests pass locally. live two-client testing and the full E2E suite were not run.

PR #3546, opened 2026-09-07
Prune dead broker subscriptions on subscribe and unsubscribe

### LINKED ISSUE #3115: [ISSUE] Prune dead MessageBroker weak subscriptions without requiring another publish
# For New Contributors

## Description

`MessageBroker` stores subscriptions in a process-wide dictionary keyed by message type. Each list contains `WeakDelegate` objects, so the broker intentionally does not strongly root the subscriber target.

When a target has been garbage-collected, the dead `WeakDelegate` entry is removed only while publishing that same message type.

If a mission or session leaves behind a weak subscription and that message type is never published again, the broker retains:

- The message-type dictionary key.
- The subscription list.
- The `WeakDelegate`.
- Its `WeakReference`.
- Its `MethodInfo`.

The dead entry no longer retains the original handler object, so this is smaller than the mission-scope leak, but repeated missed unsubscriptions can cause monotonic process-lifetime growth in broker metadata.

This is distinct from `#1648` / PR `#1758`, which added deterministic disposal/unsubscription for a known set of mission handlers. The broker should still clean up dead weak entries when a subscriber misses deterministic teardown or when a short-lived target is collected.

Any change must respect the weak-delegate behavior documented in `#1317`: wrapping subscriptions in an unrooted closure can make a live handler disappear after GC. Do not "fix" dead entries by strongly rooting every target or anonymous wrapper indefinitely.

## Intended Design

Provide deterministic dead-entry pruning that does not depend solely on a later `Publish<T>`.

Acceptable designs include:

- Prune dead entries for a message type during `Subscribe<T>` and `Unsubscribe<T>`.
- Return an `IDisposable` subscription token and use it from owned components, while retaining weak safety as a fallback.
- Add an explicit broker maintenance/prune operation invoked at session/mission boundaries.
- Another design that removes dead entries and empty type keys without introducing strong roots.

Preserve:

- Synchronous message delivery.
- Existing ordering among live subscribers.
- Weak target semantics unless an explicit owned subscription token is used.
- Correct `Unsubscribe` behavior.
- No duplicate invocation.
- Safe removal while publishing.

Do not broaden this issue into a full asynchronous event-bus rewrite.

## Location

Primary locations:

- `source/Common/Messaging/MessageBroker.cs`
- `source/Common/Messaging/WeakDelegate.cs`
- MessageBroker tests under `source/Common.Tests`
- Mission/session teardown paths if an explicit prune operation or subscription token is introduced

## Related Issues

- Mission handlers that previously missed deterministic unsubscribe: `#1648`, PR `#1758`
- Weak closure/subscription ownership warning: `#1317`

## Requirements

- A dead weak subscription can be removed without publishing that message type.
- Empty subscription lists remove their dictionary key.
- A live subscriber is never removed merely because a temporary wrapper delegate was collected.
- The broker does not introduce a permanent strong reference to ordinary weak subscriber targets.
- `Subscribe<T>` still prevents duplicate registration for the same target and method.
- `Unsubscribe<T>` remains reliable.
- Publishing invokes every live subscriber exactly once and in the same order as before.
- Pruning during or around publish cannot skip a live next entry or throw collection-modification exceptions.
- Broker/session disposal clears all remaining entries.
- Add diagnostics or test-only observability sufficient to verify subscriber/type counts.

## Verification

Add tests covering:

- Subscribe an instance method, drop the target, force GC, and prune without publishing; the entry is removed.
- Removing the last dead entry removes the message-type key.
- A live target survives pruning and receives the message.
- Two live subscribers retain registration order.
- Duplicate subscribe remains suppressed.
- Explicit unsubscribe removes the correct target/method.
- A dead subscriber adjacent to live subscribers does not cause a live subscriber to be skipped.
- Lambda/closure behavior is pinned so a weak wrapper is not silently mistaken for a durable subscription.
- Static or targetless delegates are either supported correctly or rejected/documented explicitly.
- Concurrent publish/prune behavior is safe under the broker's intended threading model.
- Repeated creation/collection of short-lived subscribers does not grow broker entry counts.

## Definition of Done

- [ ] Dead weak entries can be pruned without a later publish.
- [ ] Empty message-type keys are removed.
- [ ] Weak ownership and synchronous ordering remain correct.
- [ ] Repeated short-lived subscriptions no longer grow broker metadata indefinitely.
- [ ] Common, integration, and E2E tests remain green.

Thomas-at-sb:
## PR Checklist

## PR Type

`MessageBroker` keeps subscriptions in a dictionary keyed by message type
(`MessageBroker.cs:21`) holding `WeakDelegate` entries, so a subscriber target is
not rooted. A collected target is dropped only while that same message type is
published (`MessageBroker.cs:59`). `Subscribe` (`MessageBroker.cs:75`) and
`Unsubscribe` (`MessageBroker.cs:86`) never look at `IsAlive`, so a message type
that is not published again keeps its dictionary key, its list, the
`WeakDelegate`, its `WeakReference` and its `MethodInfo`. `Unsubscribe` removes
the type key only on an empty list (`MessageBroker.cs:93`), and a dead entry
keeps that count above zero.

#1758 disposes a named set of per-mission handlers at mission teardown, so those
unsubscribe deterministically. Anything outside that set still waits for a later
`Publish<T>` of the same type.

## What is the new behavior?

Resolves #3115

`Subscribe` and `Unsubscribe` drop entries whose target was collected, so a dead
entry goes away on the next subscribe or unsubscribe for that message type, and an
empty list removes its dictionary key.

`Publish` walks its list by index and only checks each entry as it reaches it, so an
entry it already passed can still die during a later callback of the same publish.
Removing that entry shifts the rest down and the next live subscriber is never
invoked. `Publish` therefore records the list it is walking (`MessageBroker.cs:59`),
one entry per running publish, and `RemoveDeadSubscribers` notes that list instead of
pruning it (`MessageBroker.cs:135`). The publish that leaves that list last catches
the pruning up in its `finally` (`MessageBroker.cs:96`).

The record is per list rather than per broker, because most subscribe and unsubscribe
calls here run inside a callback: the state machines switch from a handler, and the
old state unsubscribes while the new one subscribes. A broker-wide guard would skip
pruning on exactly those paths, and a publish of one message type would also block
pruning of an unrelated one.

`Unsubscribe` already removed the message type key once its list ran empty, so the
deferred prune removes that key as well; otherwise `Count` never reaches zero and an
empty list is left under a live key.

The two lists are guarded by a lock of their own (`MessageBroker.cs:25`), because they
replace an `Interlocked` counter and a counter cannot be left unbalanced by an
interleaving. Two threads do publish here: the poller task reaches `Publish` through
`CoopClient.cs:232` and `:243`, and the game thread through `CoopClient.cs:129`. A
lost list update would mark a list as walked for good and stop pruning that message
type for the lifetime of the broker. No lock is held while a subscriber runs, and
nothing else is guarded by it: every other access to the subscriber lists and the
dictionary stays outside the lock, exactly as on `development`.

The delivery loop is unchanged apart from indentation, weak ownership stays as it is,
and no strong root is added. The publish warning at `MessageBroker.cs:72` fires less
often, because an entry can be gone before the next publish of that type.

`Unsubscribe` still removes its own entry directly (`MessageBroker.cs:122`) and
`Publish` still removes entries it walks past (`MessageBroker.cs:74`), so a handler
that unsubscribes itself, and a nested publish of the same type, can still make a
running loop skip the next entry. Both predate this change and behave the same on
`development`. The broker is not thread safe before or after this change; the lock
covers only the bookkeeping this change introduces.

`MessageBrokerTests` compiles against `development`, where seven of its fourteen
tests fail.

### DIFF
--- a/source/Common/Messaging/MessageBroker.cs
+++ b/source/Common/Messaging/MessageBroker.cs
@@ -2,7 +2,6 @@
 using Serilog;
 using System;
 using System.Collections.Generic;
-using System.Threading;
 
 namespace Common.Messaging;
 
@@ -20,8 +19,11 @@ public class MessageBroker : IMessageBroker
     private static readonly ILogger Logger = LogManager.GetLogger<MessageBroker>();
     protected static MessageBroker instance;
     protected readonly Dictionary<Type, List<WeakDelegate>> subscribers;
-    // Publishes currently running on this broker, counted with Interlocked because the network thread and the game thread both publish
-    private int publishDepth;
+    // Subscriber lists a publish is currently walking, one entry per running publish, so pruning a
+    // list holds off while a loop is stepping through it by index
+    private readonly List<List<WeakDelegate>> walkedByPublish = new List<List<WeakDelegate>>();
+    // Lists whose pruning was skipped for that reason, caught up by the publish that leaves last
+    private readonly List<List<WeakDelegate>> prunePending = new List<List<WeakDelegate>>();
     public static MessageBroker Instance { 
         get
         {
@@ -48,7 +50,7 @@ public class MessageBroker : IMessageBroker
         var delegates = subscribers[typeof(T)];
         if (delegates == null || delegates.Count == 0) return;
         var payload = new MessagePayload<T>(source, message);
-        Interlocked.Increment(ref publishDepth);
+        walkedByPublish.Add(delegates);
         try
         {
             for (int i = 0; i < delegates.Count; i++)
@@ -79,7 +81,13 @@ public class MessageBroker : IMessageBroker
         }
         finally
         {
-            Interlocked.Decrement(ref publishDepth);
+            walkedByPublish.Remove(delegates);
+            if (!walkedByPublish.Contains(delegates) && prunePending.Remove(delegates))
+            {
+                delegates.RemoveAll(weakDelegate => !weakDelegate.IsAlive);
+                if (delegates.Count == 0 && subscribers.TryGetValue(typeof(T), out var current) && current == delegates)
+                    subscribers.Remove(typeof(T));
+            }
         }
     }
 
@@ -108,11 +116,16 @@ public class MessageBroker : IMessageBroker
     }
 
     // Entries whose target was collected otherwise sit here until this message type is published again.
-    // A running publish walks its list by index, so removing an entry it already passed would make it
-    // skip the next live subscriber; that list is left alone until the publish is done.
+    // Pruning a list that a publish is stepping through by index would move entries past its position,
+    // so that list is noted here and pruned in the finally of that publish instead.
     private void RemoveDeadSubscribers(List<WeakDelegate> delegates)
     {
-        if (publishDepth > 0) return;
+        if (walkedByPublish.Contains(delegates))
+        {
+            if (!prunePending.Contains(delegates))
+                prunePending.Add(delegates);
+            return;
+        }
         delegates.RemoveAll(weakDelegate => !weakDelegate.IsAlive);
     }
 
--- a/source/Common/Messaging/MessageBroker.cs
+++ b/source/Common/Messaging/MessageBroker.cs
@@ -19,6 +19,10 @@ public class MessageBroker : IMessageBroker
     private static readonly ILogger Logger = LogManager.GetLogger<MessageBroker>();
     protected static MessageBroker instance;
     protected readonly Dictionary<Type, List<WeakDelegate>> subscribers;
+    // Guards the two lists below. They replace an Interlocked counter, and a counter cannot be left
+    // unbalanced by an interleaving, while a lost list update would mark a list as walked for good
+    // and stop pruning it for the lifetime of the broker
+    private readonly object bookkeepingLock = new object();
     // Subscriber lists a publish is currently walking, one entry per running publish, so pruning a
     // list holds off while a loop is stepping through it by index
     private readonly List<List<WeakDelegate>> walkedByPublish = new List<List<WeakDelegate>>();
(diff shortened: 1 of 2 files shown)


source/Common/Messaging/MessageBroker.cs
```diff
@@ -90,10 +91,17 @@ public virtual void Unsubscribe<T>(Action<MessagePayload<T>> subscription) where
...
         if (delegates.Count == 0)
             subscribers.Remove(typeof(T));
     }
 
+    // Entries whose target was collected otherwise sit here until this message type is published again
+    private static void RemoveDeadSubscribers(List<WeakDelegate> delegates)
+    {
+        delegates.RemoveAll(weakDelegate => !weakDelegate.IsAlive);
+    }
+
```
garrettluskey, line 144: `RemoveAll` can shift entries behind `Publish`'s current index. with subscribers A, B, C, if A is collected after its callback and B calls `Subscribe`, pruning removes A and shifts C to index 1; the publish loop advances to index 2 and skips C. this reproduces on this branch and passes on its parent. keep pruning from shifting the list beneath an active publish, and add a regression where the collected target is before the current callback, not after it.

garrettluskey (changes requested):
the cleanup is useful and the scope is proportionate: eight production lines, no new dependencies or public API, and tests for the weak-subscription behavior. all 174 Common tests pass.

changes are needed because pruning during a callback can now skip a live subscriber. a targeted regression test fails on this head and passes on its parent. fixing that within the current scope is worthwhile; this does not need a broader broker redesign.

gamingmenacers-svg (commented):
reviewed b0aaf6e3560819d390e2fdd9c280da19bb4f5da9. the deferred per-list cleanup addresses the earlier skipped-subscriber finding, and the finally block checks that the dictionary still points to that list before removing its key. all 178 Common tests passed locally, including the 14 broker tests; no new blocking findings. the existing self-unsubscribe/nested-publish behavior and broader thread safety remain outside this change.

gamingmenacers-svg (commented):
Completed change, review found 0 findings

garrettluskey (approved):
the deferred cleanup fixes the earlier skipped-subscriber regression without changing weak ownership or callback delivery. no blocking findings on this head. all 183 Common tests pass locally, including the 14 broker tests; the build check passed and the remaining CI tests are still running.

PR #3552, opened 2026-09-08
reduce campaign join replay stalls

gamingmenacers-svg:
## PR Checklist

- [x] All new classes have class-level documentation comments
- [x] Tests for the changes have been added

## PR Type

- [x] Bug fix

a loading player's replay backlog can pause the whole campaign after the join grace period. replay also drains in large bursts, putting pressure on the reliable channel and the joining client's game thread.

## What is the new behavior?

loading players catch up without triggering the global overload pause. their replay is bounded by packet count, serialized bytes and server frame time, with at most four batches awaiting ordered client application. the joining client applies queued work under an 8 ms frame budget. stalled or oversized joins are disconnected individually; overload protection for players already in the campaign is preserved. failed joins show whether replay timed out, exceeded its queue limit, or timed out loading the save. queue-order tests drain every bounded batch before checking FIFO.

adjacent positive item additions are combined, and adjacent troop updates share bounded batches while retaining every operation in order. item removals, clears, object lifetimes and other intervening events remain barriers. the native roster regression test checks original and compacted replay at intermediate observations, including over-removal, clear/repopulate and troop XP changes.

the existing party and kingdom baselines, join lifecycle events, save metadata and exact-build validation remain in place. clients and servers must use the matching new build for the added replay acknowledgements. no required members were added to the public network or configuration interfaces.

## Bot Changelog Entry

- Reduced campaign join catch-up stalls and prevented slow-loading players from pausing the campaign for everyone.

## Other information

validated the review repair with a release build, all unit/integration suites (2,553 passed; 14 skips), and 28 focused disconnect/queue tests. seven loopback cases cover the actual server-to-client disconnect payload, including missing, malformed and unknown reasons. an independent agent reviewed commit a185d5686 with no unresolved findings. simulated gameplay E2E tests were not rerun for this repair.

native save capture remains synchronous and can still cause a brief hitch. this change targets replay workload and the global catch-up pause; live campaign load times have not been measured.

### DIFF
--- a/source/Coop.Core/Server/Connections/JoinPeerTerminator.cs
+++ b/source/Coop.Core/Server/Connections/JoinPeerTerminator.cs
@@ -1,0 +1,15 @@
+﻿using LiteNetLib;
+
+namespace Coop.Core.Server.Connections;
+
+/// <summary>Disconnects a failed join after its replay has been made terminal.</summary>
+public interface IJoinPeerTerminator
+{
+    void Disconnect(NetPeer peer, string reason);
+}
+
+/// <inheritdoc cref="IJoinPeerTerminator"/>
+internal sealed class JoinPeerTerminator : IJoinPeerTerminator
+{
+    public void Disconnect(NetPeer peer, string reason) => peer.Disconnect();
+}
--- a/source/Coop.Core/Server/Connections/JoinPeerTerminator.cs
+++ b/source/Coop.Core/Server/Connections/JoinPeerTerminator.cs
@@ -1,4 +1,5 @@
 ﻿using LiteNetLib;
+using LiteNetLib.Utils;
 
 namespace Coop.Core.Server.Connections;
 
@@ -11,5 +12,10 @@ public interface IJoinPeerTerminator
 /// <inheritdoc cref="IJoinPeerTerminator"/>
 internal sealed class JoinPeerTerminator : IJoinPeerTerminator
 {
-    public void Disconnect(NetPeer peer, string reason) => peer.Disconnect();
+    public void Disconnect(NetPeer peer, string reason)
+    {
+        var data = new NetDataWriter();
+        data.Put(reason);
+        peer.Disconnect(data);
+    }
 }
--- a/source/Coop.Tests/Server/Connections/ConnectionMessageQueueTests.cs
+++ b/source/Coop.Tests/Server/Connections/ConnectionMessageQueueTests.cs
@@ -51,6 +51,11 @@ public class ConnectionMessageQueueTests
 
     private bool NothingSentTo(NetPeer peer) => network.SentPayloads.ContainsKey(peer.Id) == false;
 
+    private void Drain(NetPeer peer)
+    {
+        while (queue.FlushBatch(peer).HasMore) { }
+    }
+
     [Fact]
     public void PeerVisibleBeforePlayerConnected_DropsWorldBroadcasts()
     {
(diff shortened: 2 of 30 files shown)


source/Coop.Core/Server/Connections/JoinPeerTerminator.cs
```diff
@@ -0,0 +1,15 @@
...
+/// <summary>Disconnects a failed join after its replay has been made terminal.</summary>
+public interface IJoinPeerTerminator
+{
+    void Disconnect(NetPeer peer, string reason);
+}
+
+/// <inheritdoc cref="IJoinPeerTerminator"/>
+internal sealed class JoinPeerTerminator : IJoinPeerTerminator
+{
+    public void Disconnect(NetPeer peer, string reason) => peer.Disconnect();
```
garrettluskey: `reason` is discarded here, so a player disconnected by either join watchdog only sees "You have been Disconnected" from `DisconnectHandler`, with no indication that replay timed out or exceeded its limit. send the reason and display it through the existing client disconnect handler so a failed join has a useful explanation.

source/Coop.Tests/Server/Connections/ConnectionMessageQueueTests.cs
```diff
@@ -91,7 +94,7 @@ public void Queueing_HoldsBroadcasts_ThenFlushesFifoWhileRemainingQueued()
         // Held, not sent, while the peer loads.
         Assert.True(NothingSentTo(peer));
 
-        queue.Flush(peer);
+        queue.FlushBatch(peer);
```
garrettluskey: `FlushBatch` can stop after one packet when its 4 ms budget expires, but this test still expects all three after one call. this is already failing in [CI](https://github.com/Bannerlord-Coop-Team/BannerlordCoop/actions/runs/34247253790/job/102132820571), drain until `HasMore` is false before checking FIFO order, including the other multi-packet assertions below.

garrettluskey (changes requested):
the replay gate and application acknowledgements have a separate job from the existing send coalescer and packet batcher, and the scoped game-thread budget reuses the existing queue. no redundant replacement component found. changes requested for the discarded join failure reason and a timing-dependent test that is currently failing CI. targeted local tests passed: 102 passed, 1 skipped; live join performance was not measured.

source/Coop.Core/Server/Connections/JoinPeerTerminator.cs
gamingmenacers-svg: fixed in a185d5686. the server sends the join failure reason as disconnect data, and the client copies it before recycling the packet reader. the existing disconnect handler now shows separate messages for replay timeout, queue overflow and save-loading timeout. seven loopback cases cover delivery and fallback behavior; all 2,553 unit/integration tests passed, with 14 skips.

source/Coop.Tests/Server/Connections/ConnectionMessageQueueTests.cs
gamingmenacers-svg: fixed in a185d5686. both multi-packet assertions now drain FlushBatch until HasMore is false, then check the original FIFO order. the frame budget remains unchanged. all 28 focused queue/disconnect tests passed, and the full unit/integration suite passed.

garrettluskey (approved):
both findings are fixed. join failure reasons now reach the existing disconnect handler, and the FIFO tests drain all bounded batches. the repair stays within the existing components without adding a second disconnect path. all 28 focused tests passed locally, including the loopback payload cases. no new blockers found. CI build and unit tests pass; three E2E jobs are still pending. live join performance remains unmeasured.

PR #3553, opened 2026-09-08
Fix rejected puppet deaths leaving unkillable agents

garrettluskey:
replicated deaths can remove a puppet's network registration even when a local death guard leaves it alive, making it unkillable.

Related to https://github.com/Bannerlord-Coop-Team/Bannerlord-Coop-Issues/issues/1688. the reporter's log is missing, so this failure path is reproduced by tests rather than confirmed from that run.

## What is the new behavior?
keep the puppet registered and retry when local death guards clear. only remove registration and casualty attribution after terminal application, and ignore completed duplicate deaths. blocked retries dont repeat native hit effects.

## How to test
`dotnet test source/E2E.Tests/E2E.Tests.csproj -c Release -p:PostBuildEvent= --filter FullyQualifiedName~E2E.Tests.Services.Missions`

812 mission E2E tests pass, including 23 death-mirror cases covering two-client broadcast, blocked retries, killed/wounded outcomes, duplicate deaths, and health boundaries. regressions failed before the fixes. independent review passed. these are headless mock-engine tests, not a live-game reproduction.

## Bot Changelog Entry
- Fixed a case where enemies could remain unkillable after their death was received during a protected mission state.

### DIFF
--- a/source/E2E.Tests/Environment/MockEngine/MissionEngineFixture.cs
+++ b/source/E2E.Tests/Environment/MockEngine/MissionEngineFixture.cs
@@ -725,7 +725,17 @@ public sealed class MissionEngineFixture : IDisposable
 
         registrationMock?.RegisteredBlow?.Invoke(__instance, blow);
 
-        victim.Health -= blow.InflictedDamage;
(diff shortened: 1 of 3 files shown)


gamingmenacers-svg (approved):
All good

PR #3608, opened 2026-09-15
Add "instant" option for move to settlement command

G07cha:
## PR Checklist

- [x] Tests for the changes have been added (for bug fixes / features) - verified manually as well

## PR Type

Movement commands are just giving orders, making them hard to use via MCP as someone needs to toggle the time, agent needs to wait an appropriate amount of time for character to arrive and hope that no looters would want to drink from their skull on their way there.

## What is the new behavior?

Add "instant" argument to the command allowing to instantly teleport to the desired settlement, also works when the party is already in the settlement.

## Other information

Fix handler to correctly handle client teleporting between settlements.


garrettluskey (approved):
the instant move reuses the existing movement sync and keeps server-side patches active. no actionable correctness or simplicity findings. ci checks pass; local builds and live gameplay were not run.

